Issue #21, opened 2016-05-06
getABIVersion() not defined in Object-nt.h


pefoley2:
Broken by https://github.com/dyninst/dyninst/commit/6dd97f53de96d9bbe38f23626c84811624d0d74e
getABIVersion() is only implemented for elf objects.

mcfadden8:
Argh.  I renamed the method and forgot to change the virtual one in Object.h.  Patch is here: [https://github.com/dyninst/dyninst/pull/24](https://github.com/dyninst/dyninst/pull/24)

Unfortunately, I do not have a Windows machine to test this.  Can someone try it to make sure it works?

pefoley2:
https://ci.appveyor.com/project/wrwilliams/dyninst/build/1.0.6 shold let you know if the fix compiles properly.

mcfadden8:
Please do not merge this patch yet.  I have discovered some (PPC64LE) proccontrol test problems and I need to determine whether it is related to this change or to others that have gone in to the tree.

pefoley2:
I'm assuming you mean this error?
[  6%] Building CXX object CMakeFiles/testproccontrol.dir/src/proccontrol/proccontrol_comp.C.o
/home/peter/testsuite/src/proccontrol/proccontrol_comp.C: In member function ‘uint64_t ProcControlComponent::adjustFunctionEntryAddress(Dyninst::ProcControlAPI::Process::const_ptr, uint64_t)’:
/home/peter/testsuite/src/proccontrol/proccontrol_comp.C:343:16: error: ‘class Dyninst::SymReader’ has no member named ‘getABIVersion’
       if (rdr->getABIVersion(major, minor)) {

PR #24, opened 2016-05-06
Provided base class virtual for getABIVersion() that returns false when not implemented

mcfadden8:
Fixed forgotten rename of isPPC64V2API to getABIVersion in
Object.h.  The virtual function in this base class now returns
false indicating that it is not implemented (which is the case
for Object-nt).


pefoley2:
Fixes #21

mcfadden8:
When I was running some tests locally, I noticed some strange proccontrol test failures.  I found that these failures were not due to my change, but were instead due to the fact that I was running against the wrong branch of testsuite.

mcfadden8:
This pull request is now ready.  I believe that it fixes #21.

PR #37, opened 2016-05-10
Parsing fixes for ARM

mxz297:
1. For ARM, add br as a branch instruction category and blr as a call instruction category.
2. Need to skip the first two PLT entry sizes on ARM
3. Only invoke new jump table parsing on x86. On power, we use the old heuristics. On ARM, we currently give up
4. Properly initialize decoder states when decoding operands lazily


wrwilliams:
These changes need to go to master, not arm64.

mxz297:
They are in master now.

PR #43, opened 2016-05-12
Remove low-level warnings from ptrace read/write failures

### LINKED ISSUE #36: ptrace_peektext failing and producing spam in thread tests
Tests are passing, though. Definitely happens on ppc64, and IIRC also on x86.

@cuviper, mind taking a look at this one and seeing whether we should swallow the warning, handle the failure better, or both?

cuviper:
The error cases all return false anyway, so the caller can deal with
failure, which makes this stderr chatter just spam.

Fixes #36.


cuviper:
#39 redux

PR #46, opened 2016-05-13
symtabAPI: comment out some debug chatter

cuviper:
These cout writes were added by commit 758aa22669b4a, but we don't want
such output in general use.


wrwilliams:
Yes, good. Didn't mean for that to go into master.

PR #69, opened 2016-06-01
Truncate PTRACE_GETEVENTMSG exit status to int

### LINKED ISSUE #35: Test 4_1, 4_2, and 4_4 fail on ppc64
Reproduced on vulcan. All modes. 4_3, oddly, passes.

cuviper:
The sys/wait.h macros are expecting to operate on an int, and they have
some ugly pointer macros to deal with old BSD compatibility.  But we get
an unsigned long from PTRACE_GETEVENTMSG.  Normally this work out ok,
but on big-endian ppc64 those macros end up reading the int from the
most significant bits only, which are zero.

Cast the long down to a local int first, so WEXITSTATUS works properly.

Fixes #35.

### DIFF
--- a/proccontrol/src/linux.C
+++ b/proccontrol/src/linux.C
@@ -430,9 +430,9 @@ bool DecoderLinux::decode(ArchEvent *ae, std::vector<Event::ptr> &events)
                                   proc->getPid(), thread->getLWP());
                     if (thread->getLWP() == proc->getPid())
 		            {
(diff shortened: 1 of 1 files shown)


cuviper:
The CI failure appears to be a problem with Travis itself.

pefoley2:
More specifically the upstream llvm apt repository.
Hopefully they'll fix it at some point.

PR #74, opened 2016-06-08
Fix rtlib's os_malloc

wrwilliams:
DYNINSTos_malloc in the RTlib would malloc its heap data structures and string processing data for /proc/maps. Both of these could lead to deadlocks with newer glibc versions. Fixed as follows:

1) mmaps are no longer trying to be smart, but instead asking the OS for something valid and bounds-checking it against the request
2) heap data structures are no longer malloced, but instead tacked onto the end of each heap segment


wrwilliams:
...let's try this again with a rebase, shall we?

PR #75, opened 2016-06-08
release9.2/bugs/rtheap_mmap_only

wrwilliams:


wrwilliams:
Still some spurious commits; once more with feeling.

cuviper:
FWIW, to update a PR you can just push extra commits to the branch, or even force push a rewrite.  Although once there are any review comments, it's better not to force any more because github doesn't show that history well.

Hmm, perhaps you disabled force-pushing on the main repo?  That's one reason why it's a bit nicer to open PRs from a personal github clone.  Another is that people won't get littered with your temporary branches that way.

And should someone *want* to fetch your PR branch directly:

    git fetch origin pull/75/head:pr75

Easily aliased something like:

    git config --global alias.fetch-pr '!f() { git fetch origin pull/$1/head:pr$1; }; f'
    git fetch-pr 75

Issue #50, opened 2016-05-14
Master timeout on test4_2 and test4_4 on Fedora23


jdetter:
`test4_1` and `test4_3` pass and finish quickly, however `test4_2` times out:

```
*** Process exceeded time limit.  Reaping children.
```

wrwilliams:
So fork and/or exit callbacks may be more generally broken than just PPC. Good to know.

Just to double-check: this does, or does not, happen on the same system on master?

jdetter:
I am getting the same behavior with `test4_2` and `test4_4` on the master branch on my machine. It also looks like master is missing a lot of the addressing mode fixes because a lot of bad addressing mode asserts are thrown during the other tests.

jdetter:
Gist of master run on `5/16/2016 at 10:00 AM`: https://gist.github.com/jdetter/7f048ffcef511969e4e248d2e55c9e2d

wrwilliams:
Seems likely there's something non-trivially different on your development box, then. What's the diff from RHEL6, gcc5.3, glibc 2.12, kernel 2.6.32?

jdetter:
Basically everything:

```
Fedora 23
GCC 5.3.1
glibc 2.22
kernel 4.4.8
```

I will retry both VEX and master on the instructional machines.

jdetter:
On Fedora 23, it wouldn't terminate even after 20 minutes of run time:

```
[root@localhost testsuite]# ./test_driver -test test4_2 -log -verbose
Commencing test(s) ...
Tue May 17 12:13:19 CDT 2016
Linux localhost.localdomain 4.4.8-300.fc23.x86_64 #1 SMP Wed Apr 20 16:59:27 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
TESTDIR=/usr/local/bin/testsuite
[Tests with test4_2.dyn_g++_64_none_none]
-rwxr-xr-x. 1 root root 60944 May 10 17:17 test4_2.dyn_g++_64_none_none*
        linux-vdso.so.1 (0x00007ffd42355000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f19ae0a6000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f19adda3000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f19adb8c000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f19ad7cb000)
[...]

PR #76, opened 2016-06-08
release9.2/bugs/rtheap_mmap_only

wrwilliams:
DYNINSTos_malloc in the RTlib would malloc its heap data structures and string processing data for /proc/maps. Both of these could lead to deadlocks with newer glibc versions. Fixed as follows:

1) mmaps are no longer trying to be smart, but instead asking the OS for something valid and bounds-checking it against the request
2) heap data structures are no longer malloced, but instead tacked onto the end of each heap segment

### DIFF
--- a/dyninstAPI_RT/src/RTheap.c
+++ b/dyninstAPI_RT/src/RTheap.c
@@ -283,7 +283,7 @@ void *DYNINSTos_malloc(size_t nbytes, void *lo_addr, void *hi_addr)
     }
 
     /* define new heap */
-    node = heap + size;
+    node = ret_heap + size;
     node->heap.ret_addr = (void *)ret_heap;
     node->heap.addr = heap;
     node->heap.len = size_heap;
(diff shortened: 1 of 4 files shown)


wrwilliams:
Fixes #50; posted for review.

dyninstAPI_RT/src/RTheap.c
```diff
@@ -279,75 +276,31 @@ void *DYNINSTos_malloc(size_t nbytes, void *lo_addr, void *hi_addr)
...
       free(heap);
-      free(node);
 #ifdef DEBUG
       fprintf(stderr, "MALLOC'd area fails range constraints\n");
 #endif 
       return NULL;
     }
 
     /* define new heap */
+    node = heap + size;
```
cuviper: I think it should be `node = ret_heap + size`, but in practice `malloc` probably aligns sufficiently anyway.

cuviper:
Small malloc nit, otherwise looks good to me!

Issue #16, opened 2016-05-06
Missing htobe32 function under Visual Studio


pefoley2:
Broken by https://github.com/dyninst/dyninst/commit/080b3fac7422dc84c6e0d8a163fec12d8a3f002f#diff-79cd8ea5743e12f03821c4269fae3431R276
Visual Studio does not define a htobe32 function as far as I can tell.

wrwilliams:
I can't find a clean endian swap outside of the network code either; boost has one but it's new in 1.58 (which is too new for us to depend on IMO). htonl will IIRC drag in a WinSock link dependency, which is absurd for an endian swap. Probably best addressed with the byte shuffling code that was there before, as Windows (in versions we support, at least) is guaranteed little-endian for the foreseeable future.

pefoley2:
It looks like the "before" was simply not calling that function.
Maybe just ifdef it out if os==windows?

pefoley2:
A cherry-pick of https://github.com/dyninst/dyninst/pull/19/commits/154b0307ac4d1c1a21f8cb4d02634b1ed5d35270 should resolve this issue.

wrwilliams:
Fixed on master, as we've reverted to byte-swapping.

pefoley2:
Still seems to be an issue at https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/RoseInsnFactory.C#L276

wrwilliams:
Fixed by #77.

PR #77, opened 2016-06-08
Define htobe on Windows/MSVC as a wrapper for _byteswap_ulong

wrwilliams:
Define htobe on Windows/MSVC as a wrapper for _byteswap_ulong


wrwilliams:
Should finish off fixes for #16

Issue #72, opened 2016-06-01
walkSingleFrame against local process crashes on Windows


jgerie:
This is a variation on issue #63 which I reported. My ultimate goal is to walk the stack in the local process, not in an external process. When I try my test program against the local process, the first call to walkSingleFrame dies with:

(1fd0.38c0): Access violation - code c0000005 (first chance)

My test program is below. Base on the response on issue #63, I am calling Walker::newWalker(std::string exec_name): 
[self_stackwalk.txt](https://github.com/dyninst/dyninst/files/294167/self_stackwalk.txt)

The call stack from the debugger is:

stackwalk!Dyninst::Stackwalker::LookupFuncStart::allocatesFrame+0x134
stackwalk!Dyninst::Stackwalker::FrameFuncStepperImpl::getCallerFrame+0x3b
stackwalk!Dyninst::Stackwalker::Walker::walkSingleFrame+0x2b4
StackTest!main+0x30f
StackTest!__tmainCRTStartup+0x199
StackTest!mainCRTStartup+0xd
KERNEL32!BaseThreadInitThunk+0x24
ntdll!__RtlUserThreadStart+0x2f
ntdll!_RtlUserThreadStart+0x1b

The failure point (allocatesFrame+0x134) points to line 287 of x86_swk.C:

   result = proc->getLibraryTracker()->getLibraryAtAddr(addr, lib);

The getLibraryTracker() function is returning NULL. I can see in the debugger that the proc->library_tracker member is NULL.

My build and test environment for DynInst is:

Windows 10, 64-bit
Visual Studio 12
Cmake 3.5.2 (cmake-3.5.2-win32-x86)
Boost 1.59 (I downloaded a pre-built Windows version, boost_1_59_0-msvc-12.0-32)
DynInst 9.1.0
I'm building exclusively 32-bit binaries, even though the OS is 64-bit
For most of my testing and debugging, I'm using a Debug build of DynInst. I tried a Release build first and then moved to Debug to try to figure out the problems.

PR #78, opened 2016-06-08
Set defaults for Windows first-party stack walking: library tracker

wrwilliams:
Set defaults for Windows first-party stack walking: library tracker


wrwilliams:
Partial fix for #72, though the symbol reader default needs to get set as well.

PR #82, opened 2016-06-09
Fix for square root floating point instructions

### LINKED ISSUE #79: Problems with Instruction API parsing x86-64 binaries: sqrtsd
The current master version of dyninst has problems parsing several instructions in x86-64 binaries. Below, I provide an example of an instruction that failed to parse (the first instruction) and the instruction following. This example was uncovere using the checkparse regression test developed at Rice and provided to Wisconsin.

**Application example: soliton**

problematic instructions:
`         sqrtsd    instruction length mismatch (xed length=5, instruction api length=4)`

example in context:
 ```
 404203:       f2 0f 51 04 d0          sqrtsd (%rax,%rdx,8),%xmm0 <<<----------------
 404208:       66 0f 2e c0             ucomisd %xmm0,%xmm0
```

jdetter:
Includes small fixes for John Mellor-Crummey's binary parsing issues as well. I have tested the square root instructions and they should all be the correct length now.

Fixes #79


jdetter:
@wrwilliams good to merge to master?

PR #84, opened 2016-06-11
Simplify INTERP logic for better consistency

### LINKED ISSUE #71: parseThat not outputting executable binary (Exec format error)
Mohamed is having an issue where he runs:

```
parseThat --binary-edit=ssh-dyn -i 0 /usr/bin/ssh
```

and the output binary `ssh-dyn` will not execute due to an exec format error. The header information looks correct to me so I'm not sure why exec won't take it:

```
detter@ubuntu-16:~/Mohamed$ readelf -h ./ssh-dyn
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0xdea0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          768992 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         10
  Size of section headers:           64 (bytes)
  Number of section headers:         40
  Section header string table index: 39
```

Ubuntu's version of `/usr/bin/ssh` can be found [here](http://cs.wisc.edu/~detter/ssh)
The `ssh-dyn` binary generated by parseThat can be found [here](http://cs.wisc.edu/~detter/ssh-dyn)

cuviper:
When .interp was in the first page, keep it there like the comment about
FreeBSD requires.  Just shift it for the new PHDR.  But if .interp is
already past the first page, as seen in libc.so for instance, then move
its offset a whole page like we do for every other section.  Then
program headers follow the same logic.

The conditional around startMovingSections and object type is totally
removed, as it did the wrong thing for PIE binaries which otherwise look
like shared libraries.  We now just move all sections except low interp.

Fixes #71.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -573,7 +573,6 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
     unsigned scncount;
     unsigned sectionNumber = 0;
 
-    bool startMovingSections = false;
     for (scncount = 0; (scn = elf_nextscn(oldElf, scn)); scncount++) {
         //copy sections from oldElf to newElf
         shdr = ElfTypes::elf_getshdr(scn);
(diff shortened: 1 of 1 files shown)


cuviper:
While we do test `PIC` and `nonPIC` executables, they're both still `ET_EXEC`.  The problem here arose because the executable was built `PIE`, which is `ET_DYN` like a shared library.

Also, regarding CI failures, if Travis and LLVM can't get along then we should just disable the clang build.  This is travis-ci/travis-ci#6120, which also has a workaround if you like.

pefoley2:
Clang on travis is disabled by #85

wrwilliams:
Thanks, Josh. 

I've got WIP from earlier this year on generating our PIC tests as PIE; the catch is that I need to convince the prolog spec to properly exclude all tests with assembler sources from the PIC/PIE variation. That shouldn't be hard in theory, but prolog.

Final question/comment: is there any good way to generate executables with arbitrarily weird-but-legal ELF headers for testing? (Actually, if we were to do something with directed fuzzing of ELF headers and compared the loader's gripes on pre- and post-rewritten binaries, that might be better/simpler, but there's something to be said for just checking that legal->legal is an invariant.)

cuviper:
I know that elfutils uses [afl-fuzz](http://lcamtuf.coredump.cx/afl/) a lot, but the directed fuzzing there is trying to crash the program under inspection.  It's not oriented around producing bad output AFAIK.  It still might be worthwhile to see how Dyninst fares under such input though -- AFL is good at finding really weird corner cases.

For verifying rewritten output, `eu-elflint --gnu-ld` can probably help.
(That option is necessary because gnu-ld apparently has some [questionable behavior](https://fedorahosted.org/elfutils/wiki/ElflintGNU).)

PR #99, opened 2016-06-16
RTlib: use static TLS for the tramp guard

### LINKED ISSUE #98: Infinite recursion in TLS tramp guard
Using systemtap, I have tried to instrument the `mutex_entry` static probe point in `libpthread.so`.  When any new thread tries to take a lock, it recurses on itself until the stack overflows.  The sequence looks something like:

- Thread reaches the `mutex_entry` address and starts the instrumentation.
- We call `DYNINST_lock_tramp_guard()` to make sure we aren't recursing.
- This accesses the TLS `DYNINST_tls_tramp_guard` as an offset from `__tls_get_addr()`.
- Since it's a new thread, this has to set up the TLS addr, calling `tls_get_addr_tail()`.
- This takes the rtld lock, re-entering `__pthread_mutex_lock()`.
- Hey there, `mutex_entry`... repeat until the stack runs out and we hit SIGSEGV.

I don't think there's anything actually special about the exact `mutex_entry` point here -- it's just a conveniently low-level place to probe.  Probably any point within `__tls_get_addr()` or below will have similar hazards.

cuviper:
With dynamic TLS, the call to __tls_get_addr() could reach additional
instrumented code, infinitely recursing back to check the tramp guard.
Static TLS is a limited resource, but this case in RTlib is special
enough to warrant it for safety alone, nevermind performance.

Fixes #98.

### DIFF
--- a/dyninstAPI_RT/src/RTcommon.c
+++ b/dyninstAPI_RT/src/RTcommon.c
@@ -117,10 +117,15 @@ int fakeTickCount;
 #if defined(_MSC_VER)
 #define TLS_VAR __declspec(thread)
 #else
-#define TLS_VAR __thread
+// Note, the initial-exec model gives us static TLS which can be accessed
+// directly, unlike dynamic TLS that calls __tls_get_addr().  Such calls risk
+// recursing back to us if they're also instrumented, ad infinitum.  Static TLS
+// must be used very sparingly though, because it is a limited resource.
+// *** This case is very special -- do not use IE in general libraries! ***
+#define TLS_VAR __thread __attribute__ ((tls_model("initial-exec")))
 #endif
 
-TLS_VAR int DYNINST_tls_tramp_guard = 1;
+static TLS_VAR char DYNINST_tls_tramp_guard = 1;
 
 DLLEXPORT int DYNINST_lock_tramp_guard()
 {


wrwilliams:
The restriction here is on the quantity of static TLS space that can be allocated (per DSO or for the whole address space?), correct?

cuviper:
The restriction is on the quantity across the whole address space, yes, and older glibc were also limited to the first `DTV_SURPLUS` (14, sometimes patched higher) dynamic libraries.  I think surplus is only referring to `dlopen` libraries, but I'm not sure on that point.

This also becomes a hard requirement, so the library will fail to load at all if its static TLS can't be fulfilled.  You can see it in `eu-readelf -d $DYNINSTAPI_RT_LIB`: `FLAGS STATIC_TLS`.

wrwilliams:
Sounds good; Bart passes along his thanks for this one.

Do you know if icc respects the attribute declaration per-variable, off the top of your head? Clang 3.2+ does, but my quick search has only found icc supporting this as a whole-file -f option.

cuviper:
I don't know about icc, sorry.  They generally seem to strive for gcc compatibility though, so given that they do support `-ftls-model`, I'd be really surprised if they didn't also support the attribute.  I can flip to a command line option if necessary though -- it was just more convenient inline where there's already a !Windows conditional.

cuviper:
**meta**: Do you have some GitHub setting to rebase or squash merges?  I noticed that the merged commit has an updated author date, which isn't a big deal, but it also rewrote my author email to my personal gmail address.  My gmail happens to be the primary on this github account, and redhat.com is secondary, but I would expect this to be left alone in the commit itself.

wrwilliams:
I've been squashing them, but that's unnecessary for single-commit merges, for sure. Didn't realize it was screwing with the author info.

Issue #96, opened 2016-06-15
Rewritten binary dies with SIGILL


melsabagh:
On Ubuntu 16 x64, rewriting whetstone-double (from UnixBench) with anything besides null instrumentation results in illegal instructions:

```
$ parseThat --binary-edit=whetstone-double.dyn -i 3 whetstone-double
Write File Path is whetstone-double.dyn
[ Processing whetstone-double ] ===============================================
Creating new BPatch object.... done.
Retrieving BPatch_image object.... done.
Parsing image for module information.... Found 4 module(s).
Parsing function data from whetstone-double module... Found 12 function(s).
Parsing CFG data from functions in whetstone-double module... done.
Parsing function data from crtfastmath.c module... Found 1 function(s).
Parsing CFG data from functions in crtfastmath.c module... done.
Parsing function data from whets.c module... Found 1 function(s).
Parsing CFG data from functions in whets.c module... done.
Parsing function data from crtstuff.c module... Found 2 function(s).
Parsing CFG data from functions in crtstuff.c module... done.
Instrumenting functions in module whetstone-double... done.
Instrumenting functions in module crtfastmath.c... done.
Instrumenting functions in module whets.c... done.
Instrumenting functions in module crtstuff.c... done.
Waiting for mutator to exit...
Dyninst mutator exited normally and returned 0.
[ Done processing whetstone-double ] ------------------------------------------
Analysis complete.
```

Running the rewritten binary fails with SIGILL:
```
$ ./whetstone-double.dyn 10
Calibrate
Illegal instruction (core dumped)
```
```
$ gdb ./whetstone-double.dyn
Reading symbols from ./whetstone-double.dyn...done.
(gdb) r 10
Starting program: whetstone-double.dyn 
Calibrate

Program received signal SIGILL, Illegal instruction.
0x00000000007025b2 in ?? ()
(gdb) bt
#0  0x00000000007025b2 in ?? ()
#1  0x00007ffff7ac9620 in _IO_2_1_stderr_ () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x000000000000000a in ?? ()
#3  0x0000000000708010 in ?? ()
#4  0x00007fffffffdeb0 in ?? ()
#5  0x0000000000000000 in ?? ()
```

Valgrind also fails to uplift the assembly to its IR, at the same offset. The sequence of offending instruction bytes are:
```
[...]

PR #107, opened 2016-06-24
Rewriter fixes, instruction decoding fixes.

jdetter:
Includes the fix for #96. This also contains several fixes for instruction decoding.

This also includes the `ia32_decode` cleanup. Error checking has been improved. Decoding functions can be called individually have return values to check for decoding errors. Arguments to decoding functions have been cleaned up.

All 64 bit tests are passing. All 32 bit tests pass on Windows. The 32 bit test failures are on both VEX and master, so this will not introduce any new issues.


jdetter:
@mxz297 could you check to see if the latest VEX compiles on 32 bit Windows?

pefoley2:
I found at least one build failure on windows: https://github.com/dyninst/dyninst/pull/108
The build is still running, so there may be more fixes incoming.

pefoley2:
Looks like the vex branch builds successfully under VS2015 with the above PR.

jdetter:
This pull request now includes #108

PR #132, opened 2016-07-27
Fixes for test_pt_ls

jdetter:
`findMain` now uses dataflowAPI to calculate the address of main in a register.


cuviper:
<meta> Is `v9.2_patches` supposed to be a stable branch for the next `v9.2.1`?  It seems weird that you would push non-trivial commits there without review, then PR it to `master`.

Anyway, I'll try this out on Fedora later today.

jdetter:
Hmm I think we've been treating `v9.2_patches` more as just a staging branch, then doing PR's to keep master up to date with the stable commits. @wrwilliams any input here?

cuviper:
That question is also important for API/ABI tracking.  If there's a branch intended for a 9.2.x point release, it needs to be very careful about compatibility.

wrwilliams:
v9.2_patches is intended to be a stable branch for 9.2.1.

jdetter:
I've found more issues with finding main on small binaries that I would like to take a swing at, I'll add to this when I have more optimizations.

PR #135, opened 2016-07-29
Line info optimizations

wrwilliams:
Add a cached DIE and interval tree to Modules to speed up line info lookup.
Breaks ABI, so must be delayed until 9.3 in current form.


wrwilliams:
Resubmitting to Jenkins

wrwilliams:
Resubmitting to Jenkins

wrwilliams:
Resubmitting to Jenkins

wrwilliams:
This is now a significant stack of fixes, optimizations, and refactorings for DWARF information. It probably doesn't build on Windows yet.

The overview of the new state of the world:

* Symtab can now find the set of modules that contain information for a given address. (There may be multiple such modules, for instance when two compilation units provide debugging information for the same location.) This information is properly used to initialize symbols and modules during openFile, and allows more precise determination of which modules need their line information parsed during a query.
* Lists of source files are now represented by a StringTable class and passed around by shared pointer. StringTable is likely to become a real class and not just a typedef around a multi-index container; interface thoughts are welcome.
* Statements are presently only returned by shared pointer, not by value. This may be something we can optimize, but initial profiling looks like statements are fairly heavyweight at present, so reference-counting semantics are likely preferable.
* Fixed a bug where inline functions were looking at an off-by-one element of the string table for their filename.
* Inlines and statements both share their containing Module's string table
* Line information now exposes iterator interfaces (not a complete set). Feedback on things that are missing/unintuitive would be great.
* Statements will allow you to get their filename's string table index, as well as querying the name straight through the string table.
* We do not yet lazily parse line information on file/line based queries, but we now can. (We have been lazily parsing line information on address based queries for some time, and that should be more robust as well.)

I have pretty well murdered ABI/API compatibility here in the name of performance, but I'm keeping things on this PR so that folks can take a look at the progress and provide feedback on the interface/performance tradeoffs.

Also fair warning: requires std::lambda/bind compatibility and a decently recent boost to build without errors (1.58 is sufficiently new). Reverting to boost for lambda/bind where it's absent should be a straightforward swap (and is something we should do at project scope).

wrwilliams:
ok to test

wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
The one failure here is a boost download failure (which is being a transient). Boost as external needs some cleanup for speed in our Jenkins setup but we're finally good here.

Issue #136, opened 2016-08-01
ABI changes from v9.2.0 to v9.2_patches


cuviper:
Just looking at public headers, there are two obvious groups of changes, both from `common/`.  In `dyn_regs.h`, many register variables have been added.  In `entryIDs.h`, a handful of new enum values are added, but since these are added in the middle, they will change all of the values that follow.
[diff-headers-dyninst-v9.2.diff.gz](https://github.com/dyninst/dyninst/files/394528/diff-headers-dyninst-v9.2.diff.gz)

Using [libabigail](http://www.sourceware.org/libabigail/)'s `abipkgdiff`, we can confirm that these are the only real changes, as well as see the function paths where these are exposed.
[abipkgdiff-dyninst-v9.2.log.gz](https://github.com/dyninst/dyninst/files/394527/abipkgdiff-dyninst-v9.2.log.gz)

The plain additions are generally fine for backward compatibility, but not forward.  That is, something which links against the new library may not be able to run with the older library, if it used new symbols.

The existing enum entries which changed their values are a bigger problem, both forward and back.

jdetter:
Are the `entryIDs` a big deal? Just instruction names will be off but that should affect their decoding besides the fact that `format()` will return the wrong name. I could move them to the end of the enum if it's something we should be worried about.

jdetter:
We could also move the x86 register declarations, those were added because a user was able to generate 32 bit code that used YMM registers. We don't necessarily need the 32 bit version of `k0` through `k7` and `ZMM0` through `ZMM31`, I just put them in there to be safe. I don't think AVX512 instructions are valid in 32 bit mode and there will never be a 32 bit intel chip that implements AVX512. The YMM register declarations should be the only ones we for sure need there.

cuviper:
It depends how they are used.  If a library user is ever producing these `entryID` values, then they certainly need to numerically agree with the library itself.  If they're only ever consumed by the user, and especially if only indirectly as something like formatted strings, then it probably doesn't matter.

On the registers, I don't really have any objection to specific additions, just that ABI needs to be considered when *anything* is added.

jdetter:
16fce865a6729b90f4a2443e0c2a6d7977010bca and b47e64d880de6a39375d56f37d8506039d24e86d would need to be reverted to fix the `entryIDs` ABI issue.

cuviper:
[...]

PR #137, opened 2016-08-01
Partial fix for ABI issues on v9.2_patches

jdetter:
This is a partial fix for #136. I'm not sure if the `entryIDs` issue can be cleanly fixed.


jdetter:
5a83500 can be cherry picked from this, I accidentally started the patch branch from master instead of v9.2_patches.

cuviper:
I don't understand what you're doing here.  It looks like you just shuffled them around?  How does this address ABI at all?

jdetter:
Idk why the diff looks so weird, but I moved the YMM*, ZMM* and K* register definitions below the older definitions, which should give them larger register numbers, which should give backward compatibility right?

cuviper:
AFAICS `DEF_REGISTER` yields independent values.  I don't think order matters.  The ordering concern was related to the other ABI issue, `enum entryID`, where their values are implicit.

jdetter:
Ah, okay, so I'll delete this PR and then do you want me to revert those two commits I mentioned? Those two commits just added support for a couple of AES instructions that can just go into 9.3.0 instead of 9.2.1.

Issue #114, opened 2016-07-01
Crash during liveness analysis


ea:
Hi 

I just updated to the latest release and spoted a crashing bug with a certain binary. 
The crash happens on :  https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/liveness.C#L473

and again on https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/liveness.C#L510

Both times the similar code
```
    else{
      base = changeIfMMX(base);
      ret.read[getIndex(base)] = true;
    }
```

For some cases , getIndex(base) will return -1 , which will cause an out of bounds access leading to segfault. 
This is the crash i get:
```
Program received signal SIGSEGV, Segmentation fault.
LivenessAnalyzer::calcRWSets (this=0xb7fbf040 <instPoint::liveRegisters()::live1>, curInsn=..., blk=0xc7e35d4, a=158710)
   at dyninst/dyninst-9.2.0/dataflowAPI/src/liveness.C:473
473           ret.read[getIndex(base)] = true;
(gdb) x/i $pc
=> 0xb744d95c <LivenessAnalyzer::calcRWSets(boost::shared_ptr<Dyninst::InstructionAPI::Instruction>, Dyninst::ParseAPI::Block*, unsigned long)+1308>:   or     %edx,(%eax,%edi,4)
(gdb) i r eax edi
eax            0x162ccc50       372034640
edi            0x7ffffff        134217727
(gdb)
```

I tried to enable liveness debug output, but the offending binary is big and the liveness debug log is getting huge. I'll update the issue once I isolate the problematic instruction/code. 

Cheers

ea:
Here's a sample liveness debug log:

```
---------------------
        summarize block info at block 8089b2d
liveness.C[145] After instruction vmovdqa <INVALID_REG>, <INVALID_REG>, <INVALID_REG> at address 0x8089b2d:
calcRWSets for vmovdqa <INVALID_REG>, <INVALID_REG>, <INVALID_REG> @ 8089b2d
Read registers:
        <INVALID_REG>
        <INVALID_REG>
Write Registers:
        <INVALID_REG>
liveness.C[157] After instruction at address 0x8089b2d:
-----------------------
```

So this all seems to be related t o AVX instructions... 
These aren't supported? Can they be handled gracefully ?

jdetter:
Could I have access to the binary you're getting this crash on?

jdetter:
[...]

PR #138, opened 2016-08-02
Fixes significant ABI issues on v9.2_patches.

jdetter:
This fixes the instruction entry enum issues which break the ABI. The YMM declarations can also be removed, which we are currently discussing. This however would break #114.

The instructions removed here will be readded during 9.3.0. These instructions are rare and should be unlikely to cause any user problems.


jdetter:
This fixes the `entryIDs` issue, which definitely should be pushed into 9.3.0 instead of 9.2.1. @cuviper ok to merge?

cuviper:
Works for me, thanks!

PR #140, opened 2016-08-02
Restored dyn_regs.h to v9.2.0 version

### LINKED ISSUE #136: ABI changes from v9.2.0 to v9.2_patches

jdetter:
Fixes #136. Reverts dyn_regs.h to the exact version of v9.2.0.


cuviper:
Travis will tell, but doesn't dataflowAPI need to stop referencing these too?

jdetter:
I probably should have checked, sorry I'm not used to getting up this early lol

jdetter:
Another patch will be added to this PR, I'm waiting for the compilation to complete.

jdetter:
My testsuite run is still going but so far it looks like it's good. We'll see what Travis says as well.

cuviper:
@jdetter, is there a weaker fix you can put in place for #114?  Just to avoid that out-of-bounds access on `ret.read[getIndex(base)]` where the index is `-1`.  If there's no reasonable action, even an assert would be better than undefined behavior, and that safety net might even be nice for master, in case of future surprises.

jdetter:
There really really should be a check of the return value of `getIndex()` if it's used as an indexer into an array. I can definitely put more checks in there and either return failure or throw an assert.

jdetter:
Asserts added in `liveness.C` where `getIndex` could cause a buffer underread when it's result is used as an indexer for an array.

jdetter:
Is there anything else you think we need to sneak in here?

cuviper:
LGTM

PR #141, opened 2016-08-03
Added asserts in liveness.C to prevent buffer underreads for

jdetter:
Cherry pick from `v9.2_patches`. This prevents buffer underreads if we have another situation where there is a register defined in `dyn_regs.h` for x86_64 that isn't defined in x86 but can somehow be used in 32bit mode.


jdetter:
@cuviper this is just added safety asserts, is this good to merge?

cuviper:
Yeah, looks good.

PR #142, opened 2016-08-04
findMain improvements

jdetter:
Less false positives on `findMain` dataflowAPI analysis. `findMain` now returns an int (0 on success, -1 on failure) instead of void. Some asserts have been taken out and replaced with return -1 (mutatee issues, not internal issues).

### DIFF
--- a/dyninstAPI/src/image.C
+++ b/dyninstAPI/src/image.C
@@ -1513,7 +1513,14 @@ image::image(fileDescriptor &desc,
 
    //Now add Main and Dynamic Symbols if they are not present
    startup_printf("%s[%d]:  before findMain\n", FILE__, __LINE__);
-   findMain();
+   if(findMain())
+   {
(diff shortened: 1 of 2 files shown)


jdetter:
@wrwilliams can this be merged? This doesn't fix the issue on RHEL 6 but it is an improvement on the patch that fixed the issue on Fedora 23 and Ubuntu.

wrwilliams:
Can one of the admins verify this patch?

wrwilliams:
Resubmitting to Jenkins.

jdetter:
@wrwilliams `dyninst-build` is passing but the PR status isn't getting set. It ends with:

```
Finished: SUCCESS
```

jdetter:
Jenkins in complaining because `test_pt_ls` isn't passing, but we already know `test_pt_ls` has issues on RHEL 6

Issue #123, opened 2016-07-14
test_pt_ls fails with Dyninst master


ssunny7:
Both `create` and `rewriter` `test_pt_ls`s fail with Dyninst's master branch. Branch used for testsuite was also master.

Output from running the tests using the `-log`  and `-verbose` options:
```
Commencing test(s) ...
Thu Jul 14 16:21:33 CDT 2016
Linux bigking.cs.wisc.edu 2.6.32-573.7.1.el6.x86_64 #1 SMP Thu Sep 10 13:42:16 EDT 2015 x86_64 x86_64 x86_64 GNU/Linux
TESTDIR=/afs/cs.wisc.edu/p/paradyn/development/ssunny/dyninst/testsuite-install/bin/testsuite
[Tests with none]

Enabling DyninstAPI parsing debug
ParseThat.C[79]:  resolved parseThat to /u/s/s/ssunny/dev-home/dyninst/dyninst-install/bin/parseThat
ParseThat.C[79]:  resolved parseThat to /u/s/s/ssunny/dev-home/dyninst/dyninst-install/bin/parseThat
ParseThat.C[304]:  parseThat: /u/s/s/ssunny/dev-home/dyninst/dyninst-install/bin/parseThat
ParseThat.C[343]:  about to issue command: 
		'/u/s/s/ssunny/dev-home/dyninst/dyninst-install/bin/parseThat -i 1 -p 1 -v 7 --summary -t 300 -T 0 -o ./binaries/test_pt_ls_output1 /bin/ls / 1>./binaries/test_pt_ls_stdout1 2>./binaries/test_pt_ls_stderr1'
ParseThat.C[352]:  parseThat cmd failed with code 1
```

Following is the relevant portion of the output from the debug parsing log for this test:
```
ParseThat: /u/s/s/ssunny/dev-home/dyninst/dyninst-code/dyninstAPI/src/codegen-x86.C:1207: static bool insnCodeGen::modifyData(Dyninst::Address, NS_x86::instruction&, codeGen&): Assertion `!"Couldn't decode opcode of already known instruction!\n" failed.
```

Let me know if you need any other information.

jdetter:
This was fixed by a patch in v9.2.0_patches. Could you try merging v9.2.0_patches into your branch and rerunning the testsuite?

ssunny7:
[...]

PR #143, opened 2016-08-09
Fixes for test_pt_ls

jdetter:
These changes made it into `v9.2_patches` but didn't make it into master yet.

### DIFF
--- a/dataflowAPI/src/stackanalysis.C
+++ b/dataflowAPI/src/stackanalysis.C
@@ -1077,14 +1077,20 @@ void StackAnalysis::handlePushPop(Instruction::Ptr insn, Block *block,
             long writtenSlotHeight = spHeight.height() - word_size;
             Absloc writtenLoc(writtenSlotHeight, 0, NULL);
 
-            // Get copied register
             Expression::Ptr readExpr = insn->getOperand(0).getValue();
-            assert(typeid(*readExpr) == typeid(RegisterAST));
-            MachRegister readReg = boost::dynamic_pointer_cast<RegisterAST>(
-               readExpr)->getID();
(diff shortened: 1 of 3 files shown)


jdetter:
@wrwilliams is `dyninst-pr-build` still broken or is that an actual issue?

wrwilliams:
Tests are still running, and I don't think the JUnit outputter is working on this build, but we can manually verify in the worst case.

jdetter:
@wrwilliams Can this be merged? I'm not sure why dyninst-build hasn't finished in 2 days. This fixes #123

wrwilliams:
http://follis.cs.wisc.edu:8080/job/dyninst-test/11/consoleFull (internal access only) has the test results that are relevant; do you expect the failures that are there? They're showing up on a lot of Jenkins builds, so it may be something with the configuration/environment...

wrwilliams:
In any event, if it's on 9.2 patches, it should go to master, so merging.

jdetter:
`test_pt_ls` should definitely be fixed on RHEL 6 and Fedora 23, what distro is follis running?

wrwilliams:
RHEL6.

wrwilliams:
Found the problem, btw: the Jenkins dyninst build step was still using jenkins' own archiving rather than tar, and that strips permissions. So parseThat was -x in the CI system.

I'll look at the rest of these Monday, though if anyone else is bored I won't argue.

jdetter:
Thanks @wrwilliams, it looks like there is another issue with `test_pt_ls` on RHEL 6. Somehow the mutatee is generating a SIGSEGV, however parseThat is returning success...

Issue #34, opened 2016-05-10
rewriter tests fail/crash on ppc64


wrwilliams:
Something is systemically wrong with rewriting on this platform.

wrwilliams:
The tramp guard locking functions we're executing are junk; the ones I get from disassembling the RTlib look good.

jdetter:
@wrwilliams was this fixed for ppc64? I think we fixed a similar issue on x86_64.

wrwilliams:
I haven't checked recently, but that's on today's list.

wrwilliams:
There's TOC glue code that doesn't line up with what we expect for either PPC ABI (v2 we'd expect that the local and global entries are contiguous and they're not, v1 we'd expect that there's not two separate entry points and there are). I think this is a won't fix until we work out the full implications of PPC ABI v2; removing from milestone but leaving open.

mcfadden8:
Hi @wrwilliams, 

for ABIv2, what is an example of when local and global entries are not contiguous?

wrwilliams:
...I'm not 100% positive what I'm seeing, but here are the relevant snippets from our runtime library, which exports DYNINST_unlock_tramp_guard:

```0000000000003608 <00000010.plt_call.DYNINST_unlock_tramp_guard+0>:
    3608:       f8 41 00 28     std     r2,40(r1)
    360c:       e9 62 88 90     ld      r11,-30576(r2)
    3610:       7d 69 03 a6     mtctr   r11
    3614:       e8 42 88 98     ld      r2,-30568(r2)
    3618:       28 22 00 00     cmpldi  r2,0
    361c:       4c e2 04 20     .long 0x4ce20420
    3620:       48 00 35 80     b       6ba0 <DYNINST_unlock_tramp_guard@plt>```

which is obviously a thing doing TOC setup of some sort, and:

```0000000000003b70 <.DYNINST_unlock_tramp_guard>:
    3b70:       7c 08 02 a6     mflr    r0
    3b74:       f8 01 00 10     std     r0,16(r1)
    3b78:       f8 21 ff 91     stdu    r1,-112(r1)
    3b7c:       38 62 81 58     addi    r3,r2,-32424
    3b80:       4b ff fb bd     bl      373c <00000010.plt_call.__tls_get_addr@@GLIBC_2.3+0>
    3b84:       e8 41 00 28     ld      r2,40(r1)
    3b88:       38 21 00 70     addi    r1,r1,112
[...]

PR #156, opened 2016-08-18
symtabAPI: use the known type for new relocations

cuviper:
In emitElf::createRelocationSections(), new relocations were calling
getGlobalRelType to determine their type.  However, this didn't take the
symbol type into account, so functions on PPC64 were getting GLOB_DAT
instead of JMP_SLOT, and they didn't work at all when called.
    
Each new relocationEntry already knows its type, so use getRelType().


cuviper:
This appears to fix the tramp guard issues of #34 for me.

wrwilliams:
Resubmitted to Jenkins

PR #162, opened 2016-08-26
Replaced nullptr with NULL

jdetter:
@cuviper pointed out that master won't compile on RHEL 6 because nullptr isn't supported. This PR replaces the nullptrs with NULLs to restore compatibility.


jdetter:
This is now included by #164

PR #167, opened 2016-09-02
Require users to enable C++11 too, and use it publicly

cuviper:
This is a proof-of-concept following my [RFC](https://www-auth.cs.wisc.edu/lists/dyninst-api/2016/msg00327.shtml) to fully require C++11, so we may use those features in public headers.

The first commit adds an `#error` message if we can't detect C++11.  The second commit converts `dyn_hash_set` and `dyn_hash_map` to always use `std::unordered_set` and `std::unordered_map` respectively.  The remaining commits get rid of a few `boost::assign` chains and large tables, in favor of initializer lists.

For a sample of motivation, take `arch-x86.C` in particular.  That one file used to take 108 seconds to compile for me, and now it takes just over 10.  In fact, the _entire_ build of `common` only takes 30 seconds now, even serially.

These numbers are from Fedora 24's gcc 6.1.1.  I also confirmed that this still compiles on RHEL6's gcc 4.4.7 with its partial C++0x support.


cuviper:
Huh, Windows failed the basic C++11 check.  AppVeyor has VS 2015 -- I thought MSVC just implicitly enabled C++11?  Does it not set `__cplusplus` properly?  Or did I get the check wrong?

```
#if !((__cplusplus >= 201103L) || defined(__GXX_EXPERIMENTAL_CXX0X__))
#error "Dyninst requires C++11 features -- try adding the -std=c++0x or -std=c++11 flag."
#endif
```

cuviper:
Multiple search results confirm MSVC has not updated `__cplusplus`.  I suppose it's being ultra conservative in claiming *full* coverage of the standard.

I can check `_MSC_VER` instead -- is there a minimum you all would like to require?  What version is already required to build Dyninst itself?  Then I can check MSDN to be sure that supports the features I want to use.

cuviper:
It looks like initializer lists in particular will require VS 2013, i.e. `_MSC_VER >= 1800`.
https://msdn.microsoft.com/en-us/library/hh567368.aspx

pefoley2:
Is anything happening with this?

cuviper:
Dunno -- you're the first to break the deafening silence...

wrwilliams:
I think definitionally we have to hold this for 10.0 if we're going to make users build with C++11; I can see it being a grey area but this meets the spirit of an API-breaker to me.

cuviper:
I'm going to just close this, as it's out of date and I don't have time to get back to it.  I'll leave my branch around, so feel free to take any of the commits later if you like!

Issue #53, opened 2016-05-19
Segfault when parsing binary with no functions


jdetter:
This is a low priority issue, however parsing `crtend.o` on Fedora23 causes Dyninst to segfault. It is caused by the Block deconstructor:

```
Block::~Block()
{
    // nothing special
    if (_obj && _obj->cs()) {
        _obj->cs()->decrementCounter(PARSE_BLOCK_COUNT);
        _obj->cs()->addCounter(PARSE_BLOCK_SIZE, -1*size());
    }
}
```

It would appear `_obj` contains junk here.

**Valgrind output**

```
[detter@localhost dyninst]$ valgrind dyn_instruct_test /usr/lib/gcc/x86_64-redhat-linux/5.3.1/crtend.o
==11250== Memcheck, a memory error detector
==11250== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==11250== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==11250== Command: dyn_instruct_test /usr/lib/gcc/x86_64-redhat-linux/5.3.1/crtend.o
==11250==
==11250== Invalid read of size 8
==11250==    at 0x5BAF793: Dyninst::ParseAPI::Block::~Block() (Block.C:63)
==11250==    by 0x5BAF84F: Dyninst::ParseAPI::Block::~Block() (Block.C:66)
==11250==    by 0x5B9B007: Dyninst::ParseAPI::CFGFactory::free_block(Dyninst::ParseAPI::Block*) (CFGFactory.C:166)
==11250==    by 0x5B9AFD5: Dyninst::ParseAPI::CFGFactory::destroy_block(Dyninst::ParseAPI::Block*) (CFGFactory.C:161)
==11250==    by 0x5B9B189: Dyninst::ParseAPI::CFGFactory::destroy_all() (CFGFactory.C:193)
==11250==    by 0x5B9AAE9: Dyninst::ParseAPI::CFGFactory::~CFGFactory() (CFGFactory.C:83)
==11250==    by 0x5B9AB49: Dyninst::ParseAPI::CFGFactory::~CFGFactory() (CFGFactory.C:84)
==11250==    by 0x5BB2876: Dyninst::ParseAPI::CodeObject::~CodeObject() (CodeObject.C:103)
==11250==    by 0x40257D: analyze_bin(char*) (main.C:95)
==11250==    by 0x40238B: main (main.C:56)
[...]

PR #173, opened 2016-09-06
symtabAPI: dont expect a data segment in elf; eliminate dead code

rafzi:
Simple bare metal applications may not have a data segment.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -1538,12 +1538,6 @@ void Object::load_object(bool alloc_syms)
                 //bpfatal( "no text segment\n");
                 goto cleanup;
             }
-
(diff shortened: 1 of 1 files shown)


wrwilliams:
Does this fix #53?

rafzi:
That issue looks unrelated to me.

Without the fix in this PR dyninst would fail to load a binary without a data section cleanly without a crash.

As for safety of this change, the ``data_ptr_`` memeber is only dereferenced in ``Object::elf_vaddr_to_ptr`` and this one is checked. Outside of ``AObject`` I only found a use in ``Symtab``, but it's not actually used or exposed there.

wrwilliams:
Despite not being a fix for #53, this should go to v9.2_patches as well.

Issue #175, opened 2016-09-08
PPC64 generateBranchViaTrap: Assertion `isCall == false' failed.


cuviper:
This test tries to attach-mutate-detach twice on the same process: [dynamic-double.zip](https://github.com/dyninst/dyninst/files/462890/dynamic-double.zip)

With Dyninst 9.2 on PPC64 RHEL7.2, this crashes with an assertion the second time around.  It's fine on x86_64, and also fine on PPC64 with 9.1.

```
# ./mutatee & ./mutator $! ; wait
[1] 4194
TAKE1 A
TAKE1 B
TAKE1 C
TAKE1 D
TAKE1 E
TAKE1 F
TAKE1 G
TAKE1 H
TAKE1 I
FUNCTION EXECUTED. VALUE = 1
FIRST MUTATION DONE. MUTATOR IS DOING THE SECOND ONE...
FUNCTION EXECUTED. VALUE = 2
TAKE2 A
TAKE2 B
TAKE2 C
TAKE2 D
TAKE2 E
TAKE2 F
TAKE2 G
mutator: /root/dyninst/dyninstAPI/src/codegen-power.C:259: static void insnCodeGen::generateBranchViaTrap(codeGen&, Dyninst::Address, Dyninst::Address, bool): Assertion `isCall == false' failed.
Aborted
FUNCTION EXECUTED. VALUE = 3
FUNCTION EXECUTED. VALUE = 4
FUNCTION EXECUTED. VALUE = 5
FUNCTION EXECUTED. VALUE = 6
FUNCTION EXECUTED. VALUE = 7
FUNCTION EXECUTED. VALUE = 8
FUNCTION EXECUTED. VALUE = 9
FUNCTION EXECUTED. VALUE = 10
[1]+  Done                    ./mutatee
```

wrwilliams:
Pretty sure the branch format is cosmetic; @jdetter can you confirm?

I suspect that this is a side effect of the RT heap fixes in 9.2, and that we're now needing long branches where we didn't before. The good solution is to have modifyCall (and also modifyJump while we're in here) generate something like the following:

```
spill ctr if live
allocate scratch gpr
generate target in scratch
mtctr scratch
<branch/conditional branch/call> new target
restore spilled regs
```

Because seriously, to hell with traps, we're in relocated code, we have space. And generateBranchViaTrap is an acceptable thing for the springboard code to call--where it's going to generate a branch-by-god.

[...]

PR #176, opened 2016-09-09
RT: trymmap should retry if the result is out of range

cuviper:
An address passed to `mmap` is just taken as a hint, and the OS may
return something wildly different if that address is not available.
This is undesirable when we're trying to create a constrained alloc.

Now we will check that the address is in the requested range before
accepting it.  Otherwise, unmap it and try a new hint.

### DIFF
--- a/dyninstAPI_RT/src/RTheap.c
+++ b/dyninstAPI_RT/src/RTheap.c
@@ -128,8 +128,13 @@ static Address trymmap(size_t len, Address beg, Address end, size_t inc, int fd)
   /* until we get one that succeeds.*/
   for (addr = beg; addr + len <= end; addr += inc) {
     result = map_region((void *) addr, len, fd);
-    if (result)
+    if (result) {
+      /* Success doesn't necessarily mean it actually mapped at the hinted
+       * address.  Return if it's in range, else unmap and try again. */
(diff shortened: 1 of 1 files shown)


cuviper:
This fixes the root cause that led me to #175.

wrwilliams:
Looks clean, but may want to delay merging per comment on #175 for easy reproducibility of a case where a far call needs to be generated.

wrwilliams:
ok to test

wrwilliams:
retest this please

wrwilliams:
@cuviper Mind rebasing this one so we can retest?

cuviper:
Rebased.

wrwilliams:
Looks like this is missing the boost auto-build commits based on the Jenkins failure.

cuviper:
This PR is on 9.2_patches -- did the boost stuff land there too?

I can duplicate the PR on master, if you want to test it that way first.

wrwilliams:
Oh, I'm blind. Please duplicate on master; we're going straight to 9.3 given the massive delay on 9.2.1.

wrwilliams:
Duplicates #231 and we're going to 9.3 directly; closing.

PR #177, opened 2016-09-12
Fix operand types for vpand

BlairArchibald:
The intel developer manual specifies vpand as having operands "xmm1 xmm2
xmm3/m128". This change reflects this in dyninst.


jdetter:
@wrwilliams This is good to go, this is a small correction and it shouldn't cause any issues.

wrwilliams:
Are there other similar cases in this section of the tables, or was this a lone wolf?

BlairArchibald:
I had a quick look and didn't see anything immediately obvious, although I'm not an x86 expert by any means.

wrwilliams:
Approved for 9.2.1. We should write up a test that generates a bunch of random bytes and verifies no assertions and all operands are used.

jdetter:
@wrwilliams Sorry I forgot to get back to you on this. I remember I had to add this instruction by hand and it was just a simple typo. There shouldn't be any other issues related to this one.

PR #182, opened 2016-09-25
Determine the architecture of an ELF by looking at the file header in…

rafzi:
…stead of the compiled architecture

While the main point of your project is to do live instrumentation, it still provides very useful libraries for offline static analysis.

This change allows the user to load ELFs across architectures. For example you can now load and analyze an AArch64 ELF on an x86 machine.

### DIFF
--- a/symtabAPI/src/Object-nt.C
+++ b/symtabAPI/src/Object-nt.C
@@ -2353,7 +2353,15 @@ void Object::insertPrereqLibrary(std::string lib)
 
 Dyninst::Architecture Object::getArch() const
 {
-   return Dyninst::Arch_x86;
+    switch (peHdr->FileHeader.Machine)
+    {
+    case IMAGE_FILE_MACHINE_I386:
+        return Dyninst::Arch_x86;
+    case IMAGE_FILE_MACHINE_AMD64:
+        return Dyninst::Arch_x86_64;
+    default:
+        return Dyninst::Arch_none;
+    }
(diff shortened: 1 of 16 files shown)


symtabAPI/src/Object-nt.C
```diff
@@ -2336,7 +2336,7 @@ void Object::insertPrereqLibrary(std::string lib)
            getRegionPermissions() == RP_RWX);
 }
 
-Dyninst::Architecture Object::getArch()
```
wrwilliams: There should be a way to tell whether this is properly x86 or x86_64 from the file headers, no? May as well get that bit of infrastructure in.

rafzi:
Yes sure, you can add them here if you have some additions. I've been using it in this state for some time now and did not notice any issues.

I added the implementation of getArch for PE.

wrwilliams:
Thanks; looks good.

I'll rebase onto this and get my architecture changes added, and then we can merge it over.

rafzi:
Sorry I didn't get to this earlier.

The additional changes do not compile. You use Elf_X_Data::xlatetom which neither exists in this branch or master.

wrwilliams:
ok to test (triggering Jenkins)

wrwilliams:
@rafzi can you rebase this one when you get a chance?

rafzi:
Alright it's done. The CI failure seems to just be a download error.

PR #185, opened 2016-09-30
String table now includes an entry at zero for "unknown", so don't subtract from the DWARF file number.

wrwilliams:
String table now includes an entry at zero for "unknown", so don't subtract from the DWARF file number.


wrwilliams:
ok to test

PR #186, opened 2016-10-02
Exception frame rewriting fixes

### LINKED ISSUE #121: Rewrite exception handlers to adjust for relocated code
From Bill:

C++ exception handlers determine the appropriate catch blocks to handle a piece of code based on the PC where the exception was thrown. When an exception occurs in relocated code, that PC is not covered by any exception handler unless that exception handler had a range of "whole address space". This results in incorrect behavior and unexpected termination of instrumented C++ that relies on successful catch block execution.

We need to rewrite the .eh_frame section and augment or overwrite it with entries mapping relocated code ranges to the appropriate catch code. Alternately, we could hijack the input PC to the throw routine and adjust it back to an original PC. The first approach probably produces faster execution, but perturbs the original binary more. It does allow us to add exception handling for any exceptions thrown by instrumentation, if and when that proves desirable. The second approach is much simpler to implement in the dynamic case, but requires that the mutator's relocated->original translation map be encoded into the mutatee in the rewriter case; it is likely to be modestly more expensive during execution as well. It does not allow us to add new exception handling entries, and will implicitly force exceptions occuring during instrumentation to the same handlers that cover the address of the instpoint where that instrumentation was inserted.

jdetter:
Fixes #121.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ V_*
 *~
 *.he
 *_none
+.nfs*
 BuildLog.htm
 *resumelog*
 ._*resumelog
--- a/cmake/packages.cmake
+++ b/cmake/packages.cmake
@@ -69,6 +69,7 @@ if (UNIX)
 	INSTALL_COMMAND install <BINARY_DIR>/libiberty.a <INSTALL_DIR>
 	)
       set(IBERTY_LIBRARIES ${CMAKE_BINARY_DIR}/libiberty/libiberty.a)
+      set(IBERTY_FOUND TRUE)
     endif()
 
     message(STATUS "Using libiberty ${IBERTY_LIBRARIES}")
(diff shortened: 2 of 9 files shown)


jdetter:
I'm getting some failures locally -- I should have a fix soon.

wrwilliams:
ok to test

jdetter:
resubmitted to jenkins

jdetter:
Testing locally on Linux Mint x86_64 I am getting all tests passed except the `test_pt_ls` series which aren't passing on master. There is a problem remaining where after an exception is caught, the function that handled the exception doesn't return properly and causes a segmentation fault. I will see if I can find the specific cause but I might need some debugging help from @wrwilliams at some point.

jdetter:
The new issue is caused by a crash in `_Unwind_Resume`, which is called after the exception has been handled. @wrwilliams do you want to help me with this today or tomorrow?

wrwilliams:
Yeah, find me when you get a chance and I can take a look.

wrwilliams:
@jdetter Is the fencepost error fix we talked about this week coming to this PR anytime soon?

jdetter:
Well see what the CI says but on my end everything looks like it's ready to go

jdetter:
Ok this is ready to go into v9.3.0 whenever @wrwilliams

wrwilliams:
Awesome, thanks.

PR #189, opened 2016-10-04
Improved the findMain analysis significantly

jdetter:
The `findMain` analysis should be more accurate and less susceptible to asserts and failures. I don't have the 32 bit libraries installed on my workstation but it looked like the 64 bit tests were passing when the testsuite completed.

@wrwilliams ok to test?

### DIFF
--- a/dyninstAPI/src/image.C
+++ b/dyninstAPI/src/image.C
@@ -729,61 +729,6 @@ int image::findMain()
             InstructionAPI::Instruction::Ptr insn_ptr = decoder->decode(
                     (const unsigned char*)insn_raw);
 
-            /* Use dataflow analysis here to determine the value of EDI */
-            // const unsigned char* raw = p;
-            // instruction insn;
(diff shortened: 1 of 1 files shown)


jdetter:
Resubmitted to jenkins

jdetter:
ok to test

jdetter:
@wrwilliams Can you force this to run on jenkins? I'm not sure why the build isn't triggering.

wrwilliams:
add to whitelist

(The commands work for the GHPRB admin list on the job, not the Jenkins admin list. In-house pull requests *should* be whitelisted but clearly something's not quite right there.)

wrwilliams:
Jenkins results look good but they're obviously not updating the PR status right yet. But I think we're clear to merge.

jdetter:
Ok sweet

jdetter:
This will have the same results as before, only commented out code blocks were removed. I'm fine with this being merged whenever.

PR #191, opened 2016-10-12
proccontrol: check thread handle before calculating TLS

cuviper:


cuviper:
This solves a simple segfault as reported in [rhbz1366730](https://bugzilla.redhat.com/show_bug.cgi?id=1366730).  We just need to check if there's a thread handle before using it in `plat_calcTLSAddress`.

In proccontrol logs, I now see:

```
[int_thread_db.C:647-U] - Initializing thread_db library
[int_thread_db.C:663-U] - Sucessfully initialized thread_db
[int_thread_db.C:677-U] - Creating threadAgent
[int_thread_db.C:59-U] - Looking up symbol nptl_version in libpthread.so.0
[int_thread_db.C:1006-U] - Didn't yet find loaded thread library
[int_thread_db.C:684-U] - Debuggee isn't multithreaded at this point, libthread_db not enabled
[int_thread_db.C:1715-U] - Error: Attempt to read user thread info of 15208/15208 before user thread create
[int_thread_db.C:1639-U] - Error: Thread_db not supported on thread 15208/15208
```

It's unfortunate if there's really no way to read single-threaded "TLS", but as I commented on rhbz, it seems GDB also can't figure this out.

wrwilliams:
This may need a rebase; I'm seeing widespread test crashes: http://jenkins.dyninst.org/jenkins/job/dyninst-test/25/testReport/

(It's also possible something's borked in the Jenkins setup; it's certain that something is borked in its reporting.)

cuviper:
FWIW, every build since # 19 has over 300 failures, so mine appears to keep the borked status quo.

PR #194, opened 2016-10-13
pefoley2-boost_win

wrwilliams:
Refactored boost-as-external-project code to be cross-platform.
ok to test


wrwilliams:
Obsoletes #170.

PR #206, opened 2016-10-17
Fix icc warning flags

pefoley2:


wrwilliams:
May I suggest "if gnulike, base gnuish flags; if gnu, gnu-specific flags" as a more clear form here?

pefoley2:
Good point, changed.

pefoley2:
Ping?

PR #209, opened 2016-10-19
Added symbol linkage support for GNU unique linkage types

bwelton:
Added support for the GNU Unique link type to SymtabAPI.


wrwilliams:
Needs #ifdef guards, but is otherwise fine.

wrwilliams:
ok to test

Issue #208, opened 2016-10-19
Race conditions with transient threads


cuviper:
I've been staring at [rhbz1373197](https://bugzilla.redhat.com/show_bug.cgi?id=1373197) too long -- time to involve you folks here.  (I probably should have just filed this a while ago, sorry.)

The test mutator simply detaches and reattaches the target over and over.  It has a watchdog thread to make sure we're making progress.  Sometimes this watchdog triggers, indicating something stalled, and sometime the reattach just fails outright.  I've added code to write a `debug-proccontrol` file too.

The test mutatee simply spawns a thread, and that thread waits 100ms before spawning a new copy of itself.  So threads are coming and going fairly frequently, and it makes it pretty reproducible for proccontrol to race on these events while reattaching.

[proccontrol-thread-race.zip](https://github.com/dyninst/dyninst/files/539952/proccontrol-thread-race.zip)

Run the `thready` mutatee in the background or in a different terminal, then find its PID and run `dyninst-reattach-loop PID`.  The time to getting stuck/failure varies a lot, but it inevitably does hit one of the races eventually.

Broadly speaking, the "stuck" case seems to consistently be that it sends a manual SIGSTOP, and never sees an event after that.  I never see this manual SIGSTOP otherwise, until we get to this bad case.  The direct "failed" case seems to be that it's trying to read /proc/PID/stat for a thread that already exited (known from the previous attachment?), and it bails entirely.  I'm not sure how to fix either case.

cuviper:
I've tried changing the mutator to use a full `detach` and `attachProcess` loop (vs. `temporaryDetach` and `reattach`), but it still sometimes hits the failed case trying to read a dead stat.  But I also get an occasional assertion failure, which I've never seen in the `reattach` case.

```
dyninst-reattach-loop: /home/jistone/paradyn/dyninst/proccontrol/src/process.C:4418: bool int_thread::StateTracker::setState(int_thread::State): Assertion `generator_state == stopped || generator_state == exited || generator_state == detached' failed.
Aborted (core dumped)
```

[debug-proccontrol-asserted.zip](https://github.com/dyninst/dyninst/files/542922/debug-proccontrol-asserted.zip)

mplegendre:
Josh,

[...]

PR #211, opened 2016-10-25
Fix errors when thread disappears during attach

mplegendre:
Fixes one of the issues reported in #208 

If a thread exited at the wrong point in the middle of attach PC would throw errors where it should have handled it gracefully.

### DIFF
--- a/proccontrol/src/linux.C
+++ b/proccontrol/src/linux.C
@@ -1009,17 +1009,28 @@ bool linux_process::plat_getOSRunningStates(std::map<Dyninst::LWP, bool> &runnin
         snprintf(proc_stat_name, 128, "/proc/%d/stat", *i);
         FILE *sfile = fopen(proc_stat_name, "r");
 
-        if (sfile == NULL) {
+        if (*i == getPid() && sfile == NULL) {
             pthrd_printf("Failed to open /proc/%d/stat file\n", *i);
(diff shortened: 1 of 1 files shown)


mplegendre:
I respectfully disagree.  Stylistically I prefer to have the functional code be straight-line, and the error-handling code be indented away under conditions.  The 'if success do work' style leads to long indention chains in the extreme cases.

cuviper:
Regardless of style choice... I tried the patch out and confirmed that it does fix the outright failure to attach in my #208 testcase.  Thanks!

cuviper:
In defense of the "if success do work" style, it can make cleanup easier as you leave each nested level, rather than having to repeat cleanup in each early break/continue/return or use goto-fail cleanup.  Alternatively, RAII cleanup deals with either style well.

wrwilliams:
Quick comments:

RAII is an ideal that I would like Dyninst to move towards when we touch the codebase. Improves encapsulation, improves error handling, just generally a good thing. I won't force it but consider this a nudge that code should trend in that direction.

WRT this patch in particular: if a small patch requires active thought from me before I understand it, it needs revision either in the form of documentation or in the form of improved clarity. This took a couple minutes for me. Your call whether to comment or revise.

mplegendre:
Feel free to fix this is a more stylistically acceptable way.

PR #212, opened 2016-10-29
Fix errors when thread disappears during attach

cuviper:
Fixes one of the issues reported in #208, following up on #211.

If a thread exited at the wrong point in the middle of attach PC would throw errors where it should have handled it gracefully.

The actual parsing has been rewritten to hopefully be more robust, in a more modern style.


wrwilliams:
Thanks, Josh; the comments really help.

PR #214, opened 2016-11-05
proccontrol: Synchronize additional threads found during attach

cuviper:
When additional threads are found during the attach process, we should
synchronize to their stopping point, and check for new threads again,
until no new threads are found.  This keeps a more consistent state if
threads are racing to start while we're attaching.

### DIFF
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -257,41 +257,16 @@ bool int_process::attachThreads()
    return attachThreads(found_new_threads);
 }
 
-// Attach any new threads and synchronize, until there are no new threads
-bool int_process::attachThreadsSync()
+bool int_process::plat_attachThreadsSync()
 {
-   while (true) {
-      bool found_new_threads = false;
-
-      ProcPool()->condvar()->lock();
-      bool result = attachThreads(found_new_threads);
-      if (found_new_threads)
-         ProcPool()->condvar()->broadcast();
-      ProcPool()->condvar()->unlock();
-
-      if (!result) {
-         pthrd_printf("Failed to attach to threads in %d\n", pid);
-         setLastError(err_internal, "Could not get threads during attach\n");
-         return false;
-      }
-
-      if (!found_new_threads)
-         return true;
-
-      pthrd_printf("Wait again for attach from process %d\n", pid);
-      bool proc_exited = false;
-      result = waitAndHandleForProc(true, this, proc_exited);
-      if (!result) {
-         perr_printf("Internal error calling waitAndHandleForProc on %d\n", getPid());
-         setLastError(err_internal, "Error while calling waitAndHandleForProc for attached threads\n");
-         return false;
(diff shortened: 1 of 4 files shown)


cuviper:
This has so far appeared to solve the remaining problems of #208.  I'll do more testing next week.

cc @mplegendre You mentioned this point of extra `attachThreads` was originally for BG/Q.  I have no idea if my modification to sync up here will still be safe for that environment - please review.  This was basically your suggested fix though, but I couldn't reuse `waitfor_startup()` as the main process's `getState()` already got marked as running.

proccontrol/src/process.C
```diff
@@ -242,13 +244,56 @@ bool int_process::attachThreads()
...
+         setLastError(err_internal, "Could not get threads during attach\n");
+         return false;
+      }
+
+      if (!found_new_threads)
+         return true;
+
+      pthrd_printf("Wait again for attach from process %d\n", pid);
+      bool proc_exited = false;
+      result = waitAndHandleForProc(true, this, proc_exited);
```
mplegendre: Thanks for working on this, Josh. A few points: 

- This is waiting for one event to happen on any thread in the process.  It's not specifically waiting for an attach event and it's not waiting for an event from every thread.  I think we could proceed out of this loop with things still in the half-attached state.  I suspect this is more narrowing the possibility of the race rather than fixing it.

- This is unconditionally waiting for events.  I'm not sure whether every platform and situation's attachThreads() generates an event that can be waited for (maybe they all do, I'm honestly not sure).  The prior code only waited for events if the int_process's getState() wasn't 'running'.

- Finally, this is going to cause performance problems on BG/Q.  We were previously using group operations to structure attach like "trigger attach to X number of process"  and then "wait for attach on X number of processes".  That worked well with BGQ's high-latency asynchronous debug interface.  This is moving to a "attach to process 1, wait for process 1, attach to process 2, ...".

For the first item, I think you could block with loop waiting for all threads reaching neonatal state.  The second two items could be fixed by moving this code to be linux specific.

proccontrol/src/process.C
```diff
@@ -242,13 +244,56 @@ bool int_process::attachThreads()
...
+      if (!result) {
+         perr_printf("Internal error calling waitAndHandleForProc on %d\n", getPid());
+         setLastError(err_internal, "Error while calling waitAndHandleForProc for attached threads\n");
```
mplegendre: Drop the setLastError here. The lower-level code should be responsible for doing a setLastError that records a more specific message when it first detects the error.

cuviper: OK, I can do that.  Though FWIW, I copied this pattern from `stopThread`.

cuviper:
OK, it's now moved to Linux-specific code, and explicitly waits until there are no neonatal threads left.

wrwilliams:
retest this please (bad apache log config killed disk space)

PR #215, opened 2016-11-06
Consistantly style the codebase with clang-format

pefoley2:
Generated with clang-format -style='{BasedOnStyle: Google, SortIncludes: false}' -i \
`find -name '*.h' -o -name '*.C' -o -name '*.c'`


jdetter:
Is this specifically a formatting consistency issue or is this a fix for clang build issues?

pefoley2:
No build issues, should be a no-op except for people reading the source code.
I just noticed a lot of inconsistent and messy formatting and figured this would be the best way to clean things up.

rafzi:
Great that this is addressed! In some places the code is really hard to read because of wild indention.

But please consider that this will create massive amounts of conflicts in any unmerged work.

One option would be to selectively apply this on sub-projects that nobody is working on, or find a point in time where as much as possible is merged in.

pefoley2:
It should be possible to automatically resolve conflicts by running clang-format on the pending code before it is merged, but yes, this will unfortunately cause a lot of churn.

pefoley2:
@wrwilliams thoughts on this change?

jdetter:
@pefoley2 During the weekly meeting at the UW a couple of weeks ago we discussed this with Bart and he said that he didn't want such a large refactoring of the repository because it would cause a lot of noise in the git history (a large part of the code base would now be 'owned' by you). If you want to get this merged in you will have to talk to Bart about it because we can't overrule him on this.

I do agree that the different indentation amounts makes the code particularly hard to understand. All of the code is 'supposed' to be formatted as 3 or 4 space indentations. All of the other aspects of coding style are up to developer as far as I know.

pefoley2:
Fair enough, it will create a lot of churn unfortunately.

PR #228, opened 2016-11-11
Merge ARM instruction semantics and jump table parsing

mxz297:


wrwilliams:
@ssunny7 Are there commits you have outstanding re: ARM semantics that need to be merged with this PR rather than separately? (Are we failing gracefully when we fail on this branch as it stands, in other words?)

@mxz297 This was clean locally but not on Jenkins, right? I'm suspecting that this is actually a matter of Josh's changes at this point, and they should be clean with his latest. But if you can verify that, I can keep on the other miscellaneous bugs.

ssunny7:
@wrwilliams I do not have any more commits for semantics at the moment. I think the only issue I was seeing on my machine is test4_4 taking too long, but that's happening both before and after merge.

mxz297:
@wrwilliams My local testing results look fine. However, Sunny says that he still has the issue of test4_4 taking too long to finish. For this branch, we will not assert, but will still output unnecessary messages to stderr using fprintf on ARM. It should be easy to delete these fprintf. Do you mean I pull the latest master and update this pull request?

wrwilliams:
https://github.com/dyninst/dyninst/pull/259 is the candidate fix to integrate here, actually. If @ssunny7 can check locally since he's the one able to reproduce it outside of Jenkins, that would be best.

ssunny7:
I can get a build running with #259 merged locally. I should note that test4_2 never showed up as failing for me at any point (actually no tests failed either before or after merging semantics into master, except for the case of the long-running test4_4).

ssunny7:
@wrwilliams @mxz297 Alright, this is what I've done:
* Checkout dyninst master
* Merge master from Xiaozhu's fork
* Merge appropriate branch from Josh's fork
* Rebuild dyninst
* Rebuilt testsuite

Results:
* No failures as such, but test4_4 still takes the same unusually large amount of time to run
* test1_30 crashed. I remember seeing commits that addressed this and my dyninst/testsuite codebases are up to date with remote. I don't remember if it still crashed on master without any of the merges, but I can check.

wrwilliams:
This is progress. Does test4_4 complete if given long enough, or is it infinite looping? In either case, can you post a gist of the parse log if that's where it's getting stuck?

@cuviper have you seen anything dodgy with test4_4 on any of your test machines?

ssunny7:
Well, test1_30 fails for me both before and after merge.

So I ran test4_4 and stopped it after about 10 seconds. The log ended up being close to 128 MB.

I've put it under /p/paradyn/logs. File name test4_4_log.

wrwilliams:
It looks like things start going weird when we hit a ton of advance operations with `curInsn == NULL` in that log. @mxz297 does this give you any insights/is there a safe way to cut off this path? Looks to me like we're getting wildly off track here.

mxz297:
@wrwilliams Get back from travel and start to figure things out. The `curInsn == NULL` in the log is not a real issue. It is not triggered during normal control flow traversal, but triggered during jump table analysis  where I scan each basic block to look for thunk. Here I advance the current basic block after scanning each instruction and the last advance in the loop cause it to produce this message. I change the code to not advance the basic block in the last iteration, so we get rid of this  `curInsn == NULL` message. I will continue to see whether there are suspicious parsing issues in libc as it looks like the test4_4 issue and test_pt_ls issue could somehow tied together with parsing libc.

wrwilliams:
retest this please

mxz297:
retest this please

ssunny7:
I guess this needs to be re-triggered. Failed at `Cleaning build dir... /opt/bitnami/apps/jenkins/jenkins_home/jobs/dyninst-build/workspace/build
ERROR: Unable to delete '/opt/bitnami/apps/jenkins/jenkins_home/jobs/dyn`.

wrwilliams:
Jenkins is currently idle, so: retest this please.

ikitayama:
I don't seem to execute test4_4 from testsuite (master, as of this writing) on arm64.

Below is all the results I can obtain:
 
test_pt_ls                        none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
test2_1                           none 64  create   NA      dynamic nonPIC  PASSED
test2_2                           none 64  create   NA      dynamic nonPIC  PASSED

wrwilliams:
ARM64 tests should only pass on proccontrol, symtab, and instructionAPI at present; anything more is a bonus. `runTests --proccontrol` (and equivalent for the other components with tests) should give you useful results for the current state of the platform.

ikitayama:
Executing ```runTests with the --proccontrol``` gives me the same results as above.
Is this what I should get?

wrwilliams:
No, and that's a sign that you've not built the testsuite mutatees for some reason--and it shouldn't be at all related to this PR. Can you open a new issue with your testsuite's CMake output and tag me and @ssunny7 on it?

PR #231, opened 2016-11-14
RT: trymmap should retry if the result is out of range

cuviper:
An address passed to `mmap` is just taken as a hint, and the OS may
return something wildly different if that address is not available.
This is undesirable when we're trying to create a constrained alloc.

Now we will check that the address is in the requested range before
accepting it.  Otherwise, unmap it and try a new hint.

### DIFF
--- a/dyninstAPI_RT/src/RTheap.c
+++ b/dyninstAPI_RT/src/RTheap.c
@@ -91,8 +91,13 @@ static Address trymmap(size_t len, Address beg, Address end, size_t inc, int fd)
   /* until we get one that succeeds.*/
   for (addr = beg; addr + len <= end; addr += inc) {
     result = map_region((void *) addr, len, fd);
-    if (result)
+    if (result) {
+      /* Success doesn't necessarily mean it actually mapped at the hinted
+       * address.  Return if it's in range, else unmap and try again. */
+      if ((Address) result >= beg && (Address) result + len <= end)
         return (Address) result;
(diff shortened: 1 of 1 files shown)


cuviper:
This is the master-rebase of #176.

cuviper:
The AppVeyor failure is a from partial download of boost.

wrwilliams:
Waiting for Jenkins tests to actually finish; I'm going to do some config tweaking to get it to be faster and to try to ensure that the PR status *actually* reflects what's happening, rather than giving premature green.

wrwilliams:
http://jenkins.dyninst.org/jenkins/job/dyninst-test/66/testReport/dyninst.test1_29/stat_gcc_64_none_none/

You seeing this with a local test suite build, Josh?

cuviper:
I'm confused - those results say `test1_29`, but in some of the details they say from `test1_35`.

Either way - I get all PASSED for `test1_29` and all SKIPPED for `test1_35`.

wrwilliams:
Does that include the 64-bit statically linked mutatees?

cuviper:
Yes - running on Fedora 25 x86_64, `./runTests -all -test test1_29` passes everything.
(`test1_35` is not actually interesting, since it's hard-coded to skip.)

wrwilliams:
Good. IIRC you'd had some test environments that weren't full static/dynamic 32/64 capable, so I figured I'd best check. This should be okay to merge then.

Issue #64, opened 2016-05-27
walkSingleFrame run against local process on WIndows crashes


jgerie:
This is a variation on issue #63 which I reported. My ultimate goal is to walk the stack in the local process, not in an external process. When I try my test program against the local process, the first call to walkSingleFrame dies with:

Program: StackTest.exe
f:\dd\vctools\crt\crtw32\misc\dbgheap.c
Line: 1424

Expression: _pFirstBlock == pHead

My test program is: 
[self_stackwalk.txt](https://github.com/dyninst/dyninst/files/287311/self_stackwalk.txt)

The call stack from the debugger is:

MSVCR120D!_free_dbg_nolock+0x49d
MSVCR120D!_free_dbg+0x4e
MSVCR120D!operator delete+0xbc
stackwalk!std::allocator<std::_Container_proxy>::deallocate+0x10
stackwalk!std::_String_alloc<0,std::_String_base_types<char,std::allocator<char> > >::_Free_proxy+0x38
stackwalk!std::_String_alloc<0,std::_String_base_types<char,std::allocator<char> > >::~_String_alloc<0,std::_String_base_types<char,std::allocator<char> > >+0xf
stackwalk!std::basic_string<char,std::char_traits<char>,std::allocator<char> >::~basic_string<char,std::char_traits<char>,std::allocator<char> >+0x1b
stackwalk!Dyninst::Stackwalker::Walker::newWalker+0x235
StackTest!main+0xaa
StackTest!__tmainCRTStartup+0x11a
StackTest!mainCRTStartup+0xd
KERNEL32!BaseThreadInitThunk+0x24
ntdll!__RtlUserThreadStart+0x2f
ntdll!_RtlUserThreadStart+0x1b

The frame "stackwalk!Dyninst::Stackwalker::Walker::newWalker+0x235" is the return statement in "Walker* Walker::newWalker(std::string exec_name)", so it appears to be dying while deallocating the local variables in that function.

My build and test environment for DynInst is:
- Windows 10, 64-bit
- Visual Studio 12
- Cmake 3.5.2 (cmake-3.5.2-win32-x86)
- Boost 1.59 (I downloaded a pre-built Windows version, boost_1_59_0-msvc-12.0-32)
- DynInst 9.1.0
- I'm building exclusively 32-bit binaries, even though the OS is 64-bit
- For most of my testing and debugging, I'm using a Debug build of DynInst. I tried a Release build first and then moved to Debug to try to figure out the problems.

[...]

PR #232, opened 2016-11-15
Implement ARMv6-M decoding

rafzi:
This implements decoding for the ARMv6-M architecture. None of the changes should affect existing functionality.

The most intrusive change is the addition of Dyninst::stripAddrEncoding which is then called before building the buffer that is given to the decoder. Thumb mode is encoded into addresses by setting the LSB, falsifying any buffer for the decoder. It's a no-op for everything except ARM thumb addresses. I'm sure there is a better solution, but intercepting at this layer also makes sure all the CFG logic still holds.

Large parts of this patch are generated with the help of a project that is currently closed source.

Some tests have been done and most of the instructionAPI is working properly: Decoding, successors, BBs, CFGs, semantics.


wrwilliams:
ok to test

wrwilliams:
@nathanhjay Can you run fleece on this PR and provide some feedback?

nathanhjay:
Sure, I'll run fleece today and let you know how it looks.

wrwilliams:
retest this please

(Noting that the NFS failure on this build is from attempting to perform a clean build with another build active, and I need to fix that on the Jenkins side.)

nathanhjay:
Running this, I'm getting a lot of output that looks like this:
`
Mask:    0b1111'0000'0000'0000
Compare: 0b1010'0000'0000'0000
Mask:    0b1000'0000'0000
Compare: 0b0000'0000'0000
Mask:    0b1111'0000'0000'0000
Compare: 0b1010'0000'0000'0000
Mask:    0b1000'0000'0000
Compare: 0b0000'0000'0000
Mask:    0b1111'0000'0000'0000
Compare: 0b1001'0000'0000'0000
Mask:    0b1000'0000'0000
Compare: 0b0000'0000'0000
`

Does this look like something you added for development? Have you added flags to turn it off?

wrwilliams:
Integration doesn't look great here: http://jenkins.dyninst.org/jenkins/job/dyninst-test/68/testReport/

No rush on this, we'll save it for 10.0.

nathanhjay:
@rafzi Any chance those printouts can be stopped? I can't move forward with decoder comparison testing with that much printing.

rafzi:
Updated with some fixes now.

@nathanhjay Yes those were debug artifacts, sorry for that.

@wrwilliams That link is a 404 for me. I permitted the Jenkins user on github.

wrwilliams:
@rafzi Are any build results accessible from the main jenkins.dyninst.org page for you?

rafzi:
@wrwilliams I can access the "dyninst-build" and "testsuite-build" jobs.

umdhollings:
I finally got a chance to check this out.  Overall there is a lot of good work there, but I am a bit confused about what is or is not being done with thumb instruction immediate form instructions.   It looks like they are not being handled correctly.  For example bytes: 2b 40 which should be "cmp r3, #64" is missing the immediate. 

It looks like in the file InstructionDecoder-ARMv6M.C (mostly generated by your tool I gather), that local_imm32 is set, but there is a missing call to appendOperand to add the immediate.   Is this something easy to fix in your tool?

You mention that your tool is not open source, but didn't give anymore information about it.  Is there any documentation for it?   Is there a way to obtain a copy under license?

rafzi:
@nathanhjay @umdhollings  Sorry for the late reply.

Yes, the assembly that is produced by the decoder output is not valid. Do I understand correctly that this is not required for any functionality of Dyninst? This is just a result of how the architecture and model is built to be more general. For example when an instruction reads 4 bytes, it is modelled as four 1 byte reads, because memory is bytes addressable and semantically it does not make a difference. Flags are also not treated as special case, but just like registers.

As for the other issues you pointed out, I should point out how the decoder operates. The goal is to fill all the fields of the Instruction class that are important for running algorithms on them. The decoder picks out the register and memory references and builds a RegisterAST that describes the read or write as statically as possible. Each read or write becomes one operand and duplicate operands are a result of that. Immediates do not cause a read or write by themselves so they are not added as operands.

That said, we actually do have assembler syntax information in the model that could be used to built valid assembler text. However, since we do not use the assembler for anything, this was not worth the time. It would still be a great effort and not really be compatible with your way of building asm syntax from operands.

The tool used to generate the code is completely private at the moment. We are considering open-sourcing it, but there is no decision, yet.

Unrelated note: Currently an assert is triggered when a jump table candidate is found, because IA_details is not implemented. I'll push a fix soon.

umdhollings:
Thanks for getting back to us.  

The issue of immediates is important for other things besides instructionAPI.  Other APIs and higher level tools (i.e. symbolic execution tools) use the information from the operands and so we do need to provide correct operand expressions for instructions.

One of the reasons we want to have a correct assembly output has to do with a testing process we now use where we dissemble instructions re-assemble them and compare the binaries.  This allows us a way to automatically test instruction parsers.    However, without valid assembly we can't do this type of testing.

Although you have not decided to open source your tool, would you be willing to consider making it available to the UMD and Wisconsin teams (with no rights to re-distribute it)?  This would allow us to fix bugs and continue to add features to your port.   The basic problem we are having is that it is very hard to accept code into the project if there is not a clear path for at least one other developer to maintain the code.  With your code dependent on a closed tool that only you have access to, we can't really maintain the code.

Thanks for looking at the jump table issue.  I think I tripped over that when testing your code with an ARM Arduino board.

PR #244, opened 2016-11-16
Test att_syntax_formerge for merge into master

ssunny7:
Not ready for final merge into master, needs to build successfully and pass all tests


ssunny7:
@wrwilliams So this caused the testsuite-build to fail (http://jenkins.dyninst.org/jenkins/job/testsuite-build/81/console).

The reason is that the testsuite code needs to be updated because of the changes introduced by this branch. I've made the changes and committed them locally.

Now, if I push them, it would cause the testsuite-build against this PR to pass. But testsuite-build would fail against dyninst/master since this PR is not merged yet. But this failure would be temporary - as soon as the testsuite build against this PR passes and if all tests pass, we would merge it into master. So is it okay to push the commits I made locally to testsuite to GitHub?

wrwilliams:
Is the problem here just register definitions? If so, can we keep a merged set of definitions in Dyninst (moving any deprecated registers to an unused integral range) rather than breaking the build?

ssunny7:
It really is the definition of only one register, and updates to the format() calls for instruction components (these will not break the build against this PR though, since we've brought back the old format() functions).

So if I'm understanding this correctly, I add the old register to this branch (att_syntax_formerge). Then when we approve this PR and it gets into master, we remove the old definition (directly from master) and at the same time update the testsuite. Does this sound correct? If yes, I'll get this done ASAP so this PR can be closed.

ssunny7:
I should add: doing the above may cause he build to pass, but will cause all tests that use that register to fail. The test would be expected the old register, but instruction decoding in this branch uses thew new register.

wrwilliams:
Yes, that sounds right, Sunny.

ssunny7:
What about the test failures that may happen with this branch though (https://github.com/dyninst/dyninst/pull/244#issuecomment-261055628)? I can imagine they wouldn't matter on side branches as long as we get the final version on master to work, not sure if that's the right way of looking at this.

jdetter:
One of my commits was missing and I pushed a fix for the AppVeyor build issue.

PR #259, opened 2016-11-16
proccontrol: scrub newly created threads that fail to attach

cuviper:
If `int_thread::createThread` failed to actually attach to the thread,
it was leaving the thread object in the process in the `neonatal` state.
This failed assertions later when trying to stop all of the process's
threads, as it would have handler `stopped` and generator `neonatal`.

Now when a thread attach fails, it is set to `errorstate` and removed
from the active thread pools.  The assumption is that this thread simply
exited before we could attach, but we can't be sure of that without
having access to the ptrace return code (`ESRCH`).

### DIFF
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -3495,8 +3495,14 @@ int_thread *int_thread::createThread(int_process *proc,
    bool result = newthr->attach();
    if (!result) {
       pthrd_printf("Failed to attach to new thread %d/%d\n", proc->getPid(), lwp_id);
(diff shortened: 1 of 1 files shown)


wrwilliams:
This looks like it's doing better on Jenkins (and apologies for the digging required to get test results; anyone who knows better how to get everything posted cleanly can ping me with advice/requests for keys). @mplegendre can you spare the time for a quick code review?

wrwilliams:
Looks good (Legendre 2016, personal communication). Holding merge for a bit while we try to stabilize the ARM semantics and jump table revisions.

PR #260, opened 2016-11-17
Document some of the ProcControlAPI options in PlatFeatures.h.

### LINKED ISSUE #151: document proccontrol "tracking" APIs
It appears that `proccontrol.pdf` doesn't have any mention of the various tracking APIs -- pretty much everything in `PlatFeatures.h` is missing.

In particular, I found the library tracking controls useful for [rhbz1366574](https://bugzilla.redhat.com/show_bug.cgi?id=1366574), to reduce mutatee impact, but I only discovered this by tracking down that breakpoint's origin to see that it could be disabled.

morehouse:
Fixes #151.


wrwilliams:
Looks pretty good. There should probably be an intro paragraph describing the interrupt vs. polling distinction for all the trackers rather than duplicating that bit across all the classes, and in either case we should be clear that the interrupt-based mechanisms may modify the process (e.g. breakpoints) and the polling ones must not modify the process.

@mplegendre Am I missing anything or does this + above cover these interfaces decently?

morehouse:
Incorporated @wrwilliams suggestions.

PR #261, opened 2016-11-18
proccontrol: fix double-increment while erasing a dead process

cuviper:
In the attach loop over waitfor_startup(), processes which fail are
erased from the set.  However, the iterator was getting incremented
again, which will skip the next process or even cause undefined behavior
if already at the end of the list.

With GCC 6.2.1, that UB manifested as an infinite loop on a self-
referential rbtree node.

The simple solution is to `continue` the loop after `erase(i++)`, as is
done in many other places with this same pattern.

### DIFF
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -453,6 +453,7 @@ bool int_process::attach(int_processSet *ps, bool reattach)
          pthrd_printf("Error waiting for attach to %d\n", proc->pid);
          procs.erase(i++);
          had_error = true;
+         continue;
       }
       i++;
    }
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -381,8 +381,10 @@ bool int_process::attach(int_processSet *ps, bool reattach)
       pthrd_printf("Attaching to threads for %d\n", proc->getPid());
       bool result = proc->attachThreads();
       if (!result) {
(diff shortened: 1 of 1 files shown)


cuviper:
I audited for similar problems with this pattern and noticed this bit in the same function:

```cxx
      pthrd_printf("Attaching to threads for %d\n", proc->getPid());
      bool result = proc->attachThreads();
      if (!result) {
         pthrd_printf("Could not attach to threads in %d--will try again\n", proc->pid);
         procs.erase(i++);
      }
```

This loop also has an `i++` at the end, so we again should avoid a double-increment.  But I'm not sure how this is supposed to work -- if it's erased from the set, then we won't even see this `proc` later where it's supposed to try again!  And do we still want this to hit the following `reattach` block that prunes old threads?

I think perhaps it just shouldn't be erased here, but I need a second opinion.  @mplegendre?

mplegendre:
I agree that code is buggy--both the printf comment that suggests we'll try again and the actual erase code.  I suspect this is some badly converted code from prior versions of ProcControlAPI.

I looked through the error return paths on attachThreads.  Only FreeBSD and Linux can error return into this, and Linux only does so if /proc isn't mounted.  I don't think that's a case we care much about anymore.  We should probably just make this error handling behave similar to the others.

Thanks for catching these.

cuviper:
Ok, thanks, I just added a commit to make that early `attachThreads` failure a hard error too.

Issue #213, opened 2016-11-03
VEX3 and EVEX assert - decoding invalid should throw exception or return error


nathanhjay:
I don't have the exact input bytes that caused the assert, but the line number should be sufficient to track it down.

```
dyninst/instructionAPI/src/InstructionDecoder-x86.C:1019: bool 
Dyninst::InstructionAPI::InstructionDecoder_x86::decodeOneOperand(const 
Dyninst::InstructionAPI::InstructionDecoder::buffer&, const NS_x86::ia32_operand&, int&, const 
Dyninst::InstructionAPI::Instruction*, bool, bool, bool): Assertion `!"Non VEX3 or EVEX instruction with am_B 
addressing mode!"' failed.
```

Issue #200, opened 2016-10-14
Invalid register numbers for VEX3 instructions


nathanhjay:
We seem to decode invalid register operands that are recognized by every other decoder as a different register number. Here are three examples:

```
Bytes: c4 a1 6c 5e cd
Dyninst: vdivps %ymm2,%ymm21,%ymm1
Expected: vdivps %ymm5, %ymm2, %ymm1

Bytes: c4 81 39 df c5
Dyninst: vpandn %xmm8,%xmm29,%xmm0
Expected: vpandn %xmm13, %xmm8, %xmm0

Bytes: c4 22 71 03 fa
Dyninst: vpaddsw %xmm1,%xmm18,%xmm15
Expected: vphaddsw %xmm2, %xmm1, %xmm15
```

All have the error: invalid register operand for '<opcode>'

jdetter:
I was not able to reproduce this issue on master. I remember fixing a similar issue and the fix for that may have only gotten merged into master and not att_syntax. When att_syntax and master are synced this issue should be fixed.

Issue #196, opened 2016-10-14
Decode incorrect VEX3 as valid instruction


nathanhjay:
Input produces errors on all other decoders. Dyninst output produces the assembler error: ambiguous memory operand size for `crc32`, so I think we're wrong on this instruction.

Input:
c4 e2 7f f0

Dyninst Output:
crc32 (%rdx),%eax

Issue #195, opened 2016-10-14
Decode incorrect VEX2 as valid instruction


nathanhjay:
The following two inputs produce an error according to all other decoders:
c5 a9 2c
c5 1d 2d 15 4b 9d 65 68

These instructions also suffer from the double printing mentioned in #193, but don't seem valid regardless.
Dyninst outputs:
cvttpd2pi (%rax),(%rax)
cvtpd2pi 0x68659d4b(%rip),0x68659d4b(%rip)

PR #271, opened 2016-11-22
x86 decoding and syntax fixes -- tentative

### LINKED ISSUE #193: Double printing first operand
@jdetter we seem to be printing a single operand twice instead of printing the second operand properly.
```
Bytes: 0f 2c 44 81 4e
Dyninst output: cvttps2pi 0x4e(%rcx,%rax,4),0x4e(%rcx,%rax,4): 
Assembler error: too many memory references for 'cvttps2pi'
Expected output: cvttps2pi 0x4e(%rcx, %rax, 4), %mm0: ERROR:  bad register name 
```

jdetter:
This branch was started from master so it isn't affected by the `att_syntax*` branch issues. This pull request is tentative.

Fixes #193 #195 #196 #213

~~Fix for #200 is still in progress.~~

The fix for #200 is already on master. When master and att_syntax are synced this should be fixed.

I have fixes for most of the other issues on my local `att_syntax` branch but I'm waiting for all of the issues with that branch to be resolved before I merge any of those in.

These changes will change the ABI.


jdetter:
@ssunny7 do you care if I merge all of this into `att_syntax`? I was planning on merging to master but it makes more sense to just merge into `att_syntax`.

jdetter:
I'm going to wait until we have the testsuite issues figured out before I merge this in. I don't want to create any extra noise.

PR #273, opened 2016-11-23
Compiling DynInst v9.2.0 on arm64

ikitayama:
Please review this trivial name changes needed for the arm64 build.


ikitayama:
The build was attempted on an X-Gene 1 with the Fedora 24 release.

Do I have to take care of the appveyor error as well?

wrwilliams:
Nope, that's a transient boost download failure.

PR #277, opened 2016-11-29
Cleanup ppc (and old gcc) issues

wrwilliams:
Fixed ppc build issues, and brought jump table handling to the point where it's broken because of tables that don't match our heuristics and a lack of ppc64 semantics, rather than a state where it can't possibly work.


wrwilliams:
No regressions under Jenkins; this is clean to merge when @jdetter cleans up the nullptr uses. (All of those uses have a default argument of NULL, by the way, so the argument can just be deleted.)

jdetter:
Download failed for the AppVeyor.

jdetter:
@wrwilliams can we merge this? The AppVeyor failed because the download failed.

PR #279, opened 2016-11-29
Fix dependency of LibDwarf

rafzi:
Dyninst would not build on the first run when libelf and libdwarf are not installed on the system.


jdetter:
ok to test

PR #280, opened 2016-11-30
common: use ptrace if yama blocked process_vm_readv/writev

### LINKED ISSUE #274: Test4_4 seems to be in deadlock on amd64_ubu14
@cuviper We observed that current master branch will lead to a deadlock when running test4_4 on Ubuntu 14. Preliminary debugging suggests that the mutator is forever looping for checking whether two threads have finished. Merging your two pull requests about proccontrol does not seem solve this issue.

I attached the debugging log by enabling DYNINST_DEBUG_PROCCONTROL. 

Do you have any insights on this issue?

https://gist.github.com/mxz297/23be1aa22fdecee056aa9fbfbe41464a

cuviper:
Having sysctl kernel.yama.ptrace_scope=1, one may only call ptrace
attach on direct descendants.  The same restriction is also checked for
`process_vm_readv`/`writev` and certain procfs files.  However, if an
intermediate parent process already exited, we could end up with a
grandchild that we're still ptracing but isn't our descendant, so we
can't use the `process_vm` functions anymore -> `EPERM`.

We already had a fallback here for `EFAULT`, to just use `ptrace` memory
access, so use the same fallback after `EPERM` too.

Fixes #274.

### DIFF
--- a/common/src/linuxKludges.C
+++ b/common/src/linuxKludges.C
@@ -286,9 +286,11 @@ bool PtraceBulkRead(Address inTraced, unsigned size, void *inSelf, int pid)
       if (ret == -1) {
          if (errno == ENOSYS) {
             have_process_vm_readv = false;
-         } else if (errno == EFAULT) {
+         } else if (errno == EFAULT || errno == EPERM) {
             /* Could be a no-read page -- ptrace may be allowed to
-             * peek anyway, so fallthrough and let ptrace try.  */
+             * peek anyway, so fallthrough and let ptrace try.
+             * It may also be denied by kernel.yama.ptrace_scope=1 if we're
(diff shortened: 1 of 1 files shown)


cuviper:
Did Jenkins actually test this?  I found [testsuite-build #97](http://jenkins.dyninst.org/jenkins/job/testsuite-build/97/), but it doesn't look like it ran anything.

Anyway, I'm not worried about it -- I think this change has very low risk.

wrwilliams:
It did, but it failed to aggregate; the results look clean.

May as well merge the proccontrol fixes in whatever order is best, Josh; they should help clean things up for the other testing we're doing on parsing fixes etc.

cuviper:
Ok, thanks!  They're all touching distinct lines, so I'll just merge them in submission order.

PR #282, opened 2016-12-01
findMain analysis fixes

jdetter:
There must have been some miscommunication the last time I worked on `findMain` because we were analyzing the call target of the first call from the entry of the program. I changed it back to what it should be, which is analyzing the contents of `%edi`, which should be updated right before the call to `__libc_start_main@plt`.

This should work wither the executable is PIE or non-PIE.


wrwilliams:
Should be superseded by #286.

PR #287, opened 2016-12-05
Line info cleanup

wrwilliams:
Fixes that should cover test1_30 failures and give us a not-terrible fallback when we're confronted with statements that are outside of range info.


wrwilliams:
retest this please

PR #294, opened 2016-12-10
arm64: remove syscall tracing workaround

ikitayama:
After the commit 04d7e098f541769721d7511d56aea4b976fd29fd, in both 32-bit and 64-bit ARM kernels have the capability of tracing a syscall in the fast path as well (ie changing the flag during the syscall execution is allowed). Remove the workaround introduced as part of  commit 215d0f967a259cf883c903965ac490851e665170 .

cc @cuviper

### DIFF
--- a/proccontrol/src/linux.C
+++ b/proccontrol/src/linux.C
@@ -1512,34 +1512,6 @@ bool linux_thread::plat_cont()
    int result;
    if (hasPostponedSyscallEvent())
    {
-/* This should be turned on to solve the arm kernel bug.
- * When it recognize the syscall enter event, insert a bp as the fake
- * syscall exit stop.
- */
-#if defined(arch_aarch64) && defined(SYSCALL_EXIT_BREAKPOINT)
-       if( postponed_syscall_event->event_ext == PTRACE_EVENT_FORK ||
-           postponed_syscall_event->event_ext == PTRACE_EVENT_CLONE ){
-            //install bp
-            Dyninst::MachRegisterVal addr;
-            result = plat_getRegister(MachRegister::getPC(this->llproc()->getTargetArch()), addr);
-            if (!result) {
-               fprintf(stderr, "Failed to read PC address upon crash\n");
-            }
-            pthrd_printf("Got SIGTRAP at %lx\n", addr);
-            pthrd_printf("Installing breakpoint for postpone syscall at %lx\n", addr);
-
-            this->BPptr_fakeSyscallExitBp = Breakpoint::ptr();
-            //int_breakpoint *fakeSyscallExitBp = new int_breakpoint(Breakpoint::ptr());
-            int_breakpoint *fakeSyscallExitBp = new int_breakpoint(this->BPptr_fakeSyscallExitBp);
-            fakeSyscallExitBp->setThreadSpecific(this->thread());
-            fakeSyscallExitBp->setOneTimeBreakpoint(true);
-            //this->llproc()->addBreakpoint(addr, fakeSyscallExitBp);
-            this->proc()->addBreakpoint(addr, this->BPptr_fakeSyscallExitBp);
-            this->addr_fakeSyscallExitBp = addr;
-            this->isSet_fakeSyscallExitBp = true;
-       }
-#endif
-
       pthrd_printf("Calling PTRACE_SYSCALL on %d with signal %d\n", lwp, tmpSignal);
       result = do_ptrace((pt_req) PTRACE_SYSCALL, lwp, NULL, data);
    }
--- a/proccontrol/src/linux.C
+++ b/proccontrol/src/linux.C
@@ -101,9 +101,6 @@ using namespace ProcControlAPI;
 #define PTRACE_SETREGS PPC_PTRACE_SETREGS
 #endif
 
-#if defined(arch_aarch64)
-#define SYSCALL_EXIT_BREAKPOINT
-#endif
 
 static pid_t P_gettid();
 static bool t_kill(int pid, int sig);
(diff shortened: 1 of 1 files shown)


cuviper:
There are several more blocks related to `SYSCALL_EXIT_BREAKPOINT` -- shouldn't they all be removed?  And probably the `DISABLE_POSTPONE` stuff too.

That is, assuming everyone is ok with requiring kernel 4.2 for aarch64 support.  Fedora's fine, and while it's still in development preview for aarch64, RHEL 7.2 and 7.3 also look ok.  If there are other platforms you need this for, please check their kernel.

ikitayama:
@cuviper @wrwilliams after removing the code, many of the proccontrol tests, except pc_addlibrary,
pc_exec_targ, pc_irpc, and pc_tls run fine. I think the failures, as @wrwilliams said, are observed
with the current master branch as well when tested with old ARM and arm64 kernels.

wrwilliams:
A proper fix here will check the kernel version at CMake time and enable/disable the workaround as needed. I haven't had spare time to update our ARM development box lately so we will need to keep things working there until @ssunny7 or I get around to that.

cuviper:
> check the kernel version at CMake time

Please, no.  Kernels change, and compile-time version checks end up with *implicitly* unportable binaries.

If you must retain the workaround for your old system, how about an *explicit* CMake flag?  Or if it's just about that one dev box, perhaps you can update just its kernel until you have time for a full upgrade.

mplegendre:
I agree with @cuviper about not using a CMake check.  And a dynamic check of kernel versions would be even better--we've generally tried to keep #ifdef's out of ProcControlAPI for portability.

I'd think the current code would assert-fail any time it's run.  So it shouldn't a functionality loss to remove it.  I'm not sure how the dev box is currently working.

Reverting this is a little more complicated than just removing the ARM code.  If I'm reading things right, it looks like commit 80c14b69 first broke the general trap-based syscall exiting that Josh wrote.  Then 215d0f96 wrapped the general system to be ARM specific.  We need to get it back to a pre-80c14b69 state.

ikitayama:
@mplegendre given that the commit https://github.com/dyninst/dyninst/commit/80c14b69b0696a17d053e967d218dc440ffb35fc was made some time ago, going back to a state that was assumedly stable as you suggest might be difficult. Can we instead merge small patches as we discover bugs? Apparently the current master does not work for anyone on arm64 or ARM platforms.

mplegendre:
I'm curious what @wrwilliams thinks.  In order of increasing complexity we could:

* Get ARM working by removing the syscall work-around, as per this patch.
* Get the original trap-based syscall exiting working again by hand reverting us back to the state we were in before 80c14b6.  If I recall correctly, we only saw that issue on older systems (I saw it on RHEL5), so the fact it's broken may not be too relevant.
* Re-create the ARM workaround for the kernel bug, but do it correctly using handlers.

ikitayama:
Below is the results with the 4.8.10-300.fc25.aarch64 kernel

```
pc_addlibrary              g++    none 64  create   SPST    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  create   MPST    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  create   SPMT    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  create   MPMT    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  attach   SPST    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  attach   MPST    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  attach   SPMT    dynamic nonPIC  PASSED
pc_addlibrary              g++    none 64  attach   MPMT    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  create   SPST    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  create   MPST    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  create   SPMT    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  create   MPMT    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  attach   SPST    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  attach   MPST    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  attach   SPMT    dynamic nonPIC  PASSED
pc_addlibrary              gcc    none 64  attach   MPMT    dynamic nonPIC  PASSED
```

ikitayama:
With the patches upto https://github.com/dyninst/dyninst/pull/294/commits/f47d0f6d5aa5878147dac31a86eabcc22f259da2, I can get below on Mustang.

```
$ ./test_driver -test pc_irpc -all
pc_irpc                    g++    none 64  create   SPST    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  create   MPST    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  create   SPMT    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  create   MPMT    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  attach   SPST    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  attach   MPST    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  attach   SPMT    dynamic nonPIC  PASSED
pc_irpc                    g++    none 64  attach   MPMT    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  create   SPST    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  create   MPST    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  create   SPMT    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  create   MPMT    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   SPST    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   MPST    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   SPMT    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   MPMT    dynamic nonPIC  PASSED
pc_irpc                    gcc    none 64  create   SPST    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  create   MPST    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  create   SPMT    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  create   MPMT    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   SPST    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   MPST    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   SPMT    static  nonPIC  PASSED
pc_irpc                    gcc    none 64  attach   MPMT    static  nonPIC  PASSED
```

wrwilliams:
ok to test

ikitayama:
Below results obtained on Mustang  (arm64  SoC) https://gist.github.com/ikitayama/9215c08b810badaa099520a1d239034b

wrwilliams:
This hung on my x86 RHEL6 desktop. The logic in linux.C for handling errors in what may be PEEKUSER or GETREGSET should be:
```
#if defined(arch_aarch64)
   if (ret != 0) {
#else
   if (result == -1 && errno != 0) {
#endif
```

which can probably be better factored, as well.
@cuviper, @mplegendre: we should be preferring GETREGSET where it's available, right?

PR #298, opened 2016-12-13
Fix omitting leading 0's when concatenating hex bytes in stringstream

rauhfasertapete:
Dear Dyninst-team,

when building the path to separate files with debug information, the code omits leading zeros when concatenating hex bytes into a stringstream. (Problem is that setw() is not persistent). For example:

/usr/lib/debug/.build-id/d5/5c98c7fc69cfc0fc2504bc1df8a81fe7e255f9.debug <-- correct path
/usr/lib/debug/.build-id/d5/5c98c7fc69cfc0fc254bc1df8a81fe7e255f9.debug <-- path build in Elf_X::findDebugFile where the zero in the 0x04 byte is omitted.


wrwilliams:
ok to test

PR #299, opened 2016-12-20
Enhance DWARF parser to recognize (and skip) DW_TAG_member entries for static fields

rauhfasertapete:
Dear dyninst team,

since DWARF3, static fields in C++ classes are represented by DW_TAG_variable entries and (newly added) DW_TAG_member entries. (Happened with GCC 6.2.1 here.) We found this note/issue on it: http://www.dwarfstd.org/ShowIssue.php?issue=161118.1

This seems not to be covered by dyninst yet when parsing structures: they are exposed as fields but with offset -1 in SymtabAPI. If I am not mistaken, fields exposed by SymtabAPI with offset == -1 can then be either constants or static fields.

As these fields are not part of the structures but global variables instead, we added a check to skip these newly added entries during DWARF parsing. What do you think about it?

We stumbled across this when parsing DWARF generated by GCC 6.2.1 from the following (the cyclic dependency was not handled by our code):

`class A {
    ...
    static A a;
    ...
};`

### DIFF
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -1240,6 +1240,12 @@ bool DwarfWalker::parseMember() {
    int offset_to_use = locs.size() ? locs[0].frameOffset : -1;
 
    dwarf_printf("(0x%lx) Using offset of 0x%lx\n", id(), offset_to_use);
+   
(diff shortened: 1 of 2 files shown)


rauhfasertapete:
Thanks for the heads up. The logic sounds perfect.

I have removed the new function again and tried to implement it 1:1, please see the second commit.

wrwilliams:
ok to test

PR #308, opened 2016-12-24
getVsysInfo() return NULL on arm64

ikitayama:
On arm64 the vsyscall page has been deprecated for some time, just return NULL as done on 64-bit x86 platforms.


ikitayama:
PR #309 should be reviewed, instead this one.

Issue #303, opened 2016-12-21
Stackwalk issue on arm64 


ikitayama:
@wrwilliams I'm wondering whether the StackwalkerAPI is fully functional on someone's branch other than master. As of this writing I get: 

```
$ DYNINST_DEBUG_STACKWALK=1 ./a.out 
[walker.C:98] - Creating new stackwalker on current process
[libstate.C:59] - Creating a TrackLibState on pid 873
[linuxbsd-swk.C:266] - Successfully install lib tracker at 0xffff93dcf920
[walker.C:73] - Creating new Walker with proc=0x15992bb0, sym=(nil), step = 1
[steppergroup.C:83] - Constructing new AddrRangeGroup at 0x15b38330
[framestepper.C:50] - Creating FrameStepper 0x15b0e650 with walker 0x15b9cd10
[framestepper_pimple.h:74] - Constructing DebugStepper at 0x15b0e650
[framestepper.C:50] - Creating FrameStepper 0x15bfc390 with walker 0x15b9cd10
[walker.C:649] - Registering stepper DebugStepper with group 0x15b38330
[steppergroup.C:150] - Adding stepper DebugStepper to address ranges 0 -> ffffffffffffffff
[linux-aarch64-swk.C:89] - Stepper 0x15b0e650 is DebugStepper
[symlookup.C:49] - Creating SymbolLookup 0x15bacea0
[walker.C:115] - Successfully created Walker 0x15b9cd10
[frame.C:96] - Created frame at 0xffffd8b61c80
[walker.C:391] - Starting stackwalk on thread 873
Segmentation fault (core dumped)
```

wrwilliams:
It should be fully functional on master, and that code hasn't changed other than your proccontrol changes in months (at least in any arm-specific way). Can you get a mutator-side stack trace for this crash?

ikitayama:
@wrwilliams how do I do that?

wrwilliams:
Run the mutator under gdb and get the backtrace from the crash? It shouldn't be different from any other crashing program.

wrwilliams:
I've reproduced what looks like this issue locally. Third party stack walking looks okay from here; it's only first party that's broken. I would guess that we're missing -fno-omit-frame-pointer or that the gcc flag for forcing a frame pointer has changed; the actual logic here should be stable.

I'm checking the build system now but I'd suggest disabling cotire for stackwalker and testing again. The code as written should work if that flag is functional.

wrwilliams:
[...]

PR #309, opened 2016-12-26
Fix test_basic test error

ikitayama:
Calling dwarf_srcfiles with error being NULL triggers an error when
deallocating. Fix it by adding an argument to dwarf_srcfiles.
Tested with libdwarf,
```
libdwarf.aarch64                     20161001-1.fc25                     @fedora
```
on Mustang. Please see for more technical background analyzed by William Cohen at:
```
https://bugzilla.redhat.com/show_bug.cgi?id=1316695
```

Unlike x86, Fedora 25 aarch64 still has an old version of libdwarf.


wrwilliams:
Fixes #303.

PR #313, opened 2017-01-18
Suppress debug message when no vsyscall page was found on arm64

ikitayama:
`getVsysInfo()` on arm64 has recently changed to return NULL. The test_stack_1 test works on my local x86 box.


ikitayama:
The .opd section is defined as an additional special section in the 64-bit PowerPC ELF ABI.
Please refer to #113da6ff9e4291564e1c828570d9001669c0341f, for the rational behind this change.

wrwilliams:
@ikitayama can you change this to target v9.3.x branch?

PR #316, opened 2017-01-27
Use ifdef to guard x86 code

ikitayama:


ikitayama:
@wrwilliams would you pick up the second commit 64a18e5? It took a while to figure this was needed on non-Windows platforms.

wrwilliams:
@ikitayama can you change this to target v9.3.x branch?

PR #320, opened 2017-02-07
Fix 9.3.0 incompatibilities

wrwilliams:
Removed accidental requirement for `--std=c++11`/support of `auto` to use Dyninst public headers.
Updated CMake libelf check to only recognize libelf.so.1.


wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
This targets the wrong branch; resubmitting with cleaner history to correct target.

PR #323, opened 2017-02-09
Fixes for API and dependency issues in 9.3.0

wrwilliams:
Dyninst 9.3.0 has two major problems:

1) Users were accidentally required to use the C++11 standard when building tools that use SymtabAPI, directly or indirectly.
2) While we removed libelf.so.0 support, our dependency infrastructure and CI settings would still try to use libelf.so.0.

Both of these are now fixed: we require libelf.so.1 to be present, and the spurious use of `auto` in public headers has been removed (along with some dead code).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: false
 env:
     global:
         - LLVM_VERSION=3.7.0
-        - LD_LIBRARY_PATH=$HOME/clang-$LLVM_VERSION/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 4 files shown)


wrwilliams:
@pefoley2 Are the 64-bit Windows failures here just a matter of Appveyor infrastructure misbehaving? None of these changes should have broken the Windows build and it certainly looks like these aren't our fault.

cuviper:
> Users were accidentally required to use the C++11

I'm still in the position that we *should* make that leap (#167) -- maybe for v10?

wrwilliams:
Yes, the intent is to make that leap deliberately at a major version bump, and this met neither criterion. :)

pefoley2:
Yeah, it looks like appveyor is being flaky as usual :|

PR #347, opened 2017-03-13
Fix up exception handling code so that we only consider call instructions for exception sensitivity and its attendant emulation

wrwilliams:
Fix up exception handling code so that we only consider call instructions for exception sensitivity and its attendant emulation


wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
retest this please

PR #350, opened 2017-03-15
Support 32-bit builds in Jenkins

cuviper:
Jenkins now has a `dyninst-build32` target using `CFLAGS=-m32`, but it needs this cmake fix to pass the flags to external projects.  I also cherry-picked the `Address`/`Offset` change from `v9.3.x`.

### DIFF
--- a/cmake/packages.cmake
+++ b/cmake/packages.cmake
@@ -39,7 +39,7 @@ if (UNIX)
       URL http://www.paradyn.org/libdwarf/libdwarf-20130126.tar.gz
       #	GIT_REPOSITORY git://git.code.sf.net/p/libdwarf/code libdwarf-code
       #	GIT_TAG 20130126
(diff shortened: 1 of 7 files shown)


cuviper:
Hmm, I don't think I have the new triggers set correctly, but you can see the last result here:
http://jenkins.dyninst.org/jenkins/job/dyninst-build32/15/

wrwilliams:
@cuviper I've been tweaking Jenkins flow a bit:

* Builds need to be in series though 2 test jobs can execute in parallel, or we run out of memory and crash Jenkins. I should be able to convince the VM to actually use virtual memory and not just the 4GB physical, but that's our current limiting resource. So I've got build32 downstream of build.
* Aggregating test results works if upstream triggers downstream but not if downstream watches upstream, so building tests is now triggered from build.
* Setting status should come from the PRB or from the commit status setter, not both.

wrwilliams:
And Jenkins should now have 16GB swap, which should make life a good bit more reasonable.

cuviper:
OK, do you need me to do anything more?

wrwilliams:
Don't think so. I'll let you know if I need a hand.

wrwilliams:
retest this please

wrwilliams:
retest this please

wrwilliams:
Fixes #352

PR #370, opened 2017-05-16
fix a bug when generating relocation index

fengharry:
updated relocation index comes from dynsymVector, but current code is still using the index from allDynSymbols


wrwilliams:
ok to test

wrwilliams:
Looks clean to me. Thanks, @fengharry.

PR #376, opened 2017-06-23
dyninst parses some function symbol as knownData

eastmoutain:
If symbol A and B share the same Region and Offset, A's type is
FUNCTION, however B is not, it will casue bug that the B's address
will be added into @KnownData. Here we double check whether there
existing another symbol who shares the same Region and Offset with
the current, but its type is FUNCTION, we should prevent the symbol's
address from being added into @knownData.

symbol "__aesni_set_encrypt_key"(type: NOTYPE) and
"aesni_set_encrypt"(type: FUNCTION) in libcrypto.so.1.1 meet
the condition.


mxz297:
I will revisit this issue soon.

hainest:
@eastmoutain @mxz297 It's been a year since there was any activity on this. The repository containing the PR's branch is no longer present, so I'm assuming this has been effectively abandoned. I would recommend we convert this to an issue and try to find a reproducer.

mxz297:
@hainest #755 contains fixes that will resolve this problem. 

Closed via #755

PR #381, opened 2017-07-05
Add FORCE_BOOST CMake option for @lee218llnl

### LINKED ISSUE #374: force boost build and force boost install
I would like to have build options to 1) force Dyninst to build its own boost and to 2) force Dyninst to install its own boost. This came up in the Travis build environment for STAT, which has boost 1.46.1 installed, which has a bug with shared_ptr (https://stackoverflow.com/questions/18900730/boostshared-ptrshared-ptrconst-boostshared-ptr-is-implicitly-declared). Dyninst sees the boost installation, but then fails to build. It would be nice if there were a cmake flag that tells Dyninst to download/build boost even if boost is already installed. I also think that Dyninst should install boost when it builds boost (either by default or a separate cmake flag). This also came up in Travis because I hacked Dyninst to build boost, and thus Dyninst built OK, but then I tried to build STAT, which used (via inclusion of Dyninst headers) the buggy 1.46.1 system boost installation.

wrwilliams:
Should fix #374. Caveat: this will use whatever toolchain is set in your environment for a boost build. Be sure a standalone boost build works in your environment before trusting this blindly.


wrwilliams:
@lee218llnl, does this do everything you want re Boost?

lee218llnl:
I think this looks good. If I'm not mistaken, you will now install boost whenever it is built by Dyninst?

Issue #372, opened 2017-06-01
Misc. x86 Decoding Issues


nathanhjay:
Below is a gist with remaining issues on dyninst AT&T syntax. Example instructions for each issue are given, but are not exhaustive (some issues affect addressing modes, so including all opcodes would be pretty excessive).

Most remaining issues have to do with:
old VEX encodings (or VEX2 and VEX3 instructions that have been extended),
uncommon prefixes,
mod/rm,
or formatting that may be a result of the switch to AT&T syntax (operand order, size included in opcode)

https://gist.github.com/nathanhjay/5b74260849233fba2a4ff37c956641e1

jdetter:
New issue reported by rice:

```
62 e1 7f 00 2a c3       vcvtsi2sd %ebx,%xmm16,%xmm16
```

This is decoded as No_Entry.

jdetter:
The majority of these issues have been fixed and I haven't been able to reproduce the rest, most likely because the issue was generated from a version of dyninst that didn't have the fix.

PR #393, opened 2017-07-23
Decoding and instructionAPI fixes for x86 (In Progress)

jdetter:
Multiple fixes for #372.

This PR is in progress, not ready for merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,4 @@ doxyfiles/*
 */Release/
 cotire/
 .idea/*
+boost/
--- a/common/src/arch-x86.C
+++ b/common/src/arch-x86.C
@@ -6170,7 +6170,7 @@ ia32_entry sseMapMult[][3] =
   }, { /* SSE16_NO */
     { e_vmovhps, t_done, 0, true, { Vps, Hps, Wps }, 0, s1W2R3R, 0 },
     { e_vmovhps, t_done, 0, true, { Vps, Hps, Wps }, 0, s1W2R3R, 0 },
-    { e_No_Entry, t_ill, 0, false, { Zz, Zz, Zz }, 0, 0, 0 }
+    { e_vmovhps, t_done, 0, true, { Vps, Hps, Wps }, 0, s1W2R3R, 0 },
   }, { /* SSE28_66 */
     { e_vmovapd, t_done, 0, true, { Vps, Hps, Wps }, 0, s1W2R3R, 0 },
     { e_vmovapd, t_done, 0, true, { Vps, Hps, Wps }, 0, s1W2R3R, 0 },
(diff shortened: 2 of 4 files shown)


jdetter:
@wrwilliams and @mxz297 it looks like there are a ton of testsuite failures on the `att_syntax` branch, would you rather I fix these issues on `master` or do you want them on `att_syntax?`

Here is what I get from the testsuite on x86_64 Fedora:
https://gist.github.com/jdetter/70f11b856f1f7257055127b767d3c4ef

mxz297:
@jdetter @wrwilliams I would say fixing this issue on att_syntax because this decoding issue is not visible on master. 

In terms of the testsuite failures on att_syntax, I think you can focus on fixing this decoding issue. Because this decoding issue will prevent ParseAPI correctly identify jump tables, it could have unexpected impacts on the testsuites.

jdetter:
@mxz297 This now contains a fix for the AVX-512 version of  `vcvtsi2sd` that you requested.

jdetter:
@mxz297 @wrwilliams I'm having a lot of issues reproducing the rest of the issues that aren't fixed, not sure if they were produced with a build that wasn't up to date or what.

Also I am unable to get the most up to date version of Fleece to compile, has that project been reassigned to a new member of the dyninst team or does it not have a maintainer anymore?

wrwilliams:
Fleece doesn't have an active maintainer, though you can ping @nathanhjay with questions.

It's certainly possible that at least some of the instruction decoding issues in the bug database got fixed but not closed; these should be deterministic, so if you can verify that they're resolved, feel free to mark fixed/invalid as the case may be.

mxz297:
@jdetter  I tried your latest changes and they do indeed the reported decoding errors seem fixed. Unfortunately, there is still at least one more decoding error. 
Instructions like 
```
  561c34:       48 89 54 24 20          mov    %rdx,0x20(%rsp)
  561c39:       48 89 44 24 18          mov    %rax,0x18(%rsp)

```
are decoded as:
```
561c34 : mov %rdx,(%rsp)
561c34 : mov %rdx,(%rsp)
```

It seems to me that somehow the displacements are left out. Do you have additional time to fix this? Hopefully, this will be an easy one.

mxz297:
@jdetter No need to worry about the missing displacement decoding issue. It turned out to be not hard to be fixed.

jdetter:
@mxz297 Did you push the fix to `att_syntax`?

mxz297:
@jdetter No. I am on the course of merging my jump table branch, the att_syntax branch, and the arm semantics branch. After I get these all these merges working well for my jump table tests, I merge it with the master. And then I will delete att_syntax and arm semantics since all the work from these two branches are merged.

jdetter:
@mxz297 Ok that works for me

mxz297:
Closed by #401

PR #407, opened 2017-10-04
Symtab fixes

wrwilliams:
Fix a couple of outstanding Symtab bugs.


mxz297:
@sashanicolas Can you take a look at the conflicting change in symtabAPI/src/dwarfWalker.C? I am trying to determine whether we should merge this pr or not.

PR #411, opened 2017-10-18
correct dwarf debug printing

jmellorcrummey:
in dwarfWalker::setFuncFromLowest, set the current function
before calling dwarf_printf so that curFunc() in the
dwarf_printf sees the function just found


sashanicolas:
commit cherry-picked

PR #412, opened 2017-10-18
Some necessary changes for overlapping text regions

jmellorcrummey:
Dyninst doesn't fully support binaries that contain text regions
with overlapping address ranges, e.g. CUBINs. This patch adds
necessary but not complete support for this case.

### DIFF
--- a/parseAPI/h/CFG.h
+++ b/parseAPI/h/CFG.h
@@ -548,7 +548,12 @@ class PARSER_EXPORT Function : public allocatable, public AnnotatableSparse {
     {
         bool operator()(const Function * f1, const Function * f2) const
         {
-            return f1->addr() < f2->addr();
+            if (f1->region() < f2->region()) return true;
+            else if (f1->region() == f2->region() &&
+                     f1->addr() < f2->addr()) {
+               return true;
+            }
+            return false;
         }
     };
 
--- a/parseAPI/src/SymtabCodeSource.C
+++ b/parseAPI/src/SymtabCodeSource.C
@@ -447,7 +447,10 @@ SymtabCodeSource::init_hints(dyn_hash_map<void*, CodeRegion*> & rmap,
 {
     vector<SymtabAPI::Function *> fsyms;
     vector<SymtabAPI::Function *>::iterator fsit;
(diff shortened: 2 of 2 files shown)


jmellorcrummey:
Note: This patch is not needed to support HPCToolkit's current strategy for processing CUBINs, which relocates text regions so that they are not overlapping. This support was developed before I decided to relocate text regions in CUBINs so that they don't overlap. 

Dyninst would need this patch  if it aspires to handle overlapping text regions.

sashanicolas:
ok to test

jmellorcrummey:
If you look at the files changed, you will see that it simply changes a map that identifies whether code has been seen before to check whether a (region, offset) pair has been seen before rather than just identifying regions by (offset) alone.

My change is necessary for dyninst to handle overlapping regions. I see that other conflicting changes have been committed. It looks like the table got converted from the dyn_hash_map to a TBB concurrent hash table. It would be easy enough to insert (region, offset) pairs into that TBB hash table if or when such support is desired. That could either be easily done now with a few lines of code (similar to what I wrote previously). Or, a FIXME comment could be added to the code indicating that this is something needed when code addresses are not unique.

hainest:
@jmellorcrummey Will this fix the overlapping text issues we discussed in last week's meeting?

mxz297:
@hainest Yes, this is a incomplete but necessary PR for Dyninst to support overlapping text regions. This PR can serve as a placeholder to remind us to revisit this problem later.

hainest:
@mxz297 @jmellorcrummey Since this fix is necessary, but not sufficient- can we merge as-is and create an issue to handle the rest at a future time?

mxz297:
@hainest Based on my reviewing the changes, merging this PR should not cause any negative impacts. So, we should be able to merge this after the normal testing procedure.

hainest:
@mxz297 Great, thanks! Looking at the rebase conflicts, I'm not sure of the changes. Could you take a look when you have some time?

hainest:
@jmellorcrummey @mxz297  Could one of you take a look at the rebase conflicts, when you have a moment?

hainest:
Replaced by #855

PR #414, opened 2017-10-20
Add linux-vdso64.so.1 to the library blacklist.

stanfordcox:
Ignore linux-vdso64.so.1, which is the vdso variant on some ppc64 linux.


wrwilliams:
ok to test

PR #416, opened 2017-11-06
Parallel Parsing changes

jmellorcrummey:
This pull request is not meant to be accepted. It's purpose is to show you what I've done.

This version works for Mark's complete parsing test. I want to remove the lock guard in SymtabCodeSource::lookup_region, or only use it conditionally. It is needed for Mark's post-processing, but apparently not necessary in the context of a call to parse(). Note that I have already replaced the _lookup_cache used there with an atomic, which is safe.


jmellorcrummey:
To use cilkscreen, use the following Dyninst build flags
CMAKE_CXX_FLAGS:STRING=-DENABLE_RACE_DETECTION -I</path/to/cilktools/include>

Issue #399, opened 2017-08-10
ebx should be callee-saved


melsabagh:
Shouldn't ebx on x86_32 be callee-saved? It seems to be incorrectly added to callWritten due to the assignment at this line: [https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/ABI.C#L222](https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/ABI.C#L222). This results in inconsistent slicing behavior when ebx is involved.

We discussed this some while ago on the mailing list. It seems the fix never made it to master.

mxz297:
Closed by #425

PR #425, opened 2018-01-09
Update ABI.C: remove ebx from callWritten_

melsabagh-kw:
ebx is callee-saved and should not be set in `callWritten_`. This should resolve issue #399.


wrwilliams:
ok to test

sashanicolas:
commit f3756bd2c1fd89c157cbf8b55c2f40b893d24f0f cherry-picked into master.

PR #426, opened 2018-01-10
Fixes for dynamic function wrapping without PLT entries

bwelton:
Fixes dynamic function wrapping to modify the GOT directly in cases where a PLT stub does not exist. Previously, dyninst would search PLT entries for the symbol for which the original function being wrapped would be relocated to. If not found, the GOT entry for that symbol would not be modified leading to undesired results. GCC can create code that directly looks a function up within the GOT inline, any function that was attempted to be wrapped that had this optimization would silently fail.

We may want to change the mechanics of function wrapping in dyninst to replace the cloned function instead of modifying the GOT. This would allow for the wrapper function and the cloned function symbol to be in the same .so/object.


bwelton:
Windows build failure is around the lack of a definition for a new function "bool Object::getRelocationTable(std::vector<relocationEntry> & reloc)", which makes sense since this was only added to object-elf.C. Where is the appropriate place to add this function for windows? Is it AObject?

bwelton:
This patch breaks function wrapping in binary edit mode. Until this issue gets resolved, this patch should only be used by users who require dynamic function wrapping (i.e. processCreate/processAttach).

bwelton:
This fix will not be merged currently, we need to revisit function wrapping in general and this is not a solution that should be pushed up.

PR #427, opened 2018-01-12
Fix sh_info for VERNEED section

### LINKED ISSUE #405: emitElf::createLoadableSections uses hard-coded sh_info
`sh_info` should contain the number of needed versions for the verneed table, but instead always claims to have 2 needed versions. It looks like this was a regression when code in this area was refactored according to a git blame since this information is already in scope with `verneednum`.

On a shared library with more than 2 dependencies, this issue should present itself. The following patch fixes the problem for me:

```
diff --git a/symtabAPI/src/emitElf.C b/symtabAPI/src/emitElf.C
index 342e04d7e..46935067d 100644
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -1438,7 +1438,7 @@ bool emitElf<ElfTypes>::createLoadableSections(Elf_Shdr *&shdr, unsigned &extraA
             newdata->d_align = 4;
             updateStrLinkShdr.push_back(newshdr);
             newshdr->sh_flags = SHF_ALLOC;
-            newshdr->sh_info = 2;
+            newshdr->sh_info = verneednum;
             updateDynamic(DT_VERNEED, newshdr->sh_addr);
         }
         else if (newSecs[i]->getRegionType() == Region::RT_SYMVERDEF) {
```

nedwill:
This field used to contain verneednum, but now is hardcoded to 2.
This changes restores the original correct behavior.

Closes #405

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -1438,7 +1438,7 @@ bool emitElf<ElfTypes>::createLoadableSections(Elf_Shdr *&shdr, unsigned &extraA
             newdata->d_align = 4;
             updateStrLinkShdr.push_back(newshdr);
             newshdr->sh_flags = SHF_ALLOC;
-            newshdr->sh_info = 2;
+            newshdr->sh_info = verneednum;
             updateDynamic(DT_VERNEED, newshdr->sh_addr);
         }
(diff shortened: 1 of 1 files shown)


sashanicolas:
Ok to test

PR #443, opened 2018-03-24
added BPatch_arithExpr(BPatch_xor for intel

vanhauser-thc:
Hi,

I needed BPatch_xor for arithExpr so I patch dyninst to support it.
However it is only in intel, not in ppc as  I was unsure of the instructions there.

I hope my patch covers everything necessary, if not please comment.

### DIFF
--- a/dyninstAPI/h/BPatch_snippet.h
+++ b/dyninstAPI/h/BPatch_snippet.h
@@ -92,6 +92,7 @@ typedef enum {
     BPatch_ref,
     BPatch_fieldref,
     BPatch_seq,
+    BPatch_xor,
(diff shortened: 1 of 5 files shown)


sashanicolas:
@vanhauser-thc I'm writing a test for this. Did you mean BPatch_xor to be a logical operation, that's why you didn't use BPatch_bit_xor (bitwise)? 

With the follwing behavior:
non-zero ^ non-zero = 0
non-zero ^ zero = 1
zero ^ non-zero = 1
zero ^ zero = 0

sashanicolas:
Cherry picked into master.

PR #445, opened 2018-03-28
Remove unused rpc/xdr references.

stanfordcox:
Remove unused rpc/xdr references. The various xdr wrapper functions in linuxHeaders.h, freebsdHeaders.h, ntHeaders.h, and vxworksHeaders.h are unused. glibc is removing rpc; which is described here:
 [https://fedoraproject.org/wiki/Changes/SunRPCRemoval](url)
So it is a good opportunity to remove this dead code. Tested on fedora 27 and fedora 29. (I don't have access to vxworks, nt, or freebsd)


wrwilliams:
ok to test

PR #451, opened 2018-04-12
Modification to x86 emitter to support SW out of inst frames.

bwelton:
This patch contains changes to the emitter to support first party stackwalking out of inst's frames. For inst frames to be walkable, instrimentation frames must be enabled.

The following changes were made to support this:

- Insertion of the previous SP into the stack at a known location (2 slots above the FP).
- Insertion of a special word into the stack such that stackwalker can easily identify that this frame is an inst frame (and to select the correct walker). BEEFDEAD is located at +1 slot away from FP.
- The poping of these values from the stack at frame teardown.
- Fixes to the emitter to use scratch registers where hard coded ones were present before.

Total performance impact of this patch is an extra 5 instruction. 3 on frame creation, 2 on destruction.


bwelton:
Commit 5f5358637619cf6b964cffc12e6d76f673ec5be6 closes this issue.

PR #452, opened 2018-04-12
New walker to walk out of Instrimentation Frames FP

bwelton:
This patch contains a new walker that can walk out of inst frames in first party stackwalking mode.

For this to work, the emitter fixes located in pull request #451 must be applied. The walker itself creates
a Stackwalker frame based on the information saved by the emitter.


sashanicolas:
@bwelton you know have permission to merge.

Issue #455, opened 2018-04-17
InsertSnippet does not check if "when" parameter is legal


bwelton:
Currently when attempting to insert a snippet at an illegal location (such as inserting a snippet after a jump expression), insertsnippet will return a non-null value and attempt to insert the snippet at insertion set completion. The result is a crash in relocation since you obviously cannot do something after a jump instruction in the same basic block. 

InsertSnippet should return a NULL value when an illegal "when" condition is detected and not include the snippet in the insertion set.

sashanicolas:
Closing cause commit a58103f fixed it.

PR #460, opened 2018-05-01
Fix for crashing on relocating at unistrumentable points

bwelton:
This fix is to address issue #455. When an unistrimentable point is selected for instrimenation
the behaviour Dyninst will now exhibit is the following:

1. InsertSnippet will return NULL if not in an insertion set and an illegal when parameter is used

2. If in an insertion set, InsertSnippet will return an object to the instrimentation. However when the set
   is closed the insertion of that point will fail. It will either fail and continue to the next point (if
   the insertion set closure is set to continue past errors) or roll back all prior insertions.
.


bwelton:
Can be merged as-is.

bwelton:
Do not see build issues with this, mergable as is.

Issue #461, opened 2018-05-04
Instrumentation blocks not saving/restoring correct registers.


pstach:
The merged fix for issue #434 (pull req #437) does not resolve the issue of register save/restores on x86-64.

Reproduction details:
1. Modifications are required to the codeCoverage example due to other issues:
   - binary parser library incorrectly reports main executable as shared library
   - skipLibraries needs to contain libInst.C
  A diff is included in the attached zip.
2. Build dyninst off of master, as of today (May 4th)
3. Run "./codeCoverage -sbp /bin/gzip gzip.ins" where gzip is from the ArchLinux gzip-1.9-1 package and has a md5sum of 19eed64fe6cfb8324d1a7a3f9b3dabe3.
4
[repro.zip](https://github.com/dyninst/dyninst/files/1975755/repro.zip)
. run gzip.ins in your debugger of choice.

The outputted binary contains multiple places where rdi is not saved and restored as reported in #434.  Additionally there are a few locations, including the crashing one, where rax is not correctly restored.

From the gzip binary:
objdump -SDrt /bin/gzip
... snip ...
    f436:       49 c1 e5 05             shl    $0x5,%r13 **// start of first example block**
    f43a:       4a 8d 04 29             lea    (%rcx,%r13,1),%rax
    f43e:       8b 10                   mov    (%rax),%edx
    f440:       85 d2                   test   %edx,%edx
    f442:       0f 85 ce 05 00 00       jne    fa16 <stdout@@GLIBC_2.2.5-0x2059e
a>
    f448:       44 89 18                mov    %r11d,(%rax) **// start of second example block**

First example block (offset f436 in original binary):
gdb) x/58i 0x5555559f1930
   0x5555559f1930:      lea    -0xa8(%rsp),%rsp
   0x5555559f1938:      mov    %rax,0x20(%rsp)
   0x5555559f193d:      lea    0xa8(%rsp),%rax
   0x5555559f1945:      and    $0xffffffffffffffe0,%rsp
   0x5555559f1949:      mov    %rax,(%rsp)
   0x5555559f194d:      mov    -0x88(%rax),%rax
   0x5555559f1954:      push   %r10
   0x5555559f1956:      push   %r11
   0x5555559f1958:      push   %r8
   0x5555559f195a:      push   %r9
   0x5555559f195c:      push   %rcx
   0x5555559f195d:      push   %rsi
   0x5555559f195e:      movabs $0xf436,%rax
   0x5555559f1968:      push   %rax
[...]

PR #462, opened 2018-05-08
register index and size fix for modrm where mod=0 and not SIB or disp

pstach:
Tested on a handful of binaries, appears to fix the register decoding problem mentioned in #461


mxz297:
Fixed by https://github.com/dyninst/dyninst/commit/b8369b5627eec1f7c7ab8ef461dae0b63e20b23d

PR #467, opened 2018-06-01
avoid infinite loop on bad DWARF CFI record

jmellorcrummey:
This patch avoids an infinite loop by returning false when dwarf_next_cfi returns -1 and doesn't advance the offset. The patch addresses only the infinite loop symptom and doesn't attempt to explore its cause.

Since there were other commits related to this code between the commit mentioned above and the latest commit by @sashanicolas , either of the following cases seem possible

- the CFI information was always bad, but the code never encountered it all, or
- the reader got corrupted with recent changes to that good CFI info looks bad.

I did not investigate which case it is. Since Sasha was working here, I am sure that he knows more about this than me and is better equipped to address deeper semantic issues that may need attention.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -4302,8 +4302,6 @@ public:
 
 void Object::parseLineInfoForCU(Dwarf_Die cuDIE, LineInformation* li_for_module)
 {
-    std::vector<open_statement> open_statements;
-    
     /* Acquire this CU's source lines. */
     Dwarf_Lines * lineBuffer;
     size_t lineCount;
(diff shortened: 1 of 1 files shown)


jmellorcrummey:
Commit 01414a8c688336ee07a737e2def383c0023521b0 adds handling for an empty DebugSectionMap, which I ran into with some trivial cubin files that are essentially empty (no text segments). After the patch, if Object::convertDebugOffset finds an empty DebugSectionMap, it marks it sorted and returns rather than entering the loop and dereferencing an non-existent entry, which causes a SEGV.

 A compressed tar file with examples of such empty cubins can be found here: https://rice.box.com/s/3533pgcb6q6cv4bgtwohxpvqx77svkyr

sashanicolas:
@jmellorcrummey commit ffcea6a will be ignored because Elfutils 1.7.3 fixed the infinite loop. I tested it with the file you linked: https://rice.box.com/s/3533pgcb6q6cv4bgtwohxpvqx77svkyr.

I will eventually test the other commits, but they are not related to the infinite loop issue.

sashanicolas:
these were rebased with the #472

PR #469, opened 2018-06-19
symtab: fix memory error in Statement::getFile

rafzi:
When used through BPatch_statement::filename() or various other internal code, the function Statement::getFile was causing memory errors.

This is because it returned a new string object that contains a copy of the string in the string map. If this copy is then just used to call .c_str and discarded, the returned pointer is dangling at the caller.


sashanicolas:
ok to test

sashanicolas:
I see what you're saying, @rafzi. This is an old problem, we need to remove uses of char* from the interface and do only std::string.

PR #472, opened 2018-06-30
Read line map for binaries that lack a .debug_info section

jmellorcrummey:
This pull request adds new code that relies on dwarf_next_lines from elfutils 0.173 to read line maps for binaries that lack a .debug_info section. Optimized CUBIN binaries are of this form. 

In this commit, there is some code duplicated between parseLineMapInfoForCU and a new parseLineMapInfo function. This could be coalesced with refactoring, but I didn't want to refactor Dyninst until you were willing to accept this functionality.


sashanicolas:
I rebased these.

PR #478, opened 2018-08-23
Arm64 cray fixes

mxz297:
This branch contains ARM 64 stackwalking fixes for Cray. The branch was based on v9.3.x and we should merge the fixes into master.


mxz297:
Substituted by #559

Issue #436, opened 2018-03-03
processCreate - immediate crash


vanhauser-thc:
I have a very simple example program:

```
#include <stdio.h>
#include "BPatch.h"
#include "BPatch_addressSpace.h"
#include "BPatch_process.h"
#include "BPatch_binaryEdit.h"
#include "BPatch_point.h"
#include "BPatch_function.h"

using namespace std;
using namespace Dyninst;

BPatch bpatch;

int main(int argc, char *argv[]) {
  char *exec = argv[1], *env = getenv("DYNINSTAPI_RT_LIB");

  if (argc == 1 || (argc == 2 && strcmp(argv[1], "-h") == 0)) {
    printf("Syntax: %s cmd arg1 arg2 ... \n");
    exit(0);
  }

  if (env == NULL) {
    fprintf(stderr, "Error: DYNINSTAPI_RT_LIB environment not set\n");
    exit(-1);
  } else
    printf("DYNINSTAPI_RT_LIB=%s\n", env);

  argv++;
  BPatch_process *app = bpatch.processCreate( (const char *)exec, (const char **)argv); // segfault
  sleep(1);
  printf("app %p\n", (char*)app);

  return 0;
}
```

When run:
```
# ./test /bin/echo foo
DYNINSTAPI_RT_LIB=/usr/local/lib/libdyninstAPI_RT.so
foo
Segmentation fault
#
```

valgrind reports the crash here:
```
==5499== Process terminating with default action of signal 11 (SIGSEGV)
==5499==  Bad permissions for mapped region at address 0x4112A70
==5499==    at 0x85A9705: elf_compress (in /usr/lib/x86_64-linux-gnu/libelf-0.170.so)
==5499==    by 0x87B8DFB: ??? (in /usr/lib/x86_64-linux-gnu/libdw-0.170.so)
==5499==    by 0x87B919F: dwarf_begin_elf (in /usr/lib/x86_64-linux-gnu/libdw-0.170.so)
==5499==    by 0x7F69656: Dyninst::DwarfDyninst::DwarfHandle::init_dbg() (dwarfHandle.C:115)
==5499==    by 0x7F69918: Dyninst::DwarfDyninst::DwarfHandle::type_dbg() (dwarfHandle.C:223)
==5499==    by 0x6DC6652: Dyninst::SymtabAPI::Object::fix_global_symbol_modules_static_dwarf() (Object-elf.C:2506)
[...]

PR #487, opened 2018-09-10
Open debug file in read/write mode to get a writable mmap.

cspiel:
Write privilege is necessary for newer elfutils because
function elf_compress() transparently uncompresses data
and thus has to rewrite the ELF header with the new data
size for consistency.

This patch fixes dyninst issue #436,
"processCreate - immediate crash".

### DIFF
--- a/elf/src/Elf_X.C
+++ b/elf/src/Elf_X.C
@@ -1627,11 +1627,11 @@ static bool loadDebugFileFromDisk(string name, char* &output_buffer, unsigned lo
       return false;
    if (S_ISDIR(fileStat.st_mode))
       return false;
-   int fd = open(name.c_str(), O_RDONLY);
+   int fd = open(name.c_str(), O_RDWR);
    if (fd == -1)
       return false;
 
-   char *buffer = (char *) mmap(NULL, fileStat.st_size, PROT_READ, MAP_SHARED, fd, 0);
+   char *buffer = (char *) mmap(NULL, fileStat.st_size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
    close(fd);
    if (!buffer)
       return false;


sashanicolas:
@cspiel when you don't have permission to open as WRITE a file you compromise the execution.

cspiel:
@sashanicolas: Admittedly I am having trouble to understand your
statement

> when you don't have permission to open as WRITE
> a file you compromise the execution.

on a syntactical as well as on a semantical level.  So, let's
see what I've figured out...

1. In function `loadDebugFileFromDisk` we load the *debug* information,
*not* the executable part.  No code gets executed, and obviously, the
`PROT_EXEC` bit is not set anywhere in the function.

2. Even if `PROT_WRITE` is set for a *code* section, which not the case
we are talking about here, this does not change the program flow
inside this very section in any way.

3. Newer elfutils will want write to the debug section as indicated in my
addition to issue report #436.

Am I correct in assuming that you neither have reproduced van Hauser's
segfault nor tried out the patch?

sashanicolas:
Hi, @cspiel I agree now that my statement isn't clear. What I meant was:

Suppose a user running Dyninst is trying to open the file in:
    `int fd = open(name.c_str(), O_RDWR);`
(and here the filename is given as the first parameter `name.c_str()`)
This user doesn't have permission to write, only to read. The function `open` will return -1.

Now in the points you "figured out":
1. Wrong, we don't load just debug info, but the whole file. There's no such thing as load only one part of a file.
2. We don't get to this point because the `open` already failed and we return false.
3. This has nothing to do with opening a file you don't have permission to modify.

Therefore, you're not correct. 

I understand your solution might solve your specific problem. But it seems that you did not understand what `loadDebugFileFromDisk` does.

This is about separate debug info file. And a simple case where your changes cause a huge disaster is to remove your permission to write to the debug file related to your mutatee.

I did not get a segfault with the code that @vanhauser-thc provided.

hainest:
Can't reproduce.

Issue #489, opened 2018-09-21
Segfault (nullptr deref) in getCalledFunction()


cspiel:
My applications run into segfaults with code like
```c++
if (std::vector<BPatch_point*>* points = patch_function->findPoint(BPatch_allLocations))
{
    for (const auto point : *points)
    {
        if (point != nullptr && point->getPointType() == BPatch_subroutine)
        {
            if (BPatch_function* called_function = point->getCalledFunction())
            {
                // ...
            }
        }
    }
}
```
The problem is that `BPatch_point::getCalledFunction()` dereferences `_func` *before* the pointer was checked for `NULL`.  Reversed software-engineering!  :stuck_out_tongue_winking_eye:

The fix is simple, just reverse the order:
```diff
diff --git a/dyninstAPI/src/BPatch_point.C b/dyninstAPI/src/BPatch_point.C
index f616ce77e..6d5e28944 100644
--- a/dyninstAPI/src/BPatch_point.C
+++ b/dyninstAPI/src/BPatch_point.C
@@ -220,16 +220,17 @@ BPatch_function *BPatch_point::getCalledFunction()
            return NULL;
    }
 
+   assert(point->block());
    func_instance *_func = point->block()->callee();
+   if (!_func) {
+       parsing_printf("findCallee failed in getCalledFunction\n");
+           return NULL;
+   }
    if (_func->getPowerPreambleFunc() != NULL) {
        func_instance * preambleFunc = _func->getPowerPreambleFunc();
        return addSpace->findOrCreateBPFunc(preambleFunc, NULL);
    }
 
-   if (!_func) {
-       parsing_printf("findCallee failed in getCalledFunction\n");
-           return NULL;
-   }
    return addSpace->findOrCreateBPFunc(_func, NULL);
 }
```
The additional `assert()` was inspired by `BPatch_point::getCalledFunctionName()`.

Branch with fix to follow...

mxz297:
A similar fix has been committed. Dyninst 10.0.0 and master will not have this problem.

Thanks for your patches.

PR #490, opened 2018-09-21
Fix segfault in BPatch_point::getCalledFunction().

cspiel:
In function BPatch_point::getCalledFunction() move the test for a
nullptr-valued _func in front of the first dereference of the pointer.

Add an assert() in analogy to BPatch_point::getCalledFunctionName()
to guard against further surprises.

This fixes issue #489.


mxz297:
This issue has been fixed by 0417da02415ee7bacd138414b20ede72723b6a09

Issue #438, opened 2018-03-11
xdrrec_create() type cast error: char* vs. void*


afaerber:
I am trying to package dyninst 9.3.2 for openSUSE Tumbleweed and running into this build error:

```
[   69s] In file included from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/headers.h:56:0,
[   69s]                  from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/string-regex.C:34:
[   69s] /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/linuxHeaders.h: In function 'void P_xdrrec_create(XDR*, u_int, u_int, caddr_t, xdr_rd_func, xdr_wr_func)':
[   69s] /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/linuxHeaders.h:284:45: error: invalid conversion from 'int (*)(char*, char*, int)' to 'int (*)(void*, void*, int)' [-fpermissive]
[   69s]    xdrrec_create(x, send_sz, rec_sz, handle, (int(*)(char*, char*, int))read_r, (int(*)(char*, char*, int))write_f);}
[   69s]                                            [   61.648111] serial8250: too much work for irq4
[   69s]   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[   69s] In file included from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/build/common/cotire/common_CXX_prefix.cxx:30:0,
[   69s]                  from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/build/common/cotire/common_CXX_prefix.hxx:4:
[   69s] /usr/include/rpc/xdr.h:361:15: note:   initializing argument 5 of 'void xdrrec_create(XDR*, u_int, u_int, void*, int (*)(void*, void*, int), int (*)(void*, void*, int))'
[   69s]  extern void   xdrrec_create(XDR *, u_int, u_int, void *,
[   69s]                ^~~~~~~~~~~~~
[   69s] In file included from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/headers.h:56:0,
[   69s]                  from /home/abuild/rpmbuild/BUILD/dyninst-9.3.2/common/src/string-regex.C:34:
[...]

PR #496, opened 2018-10-17
Build fixes for parallel building and xdr-related issues

### LINKED ISSUE #486: Build fails on Arch Linux
I'm trying to build hpctoolkit-externals on the current master branch
(ab50b065ada2545938abae2...), and I am getting a failed dependency when
building "symtabAPI":

> make[6]: Entering directory
> '.../build/hpctoolkit-externals/BUILD/symtabAPI/work'
> [  0%] Building CXX object common/CMakeFiles/common.dir/src/Ident.C.o
> In file included from
>.../build/hpctoolkit-externals/BUILD/symtabAPI/dyninst/common/src/headers.h:56,
>                  from
>.../build/hpctoolkit-externals/BUILD/symtabAPI/dyninst/common/src/Ident.h:47,
>                  from
>.../build/hpctoolkit-externals/BUILD/symtabAPI/dyninst/common/src/Ident.C:36:
> .../build/hpctoolkit-externals/BUILD/symtabAPI/dyninst/common/src/linuxHeaders.h:62:10:
> fatal error: rpc/types.h: No such file or directory
>  #include <rpc/types.h>

I am using Arch Linux and the "types.h" is located here:

> /usr/include/tirpc/rpc/types.h

What would be the cleanest way to fix this?

I have tried modifying the <rpc/types.h> into <tirpc/rpc/types.h> in
"linuxHeaders.h", but then it fails at the system level:

/usr/include/tirpc/rpc/types.h:93:10: fatal error: netconfig.h: No such
file or directory
 #include <netconfig.h>

LER0ever:
This PR contains various fixes for the Dyninst build system. 

#### Parallel Building
- Fixes parallel building errors due to missing dependency declarations for some sub libraries.
- Boost is now added as universal dependency for all dyninst_library's

#### xdr-related issues
- Fixes building on Linux distros with newer glibc (like Archlinux, OpenSUSE Tumbleweed, Solus, etc).
- Deletes all the xdr-related functions and definitions (they are dead code)

#### Misc
- Fixes cmake error while running configure as normal user by moving the COPY instructions to the install phase
- Set soft limits for libelf versions (>= 0.173), since we need `dwarf_next_lines` function which was introduced in that version
- Set the binutils version to be the latest, as it is needed to recognize Fedora 25 on AArch64

* * *

This would therefore resolve #486, #438


bwelton:
I will be able to review this on Friday if mxz297 doesn't get to it first. 

Can you post a follow up on the issues this fixes so that the reporters can try these changes out?

PR #506, opened 2018-11-30
Install docs into target doc dirs

stanfordcox:
Install  the pdf doc files into the correct target doc directories.


mxz297:
This looks good. Thanks for the fix!

PR #507, opened 2018-11-30
Add .dir suffix to examples/{codeCoverage,unstrip}

stanfordcox:
The Fedora build system does in tree builds as part of the release process.  This change renames examples/codeCoverage to examples/codeCoverage.dir and examples/unstrip to examples/unstrip.dir (using the .dir suffix convention used in the dyninst testsuite)


mxz297:
We almost never build Dyninst in source, so thanks for catching this error and fix it!

PR #513, opened 2018-12-07
as needed, add -fopenmp flag when linking executables

jmellorcrummey:
When USE_OpenMP=on, dyninst executables (parseThat and executables in examples) don't link without using -fopenmp as a link flag. Add this in the appropriate CMakeLists.txt files.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -21,6 +21,13 @@ add_dependencies(cfg_to_dot parseAPI symtabAPI instructionAPI common dynDwarf dy
 target_link_libraries(cfg_to_dot parseAPI symtabAPI instructionAPI common dynDwarf dynElf ${Boost_LIBRARIES})
 #add_executable(retee)
 
+if (USE_OpenMP MATCHES "ON")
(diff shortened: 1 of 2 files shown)


jmellorcrummey:
@krentel this is why dyninst wasn't linking for me.

mwkrentel:
I was going to suggest changing USE_OpenMP to a true cmake boolean.
The thing about ``(USE_OpenMP MATCHES "ON")`` is that it's an exact
string match, so On, True, Yes all mean No.

mwkrentel:
@mxz297  We got to the bottom of John's linking problem.  John had a
stray symlink in his compiler module.  This made it look like -fopenmp
was the problem (but it wasn't).

Actually, there was no bug here.  This patch is unnecessary but
harmless.  (If this was a bug, then Dyninst would never build with
openmp, which isn't the case.)

You don't need to revert the patch.  All it does is add a dependency
that it will eventually pick up by transitivity, which is harmless.

At some point in the future, parseThat probably will use openmp
directly.  When that happens, then the compiler flag will be
necessary.  But until then, it's not necessary (but harmless).

PR #512, opened 2018-12-06
Adjust Dyninst to support injecting CFGs for CUBINs 

jmellorcrummey:

PR #515, opened 2018-12-08
Adjust Dyninst to support injecting CFGs for CUBINs

Jokeren:


Jokeren:
We would like to abandon the previous PR #512 that has many duplicated commits. We will also open another PR to force reading '.debug_line' section for cubin binaries.

mxz297:
@Jokeren I am going over this PR and have a few questions:

1. Since this PR was created before Dyninst 10.0.0, have you upgrade your Dyninst to 10.0.0 or after and check whether the patches in this PR still works?

2. Why do you need to make Block::getInsns(Block::Insns) as a virtual function? I thought you did not use InstructionAPI to do instruction decoding, so you would not need InstructionAPI::Instruction.

3. Why do you need to add create a new StandardParseData object (in parseAPI/src/Parser.C, line 108)? I thought you needed to avoid any ParseAPI::Parser operations, so I thought you would not need to change the Parser.

mxz297:
@Jokeren I am sorry that I mistakenly thought the PR was created before Dyninst 10.0.0. You can ignore my first question ...

Jokeren:
2. Why do you need to make Block::getInsns(Block::Insns) as a virtual function? I thought you did not use InstructionAPI to do instruction decoding, so you would not need InstructionAPI::Instruction.

The answer for this question is straightforward. Because Mark uses `block->getInsns(imap);` to fetch all the instructions:

https://github.com/HPCToolkit/hpctoolkit/blob/master/src/lib/banal/Struct.cpp#L1664

In order to keep our parsing logics, I made `getInsns` virtual and fed pseudo instructions into a instruction map.

3. Why do you need to add create a new StandardParseData object (in parseAPI/src/Parser.C, line 108)? I thought you needed to avoid any ParseAPI::Parser operations, so I thought you would not need to change the Parser.

This is because  `_parse_data` cannot be `NULL`. 

At first, we need to construct a CodeObject object. The constructor of CodeObject invokes `process_hints`

https://github.com/dyninst/dyninst/blob/master/parseAPI/src/CodeObject.C#L76

`process_hints` adds function entries by invoking `parser->factory()._mkfunc`. If a function entry exists, it will call `parser->add_hint` (Line 97), which calls `_parse_data->findFunc`:

https://github.com/dyninst/dyninst/blob/master/parseAPI/src/Parser.C#L158

If we leave `_parse_data` to be `NULL`, there will be an error.

mxz297:
@Jokeren For ```Block::getInsns(Block::Insns)```, you said that you fed pseudo instructions. Does it mean that you created InstructionAPI::Instruction objects yourself, and you fill in the fields in those objects?

mxz297:
@Jokeren For the line of ```_parse_data```, while I understand that without it, your code would crash. But for other cases, if the user passes in an empty CodeSource, we would still create an empty ```parse_data``` object. 

I feel it would be confusing to create a parse_data object when we are returning from an error case. So, personally, I think it is better to think about alternatives for your code.

Jokeren:
> @Jokeren For `Block::getInsns(Block::Insns)`, you said that you fed pseudo instructions. Does it mean that you created InstructionAPI::Instruction objects yourself, and you fill in the fields in those objects?

Yes.

https://github.com/HPCToolkit/hpctoolkit/blob/ompt-device-cfg/src/lib/banal/cuda/CudaBlock.cpp#L23

Jokeren:
> @Jokeren For the line of `_parse_data`, while I understand that without it, your code would crash. But for other cases, if the user passes in an empty CodeSource, we would still create an empty `parse_data` object.
> 
> I feel it would be confusing to create a parse_data object when we are returning from an error case. So, personally, I think it is better to think about alternatives for your code.

Add another constructor for `CodeObject` is hard.

        PARSER_EXPORT CodeObject(CodeSource * cs, 
                             CFGFactory * fact = NULL,                                                                                                                 
                             ParseCallback * cb = NULL,
                             bool defensiveMode = false);

        PARSER_EXPORT CodeObject(CodeSource * cs, 
                             CFGFactory * fact);

The arguments needed for `CudaCodeObject` are `CodeSource` and `CFGFactory`. However, because the original `CodeObject` constructor sets the default values of fact, cb, and defensiveMode to be `NULL`, there is no way to distinguish two constructors above if the given fact is not `NULL`.

mxz297:
@Jokeren I am thinking maybe you don't even need a CudaCodeSource or CudaCodeObject. Based on my understanding, Mark actually only use CodeObject to get funclist. It is easy for you to construct a funclist directly (CodeObject::funclist is public). Then mark only need to slightly change his code.

Jokeren:
> @Jokeren I am thinking maybe you don't even need a CudaCodeSource or CudaCodeObject. Based on my understanding, Mark actually only use CodeObject to get funclist. It is easy for you to construct a funclist directly (CodeObject::funclist is public). Then mark only need to slightly change his code.

Thanks! That seems a good idea. After discussing with John, I will submit my changes.

PR #517, opened 2018-12-10
Use PRIx64 macros for long long results for i686.

stanfordcox:
Fix a problem on i686 and other 32 bit machines by using PRIx64 for long long.  The other snprintf could be likewise changed to use PRIxNN but long long is the only problematic one..

### DIFF
--- a/instructionAPI/h/Result.h
+++ b/instructionAPI/h/Result.h
@@ -449,10 +449,10 @@ namespace Dyninst
 	    snprintf(hex, 20, "%x", val.s32val);
 	    break;
 	  case u64:
-	    snprintf(hex, 20, "%lx", val.u64val);
+	    snprintf(hex, 20, "%" PRIx64, val.u64val);
 	    break;
 	  case s64:
-	    snprintf(hex, 20, "%lx", val.s64val);
+	    snprintf(hex, 20, "%" PRIx64, val.s64val);
 	    break;
 	  case sp_float:
 	    snprintf(hex, 20, "%f", val.floatval);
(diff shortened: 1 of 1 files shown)


stanfordcox:
I see this failure
CMake Error at examples/CMakeLists.txt:14 (add_dependencies):
  The dependency target "dynDwarf" of target "codeCoverage" does not exist.
CMake Error at examples/CMakeLists.txt:14 (add_dependencies):
  The dependency target "dynElf" of target "codeCoverage" does not exist.

perhaps CMakeLists.txt needs:

if (UNIX)
set (DEPS ${DEPS} dynDwarf dynElf)
endif()

mxz297:
@hainest @sashanicolas We probably should consider merge this one. Red hat internal testing system needs Dyninst to compile on 32-bit platforms. They do not really need Dyninst to work on 32-bit platforms.

mxz297:
Redhat records build log at  https://koji.fedoraproject.org/koji/

The problem is that if the build for one architecture fails, it will stop building other architectures. That's why Red hat would like us to fix build issue for i686.

hainest:
We have no way of testing this. Can someone at Red Hat confirm this patch is still effective for 10.1?

mxz297:
@hainest Let's discuss them later today. I just write things down for notes.

stanfordcox:
I can verify the linux i386 bits but they should be fine as I built the latest Fedora/RHEL release with them.  The windows bits are fine I think; the patch just does  define  PRIx64  "lx"
But I cannot locally verify as I don't have a windows build environment.

PR #518, opened 2018-12-10
Don't use system header <> syntax for dyninst includes

stanfordcox:
Fedora installs dyninst includes into /usr/include/dyninst; thus any file included from within another dyninst include will not be found if the <HEADER> syntax is used since that syntax only searches system includes and /usr/include/dyninst is not a standard system include.  Using "HEADER" syntax (search same containing directory solves the problem.  There are only two cases of this so I'm guess it was an accident as opposed to an intentional decision.

### DIFF
--- a/instructionAPI/h/ArchSpecificFormatters.h
+++ b/instructionAPI/h/ArchSpecificFormatters.h
@@ -34,7 +34,7 @@
 #include <map>
 #include <vector>
 #include <string>
-#include <dyn_regs.h>
+#include "dyn_regs.h"
 
 namespace Dyninst {
     namespace InstructionAPI {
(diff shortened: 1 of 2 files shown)


stanfordcox:
I see this failure
CMake Error at examples/CMakeLists.txt:14 (add_dependencies):
  The dependency target "dynDwarf" of target "codeCoverage" does not exist.
CMake Error at examples/CMakeLists.txt:14 (add_dependencies):
  The dependency target "dynElf" of target "codeCoverage" does not exist.

perhaps CMakeLists.txt needs:

if (UNIX)
set (DEPS ${DEPS} dynDwarf dynElf)
endif()

PR #527, opened 2019-01-07
Fix several bugs in common/src/pathName.C

hainest:
This fixes several bugs, adds a couple new features, and does a bit of cleanup. 70c9c93 contains a set of unit tests based on Catch2, and is not intended to be merged. It is here for discussion only. I will rebase it out when this is ready to be merged.

FIXED:
1. Tilde expansion is done _before_ the existence check.
	Since `stat` does not do shell expansions, resolve_file_path("~")
	always returned an empty string instead of the expanded path.

2. Tilde detection logic was inverted

3. Remove possible buffer overflow when making path relative to CWD

ADDED features:
1. Removes leading and trailing whitespace

2. Add basic support for Windows file system by using Boost::filesystem
	Expansion of Windows-specific shell variables, e.g. %HOME%, is _not_
	supported.

3. Leverage move semantics to reduce copying
	This reduces the maximum number of string copies from 5 to 3. When
	resolve_file_path is invoked with an r-value, there are only two
	copies.

### DIFF
--- a/common/src/addrtranslate-linux.C
+++ b/common/src/addrtranslate-linux.C
@@ -189,7 +189,7 @@ string AddressTranslateSysV::getExecName()
    if (exec_name.empty()) {
       char name[64];
       snprintf(name, 64, "/proc/%d/exe", pid);
-      exec_name = resolve_file_path(name);
+      exec_name = std::move(resolve_file_path(name));
    }
    return exec_name;
 }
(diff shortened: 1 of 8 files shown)


hainest:
Passing testsuite (modulo known errors).

hainest:
@mxz297 @sashanicolas @bwelton Although not a primary target, this PR introduces new behavior for Windows. Previously, `resolve_file_path` would emit a runtime assert (assuming NDEBUG was disabled). Now, it will successfully parse paths, but will not expand Windows-specific environment variables (e.g., %HOME%). This certainly won't break existing workflows, but the only documentation is the comments in the code.

Does anyone object to merging this?

mxz297:
@hainest Looking good to me.

PR #530, opened 2019-01-22
Adjust Dyninst to support injecting CFGs for CUBINs

Jokeren:
Hi @mxz297 , I talked with John today. We thought that an "CodeSource provides no CodeRegions" doesn't indicate that we cannot use `CodeObject` and `CodeSource`'s interfaces to parse external sources, which is exactly we have done on the cuda code.

The changes in the newest commit perhaps makes the logic satisfy your requirement. If `_parse_state` is not unparsable, then we iterate every code region to check overlapped conditions and create a `OverlappingParseData` object. Otherwise, we create a `StandardParseData`.

### DIFF
--- a/parseAPI/h/CFG.h
+++ b/parseAPI/h/CFG.h
@@ -341,7 +341,7 @@ public:
     void getFuncs(std::vector<Function *> & funcs);
     template<class OutputIterator> void getFuncs(OutputIterator result); 
 
-    virtual getInsns(Insns &insns) const;
+    virtual void getInsns(Insns &insns) const;
     InstructionAPI::Instruction getInsn(Offset o) const;
 
     bool wasUserAdded() const;
--- a/parseAPI/h/CFG.h
+++ b/parseAPI/h/CFG.h
@@ -314,7 +314,7 @@ public:
     inline Address start() const { return _start; }
     inline Address end() const { return _end; }
     inline Address lastInsnAddr() const {  return _lastInsn; }
-    inline Address last() const {  return lastInsnAddr(); }
+    virtual Address last() const {  return lastInsnAddr(); }
     inline Address size() const {  return _end - _start; }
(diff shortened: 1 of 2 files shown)


parseAPI/h/CFG.h
```diff
@@ -341,7 +341,7 @@ class PARSER_EXPORT Block :
     void getFuncs(std::vector<Function *> & funcs);
     template<class OutputIterator> void getFuncs(OutputIterator result); 
 
-    void getInsns(Insns &insns) const;
+    virtual getInsns(Insns &insns) const;
```
hainest: Compile error: the return type is missing.

Jokeren: Fixed. Thanks!

hainest:
Passing builds and test suite. @mxz297 Can I merge?

mxz297:
@hainest Go ahead and merge it. Thanks!

mxz297:
@Jokeren John has mentioned that this PR will not work well with your tool due to lack of accesses to an edge map. Meanwhile, due to incomplete test, we merged this PR. However, I found that this PR will cause Dyninst to crash when rewriting statically linked binaries. 

’‘’
StandardParseData::StandardParseData(Parser *p) :
     ParseData(p), _rdata(&p->obj(), p->obj().cs()->regions()[0])
 { }
‘’‘

As you can see, constructing a StandardParseData object assumes the CodeSource has code regions. So, I will temporarily revert your this change by still adding a return in the Parser constructor when there is no code region.

@hainest It looks like then it will be crucial for you to install statically linking environment on your machine.

Jokeren:
@mxz297 

That's fine. For this problem, I can push an empty region into the parse data to work around, or construct code regions externally.

But due to Dyninst 10's API changes, I came across other problems. I and John will try to figure out a solution that has as little impact as possible on Dyninst.

PR #536, opened 2019-01-23
Remove undefined behavior from bit-shifting code

hainest:
Overflow of signed integer arithmetic (including shifting) is
undefined behavior in C++. Compilers are allowed to exploit
this UB for optimization purposes up to and including not
emitting any opcodes.

These were found using cppcheck.


hainest:
Do not merge until https://github.com/dyninst/dyninst/pull/543. Will need a rebase.

PR #541, opened 2019-01-24
Add some explicit casts from boost::tribool to bool for some code in

mwkrentel:
dataflowAPI/rose/util/Message.[Ch].  This was breaking the build when
using boost >= 1.69.

----------

The places where this fails is bizarre.  For example, this fails with
the message, "cannot convert 'boost::logic::tribool' to 'bool'".

```
boost::tribool tb;
bool b = tb;
```

However, adding an explict cast succeeds.  WTF!?

```
boost::tribool tb;
bool b = (bool) tb;
```

Also, g++ seems to be ok with an implicit conversion inside if().
(Again, WTF!?)  But I added an explicit cast for that case just to be
safe.

```
-      if (cs.bold)
+      if ((bool) cs.bold)
```

I think the above diff is not strictly necessary.  But it can't hurt
and it protects against some later compiler being more strict and
breaking this usage.

Also, from looking at the code in boost/logic, tribool certainly tries
to provide an automatic conversion from tribool to bool.  So, I'm not
sure why any of this fails in the first place.  I suspect a flaw in
the tribool implementation, but we're a bit outside of my C++
knowledge here.

Anyway, all the casts are harmless, so it's an easy and safe fix.
"Safe and effective," just like the label says.  :-)

I plan to submit a PR to fix this for 10.0.0 in spack.

I checked the Sawyer repo on github (matzke1).  They have no mention
of this problem, but it clearly will affect them too.  I'll alert them
after I attend to spack.

### DIFF
--- a/dataflowAPI/rose/util/Message.C
+++ b/dataflowAPI/rose/util/Message.C
@@ -212,7 +212,7 @@ MesgProps::print(std::ostream &o) const {
 
     o <<", isBuffered=";
     if (!indeterminate(isBuffered)) {
-        o <<(((bool) isBuffered) ? "yes" : "no");
+        o <<(isBuffered ? "yes" : "no");
     } else {
         o <<"undef";
     }
@@ -247,7 +247,7 @@ MesgProps::print(std::ostream &o) const {
     
     o <<", useColor=";
     if (!indeterminate(useColor)) {
-        o <<(((bool) useColor) ? "yes" : "no");
+        o <<(useColor ? "yes" : "no");
     } else {
         o <<"undef";
     }
@@ -730,7 +730,7 @@ Prefix::toString(const Mesg &mesg, const MesgProps &props) const {
     std::string separator = "";
 
     std::string endColor;
-    if (((bool) props.useColor) && props.importance) {
+    if (props.useColor && props.importance) {
         const ColorSpec &cs = colorSet_[*props.importance];
         if (!cs.isDefault()) {
             const char *semi = "";
@@ -743,7 +743,7 @@ Prefix::toString(const Mesg &mesg, const MesgProps &props) const {
                 retval <<semi <<(40+cs.background);
                 semi = ";";
             }
-            if ((bool) cs.bold)
+            if (cs.bold)
                 retval <<semi <<"1";
             retval <<"m";
             endColor = "\033[m";
@@ -1123,7 +1123,7 @@ StreamBuf::bake() {
         destination_->bakeDestinations(message_.properties(), baked_/*out*/);
         anyUnbuffered_ = false;
         for (BakedDestinations::const_iterator bi=baked_.begin(); bi!=baked_.end() && !anyUnbuffered_; ++bi)
-            anyUnbuffered_ = (bool) !bi->second.isBuffered;
+            anyUnbuffered_ = static_cast<bool>(!bi->second.isBuffered);
         isBaked_ = true;
     }
 }
--- a/dataflowAPI/rose/util/Message.C
+++ b/dataflowAPI/rose/util/Message.C
@@ -212,7 +212,7 @@ MesgProps::print(std::ostream &o) const {
 
     o <<", isBuffered=";
     if (!indeterminate(isBuffered)) {
-        o <<(isBuffered ? "yes" : "no");
+        o <<(((bool) isBuffered) ? "yes" : "no");
     } else {
         o <<"undef";
     }
@@ -247,7 +247,7 @@ MesgProps::print(std::ostream &o) const {
     
     o <<", useColor=";
     if (!indeterminate(useColor)) {
-        o <<(useColor ? "yes" : "no");
+        o <<(((bool) useColor) ? "yes" : "no");
     } else {
         o <<"undef";
     }
@@ -730,7 +730,7 @@ Prefix::toString(const Mesg &mesg, const MesgProps &props) const {
     std::string separator = "";
 
     std::string endColor;
-    if (props.useColor && props.importance) {
+    if (((bool) props.useColor) && props.importance) {
         const ColorSpec &cs = colorSet_[*props.importance];
         if (!cs.isDefault()) {
             const char *semi = "";
@@ -743,7 +743,7 @@ Prefix::toString(const Mesg &mesg, const MesgProps &props) const {
                 retval <<semi <<(40+cs.background);
                 semi = ";";
             }
-            if (cs.bold)
+            if ((bool) cs.bold)
                 retval <<semi <<"1";
             retval <<"m";
             endColor = "\033[m";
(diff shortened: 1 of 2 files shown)


hainest:
Could you convert the C-style casts to `static_cast`? They are technically safe here, but C-style casts are harder to grock in C++ code.

---

>The places where this fails is bizarre. For example, this fails with
>the message, "cannot convert 'boost::logic::tribool' to 'bool'".
> ```
> boost::tribool tb;
> bool b = (bool) tb;
> ```

Alas, this is a vestige of C. The 'type name = value;' syntax is called _copy initialization_ in C++ and adheres to the rules of C which allow for all manner of integral type conversions (e.g., `bool b = 1200;`). C++ added _direct initialization_ because copy initialization (like its namesake) allows for copies of objects to be made (although nearly every compiler worth its salt never did). Direct initialization in C++11 forbids narrowing conversions and the use of `explicit` conversion functions. Hence, `bool b{tb};` is legal (and very much the preferred initialization syntax in C++).

> 
> Also, g++ seems to be ok with an implicit conversion inside if().

Another vestige of C, but this time without the shenanigans. The contents of an `if` (except for the optional initializer added in C++17) are _always_ assumed to be directly converted to `bool`. Other examples are `while()`, the condition term of `for(init; cond; post;)`, and pointer checks (e.g., `if(ptr)` is really `if(ptr == nullptr)`).

> I think the above diff is not strictly necessary. But it can't hurt
> and it protects against some later compiler being more strict and
> breaking this usage.

Seeing an explicit cast to `bool` inside an `if` is a code smell in C++. Leaving it out is 100% guaranteed to "do the right thing."

> Also, from looking at the code in boost/logic, tribool certainly tries
> to provide an automatic conversion from tribool to bool. So, I'm not
> sure why any of this fails in the first place. I suspect a flaw in
> the tribool implementation, but we're a bit outside of my C++
> knowledge here.

The `boost::tribool` is definitely funky here as they broke the API and ABI with this change (without mentioning it in the ChangeLog, no less!). But, I can't replicate the error even when using all sorts of macro trickery for Boost and different versions of gcc (and hence different C++ standards). I think your initial idea that there is a subtle disturbance to the Boost ecosystem somewhere in the include chain in DynInst is a plausible culprit, but I couldn't find it in my cursory search. These changes here are effectively neutral, so I think that's probably as far as we need to go at this time.

> I plan to submit a PR to fix this for 10.0.0 in spack.

Sorry, I don't follow. What needs to change in the spack package?

mwkrentel:
Ah, so my superficial understanding of C++ is exposed. :-)

Sure, I'll revert the if() cases and change the others to static_cast.
You can squash the commits to leave less of an e-trail of my mistakes,
right? :-)

This will certainly shorten things, from 6 cases to two.

As for spack, I'll add a spack PR for dyninst/package.py.  v10.0.0 is
a fixed tag which currently fails to build as is with boost 1.69.
After we've settled on a dyninst patch, it's a 1-line change to
package.py (plus the patch) to add that patch to v10.0.0 so it will
build with 1.69.

hainest:
> Ah, so my superficial understanding of C++ is exposed. :-)

No worries. There are enough rules in C++ to keep us all busy for a lifetime!

> As for spack, I'll add a spack PR for dyninst/package.py. v10.0.0 is
> a fixed tag which currently fails to build as is with boost 1.69.
> After we've settled on a dyninst patch, it's a 1-line change to
> package.py (plus the patch) to add that patch to v10.0.0 so it will
> build with 1.69.

Ah, ok. I thought that was what you meant, but I wasn't sure.

---

After looking at the compiler error and your fixes more carefully, I was able to recreate the issue. It is a boost issue insofar as they didn't implement the explicit conversion fix carefully enough not to break working code (such is the nature of things...). It's definitely not an issue with DynInst, which is a relief.

```
#include <boost/version.hpp>

#if BOOST_VERSION < 106800
#	error Boost is too old
#endif

#ifdef NO_CONVERT
#	define BOOST_NO_CXX11_EXPLICIT_CONVERSION_OPERATORS
#else
#	undef BOOST_NO_CXX11_EXPLICIT_CONVERSION_OPERATORS
#endif

#include <boost/logic/tribool.hpp>
#include <iostream>

bool convert(int i, boost::tribool tb) {
	return i > 2 && !tb;
}

int main(int argc, char **) {
	boost::tribool bold;
	return convert(argc, bold);
}
```

Compiling with gcc-8 gives

```
g++-8 -Wall -Wextra -std=c++11 -I/usr/local/lib/boost-1.68/include -DNO_CONVERT test.cpp
g++-8 -Wall -Wextra -std=c++11 -I/usr/local/lib/boost-1.69/include -DNO_CONVERT test.cpp
g++-8 -Wall -Wextra -std=c++11 -I/usr/local/lib/boost-1.68/include  test.cpp
g++-8 -Wall -Wextra -std=c++11 -I/usr/local/lib/boost-1.69/include  test.cpp
test.cpp: In function ‘bool convert(int, boost::logic::tribool)’:
test.cpp:17:19: error: cannot convert ‘boost::logic::tribool’ to ‘bool’ in return
  return i > 2 && !tb;
```

The issue is not the `!tb` term, but that the expression `i > 2 && !tb` actually returns a `boost::tribool` because `operator&&(bool, tribool)` is overloaded. This is a great example of why overloading the conjunction/disjunction operators is strongly contraindicated...

hainest:
It looks like it _is_ in the Release Notes for 1.69, but I was looking in the wrong spot. It's under "Logic", not "tribool". [Here](https://github.com/boostorg/logic/pull/5) is their lengthy discussion about the breaking change.

mwkrentel:
I see.  So it's the 'explicit' in the 'operator bool()' that's
breaking the usage in Message.[Ch].

I would call that "controversial" because I'd think one of the main
uses for tribool is to be able to freely mix it with bool.  Either
that or else make it totally separate and remove the && and ||
operators.  But here, they have a partial mixing, so it's confusing.

So I guess if() and while() have an implicit 'explicit' status?  :-)

hainest:
> I see. So it's the 'explicit' in the 'operator bool()' that's
> breaking the usage in Message.[Ch].

Yup.

> I would call that "controversial" because I'd think one of the main
> uses for tribool is to be able to freely mix it with bool. Either
> that or else make it totally separate and remove the && and ||
> operators. But here, they have a partial mixing, so it's confusing.

In the comment above, I linked to a lengthy [conversation](https://github.com/boostorg/logic/pull/5) the Boost devs had about this. In the end, I agree that the benefits outweigh the costs, even though the cost is that it most definitely breaks previously valid code. This is because Boost::tribool was designed long, long before explicit user-defined conversions were even thought about. That said, they should never have overloaded the conjunction/disjunction operators. That's always been a bad idea- since C++98.

> So I guess if() and while() have an implicit 'explicit' status? :-)

For as confusing as that sounds, yes, pretty much.

hainest:
Building with master against Boost-1.68 and Boost-1.69 is working.

PR #543, opened 2019-01-29
Ppcle rhel8 fix

mxz297:
Fixes most test suite failures on ppc running rhel8/fedora 28


mxz297:
Results are looking good on rhel8 testing machines.

PR #545, opened 2019-01-30
Implement ARMv8 Rewriter IMFC

LER0ever:
This PR implements the Inter-Module Function Calls for Dyninst Rewriter mode. 

- Currently 28 out of 41 tests of **group 1** pass, (with 8 fail, 5 crash)
- ~~still need to check rewriter mode to prevent regression for dynamic mode~~
- Currently the testsuite cannot run multiple tests, otherwise all of them would fail

The test script I'm using to run individual tests: 
```bash
for i in `seq 1 41`; do ./runTests -rewriter -test test1_$i; done
```

image
image

### DIFF
--- a/dyninstAPI/src/inst-aarch64.C
+++ b/dyninstAPI/src/inst-aarch64.C
@@ -661,8 +661,8 @@ Register EmitterAARCH64::emitCall(opCode op,
         // Register s1 = gen.rs()->getScratchRegister(gen, noCost);
         assert(scratch != REG_NULL);
         gen.markRegDefined(scratch);
-        if (gen.func()->obj() != callee->obj()) {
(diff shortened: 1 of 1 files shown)


mxz297:
@sashanicolas Should this be merged to ARMv8 branch?

sashanicolas:
@mxz297 not yet. @LER0ever was still trying to fix the last tests of group 1.

LER0ever:
> @mxz297 not yet. @LER0ever was still trying to fix the last tests of group 1.

@sashanicolas actually you can merge it. The `test store shared` is not related to the function implemented in this PR, and therefore won't be added to this one.

sashanicolas:
These has been moved to branch https://github.com/dyninst/dyninst/tree/ER/ARMv8.

Issue #529, opened 2019-01-15
After instrument with dyninst binary exports functions from libc


alvarofe:
I was testing out dyninst more specifically with version 10. I took as an example afl-dyninst to get acquaintance with the API and learn how it works - implementing minimal logic to instrument and print random id for each basic block visited. However, after instrumenting `/bin/ls` I noticed an infinite loop with the binary instrumented.

After a bit of researching I noticed that the binary instrumented somehow is exporting libc functions - `ls-inst` is `/bin/ls` instrumented.

```
ryzen:dyninst-fun alvaro$ rabin2 -E ls|grep -E "(malloc|free|__cxa_finalize)"
125 0x00016310 0x00016310 GLOBAL   FUNC  101 _obstack_free

ryzen:dyninst-fun alvaro$ rabin2 -E ls-inst|grep -E "(malloc|free|__cxa_finalize)"
125 0x00017310 0x00017310 GLOBAL   FUNC  101 _obstack_free
137 0x000056d0 0x000056d0 GLOBAL   FUNC    0 free
138 0x000056d8 0x000056d8 GLOBAL   FUNC    0 malloc
139 0x000056e0 0x000056e0 GLOBAL   FUNC    0 __cxa_finalize
345 0x00077b1e 0x00155b1e GLOBAL   FUNC 2493 _obstack_free_dyninst
```
Those exported symbols points to the `.plt.got` section therefore there will be an infinite loop when the dynamic loader resolve the relocs.

```
ryzen:dyninst-fun alvaro$ r2 ls-inst
 -- Thanks for using radare2!
[0x00007130]> s 0x000056d8
[0x000056d8]> pd 2
            ;-- malloc:
            0x000056d8      ff250ad90100   jmp qword [reloc.malloc_232] ; [0x22fe8:8]=0
            0x000056de      6690           nop
[0x000056d8]> pxw 20 @0x22fe8
0x00022fe8  0x00000000 0x00000000 0x00000000 0x00000000  ................
0x00022ff8  0x00000000 
[0x000056d8]> iS~.plt.got
13 0x000056d0    24 0x000056d0    24 -r-x .plt.got
```
The addr `0x00022fe8` will be patched by the dynamic loader with the value of `0x56d8` leading to the infinite loop. 

If I patch those things manually with gdb by the actual values, everything works as expected. 
Do you know which might be the reason of this? I don't know if I am missing something or might be a bug with dyninst. 

[...]

PR #548, opened 2019-02-06
Do not put newly added functions to .dynsym, which may cause the load

mxz297:
to behave abnormally due to duplicated symbols.

Note that adding symbols to .dynsym can be a legit use case, where
we expose hidden functions in shared libraries. But this use case
needs more development to support.


mxz297:
This fixes issue #529

mxz297:
Test suite runs well on x64 and power

Issue #551, opened 2019-02-11
Springboards can trample function data due to incorrect range


bwelton:
When generating springboards within a single function (i.e. at entries to basic blocks),  the valid range calculation makes the assumption that any gaps between basic blocks within the function (i.e. ones that that do not contain valid instructions) are free to be used as spillover when writing a springboard (the springboard can extend into that region if necessary). 

However, this behavior can cause problems if the gap space contains data. An example from PPC:

Block Begin:
    101c9b4c:   5d ff 22 3d     addis   r9,r2,-163
    101c9b50:   68 17 29 39     addi    r9,r9,5992
    101c9b54:   64 17 07 79     rldicr  r7,r8,2,61
    101c9b58:   aa 3a e9 7c     lwax    r7,r9,r7
    101c9b5c:   14 4a 27 7d     add     r9,r7,r9
    101c9b60:   a6 03 29 7d     mtctr   r9
    101c9b64:   20 04 80 4e     bctr
End Block / Start Data:
    101c9b68:   38 01 00 00     .long 0x138
    101c9b6c:   18 00 00 00     .long 0x18
    101c9b70:   18 00 00 00     .long 0x18
    101c9b74:   48 00 00 00     .long 0x48
    101c9b78:   48 00 00 00     .long 0x48
Next Block in same function:
    101c9b7c:   00 00 42 60     ori     r2,r2,0
    101c9b80:   78 fb e3 7f     mr      r3,r31
 
The springboard for 101c9b4c extends into the region of 101c9b68 corrupting the data in that region (101c9b68 is read by the block 101c9b4c and used as the jump address to 101c9b7c, why the compiler chose to do this is another story). 

The simplest fix is to enforce strict bounds on springboard generation to fall within the block only (or another blocks free space). However this has the downside where single block functions with empty gaps within them might be required to rely on traps.

mxz297:
On power, jump table data are typically put immediately after the indirect jump; so the jump table data are embedded in .text. This is very common on power, but not so for x86 or ARM. 

I took at the related code in Dyninst. The code comments are describing the right thing to do, however the code after the comment does not match the comments.

https://github.com/dyninst/dyninst/blob/05cc087cf5adb3c5e96f47f6d47c75f7182917e1/dyninstAPI/src/Relocation/Springboard.C#L182

[...]

PR #552, opened 2019-02-13
Correctness fixes for springboard generators

bwelton:
This patch deals with issues where incorrect springboards could be generated.
Specifically addresses three issues:

1. Springboards trampling function data (addresses github issue #551)

2. Required springboards being spilling over block bounds onto
   other required springboards. Prior a if two required springboards were
   located near each other, there was a chance that one of the springboards
   could be partially overwritten resulting in broken behavior.

3. Unsafe reusage of a codegen object where if a trap was needed,
   both a branch and a trap would be written causing bad behavior.

### DIFF
--- a/dyninstAPI/src/Relocation/Springboard.C
+++ b/dyninstAPI/src/Relocation/Springboard.C
@@ -185,12 +185,17 @@ bool InstalledSpringboards::addBlocks(func_instance* func, BlockIter begin, Bloc
     ParseAPI::CodeRegion* cr = func->ifunc()->region();
     std::set<ParseAPI::Block*> blocks;
     co->findBlocks(cr, end, blocks);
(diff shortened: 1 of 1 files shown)


bwelton:
This patch needs ARM and X86 testing (as well as Power testsuite test).

hainest:
This is passing on x86 and ARM, but introduces regressions on Power.

```
test1_10
test1_11
test1_13
test1_16
test1_25
test1_3
test_callback_2
test_fork_9
test_pt_ls
test_stack_3
```

They appear to be dynamic instrumentation issues
```
attach	24
create	37
```
and all are dynamically-linked executables.

mxz297:
@bwelton After changing the size back to 1, it still has about the same amount of regressions. I will look into the springboard code in more details.

mxz297:
Replaced by #592

PR #555, opened 2019-02-21
Add --enable-install-elfh when building elfutils from source

hainest:
When building elfutils from source n systems that have old elfutils,
Dyninst couldn't find the correct version of elf.h due to it not
being copied from the elfutils build source directory to the include
directory. In the Feb 15 2019 release of elfutils, this flag was
added to fix this.


hainest:
Building correctly on x86 systems with and without system elfutils 0.173.

PR #556, opened 2019-02-24
Support parsing cubin in dyninst-10.0

Jokeren:


parseAPI/h/CFG.h
```diff
@@ -181,6 +183,8 @@ class PARSER_EXPORT Edge : public allocatable {
          EdgeTypeEnum type);
      virtual ~Edge();
 
+    void ignore_index() { _from_index = false; }
```
Jokeren: A temporary solution for avoiding read blocks from index

parseAPI/src/Parser.C
```diff
@@ -101,12 +101,12 @@ Parser::Parser(CodeObject & obj, CFGFactory & fact, ParseCallbackManager & pcb)
         plt_entries[lit->first] = lit->second;
     }
 
-    if(obj.cs()->regions().empty()) {
+    if(obj.cs()->regions().empty() ||
+      (obj.cs()->regions()[0] == NULL && obj.cs()->regions().size() == 1)) {
```
Jokeren: By default, a `StandardParseData` parser must initialize its `_rdata` member with the first element in code regions. Thus, I pushed a NULL region into the code region just to avoid getting crashed and still want an `UNPARSEABLE` parser.

```
StandardParseData::StandardParseData(Parser *p) :
ParseData(p), _rdata(&p->obj(), p->obj().cs()->regions()[0])
{ }
```

parseAPI/h/CFG.h
```diff
@@ -481,6 +485,7 @@ class PARSER_EXPORT Function : public allocatable, public AnnotatableSparse, pub
     CodeObject * _obj;
     CodeRegion * _region;
     InstructionSource * _isrc;
+    bool _cache_valid;
```
Jokeren: Set `_cache_valid` to true to avoid calling `finalize()` in `CudaFunction` that derived from `Function`

Jokeren:
Hi @mxz297 , 

I have updated changes to dyninst and tested on hpcstruct. Apart from the changes made to the `Edge` class that avoid using index to get target blocks, I removed some modifications which I found unnecessary, such as the `virtual` qualifiers for `getFuncs` and `containingFuncs`.

Could you take a look and let me know if there are some problems? Thank you!

@jmellorcrummey FYI

parseAPI/src/Parser.C
Jokeren: Instead of adding the NULL region externally, maybe we can add it here when we see `obj.cs()->regions().empty()`? @mxz297 , what is your suggestion on this issue?

hainest:
This has introduced a regression on x86 (not tested elsewhere). All static rewriters are failing.

mxz297:
@Jokeren I will follow up with more detailed action items

mxz297:
@Jokeren  the following patch does what we discussed before, so that you don't need to push a NULL region anymore. Can you apply the patch on your end and make sure it works with your CUDA  parsing application, and then update your PR accordingly.

```
diff --git a/parseAPI/src/ParseData.C b/parseAPI/src/ParseData.C
index 62cbe235f..a0c12b89f 100644
--- a/parseAPI/src/ParseData.C
+++ b/parseAPI/src/ParseData.C
@@ -93,7 +93,7 @@ ParseWorkElem * ParseFrame::mkWork(
 /**** Standard [no overlapping regions] ParseData ****/
 
 StandardParseData::StandardParseData(Parser *p) :
-    ParseData(p), _rdata(&p->obj(), p->obj().cs()->regions()[0])
+    ParseData(p), _rdata(&p->obj(), p->obj().cs()->regions().empty() ? NULL : p->obj().cs()->regions()[0])
 { }
 
 StandardParseData::~StandardParseData() 
diff --git a/parseAPI/src/Parser.C b/parseAPI/src/Parser.C
index fbad680cb..3aaad6b04 100644
--- a/parseAPI/src/Parser.C
+++ b/parseAPI/src/Parser.C
@@ -101,8 +101,7 @@ Parser::Parser(CodeObject & obj, CFGFactory & fact, ParseCallbackManager & pcb)
         plt_entries[lit->first] = lit->second;
     }
 
-    if(obj.cs()->regions().empty() ||
-      (obj.cs()->regions()[0] == NULL && obj.cs()->regions().size() == 1)) {
+    if(obj.cs()->regions().empty()) {
         parsing_printf("[%s:%d] CodeSource provides no CodeRegions"
                                " -- unparesable\n",
                        FILE__,__LINE__);
```

parseAPI/src/Parser.C
mxz297: The patch I attached below does what you suggested.

Jokeren:
@mxz297 , thanks!

How about the `trg` method in the `Edge` class? Do you still want to read a target from ParseData index?

mxz297:
> @mxz297 , thanks!
> 
> How about the `trg` method in the `Edge` class? Do you still want to read a target from ParseData index?

I don't want to cause any potential side effect at this point. So, I think your changes on short-cutting the index lookup is fine.

hainest:
Passing on x86 and ARM, but introduces regressions on Power.

```
test1_10
test1_11
test1_13
test1_16
test1_25
test1_3
test_callback_2
test_fork_9
test_pt_ls
test_stack_3
```

They are all appear to be dynamic instrumentation issues
```
attach	24
create	37

```
and all are dynamically-linked binaries.

Jokeren:
@mxz297 , @hainest 

I have verified hpctoolkit's results. Please go ahead and check if there are still problems with Dyninst.

mxz297:
@Jokeren This PR now seems to trigger a few non-deterministic test failures. I don't think your PR is doing anything wrong, but just exposes problems that were hard to reproduce. So, I will need to continue to do more debugging to fix those problems (which are totally unrelated to what you are doing) before I can merge this PR.

hainest:
@Jokeren @mxz297 Could one of you rebase this onto dyinst/master (1bd663ff8)? There is a rebase conflict that I'm not sure how to resolve.

Jokeren:
@hainest , sure. I also have another tiny change to add.

Jokeren:
@mxz297 , @hainest , I have done rebase and local tests.

hainest:
No regressions detected on x86, Power9, or ARM (leela)

PR #558, opened 2019-03-01
Elfutil version check

hainest:
Enforce the version check for the elfutils shared library. Add the Ubuntu-specific directories /usr/{lib,include}/x86_64-linux-gnu. Search for libelf.so instead of libelf.so.1 to increase likelihood of finding a symlink.


hainest:
Builds are now working on Leela

PR #559, opened 2019-03-02
Fix ARM stack walking

mxz297:


mxz297:
@hainest I pushed a commit that should fix the regression and the remaining failed test_stack_* for PIC code

hainest:
@mxz297 Thanks! I'll run it now.

PR #567, opened 2019-03-15
supporting line map information for dyninst instrumented binaries 

zygyz:
Line map information for dyninst instrumented binaries.  The modification basically does the following:
At instrumentation phase,
1. Get the code relocation information from CodeTracker. 
2. Get the line map information of an instruction being relocated. Associate the line map with the relocated instruction. If the relocated instruction is added for instrumentation, add an offset to the line number. By doing this, we could know that the instruction is not the original instruction from user program, but is added by Dyninst.  
3. Collect the file names and make our own string table. Because file ids are not global and are related to each module. This makes life easier. 
4. Write the dyninst line map into a ".dyninstLineMap" region in the instrumented binary, encoded as a list of  [start_addr, file_id, line_number, column_number] 
5. Write our own string table into a ".dyninstStringTable" region.

For line map information queries, hpcstruct as an example:
1. Upon construction of class Symtab, extract the dyninst line map and string table by reading region ".dyninstLineMap" and ".dyninstStringTable"
2. Add the extracted line information into LineInformation , so that we can leverage the query function inside LineInformation::getSourceLines().

### DIFF
--- a/common/src/headers.h
+++ b/common/src/headers.h
@@ -62,6 +62,7 @@
 #include "common/src/vxworksHeaders.h"
 
 #endif  /* architecture specific */
+#define DYNINST_STR_TBL_FID_OFFSET 2000000 // anyone creates more than 2 million files in a project?
 
 typedef enum {
    RRVsuccess,
--- a/common/src/headers.h
+++ b/common/src/headers.h
@@ -62,7 +62,6 @@
 #include "common/src/vxworksHeaders.h"
 
 #endif  /* architecture specific */
-#define DYNINST_STR_TBL_FID_OFFSET 2000000 // anyone creates more than 2 million files in a project?
 
 typedef enum {
    RRVsuccess,
--- a/dataflowAPI/rose/ExtentMap.C
+++ b/dataflowAPI/rose/ExtentMap.C
@@ -6,6 +6,7 @@
 #include "util/StringUtility.h"
 #include "ExtentMap.h"
 #include <boost/foreach.hpp>
+#include <stdio.h>
 
 #define DUMP_FIELD_WIDTH        64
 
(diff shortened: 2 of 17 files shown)


mxz297:
Thanks for the descriptions. A few questions below:

> 2. If the relocated instruction is added for instrumentation, add an offset to the line number. By doing this, we could know that the instruction is not the original instruction from user program, but is added by Dyninst.

What's the offset added to the line number for instrumentation? With your current set up, can we profile Dyninst's instrumentation? Suppose address ranges [a1-a2), [b1-b2), and so on are generated by Dyninst instrumentations. Can we accumulate all the PC samples that fall in Dyninst instrumentation ranges and see how much overhead is caused by Dyninst's instrumentation?

> 4. Write the dyninst line map into a ".dyninstLineMap" region in the instrumented binary, encoded as a list of  [start_addr, file_id, line_number, column_number]

Can you explain a little bit more about the format of the `.dyninstLineMap` region? How do you encode the length of the list? Is the format a binary format, or a more like a text string?

Also, do we know the size impact of `.dyninstLineMap`, compared to original binary size? 

> 5. Write our own string table into a ".dyninstStringTable" region.

Is `.dyninstStringTable` storing full paths for each source file? Do we know the size impact of this region?

hainest:
@zygyz Is this ready to be tested in Dyninst? If so, could you rebase it onto the current HEAD of master?

hainest:
@zygyz @mxz297 It's been more than a year since this was last updated. We should either rebase and test so we can merge, or we should abandon it.

mxz297:
This is still an ongoing project, which is developed in another branch. When it is ready to merge, we can post new PR.

PR #569, opened 2019-03-20
Should set errno to zero before calling ptrace

mxz297:
This PR fixes a proccontrol hang on leela, but does not resolve a hang on ppcle. 

Even though this is not a complete fix, people are seeing test suite hang on all architectures. We should merge this PR soon and re-test all architectures.

### DIFF
--- a/proccontrol/src/linux.C
+++ b/proccontrol/src/linux.C
@@ -914,6 +914,7 @@ bool linux_process::plat_create_int()
       ProcPool()->condvar()->unlock();
 
       //Child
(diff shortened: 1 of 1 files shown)


mxz297:
@hainest Please prioritize testing this PR, and see whether this helps reduce any time limit exceeding issue

hainest:
I see no regressions on x86 wrt master (12c61fc44). Waiting for tests on ARM and Power.

mxz297:
Power has no regression as well.

PR #571, opened 2019-03-27
Fix compilation of thread_db_process

sashanicolas:
On a machine without libthread_db.so, the placeholder class
had missing member functions and the contructors were calling
wrong base constructors.


mxz297:
No regression on po.rice.edu (Power 8)

hainest:
No regressions on x86 or ARM

PR #572, opened 2019-03-27
A few fixes for parallel parsing

mxz297:
This PR should fix the test suite hang observed on power 8.

I have run the hanging tests 1000 times without hanging.

@hainest Please create a dev branch for this PR, and it is ready to test on all platforms.


hainest:
@mxz297 could you post the Testsuite results for Power?

mxz297:
@hainest I did not run the full testsuite on power for this PR. I only ran the hanging tests repetitively to see whether it hangs or not.

@sashanicolas Since you have already set up testing environment on po.rice.edu, can you run the full test suite on po?

Obviously, I have the instruction decoding bugs to fix now.

hainest:
@sashanicolas It's going to be a couple of days before I am ready to submit jobs on LLNL. Could you run this on po?

PR #578, opened 2019-04-04
Boost cleanup

### LINKED ISSUE #563: Make boost available after 'make install'
This was previous discussed and (possibly) fixed in https://github.com/dyninst/dyninst/pull/381, but is no longer in the codebase. In any case, we need to make the libs and headers of the version of Boost used to build Dyninst available after 'make install'. If built from source, this can be done with an 'install' target for Boost. If the system libraries were used, then we need to symlink to them.

This is important for getting the same version of Boost to work for the Testsuite.

hainest:
This substantially changes how we handle Boost builds in Dyninst. The goal is to have a coherent Boost build between Dyninst and Testsuite.

Notable changes:

1. Tagged build layouts are now fully supported
2. Boost is exported to the same install structure as Dyninst (fixes #563)
3. All cmake variables that control Boost's build options are now available for the user to change via cache variables (e.g., -DBoost_USE_MULTITHREADED)
4. Substantial error handling has been added
5. Consolidates all Boost-related activities into cmake/Boost.cmake
6. Makes Boost-1.61 the new minimum version (needed for Filesystem)
7. Builds Boost-1.69 when building from source


hainest:
Passing on x86, ARM (Leela), and Power9.

@sashanicolas could you test on Kachina to see if this resolves the Boost build issues you were experiencing?

hainest:
"Passing" meaning building successfully. I can't test with the Testsuite until I change the Boost building there, as well.

PR #584, opened 2019-04-16
TBB Cleanup

hainest:
This substantially changes how we handle TBB builds in Dyninst. The goal is to have a coherent TBB build between Dyninst and Testsuite.

Notable changes:

1. Remove FindTBB submodule
2. Consolidate all TBB stuff into cmake/ThreadingBuildingBlocks.cmake
3. Explicitly die on non-UNIX platforms when building from source (previously, it did nothing)
4. Minimum TBB version is 2018.0 (I'm not sure this is necessary or sufficient)
5. Export include/library directories into the CMake cache for consumption by Testsuite
6. Default version to build from source is 2019_U5
7. Libraries and headers are exported to CMAKE_INSTALL_PREFIX when building from source
8. Correctly handles broken TBB install (https://github.com/01org/tbb/issues/98)

These are orthogonal to #570


anton-malakhov:
@tbbdev, FYI people keep insisting on symlinks instead of using official approach with linker scripts.

PR #588, opened 2019-04-24
simple working hybrid example of patchAPI 

ianamason:
Using dyninstAPI to create the PatchMgr and PatchObject.  

I should point out that I have not succeeded in getting a pacthAPI example working that
uses only the patchAPI.

### DIFF
--- a/patchAPI/example/README.md
+++ b/patchAPI/example/README.md
@@ -1,0 +1,18 @@
+To build you need to set the three environment variables `DYNINST`, `BOOST`, and
+`TBB` to the install directories of these three dependencies.
+
+
+For example, if you `spack` installed `dyninst` you could do:
+
+
+```
+export DYNINST=`spack location --install-dir dyninst`
+export BOOST=`spack location --install-dir boost`
+export TBB=`spack location --install-dir tbb`
+```
+
+Then to build and test:
+
+```
+make test
+```
--- a/patchAPI/example/README.md
+++ b/patchAPI/example/README.md
@@ -16,3 +16,7 @@ Then to build and test:
 ```
 make test
 ```
+
+To run the rewritten mutatee program you will also need to set the 
+`DYNINSTAPI_RT_LIB` environment variable to 
+`${DYNINST}/lib/libdyninstAPI_RT.so.1`
--- a/patchAPI/example/main.cpp
+++ b/patchAPI/example/main.cpp
@@ -1,0 +1,115 @@
+// DynInst
+#include "BPatch.h"
+#include "BPatch_binaryEdit.h"
+#include "BPatch_image.h"
+#include "BPatch_function.h"
+#include "BPatch_object.h"
+#include "BPatch_point.h"
+#include "BPatch_Vector.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <iterator>
+#include <string>
+#include <vector>
+#include <iterator>
+
+// patchAPI
+#include "PatchMgr.h"
+
+using namespace std;
+using namespace Dyninst;
+using namespace PatchAPI;
+
+
+// a simple example, just insert a bunch of no ops
+class NoopSnippet : public Snippet {
+
+public:
+  bool generate(Point *pt, Buffer &buffer){
+    uint8_t byte = 0x90;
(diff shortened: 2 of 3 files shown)


patchAPI/example/README.md
```diff
@@ -0,0 +1,18 @@
+To build you need to set the three environment variables `DYNINST`, `BOOST`, and
+`TBB` to the install directories of these three dependencies.
+
+
+For example, if you `spack` installed `dyninst` you could do:
+
+
+```
+export DYNINST=`spack location --install-dir dyninst`
```
hainest: What are these being used for? Do we need one for _all_ dependencies (e.g., elfutils, libiberty, etc.)?

patchAPI/example/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+  Patcher patcher(patchMgr);
+
+  NoopSnippet::Ptr snippet = NoopSnippet::create(new NoopSnippet);
+
+  vector<PatchFunction*> functions;
+
+  binobj->funcs(back_inserter(functions));
+
+
+  for(vector<PatchFunction*>::iterator funIter = functions.begin(); funIter != functions.end(); funIter++){
```
hainest: C++11's range-based `for` makes this much easier to use.

patchAPI/example/mutatee/main.c
```diff
@@ -0,0 +1,34 @@
...
+int main(int argc, const char *argv[])
+{
+  if (argc > 2){
+    foo0();
+  } else {
+    foo1();
+    foo2();
+  }
+  message("Done\n");
+  return status;
```
hainest: Returning non-zero indicates failure to the shell. Is that relevant here?

patchAPI/example/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+    PatchFunction *fun = *funIter;
+
+    vector<Point*> f_entryPoints;
```
hainest: Destruction of this vector leaks all of its containing objects because they are pointers.

hainest:
@mxz297 I have a couple of comments about the code, but we can merge this without testing once those are addressed.

patchAPI/example/README.md
mxz297: I think without your RPATH fixes, users will have to set these environment variables to specify where to find the installed dependencies. And depending on what are pre-installed on the system, you may need to specify paths to find BOOST, TBB, ELFUTILS, and so on.

I don't think it is worthwhile to tweak these environment variables right now. After your CMake & RPath fixes, we can clear the README file.

patchAPI/example/main.cpp
mxz297: These objects pointed by the pointers are constructed inside Dynins  and the objects are stored in a map. These objects will be destructed when Dyninst destructs the map. Users should not explicitly destruct these objects.

ianamason: I did valgrind it, but got so much noise that I left it at that.
```
==1932== LEAK SUMMARY:
==1932==    definitely lost: 139,464 bytes in 449 blocks
==1932==    indirectly lost: 218,810 bytes in 2,894 blocks
==1932==      possibly lost: 110,365 bytes in 1,721 blocks
==1932==    still reachable: 1,284,088 bytes in 16,184 blocks
==1932==                       of which reachable via heuristic:
==1932==                         newarray           : 528 bytes in 1 blocks
==1932==                         multipleinheritance: 239,304 bytes in 767 blocks
==1932==         suppressed: 0 bytes in 0 blocks
==1932== Reachable blocks (those to which a pointer was found) are not shown.
==1932== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1932==
==1932== For counts of detected and suppressed errors, rerun with: -v
==1932== ERROR SUMMARY: 62 errors from 62 contexts (suppressed: 0 from 0)

```

patchAPI/example/mutatee/main.c
ianamason: I fixed this, plus added a remark about what the rewritten program depends on
(i.e. the dyninst runtime library).

mxz297:
@hainest Are you OK with the new changes? We can update the C++11's range-based `for` later.

Issue #577, opened 2019-04-04
Symtab should not always demangle every function name


mwkrentel:
When Symtab demangles symbol names in `P_cplus_demangle()` in
`common/src/linuxKludges.C`, it looks like Symtab automatically calls
`__cxa_demangle()` to demangle every name, whether it is mangled or
not.  This is not correct.

Instead, Symtab should lookup what module the name is from and
demangle if the source language is C++, ie, `lang_CPlusPlus` or
`lang_GnuCPlusPlus` in module-speak.  (I don't know what
`lang_Fortran_with_pretty_debug` is, but I'm guessing it isn't
mangled.)

I think the infrastructure for this is already in place, so hopefully
this is a fairly small fix.

----------

Here's how to reproduce the problem with a .c program.

```
#include <stdio.h>

void a(void)
{
    printf("hello\n");
}

int main(int argc, char **argv)
{
    a();
    return 0;
}
```

Compile this with a C compiler (not C++) without optimization or -O0
so the compiler doesn't inline "a" and eliminate it.

Then, see what Symtab produces for `mangled_names_begin()`,
`pretty_names_begin()` and `typed_names_begin()` for function "a".
I used debug output from hpcstruct for this, but by whatever means,
what I get is:

```
mangled name:  a
pretty name:   signed char
typed name:    signed char
```

The problem, of course, is that `__cxa_demangle("a") --> "signed char"`.
This is a valid demangling (of a type encoding, not a symbol name),
but in the wrong context.  The source file is C, so the symbol is not
mangled.

ping @mxz297

mxz297:
@mwkrentel The problem is reproduced on my end. Thanks for the example. I will fix it soon.

mxz297:
@mwkrentel @jmellorcrummey After more thoughts, I feel this problem is more complicated and I don't think there is a perfect solution for it. 

## Source language of the module did not really help 

Let me start with Mark's suggested solution, which is to look at the source language of the module. 

1. The binary is not guaranteed to be compiled with `-g`. In this case, symtabAPI says that the source language is `lang_Unknown`. 

2. Even if the binary is compiled with `-g`. Module level information is not sufficient for dealing with `extern "C"`. I modified Mark's example as follows:

```
#include <stdio.h>

extern "C" {

void a()
{
    printf("hello\n");
}

}

int main(int argc, char **argv)
{
    a();
    return 0;
}
```

SymtabAPI tells me that

```
Symbol at 400507
	getMangledName(): a
[...]

PR #592, opened 2019-04-29
More fixes for parallel parsing, spring boards, and ARM

mxz297:


mxz297:
Fixes #577

mxz297:
Fixes #551

hainest:
Regressions on x86 for test5 series with static linking in rewriter mode (PIC and nonPIC).

mxz297:
Re-test the PR with the latest master on all three platforms and no regressions.

PR #599, opened 2019-05-10
Update version number to 10.1.0

mxz297:
This should be last PR to merge for 10.1


mwkrentel:
@mxz297  @hainest 
I'm currently finishing my spack recipe and walking through the
platforms to smoke test.

We have a phone call tomorrow, May 16.  How about we agree to hold off
tagging until after the phone call.  I'll either say Ok, or else give
a compelling reason why not.

mxz297:
@mwkrentel Yes, we can hold off tagging to at least tomorrow.

hainest:
@mwkrentel It's not needed for 10.1, but I'd like to chat with you about adding some documentation about building with spack to the [wiki](https://github.com/dyninst/dyninst/wiki/Building-Dyninst#spack).

mwkrentel:
@mxz297  @hainest 
Ok, I'm ready to tag and release 10.1.0.
Thanks for waiting on my tests.

Issue #585, opened 2019-04-17
Cobwebs on the Documentation and boost 1.70.0 issues.


ianamason:
@mxz297 is there any updated documentation for the patchAPI classes?
The doc area looks quite old, and the only example findPoint looks to be
quite out of date.

As an aside, I tried spack installing on a clean ubuntu 18.0.4 box again yesterday,
but it failed this time. Seems boost has gone from 1.69.0 to 1.70.0 and dyninst no longer 
compiles against this version.  I got around it with the `^boost@1.69.0` hack, but ...

hainest:
> As an aside, I tried spack installing on a clean ubuntu 18.0.4 box again yesterday,
> but it failed this time. Seems boost has gone from 1.69.0 to 1.70.0 and dyninst no longer
> compiles against this version. I got around it with the ^boost@1.69.0 hack, but ...

We just found this a couple of days ago. Someone is working on a patch. I hope to have it merged soon (definitely before the 10.1 release).

ianamason:
Thanks @hainest

PR #600, opened 2019-05-13
Compilation fix for boost 1.70

mxz297:
Tested with boost 1.70 and boost 1.61


mxz297:
This should fix #585.

Issue #582, opened 2019-04-15
RAX corruption in EmitterAMD64::emitStackAlign


mothran:
In emitStackAlign() the LAHF instruction is used to save flags to AH, before the registers are pushed onto the stack and saved.  This means that the RAX register is corrupted between original basic block executions.

If the scratch register used in emitStackAlign() is not RAX, then the register does not get saved correctly.  I was able to avoid the corruption by forcing the register to RAX if saveFlags was true.  I am not sure how to address the corruption short of forcing the scratch register to RAX which seems to prevent the register from corruption in testing.

Input basic block:
```
lea     rdx, [rax+0x1]
mov     rcx, rdx
sub     rcx, rbx
cmp     rcx, 0x6
jle     0x1d37
```

Example output:
```
0x5598da9f0f86  lea    rsp,[rsp-0xb0]
0x5598da9f0f8e  mov    QWORD PTR [rsp+0x20],r10  <- r10 picked as a scratch register
0x5598da9f0f93  lahf                             <- AH corrupted here
0x5598da9f0f94  seto   al
0x5598da9f0f97  mov    QWORD PTR [rsp+0x28],r10
0x5598da9f0f9c  lea    r10,[rsp+0xb0]
0x5598da9f0fa4  and    rsp,0xffffffffffffffe0
0x5598da9f0fa8  mov    QWORD PTR [rsp],r10
0x5598da9f0fac  mov    r10,QWORD PTR [r10-0x88]
0x5598da9f0fb3  add    al,0x7f
0x5598da9f0fb6  sahf
0x5598da9f0fb7  mov    r10,QWORD PTR [rsp]
0x5598da9f0fbb  mov    r10,QWORD PTR [r10-0x90]
0x5598da9f0fc2  push   rax                      <- RAX (corrupted) saved here
0x5598da9f0fc3  push   rdx
```

mxz297:
@mothran Based on your post, it definitely looks like Dyninst is doing something wrong. Can you create a reproducer? I have not seen this issue before and a reproduce will be extremely helpful for debug.

mothran:
@mxz297 Sure, I have a small target .c program that expresses the issue when instrumented with a per-basic-block AOT instrumentation with dyninst.  I can provide the target but can't provide the instrumentation tooling.  I expect that there is nothing in the tooling that specifically causes the issue.  Any dyninst tooling that instruments each basic block should induce the issue.

Below is the target program. 

target.c
```c
#include <stdio.h>
[...]

PR #603, opened 2019-05-23
Resolves emitCall corrupting AMD64 calling registers

mothran:
When a basic block is instrumented with a callback, the arguments are passed to the callback via the AMD64_ARG_REGS.  These registers are not always correctly marked as live by registerSpace and thus don't get saved during a emitCall or emitBTSaves.  Commonly this will be RDI (arg0) being corrupted by a instrumentation block and causing problems/craches in the original code.

This issues was mentioned in the following issues / external projects:
https://github.com/dyninst/dyninst/issues/461
https://github.com/talos-vulndev/afl-dyninst/blob/master/README.txt#L61

Notes on my patch:
This patch is very aggressive in how often and how much it saves.  For a single function callback it will push/pop the args multiple times depending on the lock/unlock guards.  It also seems to mess with the refcounting / keptvalue system for the register tracking and can cause spurious saves within a single injected trampoline.  Ideally this would be resolved in the registerSpace / AST engine.  Then the emitBTSaves function would correctly identify the AMD64_ARG_REGS which correspond to the number of arguments for a emitCall but I was unable to work back the root cause well enough to make a patch.  I hope this pull request is at least enough to start the process of identifying the root issue / providing a hotpatch for anyone who might need it.

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -2439,7 +2439,6 @@ void EmitterAMD64::emitStackAlign(int offset, codeGen &gen)
    if (gen.rs()->checkVolatileRegisters(gen, registerSlot::live)) {
       saveFlags = true;   // We need to save the flags register
       off += 8;           // Allocate stack space to store the flags
-      scratch = REGNUM_RAX;
    }
 
    emitLEA(REGNUM_RSP, Null_Register, 0, -off, REGNUM_RSP, gen);
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -1812,9 +1812,19 @@ Register EmitterAMD64::emitCall(opCode op, codeGen &gen, const pdvector<AstNodeP
                          reg->keptValue,
                          (reg->liveState == registerSlot::live) ? "live" : ((reg->liveState == registerSlot::spilled) ? "spilled" : "dead"));
 
+         bool arg_reg = false;
+         if (operands.size() > 0) {
+            for(unsigned int j = 0; j < std::min(operands.size(), AMD64_ARG_REGS); j++) {
+                if (reg->number == amd64_arg_regs[j]) {
+                    arg_reg = true;
+                    break;
+                }
+            }
+         }
+
          if (reg->refCount > 0 ||  // Currently active
              reg->keptValue || // Has a kept value
-             (reg->liveState == registerSlot::live)) { // needs to be saved pre-call
+             (reg->liveState == registerSlot::live) || arg_reg) { // needs to be saved pre-call
             regalloc_printf("%s[%d]: \tsaving reg\n", FILE__, __LINE__);
             pair<unsigned, unsigned> regToSave;
             regToSave.first = reg->number;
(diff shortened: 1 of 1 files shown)


hainest:
Is this a fix for #582?

mothran:
@hainest No but I did make a mistake and accidentally include my patch to work around #582 I can remove that if needed.  They are two distinct corruptions.

hainest:
Since they are fixes for two separate issues, could you split the code changes here? That will keep us from getting confused as we work on 582. Thanks!

mothran:
@hainest I updated my branch with a commit that removes the inclusion of the scratch overwrite.  Should be just related to AMD64 regs now.

mothran:
Minor update:
While working with my above patch, I found that just saveing/restoring the registers used to make the call operation was insufficient.  If the callback uses any of the AMD64_ARG_REGS (aka by calling external libraries) it can leak the values back after the callback and corrupt the state of the original code.

I changed things so that the shouldSaveReg() checks if the reg->number is in the amd64_arg_regs array, and if true then it will return true.  This also means that there is less saving/restoring of the registers within a Trampoline, instead saving all arg regs before a trampoline.

I am happy to update my repository to reflect the changes into this pullreq if needed.

mxz297:
@mothran I have been moving and now I am back to work. Here is my thoughts on the root problem:

Based on your changes ```(reg->liveState == registerSlot::live) || arg_reg)```, it looks to me that ```reg->liveState==registerSlot::live``` was false. However, let's say `reg` is RDI; even if RDI was overwritten by instrumentation (through multiple layers of function calls), as long as RDI was not used after the instrumentation by the original program (before a new value is assigned to RDI), the program should not crash. Obviously, I re-stated the definition of a dead register. So, To me, the root cause is that Dyninst liveness analysis is wrong and wrongly claims some live registers as dead. 

In my experiences, liveness analysis can go wrong if Dyninst wrongly decodes instruction operands. You mentioned issue #461, whose root cause is indeed an instruction operand decoding problem. My guess is you are seeing a similar instruction decoding problem.

In the short term, can you compare the InstructionAPI::Instruction for the problematic blocks with the instructions you got from another disassembler? If you can find the bytes and the corresponding wrong instructions, I can fix it.

In the long term (for the next release), we are planning to adding instruction decoding support by leveraging intel xed or/and LLVM, which should be much more stable and complete compared to the instruction decoder in instructionAPI.

mothran:
@mxz297 The issue I ran into seemed to stem from the liveness analysis breaking when it came to handling the AMD64 abi in a binary.  For example I found that a basic block containing a single 'ret' would leak registers from the instrumentation.  As I mentioned in my recent comment this seemed to be more than the operand registers used for a instrumentation callback but would leak other registers from the amd64 ABI.

What debug env-vars can I use to debug the liveness during instruction Decoding?

mxz297:
> For example I found that a basic block containing a single 'ret' would leak registers from the instrumentation. 

Consider the following code:

```
A: 
Instrumentation overwriting rax and rdi
ret
```

If the instrumentation code does not save & restore rax or rdi. Overwritten values for rax or rdi would be passed back to the caller of A. However, if the original program is abi-compliant, the caller of A should not expect rdi to preserve its value. Similarly, any other parameter passing registers are preserved across calls. 

On the other hand, rax is different because rax passes return values. And Dyninst liveness analysis will mark rax is read by the return instruction. 

If the original program is not abi-compliant, then everything is off.

Therefore, I feel there are still quite some missing pieces to discovery. Can you post a minimal example for debugging? 
 
> What debug env-vars / other debugging output can I use to debug the liveness during instruction Decoding?

You can use this tool (https://github.com/mxz297/dyninst-tools/tree/master/CodeDump) to print instructions from Dyninst. I am gradually uploading Dyninst based tools that I have been using to my github page.

mothran:
Hmm yeah you're right.  I think without a reproduction sample we are going to go in circles.  I am closing this pull request and I will work on my end to create a reproduction that I can upload.  Once I have something more concrete I will open an issue ticket.

Thanks for the pointer to those tools, I will check them out.

PR #610, opened 2019-06-13
C++ cleanup

hainest:
Use the C and C++ language feature-detection support built into CMake (>3.1). This is more accurate and faster than the previous method which only checked if `auto` was supported.


hainest:
This breaks the Testsuite(https://github.com/dyninst/testsuite/issues/72) build because there is lots of code there that doesn't adhere to strict C99 conformance. I've been adding back the necessary GNU-specific macros to get it to compile under -std=c99, but it's incomplete. DO NOT MERGE.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR610&testsuite_branch=PR71&since_date=2020-05-30

No regressions (those listed are unrelated).

PR #612, opened 2019-06-14
ARMv8 initial work on rewriter

sashanicolas:


dyninstAPI/src/inst-aarch64.C
```diff
@@ -1424,13 +1424,139 @@ bool EmitterAARCH64Dyn::emitTOCCommon(block_instance *block, bool call, codeGen
 
 // TODO 32/64-bit?
 bool EmitterAARCH64Stat::emitPLTCall(func_instance *callee, codeGen &gen) {
-    assert(0); //Not implemented
-    return emitPLTCommon(callee, true, gen);
+    /*
```
hainest: If the commented-out code is acting as documentation, please add more comments to explain. If it's just dead code, please remove it.

dyninstAPI/src/inst-aarch64.C
```diff
@@ -1424,13 +1424,139 @@ bool EmitterAARCH64Dyn::emitTOCCommon(block_instance *block, bool call, codeGen
...
+
+    //Set other bits. Basically, these are the opcode bits.
+    //The only difference between BR and BLR is that bit 21 is 1 for BLR.
+    INSN_SET(branchInsn, 16, 31, BRegOp);
+    INSN_SET(branchInsn, 21, 21, 1);
+    insnCodeGen::generate(gen, branchInsn);
+    //insnCodeGen::generateBranch(gen, gen.currAddr(), lr, true);
+    //insnCodeGen::generateBranch(gen, gen.currAddr(), gen.currAddr() +varOffset, true);
+
+    return true;
```
hainest: No branch returns false. Should this function always succeed?

dyninstAPI/src/inst-aarch64.C
```diff
@@ -1424,13 +1424,139 @@ bool EmitterAARCH64Dyn::emitTOCCommon(block_instance *block, bool call, codeGen
...
+
+    //assert(0); //Not implemented
+    //return emitPLTCommon(callee, true, gen);
 }
 
 bool EmitterAARCH64Stat::emitPLTJump(func_instance *callee, codeGen &gen) {
-    assert(0); //Not implemented
-    return emitPLTCommon(callee, false, gen);
+    /*
```
hainest: Same comments as above.

dyninstAPI/src/inst-aarch64.C
sashanicolas: In a perfect world, not. Many functions in Dyninst will always "succeed", meaning the return value will not be checked. Either because originally someone designed the return value to be eventually used in a not foreseen case or because the the function should actually run without interrupting the course of the execution and succeeding, meaning that if it does not do what it is supposed to do there should not even be a failure case.

dyninstAPI/src/inst-aarch64.C
sashanicolas: It's praxis to keep old cold commented out or disabled in case you find a bug in a the new implementation or you need to revert something. Searching back to git history is not as practical, and even worse when you are using non-git functionality that are dependent on a single vendor and not supported by git compliant clients, such as PRs.

hainest:
Do we need to make any changes to the Testsuite to activate these codepaths for testing?

PR #614, opened 2019-06-14
Updates TLS descriptors values in .dynamic section

sashanicolas:
when .got and .plt offsets are modified for ARMv8.

PR #618, opened 2019-06-27
Fix overlapping bug of program header segments on fixPhdr.

sashanicolas:
This fix is making the update of the program headers in two passes. First,
we need to update the addresses, and then insert the new created segment.

This needs to be tested on all supported platforms on rewriter mode.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -66,54 +66,6 @@ unsigned int elfHash(const char *name) {
 }
 unsigned long bgq_sh_flags = SHF_EXECINSTR | SHF_ALLOC | SHF_WRITE;;
 
-
-
-static bool libelfso0Flag;
-static bool libelfso1Flag;
-static int libelfso1version_major;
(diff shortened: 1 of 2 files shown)


hainest:
Do you remember which test failure triggered this?

mxz297:
@hainest IIRC, this is caused by a rewriter test failure involving rewriting a shared library. The issue fixed by Sasha here however is not sufficient to make that test passing. 

So, from a test suite perspective, we are not going to see any more tests passing. It will be a success if there is no regression.

hainest:
This depends on #614 and #612.

sashanicolas:
@hainest the tests that triggered this involved the shared libraries A and B. This fix is not sufficient as @mxz297  said, but with https://github.com/dyninst/dyninst/pull/612 and https://github.com/dyninst/dyninst/pull/614 it is.

hainest:
https://bottle.cs.wisc.edu/branch/PR618

No regressions.

PR #624, opened 2019-07-18
Private writable mmap for opening binaries and debug files

blue42u:
Ran into #436 myself, on a Debian system when opening `libc.so` with `Symtab::openFile`. This patch is essentially a corrected version of #487 (this version still opens the file ro, but also allows the mapped region to be rw).

Also fixes a similar issue that pops up when endianness conversions are needed (done in-place).

Steps to reproduce the issue:

0. Obtain a binary (`.../example.so`, although executables should work too) compiled with debugging info, and a small program (`./symopen`) that calls `Symtab::openFile` on said binary.
1. Generate a debug info file for the library: 
   `objcopy --only-keep-debug --compress-debug-sections .../example.so borked.debug`
2. Add the debug link into (a copy of) the library: 
   `objcopy --add-gnu-debuglink=borked.debug .../example.so borked.so`
3. Try out the little program (`./symopen borked.so`) and look for a segfault in `elf_compress`.

Notes:

- Not compressing sections (removing `--compress-debug-sections` above) runs fine, apparently libelf is read-only when dealing with decompressed sections.
- Running on a library with included compressed debug sections also runs fine, libelf manages its own mmap'd region to account for this.

### DIFF
--- a/common/src/MappedFile.C
+++ b/common/src/MappedFile.C
@@ -361,19 +361,8 @@ bool MappedFile::map_file()
 
 #else
 
-   int mmap_prot  = PROT_READ;
-   int mmap_flags = MAP_SHARED;
-
-#if defined(os_vxworks)   
-   // VxWorks kernel modules have relocations which need to be
-   // overwritten in memory.
-   //
-   // XXX - We don't overwrite our memory image with relocations from the
-   // target memory space yet due to performance concerns.  If we decide
-   // to go this route, it would simplify a lot of the Dyninst internals.
-   mmap_prot  = PROT_READ | PROT_WRITE;
-   mmap_flags = MAP_PRIVATE;
-#endif
+   int mmap_prot  = PROT_READ | PROT_WRITE;
+   int mmap_flags = MAP_PRIVATE;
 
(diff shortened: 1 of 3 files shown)


hainest:
This fix has been proposed [before](https://github.com/dyninst/dyninst/pull/487), but rejected. @sashanicolas could you double-check that this fix demonstrates the same issues?

mxz297:
@hainest @sashanicolas The fix in this PR looks similar but is different from #487 .

In this PR, we still just open a file in read-only mode, so no file permission is messed up. However, we call mmap with MAP_PRIVATE and PROT_READ|PROT_WRITE, so the mmap memory region can be written, but the changes are not going to be propagated to the original file due to MAP_PRIVATE.

mxz297:
@hainest This PR is renamed to reflect its new scope.

hainest:
@mxz297 is this ready to test?

mxz297:
@hainest yes, ready to test

hainest:
https://bottle.cs.wisc.edu/branch/PR624

hainest:
No regressions detected. This also seems to have fixed the test suite errors I was previously seeing on ray at LLNL (power8), so double-good!

Issue #626, opened 2019-08-07
Missing PRIx64 macro on ppcle64 before Power9


hainest:
Regarding #517:

One of my cron jobs tripped an alarm for Dyninst master on powerpc this morning. This one is bizarre, but the last commit about the PRIx64 macros breaks the build on some systems, depending on the exact version of the <inttypes.h> header file. Here is a unit test that reproduces the problem.  This is a snippet from instructionAPI/h/Result.h, near line 453.

```
#include <string.h>
#include <inttypes.h>
int main(int argc, char **argv) {
     char hex[20];
     int64_t val = 123;
     snprintf(hex, 20, "%" PRIx64, val);
     printf("ans = %s\n", hex);
     return 0;
}
```

On one power8 system with glibc-headers-2.17-157.el7.ppc64le,

> $ g++ -g -O -std=c++11 print.cpp
>
> print.cpp: In function 'int main(int, char**)':
> print.cpp:11:27: error: expected ')' before 'PRIx64'
>       snprintf(hex, 20, "%" PRIx64, val);
>                             ^~~~~~

Building Dyninst, I get the same error in Result.h near line 453. I had to diff <inttypes.h> across systems to figure out what's going on. The powerpc version has the following. The x86_64 version does not have the #if on lines 44-46.

>    44  /* The ISO C99 standard specifies that these macros must only be
>    45     defined if explicitly requested.  */
>    46  #if !defined __cplusplus || defined __STDC_FORMAT_MACROS
>    47
>    48  # if __WORDSIZE == 64
>    49  #  define __PRI64_PREFIX        "l"
>    50  #  define __PRIPTR_PREFIX       "l"
>    51  # else
>    52  #  define __PRI64_PREFIX        "ll"
>    53  #  define __PRIPTR_PREFIX
>    54  # endif

So, on powerpc, things like PRIx64 are either not defined or else not
defined the way you want them. I'm not sure if this is a glibc-headers version issue, or an x86 vs ppc issue.  It seems to be more common on powerpc, but older RH 6.x
x86_64 systems also have the same problem. Anyway, I'll let you decide the best and most portable way of handling this.  Maybe you just want to define __STDC_FORMAT_MACROS and test that it doesn't break anything else.

mwkrentel:
To answer one of your questions, here's a quick inventory of a few
systems, where "broken" means that PRIx64 is not defined.

```
nancy    x86_64    RH 6.x       glibc 2.12    broken
[...]

PR #627, opened 2019-08-07
Convert PRIx64 -> std::to_string

hainest:
This is a fix for #626 as the PRIx64 macros don't work on PPC before Power9.


hainest:
https://bottle.cs.wisc.edu/branch/PR627

mwkrentel:
Ok, I tested your string_format_fix branch on power8 with gcc 4.8.5
and it builds cleanly.  This is the same machine (but later compiler)
that threw the alarm.

So, I approve.

I'll look forward to my cron job passing in the near future.  :-)
Thanks!

Issue #497, opened 2018-10-21
build error (clang6 on ubuntu 18.04)


pbst:
Building with clang/clang++ (version 6.0) fails on x86_64 Ubuntu 18.04 system.

I configured it using:
`cmake .. -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DCMAKE_INSTALL_PREFIX=$HOME/.local`

After a while it fails with:
```
dyninst/dyninstAPI/src/codegen.C:325:18: error: arithmetic on a pointer to void                                                                                                                  
    memmove(temp + size, temp, used()-index);                                                                                                                                                    
            ~~~~ ^
```
Inserting explicit casts solves the error-throwing part, but there is probably a nicer looking way to patch this.

```patch
-    memmove(temp + size, temp, used()-index);
+    memmove((void *)((uint64_t)(temp) + size), temp, used()-index);
```
```
$ clang++ --version                                                                                                                                                     
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)                                                                                                                                            
Target: x86_64-pc-linux-gnu                                                                                                                                                                      
Thread model: posix                                                                                                                                                                              
InstalledDir: /usr/bin
```

hainest:
Fixed in clang branch.

PR #631, opened 2019-08-12
Add Clang compiler support

blue42u:
Two minor changes that let Dyninst compile with Clang. Comments and adjustments welcome.

Ping @hainest

### DIFF
--- a/common/h/dyntypes.h
+++ b/common/h/dyntypes.h
@@ -124,6 +124,8 @@
 // TODO: when should we use thread_local?
 #if defined(_MSC_VER)
   #define dyn_tls __declspec(thread)
+#elif defined(__clang__)
+  #define dyn_tls thread_local
 #else
   #define dyn_tls __thread
 #endif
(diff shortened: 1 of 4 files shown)


hainest:
@blue42u thanks! It's interesting that the original problem (#497) didn't need to be fixed for Dyninst to build with clang now.

blue42u:
It does need fixed, I just used a different method to do it (casting to `char*` rather than to `uint64_t` and back). I figured this way would be cleaner and more portable.

hainest:
LLVM supports a very limited number of options for the '.machine' directive on PPC ([code](https://github.com/llvm-mirror/llvm/blob/master/lib/Target/PowerPC/AsmParser/PPCAsmParser.cpp#L1705)). This breaks DyninstAPI_RT ([code](https://github.com/dyninst/dyninst/blob/master/dyninstAPI_RT/src/RTthread-powerpc-asm.S#L8)). The fix is just to ignore this directive in RThread-powerpc-asm.S when compiling with clang.

```
#ifndef __clang__
# if defined(arch_64bit)
   .machine "ppc64"       
# else
   .machine "ppc"       
# endif
#endif
```

blue42u:
Folded that new change in (and the equivalent in RTthread-power-asm.s). Untested but should work.

hainest:
Closing to migrate to public Dyninst branch.

PR #637, opened 2019-08-16
Fix detach for trap springboards

### LINKED ISSUE #634: ppcle detach removes snippets
The Red Hat release process has a smoke test which essentially does:
insertSnippet
continueExecution
detach
This works fine on x8664 and aarch64 but not on a ppcle box (POWER8 kernel-4.1 8.0-112.el8 glibc-2.28-66.el8)
On ppcle there is no instrumented call to incr; the call to detach removes the instrumented call. If you add a waitForStatusChange before detach (or simply just sleep) then ppcle hits the snippet call to incr just fine.

Here is the example run:

./mutatee -v 10 & LD_LIBRARY_PATH=/usr/lib64/dyninst DYNINSTAPI_RT_LIB=/usr/lib64/dyninst/libdyninstAPI_RT.so ./mutator $!
[1] 106821
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
MUTATION DONE. MUTATOR IS GOING...
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0
FUNCTION EXECUTED. VALUE = 0

(adjust DYNINST in Makefile, make all ; make test)

[smoke-test.zip](https://github.com/dyninst/dyninst/files/3506845/smoke-test.zip)

mxz297:
This PR should fix #634 .

@stanfordcox Can you test whether this PR fixes the problem on your end?

### DIFF
--- a/dyninstAPI/src/dynProcess.C
+++ b/dyninstAPI/src/dynProcess.C
@@ -3243,25 +3243,13 @@ bool PCProcess::continueSyncRPCThreads() {
 }
 
 void PCProcess::addTrap(Address from, Address to, codeGen &gen) {
-    map<Address, Breakpoint::ptr>::iterator breakIter =
-       installedCtrlBrkpts.find(from);
-
-    if( breakIter != installedCtrlBrkpts.end() ) {
-        if( !pcProc_->rmBreakpoint(from, breakIter->second) ) {
-	  // Oops? 
-        }
-        installedCtrlBrkpts.erase(breakIter);
-    }
-    
-    Breakpoint::ptr newBreak = Breakpoint::newTransferBreakpoint(to);
-    newBreak->setSuppressCallbacks(true);
-
-    if( !pcProc_->addBreakpoint(from, newBreak) ) {
-      // Oops? 
-    }
-
-    installedCtrlBrkpts.insert(make_pair(from, newBreak));
-    gen.invalidate();
+   gen.invalidate();
+   gen.allocate(4);
+   gen.setAddrSpace(this);
+   gen.setAddr(from);
+   insnCodeGen::generateTrap(gen);
+   trapMapping.addTrapMapping(from, to, true);
+   springboard_cerr << "Generated springboard trap " << hex << from << "->" << to << dec << endl;
 }
 
 void PCProcess::removeTrap(Address from) {


stanfordcox:
@mxz297 It fixes the problem on ppcle (and x8664 still works as expected)

mxz297:
@hainest Ready to test.

hainest:
@mxz297 It will be a couple of days before I get to this. I'm making some changes to the build script and dashboard to track compiler information so I can better test Jonathan's ARM decoder thread race fix first.

hainest:
https://bottle.cs.wisc.edu/branch/PR637

hainest:
I've run this twice on Leela, and it's timed out both times. I've made sure there aren't any errant processes, but it still won't run to completion. Could someone try this on a different ARM machine?

mxz297:
@hainest I run both master and this PR on arm1.cs.rice.edu. Both branches timed out at the same problem.

I ran this PR on leela. It timed out at test_thread_8, but then I ran the test individual many times, it never hang...

hainest:
@mxz297 Does aarch64 support traps? If so, does test_thread_8 exercise that code path? I'm inclined to believe this is another issue of problems with `runTests` rather than this patch causing something to fail. I just want to make sure we didn't accidentally break something on ARM.

hainest:
No regressions on Power9 (lassen.llnl.gov)

mxz297:
> @mxz297 Does aarch64 support traps? 

Yes, trap based trampolines are implemented mostly in an architecture independent way. Only the actual trap instruction is platform specific.

> If so, does test_thread_8 exercise that code path? 

I don't test_thread_8 is specifically designed to test this functionality but trap trampolines can be used. 

> I'm inclined to believe this is another issue of problems with `runTests` rather than this patch causing something to fail. I just want to make sure we didn't accidentally break something on ARM.

I agree that this is more likely a different issue from this PR. Non-deterministic hang has been there for a long time in the test suite.

hainest:
@mxz297 Ok. Since it's passing on PPC where the problem was initially described and the ARM hangs are the usual affair, then I'm comfortable merging this. Thanks!

PR #638, opened 2019-08-26
Capstone integration for x86 and x86-64

### LINKED ISSUE #402: More vector instruction decoding issues
@jdetter I carefully checked the binary that contains many vector instructions and caused many decoding errors. There are still quite some decoding mistakes, but we are much better than before. 

Below are a set of instructions that InstructionAPI reported as "No_Entry", but objdump shows otherwise.

```
5718aa:	62 f1 ed 48 fb d6    	vpsubq %zmm6,%zmm2,%zmm2
5c6972:	62 b1 3d 48 72 34 8e 03	 vpslld $0x3,(%rsi,%r9,4),%zmm8
61e150:	62 e2 7e 48 3a cb    	vpbroadcastmw2d %k3,%zmm17
62d197:	62 f2 6d 4a 64 c1    	vpblendmd %zmm1,%zmm2,%zmm0{%k2}
678edb:	c4 e3 79 0b c8 02    	vroundsd $0x2,%xmm0,%xmm0,%xmm1
6d3246:	62 b1 fd 08 2e c5    	vucomisd %xmm21,%xmm0
921129:	66 0f 3a 0a c0 02    	roundss $0x2,%xmm0,%xmm0
a1cfbb:	62 d2 f5 78 b8 d1    	vfmadd231pd {rz-sae},%zmm9,%zmm1,%zmm2
```

I think I have merged all your commits, but let me know if you are not able to reproduce any of these decoding issues.

mxz297:
@hainest Before testing this PR, there are a few issues that I would like discuss:

1. You would like the user to specify where to find Capstone, which indicates that the user also control which version of Capstone to provide. I think this means we only need to create a CMake option `CAPSTONE_ROOT` to find the installation of Capstone. Do you agree?

2. In Dyninst's `CMakeList.txt`, at line 27, I hard-coded that capstone's library install location is `${CAPSTONE_ROOT}/lib64`. On Red Hat Linux, capstone will generate a directory named `lib64`. 
On other linux distributions, capstone may generate a directory named `lib`. What is the best way to handle this problem?

3. This PR currently treats Capstone as a required dependency. It is reasonable that if the user does not want to use Capstone at all, so the user may not want to build Capstone. Do we want to make Capstone an optional, but recommended dependency? 

Fixes #402


hainest:
>Dyninst instrumentation on x86-64 needs some low level instruction decoding interfaces implemented in libcommon for determining where are the register and immediate operands in an instruction, then change these operands accordingly.

Is there possibly a way to provide this through Capstone or a wrapper around it?

>The x86-64 decoding code in instructionAPI is still reachable if the user explicitly constructs an instruction decoder with Dyninst as the backend.

This is what Bart didn't want- having two possible means of decoding x86 instructions since it would diverge our maintenance and testing efforts. I would be ok with marking it as deprecated now and removing it later if we really need that code to work in order to support the Capstone interface.

mxz297:
@hainest Ready for testing.

hainest:
https://bottle.cs.wisc.edu/branch/PR638

mxz297:
@hainest It looks like the regressions are with libc-2.28 & 2.29. I don't have accesses to a system with these glibc versions. Since you are able run tests with them, can you have me some accesses?

Meanwhile, do you happen to find out whether we can use spack to create a multi-version glibc system?

hainest:
@mxz297 Sorry, that interface is a bit confusing (it's on my list of things to fix... eventually). That first result indicates that tests which _passed_ on a machine with glibc-2.29 are now failing on coriander with glibc-2.27. The detailed [logs](https://bottle.cs.wisc.edu/logs/3c6d27d2-206b-400d-bd32-5f871e2f633f.tar.gz) are in <hash>/testsuite/tests.

>Meanwhile, do you happen to find out whether we can use spack to create a multi-version glibc system?

I am still working on this. My idea is to build the toolchains external to Spack and then use compilers.yaml to have it build the dependencies with each compiler/glibc combination. I am pretty sure I can do that, but the trick will be creating a module file to reflect which dependencies were built with which compiler.

hainest:
@mxz297 I just pushed a fix for how the regressions are calculated. If you go back to the regressions page for the latest run on coriander, you can search for "coriander.cs.wisc.edu" and see the last _successful_ run on that machine. For which now there are regressions. Sorry about that. It's a work in progress!

mxz297:
@hainest I see. However, I actually just ran the testsuite with the capstone_integration branch on coriander. It did not show these regressions. I will do a merge myself locally and see what's going on. 

Meanwhile, can you double-check your setting of testsuite? One question is that I cloned your latest test suite branch. The auto testing script will try to build Dyninst but I am not sure how to specify where to find Capstone in this case.

hainest:
@mxz297 That's interesting. I'll try running it again.

>The auto testing script will try to build Dyninst but I am not sure how to specify where to find Capstone in this case.

It's `-DCapstone_ROOT_DIR`. https://github.com/dyninst/dyninst-wiki/pull/1/files

hainest:
@mxz297 I re-ran both PR638 and the head of master on coriander, and found the following regressions: https://bottle.cs.wisc.edu/regressions/run?curid=105&previd=108

mxz297:
@hainest I pushed a few more commits that deal with 32-bit registers, invalid instructions, and so on.

On my end, I cannot reproduce the failure for test1_12. test1_12 is always passing on my end. My new commits should fix most of the regressions except for the following two:

1. fucompp still fails but it looks like a Capstone bug. I have reported the issue to Capstone.

2. test_instruction_farcall is weird. It has code 

```
if(decodedInsns.empty() || !decodedInsns[0].isValid() || decodedInsns[0].isLegalInsn())
   logerror("FAILED: %s\n", decodedInsns.empty() ? "no instructions decoded" : "first instruction was valid");
   return FAILED;
```

The comment suggest that we should expect an invalid instruction. But the code says if `isValid()` is false it will report test failure.

hainest:
>On my end, I cannot reproduce the failure for test1_12. test1_12 is always passing on my end.

Did you run this on coriander? Are you using group mode or running them individually?

>fucompp still fails but it looks like a Capstone bug. I have reported the issue to Capstone.

That raises the issue of bringing this into 10.2. Assuming it is a Capston bug and they can't fix it immediately, do we want to knowingly introduce a regression into a point release?

>The comment suggest that we should expect an invalid instruction. But the code says if isValid() is false it will report test failure.

Assuming that far calls are not allowed on AMD64 (are they?), then the test passes when the decoded data stream results is an illegal instruction. Though confusing, I think the logic here is correct given the comments on the class methods.

    //Returns true if this %Instruction object is valid.
    //  Invalid instructions indicate that an %InstructionDecoder has reached the end
    //  of its assigned range, and that decoding should terminate.
    !decodedInsns[0].isValid()

    // Returns true if this %Instruction object represents a legal instruction, as specified
    // by the architecture used to decode this instruction.
    decodedInsns[0].isLegalInsn()

The question then remains as to whether or not Capstone agrees with the underlying assumption that far calls are invalid on AMD64.

mxz297:
> Did you run this on coriander? Are you using group mode or running them individually?

Yes, I ran this on coriander. I have tried both `./runTest -all -64` and running them individually.

> 
> > fucompp still fails but it looks like a Capstone bug. I have reported the issue to Capstone.
> 
> That raises the issue of bringing this into 10.2. Assuming it is a Capston bug and they can't fix it immediately, do we want to knowingly introduce a regression into a point release?

It probably won't take them too long to confirm and fix this. Previously, my issues got resolved in a timely manner.

> //Returns true if this %Instruction object is valid.
> //  Invalid instructions indicate that an %InstructionDecoder has reached the end
> //  of its assigned range, and that decoding should terminate.
> !decodedInsns[0].isValid()

I don't think this comment is strictly correct based on my experiences of using instructionAPI. `isValid()` has been used to check whether the instruction represent a valid instruction. An example is the `test_instruction_profile` test:

https://github.com/dyninst/testsuite/blob/master/src/instruction/test_instruction_profile.C#L110

This test uses the while loop to do decoding address range check and use `isValid()` inside the while loop to determine whether we encounter a valid instruction.

If the comment above the declaration of `isValid()` is correct, then `isValid()` will never be false in the while loop. Then the code will be stuck in an infinite loop because bad instruction has size 0.

> // Returns true if this %Instruction object represents a legal instruction, as specified
> // by the architecture used to decode this instruction.
> decodedInsns[0].isLegalInsn()

I think `isValid()` and `isLegalInsn()` have been overlapping but slightly different. In my current implementation, they will be equivalent. 

> 
> The question then remains as to whether or not Capstone agrees with the underlying assumption that far calls are invalid on AMD64.

Capstone says far call is not valid on AMD64. But the test says if `!isValid()` then report failure. Capstone reports that far call is not valid and my integration code translates it in a way that `isValid()` return false.

mxz297:
@hainest I guess the more important thing is that I don't believe the comment above `isValid()` is that useful. The user needs to provide the a buffer starting address and a length to construct InstructionDecoder. A typical way to keep decoding until the end of a buffer is to keep track of the current offset in the buffer and use a while loop:

https://github.com/dyninst/testsuite/blob/master/src/instruction/test_instruction_profile.C#L104

hainest:
@mxz297 I'm having trouble getting the test suite to run consistently on coriander, but looking at the diff of regressions between the last three runs, it looks like everything is fixed except for 'fucompp' and 'test_instruction_farcall'.

>isValid() has been used to check whether the instruction represent a valid instruction.

On at least aarch64, it is also used to signal the end of the decoding process ([code](https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/InstructionDecoder-aarch64.C#L118)) because the default constructor sets `m_valid` to `false` ([code](https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/Instruction.C#L124)). So it looks like it's both- which isn't helpful.

mxz297:
@hainest Following our last discussion and feedback from Bart, I believe the plan is to ignore the regression on test fucompp. For test_instruction_farcall, I will create a PR for the test suite that includes the following changes:

diff --git a/src/instruction/test_instruction_farcall.C b/src/instruction/test_instruction_farcall.C
index 8826b2d8..62b6ee2c 100644
--- a/src/instruction/test_instruction_farcall.C
+++ b/src/instruction/test_instruction_farcall.C
@@ -82,7 +82,7 @@ test_results_t test_instruction_farcall_Mutator::executeTest()
     }
     while(i.isValid());
 #if defined(arch_x86_64_test)
-  if(decodedInsns.empty() || !decodedInsns[0].isValid() || decodedInsns[0].isLegalInsn())
+  if(decodedInsns.empty() || decodedInsns[0].isLegalInsn())
   {
     logerror("FAILED: %s\n", decodedInsns.empty() ? "no instructions decoded" : "first instruction was valid");
     return FAILED;

This will make this test passing for this PR. 

I will then create another PR to clean up the Instruction class interface issue.

hainest:
Fixes #402

mxz297:
@hainest I find some unpleasant issues with Capstone. I will write an email to detail the problems. Please don't merge this until the issues resolved.

hainest:
In Capstone, the detailed register information (e.g., read/write) is not available if it is compiled with CAPSTONE_DIET. We need a way to detect this during Dyninst's compilation.

mxz297:
@hainest I can port my patches for Dyninst's use of Capstone to this branch. To get this PR merged, besides filling read/write information from Xed, do we have any other issues to resolve?

hainest:
I'm waiting to hear back from the Capstone devs about their integration of LLVM 10. If we can work around that, then we'll just update their tables directly from Xed. After that, we should be ok- assuming testing works, of course.

mxz297:
@hainest On gpu.cs.rice.edu (a x86-64 machine), the test suite has the following failures:

```
test_reloc                 g++    none 64  create   NA      dynamic nonPIC  FAILED (Group Teardown)
test_reloc                 g++    none 64  attach   NA      dynamic nonPIC  PASSED
test_reloc                 g++    none 64  rewriter NA      dynamic nonPIC  CRASHED (Group Teardown)
test_reloc                 g++    none 64  create   NA      dynamic PIC     FAILED (Group Teardown)
test_reloc                 g++    none 64  attach   NA      dynamic PIC     PASSED
test_reloc                 g++    none 64  rewriter NA      dynamic PIC     CRASHED (Group Teardown)
test_reloc                 gcc    none 64  create   NA      dynamic nonPIC  FAILED (Group Teardown)
test_reloc                 gcc    none 64  attach   NA      dynamic nonPIC  PASSED
test_reloc                 gcc    none 64  rewriter NA      dynamic nonPIC  CRASHED (Group Teardown)
test_reloc                 gcc    none 64  create   NA      dynamic PIC     FAILED (Group Teardown)
test_reloc                 gcc    none 64  attach   NA      dynamic PIC     PASSED
test_reloc                 gcc    none 64  rewriter NA      dynamic PIC     CRASHED (Group Teardown)
```

However, the current master also has the same failures...

mxz297:
@hainest Can you launch a run on other platforms for this PR?

hainest:
I'm running the head of master on coriander now for a baseline. I'll try to run this PR tomorrow, assuming no major issues. Could you run the build script on gpu.cs.rice.edu so we can have a record and see its system configuration?

hainest:
@mxz297 The instruction 'vpperm' used to be in Dyninst, but isn't after this PR. That opcode is in Capstone, so it might be a processing error. Could you double check?

hainest:
Subsumed by #1646.

PR #639, opened 2019-08-28
Allow CMake files for dependencies to be included multiple times

hainest:
Previously, multiple inclusions would result in the library being looked up or downloaded/built multiple times. This ensures these actions happen only once.


hainest:
https://bottle.cs.wisc.edu/branch/PR639

PR #653, opened 2019-10-05
Fix instrumentation regressions for libc-2.29 on ARM

### LINKED ISSUE #652: Test suite regressions on ARM64 with glibc-2.29
On ARM64 (zeroah)
----

| Name        | mode   | link     | PIC     | result   |
| ----------- | ------ | -------- | ------- | -------- |
| test_pt_ls | create | dynamic | nonPIC | FAILED |
| test5_1 | create | dynamic | nonPIC | CRASHED |
| test5_1 | attach | dynamic | nonPIC | CRASHED |
| test5_1 | create | dynamic | PIC | CRASHED |
| test5_1 | attach | dynamic | PIC | CRASHED |
| test1_23 | create | dynamic | nonPIC | CRASHED |
| test1_24 | create | dynamic | nonPIC | CRASHED |
| test1_26 | create | dynamic | nonPIC | CRASHED |
| test1_23 | attach | dynamic | nonPIC | CRASHED |
| test1_24 | attach | dynamic | nonPIC | CRASHED |
| test1_26 | attach | dynamic | nonPIC | CRASHED |
| test1_23 | create | dynamic | PIC | CRASHED |
| test1_24 | create | dynamic | PIC | CRASHED |
| test1_26 | create | dynamic | PIC | CRASHED |
| test1_23 | attach | dynamic | PIC | CRASHED |
| test1_24 | attach | dynamic | PIC | CRASHED |
| test1_26 | attach | dynamic | PIC | CRASHED |
| test1_23 | create | dynamic | nonPIC | CRASHED |
| test1_24 | create | dynamic | nonPIC | CRASHED |
| test1_26 | create | dynamic | nonPIC | CRASHED |
| test1_23 | attach | dynamic | nonPIC | CRASHED |
| test1_24 | attach | dynamic | nonPIC | CRASHED |
| test1_26 | attach | dynamic | nonPIC | CRASHED |
| test1_23 | create | dynamic | PIC | CRASHED |
| test1_24 | create | dynamic | PIC | CRASHED |
| test1_26 | create | dynamic | PIC | CRASHED |
| test1_23 | attach | dynamic | PIC | CRASHED |
| test1_24 | attach | dynamic | PIC | CRASHED |
| test1_26 | attach | dynamic | PIC | CRASHED |

Additionally, the test suite hangs before running to completion.

mxz297:
The commit that I just pushed resolved lots of the crashes on libc-2.29 on aarch64.

Fixes #652

### DIFF
--- a/dataflowAPI/src/liveness.C
+++ b/dataflowAPI/src/liveness.C
@@ -567,6 +567,9 @@ ReadWriteInfo LivenessAnalyzer::calcRWSets(Instruction curInsn, Block *blk, Addr
       if (curInsn.getOperation().getID() == power_op_svcs) {
 	isSyscall = true;
       }
+      if (curInsn.getOperation().getID() == aarch64_op_svc) {
+          isSyscall = true;
+      }
       if (isInterrupt || isSyscall) {
 	ret.read |= (abi->getSyscallReadRegisters());
 	ret.written |= (abi->getSyscallWrittenRegisters());
--- a/dyninstAPI/src/dynProcess.C
+++ b/dyninstAPI/src/dynProcess.C
@@ -836,6 +836,18 @@ bool PCProcess::loadRTLib() {
    {
 	   startup_printf("%s[%d]: DYNINSTinit not called automatically\n", FILE__, __LINE__);
    }
+
+   // Install a breakpoint in DYNINSTtrapFunction.
+   // This is used as RT signal.
+   Address addr = getRTTrapFuncAddr();
+   if (addr == 0) {
+       startup_printf("%s[%d]: Cannot find DYNINSTtrapFunction. Needed as RT signal\n", FILE__, __LINE__);
+       return false;
+   }
+   if (!setBreakpoint(addr)) {
+       startup_printf("%s[%d]: Cannot set breakpoint in DYNINSTtrapFunction.\n", FILE__, __LINE__);
+       return false;
+   }
    startup_printf("%s[%d]: DYNINSTinit succeeded\n", FILE__, __LINE__);
    return setRTLibInitParams();
 }
@@ -3198,6 +3210,14 @@ Address PCProcess::getRTEventArg3Addr() {
     }
 
     return sync_event_arg3_addr_;
+}
+
+Address PCProcess::getRTTrapFuncAddr() {
+    if (rt_trap_func_addr_ == 0) {
+        func_instance* func = findOnlyOneFunction("DYNINSTtrapFunction");
+        rt_trap_func_addr_ = func->addr();
+    }
+    return rt_trap_func_addr_;
 }
 
 bool PCProcess::hasPendingEvents() {
--- a/dyninstAPI/src/dynProcess.h
+++ b/dyninstAPI/src/dynProcess.h
@@ -353,6 +353,7 @@ protected:
           sync_event_arg2_addr_(0),
           sync_event_arg3_addr_(0),
           sync_event_breakpoint_addr_(0),
+          rt_trap_func_addr_(0),
        thread_hash_tids(0),
        thread_hash_indices(0),
        thread_hash_size(0),
@@ -393,6 +394,7 @@ protected:
           sync_event_arg2_addr_(0),
           sync_event_arg3_addr_(0),
           sync_event_breakpoint_addr_(0),
+          rt_trap_func_addr_(0),
        thread_hash_tids(0),
        thread_hash_indices(0),
        thread_hash_size(0),
@@ -436,6 +438,7 @@ protected:
           sync_event_arg2_addr_(parent->sync_event_arg2_addr_),
           sync_event_arg3_addr_(parent->sync_event_arg3_addr_),
           sync_event_breakpoint_addr_(parent->sync_event_breakpoint_addr_),
+          rt_trap_func_addr_(parent->rt_trap_func_addr_),
        thread_hash_tids(parent->thread_hash_tids),
        thread_hash_indices(parent->thread_hash_indices),
        thread_hash_size(parent->thread_hash_size),
@@ -487,6 +490,7 @@ protected:
     Address getRTEventArg1Addr();
     Address getRTEventArg2Addr();
     Address getRTEventArg3Addr();
+    Address getRTTrapFuncAddr();
 
     // Shared library managment
     void addASharedObject(mapped_object *newObj);
@@ -592,6 +596,7 @@ protected:
     Address sync_event_arg2_addr_;
     Address sync_event_arg3_addr_;
     Address sync_event_breakpoint_addr_;
+    Address rt_trap_func_addr_;
     Address thread_hash_tids;
     Address thread_hash_indices;
     int thread_hash_size;
--- a/dyninstAPI/src/emit-aarch64.C
+++ b/dyninstAPI/src/emit-aarch64.C
@@ -274,9 +274,9 @@ void EmitterAARCH64::emitLoadOrigRegister(Address register_num, Register destina
    registerSlot *dest = (*gen.rs())[destination];
    assert(dest);
 
-   if (register_num == REG_SP) {
+   if (src->name == "sp") {
       insnCodeGen::generateAddSubImmediate(gen, insnCodeGen::Add, 0,
-             TRAMP_FRAME_SIZE_64, destination, REG_SP, true);
+             TRAMP_FRAME_SIZE_64, REG_SP, destination, true);
 
       return;
    }
@@ -284,7 +284,7 @@ void EmitterAARCH64::emitLoadOrigRegister(Address register_num, Register destina
    if (src->spilledState == registerSlot::unspilled)
    {
       // not on the stack. Directly move the value
-      emitMoveRegToReg((Register) register_num, destination, gen);
+      insnCodeGen::generateMove(gen, destination, (Register) register_num, true);
       return;
    }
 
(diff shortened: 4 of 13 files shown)


mxz297:
@hainest I plan to gradually push my fixes to this PR. Currently, this PR fixes several CRASH on zeroah. At this point, can you configure your test script to run `runTests` with `OMP_NUM_THREADS=1`?

If the test suite still hand with `OMP_NUM_THREADS=1`, this means the hang comes from Proccontrol or the test suite itself. Otherwise, it means the hang is mostly likely to be caused by parallel parsing. 

Can you re-run the whole suite on zeroah and update the regression test table?

mxz297:
@hainest I encountered hang twice when running `OMP_NUM_THREADS=1 ./runTests -all` on zeroah. The first time it hanged with `test_snip_remove`, and the second time it hanged with `test2_6`. 

Running with a single OpenMP thread still needs to non-deterministic hang. This indicates that the root cause is not in parallel code parsing. 

By the way, if we manually run the test suite and we observe a hang (for example, it takes more than 5 minutes to print any test results), we can use `ctrl-c` to kill this hanging test, and the `runTests` will continue to run remaining tests. 

It is possible to write a wrapper script to monitor stdout of `runTests` and then send SIGINT to `runTests` process, so that we can at least see the test suite to finish.

hainest:
@mxz297 I found the same behavior using `OMP_NUM_THREADS=1` on Zeroah, coriander, and my home machine, GalacticCentralPoint.

>Running with a single OpenMP thread still leads to non-deterministic hang. This indicates that the root cause is not in parallel code parsing.

Agreed. Were you able to rerun the hanging tests in non-group mode?

>By the way, if we manually run the test suite and we observe a hang (for example, it takes more than 5 minutes to print any test results), we can use ctrl-c to kill this hanging test, and the runTests will continue to run remaining tests. It is possible to write a wrapper script to monitor stdout of runTests and then send SIGINT to runTests process, so that we can at least see the test suite to finish.

I agree this is a workaround, but it makes determining regressions more difficult since we don't know which ones failed due to code breaks introduced by a PR and which were killed due to the hang issue. Since we've usually had better results running the tests individually (rather than in group mode), I think it might be useful to switch the build script to using that mode until we can track down the greater issue. I'll add a threadpool to the scipt so it doesn't take 6+ hours to run.

mxz297:
@hainest I found some other deterministic failures, so I will focus on those first.

hainest:
@mxz297 I rebuilt the test suite manually, but the issue remained. I also recompiled elfutils-0.176 from source (which seems to have fixed #669), but did not see any change. I will try 4c5254edc4953a126dc3ffe29aa0e533212d219a.

hainest:
@mxz297 I should have also mentioned that not setting `-DElfUtils_ROOT_DIR` when building Dyninst (thus using the system elfutils) gets rid of the hang. This is very bizarre, and I have no explanation for it.

mxz297:
Well... Not everything happened in Dyninst can be understood... The hang we saw is a real bug.

hainest:
@mxz297 With 4c5254e and dyninst/testsuite#117, `test_thread_5` is now passing with my custom-built elfutils.

hainest:
@mxz297 Frustratingly, the test is always passing with my custom build of elfutils, but only *sometimes* passes when using the system elfutils. I don't think it's worthwhile to spend any more time on this unless you really want to. I'll post again when I get the whole test suite to run.

hainest:
@mxz297 After adding 4c5254e to the dev branch, I'm seeing a non-deterministic failure in test2_8 (see the first two [results](https://bottle.cs.wisc.edu/commit/0b1c693ae89f209bdba7b074b9cbb4acd8640b1b)- same commits, different runs). The failure message is

```
**Failed test #8 (BPatch_breakPointExpr)
    process stopped on signal 38, not SIGSTOP
```

from line 505 of dyninst/testsuite#117. For glibc, `SIGRTMIN` is really `__libc_current_sigrtmin` which is not guaranteed to be idempotent. From [signal(7)](https://linux.die.net/man/7/signal), pthreads uses the first two real-time signals which, on Linux, would be 33 and 34. This would make `SIGRTMIN+4` -> 38 which is the signal the test received, but didn't expect. Using hard-coded real-time signal values is generally bad, but I am curious if we could skirt this issue by using something like `_NSIG - 4` to get a really high signal number that doesn't change with `SIGRTMIN`. We would just have to hope that no other application Dyninst can see would be using that one.

mxz297:
I have a PR for the testsuite for this change.

hainest:
dyninst/testsuite#117? These results include that in my local testsuite dev branch (not pushed it to GH, though).

mxz297:
Sorry. Misread your message. 

It means the Dyninst compilation will have SIGRTMIN with one value, but the test suite compilation will have a different SIGRTMIN value?

hainest:
>It means the Dyninst compilation will have SIGRTMIN with one value, but the test suite compilation will have a different SIGRTMIN value?

It is possible, yes. `SIGRTMIN` is just `__libc_current_sigrtmin()` which is a function exported by glibc.

mxz297:
I see. How is `_NSIG` defined? Is `SIGRTMAX` a fixed number or a function call as well?

hainest:
>How is _NSIG defined? Is SIGRTMAX a fixed number or a function call as well?

On my x86 machine with glibc-2.23, they are defined as

```
#define	_NSIG		65	/* Biggest signal number + 1 (including real-time signals).  */
#define SIGRTMIN        (__libc_current_sigrtmin ())
#define SIGRTMAX        (__libc_current_sigrtmax ())
```

On Zeroah with glibc-2.29, they are defined as

```
#define __SIGRTMIN      32
#define __SIGRTMAX      __SIGRTMIN

/* Biggest signal number + 1 (including real-time signals).  */
#define _NSIG           (__SIGRTMAX + 1)
```

Which means this whole business of real-time signals isn't portable at all. How fun.

mxz297:
I see. I guess we then simply use SIGUSR1 and tell users that if their application also uses SIGUSR1, please change it to a different signal number.

hainest:
It looks like `SIGUSR1` is being used by [pc_detach_mutatee.C:78](https://github.com/dyninst/testsuite/blob/master/src/proccontrol/pc_detach_mutatee.c#L78) and the stackwalker test framework, [stack_signal.cc:15](https://github.com/dyninst/testsuite/blob/master/src/stackwalker/stack_signal.cc#L15). It's also used to handle a PC bug in [FreeBSD](https://github.com/dyninst/dyninst/blob/master/proccontrol/src/freebsd.C#L335). `SIGUSR2` is explicitly handled in the [FreeBSD](https://github.com/dyninst/dyninst/blob/master/proccontrol/src/freebsd.C#L400) and [Linux](https://github.com/dyninst/dyninst/blob/master/proccontrol/src/linux.C#L241) proccontrol internals.

I'm not sure if any of these would prevent us from using either signal for `DYNINST_BREAKPOINT_SIGNUM`.

mxz297:
I am not that sure either. I don't like SIGBUS, but if we dial back to use SIGBUS, at least we can get the test passing. 

With my fix of checking signal number for determining dyninst RT message, using SIGBUS is not ideal, but at least no worse than before...

hainest:
I agree. I'm running the current dev branch on Zeroah right now. I've still had a lot of intermittent hangs there. Some of them are just crazy. I can run the permutations of `test_thread_5` individually, and they'll all pass. I can run them using `test_driver -all`, and they'll all pass. But if I run every test individually, then `test_thread_5` will sometimes hang. It's maddening.

mxz297:
Well... It takes some special design and efforts to make a software failing in this way...

hainest:
Good news! Your version of `test_thread_5` can hang, too. It's not a compilation problem after all. On Zeroah in the `/home/tim/paradyn/aHD9lLAP/testsuite/tests/isolation` directory, I've repeated the previous experiment where I can swap between my files and yours and then run the test several times with different configurations. That build uses the current dev branches of both [Dyninst](https://github.com/dyninst/dyninst/commits/dev) and [testsuite](https://github.com/dyninst/testsuite/commits/dev).

This will run the test using your files and your Dyninst installation
```
sh switch.sh xiaozhu
sh run.sh xiaozhu
```
Changing your name to mine will run my version.

Unfortunately, it's never the same permutation that fails when running this. However, the version `./test_driver -64 -none -g++ -dynamiclink -create -nonpic -test test_thread_5 -log tmp.log` has been more consistently failing between yours and mine. I would start looking there- even if you have to run it 5+ times to get it to hang.

mxz297:
@hainest  With the new commit I just pushed, I don't see any serious hang on zeroah. When I was running the test suite with `./runTest -all`, I saw a couple of test hanging. But when I ran these hanging tests individually, they don't hang.

I think we are in a much better shape now.

hainest:
@mxz297 Great. I'll start testing.

hainest:
This introduces [regressions](https://bottle.cs.wisc.edu/regressions/run?curid=186&previd=183) on x86. The most common failure message is

>test_driver: /nobackup/paradyn/4TopkEBM/dyninst/src/dyninstAPI/src/ast.C:469: AstVariableNode::AstVariableNode(std::vector<boost::shared_ptr<AstNode> >&, std::vector<std::pair<long unsigned int, long unsigned int> >*): Assertion `!ast_wrappers_.empty()' failed.

mxz297:
@hainest Are you sure the root cause is this PR? The crash location and the failure of test_local_var_locations seem to point to a failure in DWARF parsing for variables, where we fail to find variables.

hainest:
@mxz297 I grabbed the wrong run. The correction regressions are [here](https://bottle.cs.wisc.edu/regressions/run?curid=160&previd=183). The difference between master and dev are the commits from this PR and #667 (see `git log --oneline master..2e0ddc4`) in Dyninst. The test suite has dyninst/testsuite#117, dyninst/testsuite#118, dyninst/testsuite#106, and dyninst/testsuite#107 (see `git log --oneline master..d8255f5`).

hainest:
Good news! test_thread_5 is now passing on both ARM platforms. The fix seems to have been a combination of #680 and #667. The test suite still hangs when running in group mode on ARM, but I'm going to call this good for now and merge this.

PR #665, opened 2019-10-14
Reorder includes to fix hidden build dependencies

### LINKED ISSUE #658: Deterministic hang when handling user events
Executing

    ./test_driver -64 -none -gcc -dynamiclink -create -pic -test pc_thread

works perfectly fine. But executing

    ./test_driver -64 -none -gcc -dynamiclink -create -nonpic -test pc_thread

causes a deterministic hang on a variety of platforms: Power8 with gcc7.3.1 and glibc-2.17, Power9 with gcc-7.3.1 and glibc-2.17, and x86 gcc-8.3 with glibc-2.23 (I haven't tried on ARM).

On x86, I see that [supportsUserThreadEvents](https://github.com/dyninst/testsuite/blob/master/src/proccontrol/pc_thread.C#L436) returns false. Later, test_driver then checks [isTerminated](https://github.com/dyninst/testsuite/blob/master/src/proccontrol/proccontrol_comp.C#L911)​​ which returns false and so it assumes the process must be blocking for a user event. However, the launched process is already terminated. The log says: "Process 11035 not terminated, is stopped, is not exited, blocking for events". It then goes on to try to block for events, but the process never responds and hangs on the call to [select](https://github.com/dyninst/testsuite/blob/master/src/proccontrol/proccontrol_comp.C#L1362).

At least one of the following must be going on:

1. proccontrol is not correctly determining if the process can handle user events
2. proccontrol is correct and the process cannot handle user events, and thus test_driver needs check this before blocking
    - But then how would the suspending/tracking process then be terminated?
3. proccontrol is not correctly determining that the process is terminated

Why this is happening with PIC and not with nonPIC, I have no idea.

Ping @sashanicolas and @mxz297.

hainest:
Fixes #658 

cmake/shared.cmake includes cmake/cap_def_arch.cmake which conditionally sets cap_thread_db only if ThreadDB has been included (but does not include it, itself). This introduces a global include ordering which was broken in #636.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,20 +17,19 @@ set(RT_BINARY_DIR ${PROJECT_BINARY_DIR}/dyninstAPI_RT)
 
 set (CMAKE_MODULE_PATH "${DYNINST_ROOT}/cmake" "${DYNINST_ROOT}/cmake/Modules" ${CMAKE_MODULE_PATH})
 
-# This needs to be done first to set the global compiler flags
-include(shared)
-
-# PLATFORM is set by including 'shared'
-if(PLATFORM MATCHES "bgq")
-  message(WARNING "Support for Bluegene/Q has been deprecated and will be removed in a future release")
-endif()
-
 find_package(ThreadDB)
 find_package(Threads)
 include(Boost)
 include(ThreadingBuildingBlocks)
 include(ElfUtils)
 include(LibIberty)
+
+include(shared)
+
+# PLATFORM is set by including 'shared'
+if(PLATFORM MATCHES "bgq")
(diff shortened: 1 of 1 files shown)


hainest:
I'm still getting [failures](https://bottle.cs.wisc.edu/results/FAILED/136) and [crashes](https://bottle.cs.wisc.edu/results/CRASHED/136) on Power8. Power9 and x86 are fine. ARM is still hanging using [dyninst/dev](https://github.com/dyninst/dyninst/tree/dev) and [testsuite/dev](https://github.com/dyninst/testsuite/commits/dev). I will investigate further.

PR #666, opened 2019-10-14
Make ThreadDB required when building

hainest:
In the test suite, proccontrol causes hangs and zombie processes when it is not present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ if(PLATFORM MATCHES "bgq")
   message(WARNING "Support for Bluegene/Q has been deprecated and will be removed in a future release")
 endif()
 
-find_package(ThreadDB)
+find_package(ThreadDB REQUIRED)
 find_package(Threads)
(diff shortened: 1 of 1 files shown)


mxz297:
@hainest It is not correct to make ThreadDB required. ThreadDB is necessary for thread-level control in proccontrol, but it is not necessary for process level control. There are platforms that do not have ThreadDB available. 

@sashanicolas I remembered that a few Cray systems do not ThreadDB available, which caused you to fix the compilation when ThreadDB is not available.

hainest:
@mxz297 Ok, I won't merge this PR. I'll create a new test suite issue, instead. Although we should definitely confirm that the test suite will work when ThreadDB is not present.

sashanicolas:
@mxz297 That is true. We have "dummy" implementation of some thread_db classes for platforms that do not have it. I fixed one issue to allow it to compile but it was for poman.

hainest:
Closing as wontfix.

PR #667, opened 2019-10-14
Implement the check of ThreadDB at BPatch level

mxz297:
Add interface for checking whether thread level control is supported at the BPatch level.


sashanicolas:
I noticed after running on poman that

sashanicolas:
I noticed that when running the testsuite on poman I see this print out 
```
cerr << "Thread DB process constructor" << endl;
```
Which comes from the constructor I fixed in the past here:
https://github.com/dyninst/dyninst/commit/4fca3d7ec39175f4afcca382dba60e4bd7346fbe

hainest:
https://bottle.cs.wisc.edu/branch/PR667

This fixes several non-deterministic failures on x86. No regressions on ARM or Power.

PR #678, opened 2019-11-06
Add explicit ElfUtils dependency for ParseThat and examples

hainest:
When linking the binaries for ParseThat and the examples, the transitive linkage between libsymtabAPI and libelf was not correctly resolved by the linker. Instead, it used the system libelf which may or may not be correct.


hainest:
https://bottle.cs.wisc.edu/branch/PR678

The failures on x86 are non-deterministic. Re-running them manually gives different results each time, but each component of the failing tests passes at least once on _some_ run. No regressions on ARM.

hainest:
When linking the binaries for ParseThat and the examples, the transitive linkage between libsymtabAPI and libelf was not correctly resolved by the linker. Instead, it used the system libelf which may or may not be correct. This was discovered when using a non-system compiler with the --sysroot switch.

PR #700, opened 2020-01-29
Update Examples.tex file in dataflowAPI

mxz297:
Update the code examples in dataflowAPI so that they can compile.


mxz297:
@hainest I did not update the PDF file because the machine I am working on does not have pdflatex installed. Can you recompile the PDF and merge this PR?

wcohen:
Thanks, the revised examples compile for me.

PR #703, opened 2020-01-31
Remove undefined behavior from Symtab::Type

hainest:
Calling a member function of a base class before the base class's constructor has completed is undefined behavior.


hainest:
Wrong branch.

PR #704, opened 2020-01-31
Do not use non-trivial types in varargs

hainest:
How this is handled is implementation-defined.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64867


hainest:
https://bottle.cs.wisc.edu/branch/PR704

I've manually verified no regressions were introduced here.

PR #706, opened 2020-01-31
Remove undefined behavior from Symtab::Type

hainest:
Calling a member function of a base class before the base class's constructor has completed is undefined behavior.

I tried not to change `Type`'s API, but the member needed to be extracted in order to be handled during base class construction.

@mxz297 @blue42u Could you double-check that I didn't change the semantics?

### DIFF
--- a/symtabAPI/h/Type.h
+++ b/symtabAPI/h/Type.h
@@ -161,8 +161,6 @@ class SYMTAB_EXPORT Type : public Serializable, public  TYPE_ANNOTATABLE_CLASS
     **/
    bool updatingSize;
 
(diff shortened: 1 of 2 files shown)


blue42u:
Looks good to me.

mxz297:
@hainest Sorry for the delay.  Looks good to me.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR706&since_date=2020-03-10

hainest:
I've manually verified that this PR does not introduce any regressions (the regressions in the dashboard are carryover from other breakages).

PR #729, opened 2020-03-24
More parallel parsing optimization and fixes

mxz297:
This PR includes the parallel parsing optimizations and fixes implemented since last merge. 

@hainest  In particular, when I debug this PR on coriander, I fixes a problem in gap parsing. Before that fix, I was seeing quite often hangs for test_thread_*. After the gap parsing fixes, I don't see these hangs any more (with 4 full `./runTest -all -64` runs). Can you test this PR on other platforms to see if this will resolve the hang issue of test_thread_*?

Meanwhile, all rewriter tests for statically linked binaries failed. I will try to fix these failures in the next a few days.

I have tested this PR on gpu.cs.rice.edu and amd2.cs.rice.edu (gcc 7.5.0 and libc-2.27.so). All tests are passing.


hainest:
I'm testing this now.

>Meanwhile, all rewriter tests for statically linked binaries failed. I will try to fix these failures in the next a few days.

I believe @sashanicolas's previous email discussion about libgcc.a symbols was related to this.

mxz297:
@sashanicolas I am seeing

```
xmeng@coriander:~/dyninstapi/install/bin/testsuite$ ./test_driver -test test1_1 -staticlink
##### Failed to link to static library code into the binary: Symbol_Resolution_Failure = failed to locate symbol '__addtf3' for object 'strtof128_l.o'
test1_1                    g++    none 64  rewriter NA      static  nonPIC  FAILED (Group Teardown)
```

Is this the same error message you are seeing?

mxz297:
@hainest @sashanicolas 

It looks like on x86, we do not load `libgcc.a`. This has been fine until now. This means that on old x86 systems, `libc.a` does not really depend on `libgcc.a`. On newer systems such as coriander, `libc.a` starts to depend on `libgcc.a`.

mxz297:
@sashanicolas @hainest With my latest commit, the whole suite is passing on coriander with the caveat that I need to set `LD_LIBRARY_PATH` to help Dyninst find `libgcc.a`.

sashanicolas:
That's exactly the problem I've been working on. Same message.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR729&testsuite_branch=master&status=OK

This has *substantially* improved the number of tests that are passing across all platforms (except Power which seems to be immune to all of this). However, there are still some outstanding items:

1. One permutation of test_reloc is still failing on [cayenne](https://bottle.cs.wisc.edu/results/FAILED/610)

2. All test_reloc tests are failing on [arm1](https://bottle.cs.wisc.edu/results/FAILED/612).

3. There are hangs on both ARM machines ([1](https://bottle.cs.wisc.edu/results/HANGED/612), [2](https://bottle.cs.wisc.edu/results/HANGED/611)), but nowhere else.

All runs of the test suite were executed with a single `test_driver` job (see [`testsuite::run`](https://github.com/dyninst/testsuite/blob/master/scripts/build/Dyninst/testsuite.pm#L178)) and OMP_NUM_THREADS=2.

mxz297:
@hainest Thanks! At least we are making some progresses! I will investigate these problems.

mxz297:
@hainest I ran the test suite on arm1. It looks to me that the test_reloc failures are caused by out-of-memory. arm1 only has 8GB memory. I was seeing error message like:

```
Failed to create process ./test_reloc.dyn_gcc_64_none_none
```

Failing to create processes can only be attributed running out of memory. 

And I do not see the test1_36 failing on arm1... So, I guess test1_36 is also impacted by non-determinism.

hainest:
> It looks to me that the test_reloc failures are caused by out-of-memory. arm1 only has 8GB memory.

That's odd. I'm seeing the same message for the same permutations on Zeroah which has 64GB of memory.

>And I do not see the test1_36 failing on arm1... So, I guess test1_36 is also impacted by non-determinism.

Weird. That functionality is definitely broken on ARM.

mxz297:
@hainest I feel we should merge this PR as it clearly improves the test suite results and performs better than the current master as reported in #737.

While it does not resolve all the test suite problems, but it is not like that the current master branch is doing anything better.

hainest:
@mxz297 I'm re-running the test suite now. I wasn't sure if this work was complete yet, given that updates were still being made. Do you want to preserve the history of these commits?

mxz297:
> @mxz297 I'm re-running the test suite now. I wasn't sure if this work was complete yet, given that updates were still being made. Do you want to preserve the history of these commits?

Understandable. This branch was started to address the performance issue mentioned by Mark. Then I started to fix other issues. It will never be fully complete... It will be better that we have small branches with more frequent merges, so the PR's names will be more specific and meaningful. 

For this PR, it is fine to squash it.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR729&since_date=2020-04-09

Quite interestingly, the `test_reloc` failures on cayenne have gone away (I ran it twice to verify). There is one regression on quartz (x86_64 at LLNL). I'm re-running there to verify.

This is also the first time I've run on ulna at LLNL (thunderx2 ARM). There are fewer failures than on Zeroah and zero hangs.

hainest:
The failure on quartz wasn't real. Everything looks good.

PR #731, opened 2020-03-26
Add search of libgcc.a to x86 and aarch64

sashanicolas:
and warn user if it's not found.


hainest:
Where does `openResolvedLibraryName` search for files? libgcc.a is very unlikely to be in a system directory.

sashanicolas:
LD_LIBRARY_PATH

PR #746, opened 2020-05-13
Don't abort for instruction semantic errors.

stanfordcox:
Define ROSE_ASSERT so that a signal is thrown for instruction
semantic errors instead of doing an assertion abort.  SymbolicExpansion.C  includes x86InstructionSemantics.h which includes rose.h which defines ROSE_ASSERT but a subsequent redefinition of ROSE_ASSERT by x86_64InstructionSemantics.h has no effect since it is bracketed by ROSE_X86INSTRUCTIONSEMANTICS_H

### DIFF
--- a/dataflowAPI/rose/x86_64InstructionSemantics.h
+++ b/dataflowAPI/rose/x86_64InstructionSemantics.h
@@ -16,12 +16,13 @@
 #include "SgAsmExpression.h"
 #include "conversions.h"
 
-#ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
-
+#undef ROSE_ASSERT
 #define ROSE_ASSERT(x) \
     if (!(x)) {\
         throw rose::BinaryAnalysis::InstructionSemantics2::BaseSemantics::Exception("", NULL); \
     }\
+
+#ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
 
 /* Returns the segment register corresponding to the specified register reference address expression. */
 static inline X86SegmentRegister getSegregFromMemoryReference(SgAsmMemoryReferenceExpression* mr) {


dataflowAPI/rose/x86_64InstructionSemantics.h
```diff
@@ -16,13 +16,14 @@
...
 
-#ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
-
+#undef ROSE_ASSERT
 #define ROSE_ASSERT(x) \
     if (!(x)) {\
         throw rose::BinaryAnalysis::InstructionSemantics2::BaseSemantics::Exception("", NULL); \
     }\
 
+#ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
```
hainest, line 25: @mxz297 I realize this was the case before this change, but why is the file guard not guarding the entire file?

mxz297: @hainest I have no clue. This is a legacy source file from the ROSE project...

hainest: @mxz297 Neat

@stanfordcox Is there a specific reason for moving the definition of ROSE_ASSERT outside the file guard?

mxz297: @hainest To be clear, even though this is a legacy file, I add the ROSE_ASSERT macro so that it will throw an exception, instead of asserting. This worked fine for me.

However, this does not work for Stan as Stan explained that what I have written may lead to a macro-redefinition. So, by moving the macro outside of the file guard, it will force the macro to be effective for this file, and ensure that ROSE_ASSERT throws an exception instead of asserting.

stanfordcox: SymbolicExpansion.C includes
 /home/scox/dyninst/src/dataflowAPI/src/../rose/x86InstructionSemantics.h
  => #define ROSE_X86_64INSTRUCTIONSEMANTICS_H includes
  =>./home/scox/dyninst/src/dataflowAPI/src/../rose/rose.h
       => #define ROSE_ASSERT assert
/home/scox/dyninst/src/dataflowAPI/src/../rose/x86_64InstructionSemantics.h
  => #ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
  => #define ROSE_ASSERT(x)
  => getSegregFromMemoryReference
  => #endif
 does the ifndef block ever get invoked?  It seems everything SymbolicExpansion.C, SymVal.C includes both such that ROSE_X86INSTRUCTIONSEMANTICS_H is always defined

stanfordcox: awkward last sentence:  the files that include x86_64InstructionSemantics.h also include x86InstructionSemantics.h:  SymbolicExpansion.C, SymVal.C.   So I'm wondering if the block at #ifndef ROSE_X86INSTRUCTIONSEMANTICS_H is actually ever invoked since x86InstructionSemantics.h defines ROSE_X86INSTRUCTIONSEMANTICS_H

mxz297:
This PR is now part of #755

PR #748, opened 2020-05-19
Expand the list of non-returning functions in CodeSource.C.

### LINKED ISSUE #150: Non-returning functions from libstdc++ are not included in the non-returning function list
When parsing a binary provided by Victor van der Veen, the non-returning function "_ZSt20__throw_system_error@plt" from libstdc++ is not identified as non-returning. If we apply our non-returning function analysis to commonly linked system libraries such as libc and libstdc++, we should be able to generate a more comprehensive list of non-returning functions.

### LINKED ISSUE #745: parseapi missing several no-return functions
It seems that ParseAPI's list of non-returning functions is
incomplete.  Off the top of my head, it's missing _exit(), err()
and errx().  But you probably want to scan /usr/include for a more
complete list.

----------

Here's one example from my dlstress.c test in the repo:
https://github.com/HPCToolkit/hpctoolkit-tests/tree/master/stress/dlstress

The mk_thread_args() function has two ways to return: normal return at
end of function (returns void), or errx() if something earlier fails.
The function immediately following this is do_loop().

Compiled with -O, objdump reports:

```
0000000000400b54 <mk_thread_args>:
  400b54:       83 3d a5 18 20 00 07    cmpl   $0x7,0x2018a5(%rip) 
  400b5b:       7e 4e                   jle    400bab <mk_thread_args+0x57>
  400b5d:       48 c7 07 38 11 40 00    movq   $0x401138,(%rdi)
  400b64:       c7 47 08 01 00 00 00    movl   $0x1,0x8(%rdi)
  ...
  400baa:       c3                      retq   
  400bab:       48 83 ec 08             sub    $0x8,%rsp
  400baf:       be a8 12 40 00          mov    $0x4012a8,%esi
  400bb4:       bf 01 00 00 00          mov    $0x1,%edi
  400bb9:       b8 00 00 00 00          mov    $0x0,%eax
  400bbe:       e8 3d fc ff ff          callq  400800 <errx@plt>

0000000000400bc3 <do_loop>:
  400bc3:       41 57                   push   %r15
  400bc5:       41 56                   push   %r14
  400bc7:       41 55                   push   %r13
  400bc9:       41 54                   push   %r12
  ...
```

So, do_loop() is a real function and nm reports a T symbol.  But dyninst
computes the last block of mk_thread_args() as a normal call.  My
struct debug output says that parseapi reports this block with two out
edges, including a call/fall-through edge to 0x400bc3.

```
block:
stmt:  0x400bab (4)  l=188  f='/home/krentel/struct/hpctoolkit-tests/stress/dlstress/dlstress.c'
stmt:  0x400baf (5)  l=190  f='/home/krentel/struct/hpctoolkit-tests/stress/dlstress/dlstress.c'
stmt:  0x400bb4 (5)  l=190  f='/home/krentel/struct/hpctoolkit-tests/stress/dlstress/dlstress.c'
stmt:  0x400bb9 (5)  l=190  f='/home/krentel/struct/hpctoolkit-tests/stress/dlstress/dlstress.c'
stmt:  0x400bbe (5)  l=190  f='/home/krentel/struct/hpctoolkit-tests/stress/dlstress/dlstress.c'
out edges:  0x400800 (call, interproc)  0x400bc3 (call-ft)
```

As a result, hpcstruct puts the entire body of do_loop() inside
mk_thread_args() and the results in the viewer look bad.

```
  <P i="123" n="do_loop" l="212" v="{[0x400bc3-0x400bc4)}">
  </P>
```

It looks like parseapi is missing errx() as noreturn.  The <err.h>
header file has:

```
extern void errx (int __status, const char *__format, ...)
     __attribute__ ((__noreturn__, __format__ (__printf__, 2, 3)));
```

This is from a recent dyninst master.

mwkrentel:
This list comes from scanning the dwarf info in libc.so, etc for
the `DW_AT_noreturn` attribute, and from scanning header files in
/usr/include for the `__noreturn__` attribute.

Fixes #745
Fixes #150


hainest:
[Evidently](https://stackoverflow.com/questions/3547445/closing-multiple-issues-in-github-with-a-commit-message), "Fixes #A #B" only mentions B instead of closing it. Fun.

Could you also check if there are any in libgcc? We now include that sometimes (see #732).

mwkrentel:
Ok, I checked libgcc_s.so.1 from gcc 9.3.0.  My script reports nothing
and there are no DW_AT_noreturn tags reported from objdump.  So, no
no-return functions.

mwkrentel:
@hainest   Ok, I'm done.  I added the script and updated a few names.
The new names are just aliases, that is, names with the same address
as a noreturn name.

For example, libc.so has a function _exit() and an alias _Exit(), but
the dwarf only mentions _exit.  But adding the aliases (from nm) pretty
much covers all the names that were missing.

hainest:
@mwkrentel Everything is testing fine (as expected). I refactored the parsing script to use modern Perl idioms and split the functionality up to make it easier to comment so the next person to use it (likely me in 6 months) can remember what I'm supposed to do to make the magic happen. I verified my version against the libs you built on coriander. Could you take a look at it and let me know if you think I've added incorrect comments? https://github.com/hainest/dyninst/blob/noreturn_parser/parseAPI/src/noreturn.pl

mwkrentel:
Well, you edited the script a bit more than what I was expecting.
I guess there's old school and there's dinosaurs. :-) I suggest you
should add yourself as co-author.

It kinda breaks the purpose of the verbose mode.  The idea was that
for every name it added, you could see the input data that caused it
to be added.  By delaying the output, it splits the names from the
input data.

Anyway, it produces the same output, so go ahead.

hainest:
>Well, you edited the script a bit more than what I was expecting. I guess there's old school and there's dinosaurs. :-) I suggest you should add yourself as co-author.

Actually, I learned about the `/w` regex modifier and `$&` from this, so there's always something new to learn. Besides, I might be the only person who actually enjoys learning about and writing perl.

> It kinda breaks the purpose of the verbose mode. The idea was that for every name it added, you could see the input data that caused it to be added. By delaying the output, it splits the names from the input data.

Fixed. I also display the address the functions point to. I'm not sure it's needed, but it might be useful.

>Anyway, it produces the same output, so go ahead.

I don't have push permissions to your branch. Could you merge it in?

mwkrentel:
When I tried the new version with -v, I get a spew of these errors on
libpthread.  Maybe because pthread has a bunch of no-return futexes
that are inlined and thus have no actual symbol.

```
Use of uninitialized value in concatenation (.) or string at ./noret-tim.pl line 105.
Use of uninitialized value in concatenation (.) or string at ./noret-tim.pl line 105.
```

It looks like you undid the sort to accomodate verbose.  You didn't
have to do that.  If you want to keep the sort, you can keep it for
the non-verbose cases.  (Or, print an extra, sorted list at the end.)
It's up to you.  (I was just using /usr/bin/sort.)

Sorry, I don't mean to be running you around in circles. :-)

hainest:
>When I tried the new version with -v, I get a spew of these errors on libpthread. Maybe because pthread has a bunch of no-return futexes that are inlined and thus have no actual symbol.

Good catch! I didn't test that part. That was exactly the issue. Should we still print out the DWARF entry for inlined functions?

>You didn't have to do that. If you want to keep the sort, you can keep it for
the non-verbose cases. (Or, print an extra, sorted list at the end.)
It's up to you. (I was just using /usr/bin/sort.)

In true Unix fashion, I think piping through sort is better. :D

>Sorry, I don't mean to be running you around in circles. :-)

No worries. It was my fault for not testing it thoroughly enough.

mwkrentel:
> Should we still print out the DWARF entry for inlined functions?

I suggest yes.  Print the Dwarf entry for everthing that has the
noreturn attribute.  Of course, this is for verbose only.  For a
noreturn symbol, it has to have an actual symbol.

Btw, the point of verbose was to justify the answers.  We pick this
symbol and not that symbol, and here is why.  AFAIK, the only noreturn
entries that don't have symbols (and thus we don't count) are all
because of inlined.

> In true Unix fashion, I think piping through sort is better. :D

As you wish.  I left it that way simply out of convenience (ie, less
work) and unix separation of concerns.

mwkrentel:
@hainest  I see where you updated your script to fix the "uninitialized value"
bug, and I added your latest changes to my repo.  (I was kinda
expecting you'd ping me, but eventually I noticed.)

Anything else?

hainest:
@mwkrentel Sorry about that. I really thought I had sent you an alert. I'll re-test this quickly and get it merged. Thanks!

hainest:
Looks good. Thanks, Mark!

PR #755, opened 2020-05-27
Libxul fixes

### LINKED ISSUE #747: ParseAPI unable to parse a firefox shared object
A user is unable to use the systemtap dyninst backend to probe firefox

The problem is parsing libxul.so.  This is a  large shared objects, over 100M; by comparison libc is 3M. 

The problem is independent of systemtap and can be easily reproduced as shown in:
 https://scox.fedorapeople.org/dyninst/libxul.readme

Hopefully once libxul.so can be parsed then dyninst will be able to probe it which should get the systemtap dyninst backend working.

I have tried the above, with identical results (see libxul.readme) on fedora 31 and rhel 8.2 both x86_64 so seems independent of libc, tbb, boost version.

mxz297:
This PR includes necessary fixes for analyzing libxul.so from FireFox. 

Fix #747

### DIFF
--- a/dataflowAPI/rose/x86_64InstructionSemantics.h
+++ b/dataflowAPI/rose/x86_64InstructionSemantics.h
@@ -16,12 +16,13 @@
 #include "SgAsmExpression.h"
 #include "conversions.h"
 
-#ifndef ROSE_X86INSTRUCTIONSEMANTICS_H
-
+#undef ROSE_ASSERT
 #define ROSE_ASSERT(x) \
     if (!(x)) {\
         throw rose::BinaryAnalysis::InstructionSemantics2::BaseSemantics::Exception("", NULL); \
     }\
+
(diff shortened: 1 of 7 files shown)


hainest:
@stanfordcox Does this fix the problems you were seeing?

stanfordcox:
The 6 commits fix the libxul analysis error while doing bpatch.processAttach

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR755

Things are looking good. I want to re-run the test suite on quartz to make sure those errors aren't real regressions (I doubt they are; test_reloc has been a pain for a long time).

@mxz297 Once I'm convinced it's testing well, can I merge?

mxz297:
@hainest Yes, you can merge this PR when you feel good about it.

PR #760, opened 2020-05-30
Retrieve dyninstAPI examples from May 2019 manual

bbiiggppiigg:
The examples in the May 2019 dyninstAPI manual appendix are now standalone c++ source code placed under the path dyninstAPI/doc/examples


hainest:
This looks to be the same as https://github.com/dyninst/examples/pull/4. If so, we can close this one.

bbiiggppiigg:
I'm not sure if they are the same. 
This one builds with cmake, while dyninst/examples#4 works with Make

hainest:
Oh, in that case, we should figure out a way to merge them together. Ultimately, we'll want them to build with CMake.

PR #768, opened 2020-06-18
Retain hint function from external source

mxz297:
In earlier PR for parallel parsing, I changed the parser to only populate the function list during parsing finalization, which makes it convenient to delete bogus functions created during parsing. However, for external code source such as CUDA, we will not invoke parsing or finalization. This causes hpcstruct to generate empty CFG.

This PR fixes the empty CUDA CFG problem by retaining functions from external source when the code object is told to not parse or cannot perform parsing.

### DIFF
--- a/parseAPI/src/CodeObject.C
+++ b/parseAPI/src/CodeObject.C
@@ -78,6 +78,12 @@ CodeObject::CodeObject(CodeSource *cs,
     process_hints(); // if any
     if (!ignoreParse)
       parse();
+    else {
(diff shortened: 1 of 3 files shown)


mxz297:
@hainest This PR is for the CUDA CFG bug mentioned in today's UW/Rice call.

hainest:
https://github.com/dyninst/dyninst/pull/768

Looking good so far; just waiting for three more to finish.

hainest:
Looks good. Merging.

PR #779, opened 2020-07-13
Remove Aggregate::setModule declaration

hainest:
This was introduced in c848409ec, but never implemented.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR779&status=OK

PR #317, opened 2017-01-27
Memory leak fixes

wrwilliams:

PR #793, opened 2020-08-06
Fix memory leak in SymEval::expandInsn

hainest:
This was originally part of https://github.com/dyninst/dyninst/pull/317


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR793&since_date=2020-08-20

PR #813, opened 2020-08-25
Make ~DynObject virtual

### LINKED ISSUE #802: DynObject needs a virtual dtor

hainest:
This is base class for dyninstAPI/mapped_object, so it should have a
virtual destructor.

Fixes #802


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR813&since_date=2020-08-29

PR #817, opened 2020-08-27
Fix cmake configuration with -pie

stanfordcox:
Remove DYNINST_RT_CMAKE_C_FLAGS which causes a mismatch when configuring dyninstAPI_RT with -pie

This is just the minimum patch.  It builds and links with the standard *_FLAGS.  I wonder about the if(BUILD_RTLIB)
loop step to allow building with a separate compiler.  I cannot foresee a situation on Linux where doing that is advantageous.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR817&status=OK

No new regressions. Looks good. Thanks, Stan!

PR #844, opened 2020-09-13
Replace BPatch_vector internal implementation with std::vector

### LINKED ISSUE #718: Remove BPatch_Vector usage
Just that! It's been deprecated.

hainest:
This keeps the BPatch_vector type, but only as an alias to std::vector.

Closes #718 

@sashanicolas I did not remove all usages of BPatch_vector since we can just alias it in a type-safe manner and then we don't have to update the documentation.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR844&status=OK

Testing fine. @sashanicolas Let me know if this is good for you. Also, check the test suite change dyninst/testsuite#166. Thanks!

PR #847, opened 2020-09-14
Remove bluegene support

hainest:


hainest:
@mplegendre 

Is [`Thread::getCallStackUnwinding`](https://github.com/dyninst/dyninst/blob/master/proccontrol/h/PCProcess.h#L642) only supported by Bluegene? If so, should we remove it here?

Was [asyncIO](https://github.com/dyninst/dyninst/blob/master/proccontrol/src/int_process.h#L444) only for Bluegene? If so, should we remove it entirely?

PR #855, opened 2020-09-19
Some necessary changes for overlapping text regions

mxz297:
A rebased replacement PR for #412


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR855

The regressions are from test_reloc hanging and are unrelated.

PR #858, opened 2020-09-21
parseThat: remove autotools build files

hainest:
It has been building with CMake since 2016, so we don't need a second
option.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR858

Building fine.

PR #859, opened 2020-09-21
Remove vxworks support

hainest:
Support for VxWorks was added by 02bc600ea557d2ad16bc442ff161d0f110bf14e3 in March 2010, and the code broken by 0bc3980d172bbb8f698b674df56cb4b9b320f829 in November 2011. It has also never been part of the CMake build system (ca. 2013).

@n738cu We had discussed this platform previously, but I've since discovered that it hasn't worked in nearly a decade. The question is do we want to leave updating this platform as a TODO or remove it entirely?


n738cu:
I think that we can cut this vxworks support.  It was a Mary project and we’re not likely to work on it.

—bart

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR859

No change in build status.

PR #861, opened 2020-09-22
Cleanup dyn_hash_{set,map}

hainest:
Require std::unordered_{map,set} for dyn_hash_{map,set}, and convert the dyn_hash_* macros to alias templates for better type safety and error messages. Remove the custom hashers, and use typed enums for IA32Regs, entryID, and prefixEntryID.

### DIFF
--- a/common/h/dyntypes.h
+++ b/common/h/dyntypes.h
@@ -53,70 +53,22 @@
 #define XLC
 #endif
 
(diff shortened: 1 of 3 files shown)


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR861

PR #867, opened 2020-09-25
Remove memory emulation from AddressSpace

hainest:
This was disabled by 1b8bed2c7 in 2010. Its presence prevents correcting
a memory corruption bug in addressSpace::copyAddressSpace.

### DIFF
--- a/dyninstAPI/src/addressSpace.C
+++ b/dyninstAPI/src/addressSpace.C
@@ -50,7 +50,6 @@
 #include "Relocation/Transformers/Include.h"
 #include "Relocation/CodeTracker.h"
 
-#include "MemoryEmulator/memEmulator.h"
 #include "parseAPI/h/CodeObject.h"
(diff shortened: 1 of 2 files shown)


mxz297:
@hainest I remember this memory emulation thing is used for malware analysis where Dyninst will intercept and emulate every memory access. The whole malware analysis thing has not been tested for several years and it is only done for Windows. You may want to ask Bart in terms of what to do with malware analysis code.

hainest:
@mxz297 Yeah, I noticed that. I was hoping to just remove the functionality in AddressSpace since it had been disabled long ago (under the assumption that I wouldn't break anything further), but I agree that we should probably take a look at the memory emulation feature as a whole.

hainest:
I've decided against this change. After more careful consideration, my fix to `copyAddressSpace` wouldn't be any better than what is currently there without possibly breaking the pre/post-conditions of the function.

PR #868, opened 2020-09-27
Remove include cycle in dyntypes.h

hainest:
This prevented using concurrent data structures in dyntypes.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR868

No build issues on all three platforms.

mxz297 (approved):

PR #870, opened 2020-09-29
Update documentation for ParseAPI::CodeObject::CodeObject

hainest:
Make an explicit note that the CodeSource is not owned by the
CodeObject. This clarifies the ownership semantics.

There was a large "fix" in #317 that tried to take ownership of the `CodeSource` passed to the `CodeObject` constructor. However, there are many places in the codebase where we do not pass a heap-allocated pointer to the constructor (e.g., https://github.com/dyninst/dyninst/blob/a1e3391edace61a18d99fe7318ea2709914d796c/dyninstAPI/src/image.C#L517). In those cases, destruction would lead to heap corruption. It is best just to leave the management of the `CodeSource` object to the user and make sure they understand that the lifetimes of the two objects *must* be the same. Pointer semantics are still needed to maintain polymorphic behavior.


hainest:
There was a large "fix" in #317 that tried to take ownership of the `CodeSource` passed to the `CodeObject` constructor. However, there are many places in the codebase where we do not pass a heap-allocated pointer to the constructor. In those cases, destruction would lead to heap corruption. It is best just to leave the management of the `CodeSource` object to the user and make sure they understand that the lifetimes of the two objects *must* be the same.

PR #871, opened 2020-09-29
Correctly destruct AddressSpace objects

hainest:
`deleteAddressSpace` is still highly suspect because its only use is `copyAddressSpace` which doesn't fully copy the object. However, this fix at least makes `deleteAddressSpace` and `~AddressSpace` correct and perform the same actions.

This was originally part of #317.

### DIFF
--- a/dyninstAPI/src/addressSpace.C
+++ b/dyninstAPI/src/addressSpace.C
@@ -154,7 +154,6 @@ void AddressSpace::copyAddressSpace(AddressSpace *parent) {
     // This is only defined for process->process copy
     // until someone can give a good reason for copying
     // anything else...
-
     assert(proc());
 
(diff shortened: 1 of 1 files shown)


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR871

No regressions (cori@NERSC is offline right now).

mxz297 (approved):

mxz297:
@hainest Looks good. I don't think this memory leak fix will make much difference as AddressSpace objects do not get destructed. But this is still good to fix.

hainest:
>AddressSpace objects do not get destructed.

Yikes. I did some digging and found that there are circular lifetimes between `BinaryEdit` and `mapped_object` which prevents correct destruction. I'll add it to my TODO list...

PR #874, opened 2020-10-02
Add arch address for intel gen9 gpus

Jokeren:


hainest:
We don't have a way of testing this. As long as it's working for you, then I'll merge.

Jokeren:
@hainest Thanks for your reply. Yes, it works for me.

PR #875, opened 2020-10-02
Add lookup by name in block_instance::callee

### LINKED ISSUE #852: test5_7 failure on coriander
This fails at [test5_7.C:82](https://github.com/dyninst/testsuite/blob/master/src/dyninst/test5_7.C#L82) when `index`==4 (i.e., the inst point is [`template_test::func_cpp`](https://github.com/dyninst/testsuite/blob/master/src/dyninst/test5_7_mutatee.C#L69)). The root cause is that the target edge in the CFG says the callee is `__stack_chk_fail` at address 0x4215744. But calling `getPltFuncs` on the image says that function is at 0x4215088. The gdb log contains the details ([test5_7.gdb.log](https://github.com/dyninst/dyninst/files/5229463/test5_7.gdb.log)).

@mxz297 I'm out of my depth here, so I have no idea where to go next. Do you have any thoughts? This is on coriander under my home directory. You have sudoer privileges there, so you can go re-run this, if you like. My gdb config is

    gdb --args ./test_driver -g++ -none -64 -create -dynamiclink -nonpic -test test5_7 -log tmp.log
    set env OMP_NUM_THREADS=2
    set env DYNINSTAPI_RT_LIB=/home/tim/paradyn/cMYYZBGM/dyninst/lib/libdyninstAPI_RT.so
    set env LD_LIBRARY_PATH=/home/tim/paradyn/cMYYZBGM/testsuite/tests:/home/tim/paradyn/cMYYZBGM/dyninst/lib:/var/software/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/perl-5.30.2-2zeizvc4oxifswn43n25of2d2mc645gk/lib:/var/software/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/libxml2-2.9.10-folxclnhb4zjtfo7xorl2o25iuswdlst/lib:/var/software/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/intel-tbb-2020.2-hqa5vzggodkis5bca2d7cudhzy3vwyxd/lib:/var/software/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/elfutils-0.179-wrifubs3rbr6lqviq266kaeiaeorji5e/lib:/var/software/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/boost-1.73.0-tkvazvirvs6mnongrmrqbkmklve6cf5n/lib
    handle SIGUSR2 nostop
    b test5_7.C:82 if index==4

This is the second half of #778

hainest:
Sometimes, the PLT address and the CFG target aren't the same (e.g., Intel's CET causes this), so we just look up by name.

Fixes #852

### DIFF
--- a/dyninstAPI/src/block.h
+++ b/dyninstAPI/src/block.h
@@ -128,7 +128,7 @@ class block_instance : public Dyninst::PatchAPI::PatchBlock {
  private:
     void updateCallTarget(func_instance *func);
     func_instance *findFunction(ParseAPI::Function *);
-
+    func_instance* callee(std::string const&);
     // edges srcs_;
     // edges trgs_;
 
--- a/dyninstAPI/src/unix.C
+++ b/dyninstAPI/src/unix.C
@@ -750,8 +750,7 @@ func_instance *block_instance::callee() {
    // get the relocation information for this image
    Symtab *sym = obj()->parse_img()->getObject();
    std::vector<relocationEntry> fbt;
-   vector <relocationEntry> fbtvector;
-   if (!sym->getFuncBindingTable(fbtvector)) {
+   if (!sym->getFuncBindingTable(fbt)) {
       return NULL;
    }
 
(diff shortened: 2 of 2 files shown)


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR875

This fixes the test5_7 failures on coriander.

hainest:
When combined with dyninst/testsuite#170, all of the breakages on coriander are fixed.

hainest:
@mxz297 This is testing fine now and it fixed the test failure. One thing I changed is `block_instance::callee(std::string const& target_name)` now will try both `dynamic_cast` checks instead of having them be mutually exclusive. I couldn't think of a reason why this would cause an issue and it *might* increase the change of finding a callee. Do you see anything wrong with that change?

mxz297 (approved):

PR #876, opened 2020-10-03
cleanup memoryTracker memory handling

hainest:
This gives explicit copy and move semantics to the class and uses a std::unique_ptr to manage the contained buffer. Construction is also simplified by always using calloc.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR876

This is showing substantial regressions across the board from memory corruptions.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR876&since_date=2020-10-04

No new regressions.

PR #878, opened 2020-10-04
Provide correct default constructor for parse_func class

hainest:
This fixes a memory corruption bug when destroying default-constructed objects of this class.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR878

No regressions (the ones on coriander and ulna are unrelated).

PR #879, opened 2020-10-04
Fix memory leaks in BinaryEdit::openResolvedLibraryName

hainest:

@wrwilliams 
1. I didn't use SymElfFactor::openSymbolReader from the original fixes in #317 because it adds a link dependency to unix.o. Instead, I used a lambda with admittedly funky semantics.

2. For the static executable case, previously only the first archive member was checked. If it was not a compatible binary, the retMap was cleared (approx. line 605 in c0acebbe7). Now, all members are checked, and the retMap is never cleared. Is this a significant algorithmic change?

@mxz297 

Could you take a close look at dca0e46f3 to make sure I didn't inadvertently change the overall behavior?

### DIFF
--- a/dyninstAPI/src/unix.C
+++ b/dyninstAPI/src/unix.C
@@ -553,6 +553,7 @@ bool PCProcess::startDebugger() {
 
 #include "dyninstAPI/src/binaryEdit.h"
 #include "symtabAPI/h/Archive.h"
+#include <memory>
 
 using namespace Dyninst::SymtabAPI;
 
@@ -570,17 +571,28 @@ mapped_object *BinaryEdit::openResolvedLibraryName(std::string filename,
                        FILE__, __LINE__, filename.c_str());
 
         Symtab *origSymtab = getMappedObject()->parse_img()->getObject();
+
+        // A little helper to fix some clunky checks
+        auto is_compatible =
+            [this](std::string const& path, std::string const& member={}) {
+                auto temp = std::unique_ptr<BinaryEdit>{
+                	BinaryEdit::openFile(path, mgr(), patcher(), member)
+                };
+                if(temp && temp->getAddressWidth() == getAddressWidth()) {
+                    return temp;
+                }
+                temp.reset(nullptr);
+                return temp;
+            };
+
 	assert(mgr());
         // Dynamic case
         if ( !origSymtab->isStaticBinary() ) {
             for(pathIter = paths.begin(); pathIter != paths.end(); ++pathIter) {
-               BinaryEdit *temp = BinaryEdit::openFile(*pathIter, mgr(), patcher());
-
-                if (temp && temp->getAddressWidth() == getAddressWidth()) {
-                    retMap.insert(std::make_pair(*pathIter, temp));
-                    return temp->getMappedObject();
+                if (auto temp = is_compatible(*pathIter)) {
+                    auto ret = retMap.insert(std::make_pair(*pathIter, temp.release()));
+                    return (*ret.first).second->getMappedObject();
                 }
-                delete temp;
             }
         } else {
             // Static executable case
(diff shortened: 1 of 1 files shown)


mxz297 (changes requested):
@hainest  Overall looks good. It looks to me that it may be better to keep the original behavior that clearing `retMap` when finding an incompatible archive member. Here is my guess:

1. Let's say we have multiple paths resolved. 
2. The first path points an archive file, where part of the members are compatible and some of the members are not compatible. TBH, I am not sure whether this is possible... 
3. In the original code, all members from the first path would be cleared and the code will move on to the next path
4. In your change, the compatible members in the first path will cause the function to return and potentially return the wrong library.

I am not sure how likely this scenario to happen...

hainest:
@mxz297 That makes sense. I wasn't sure why that change was made in the original fixes, so I wanted to highlight it. I'll change it back. Thanks!

wrwilliams:
> This was originally part of #317
> 
> @wrwilliams
> 
> 1. I didn't use SymElfFactor::openSymbolReader from the original fixes in #317 because it adds a link dependency to unix.o. Instead, I used a lambda with admittedly funky semantics.
> 2. For the static executable case, previously only the first archive member was checked. If it was not a compatible binary, the retMap was cleared (approx. line 605 in [c0acebb](https://github.com/dyninst/dyninst/commit/c0acebbe7acf8e9c8ce49e300ec15e8a4cd0ca2c)). Now, all members are checked, and the retMap is never cleared. Is this a significant algorithmic change?
> 
> @mxz297
> 
> Could you take a close look at [dca0e46](https://github.com/dyninst/dyninst/commit/dca0e46f35ebc38f997e45fb7851eb25a8e997fc) to make sure I didn't inadvertently change the overall behavior?

Assuming that we're talking about ELF archives exclusively, I believe they're required to be single-format. However, given the number of things that could legitimately be called an "archive" these days (various fat binaries for GPU code etc) I suspect that the notion of "is this archive compatible to link in" is the wrong concept: what you want is "here are functions/symbols we seek and names of places to look for them" in keeping with linker behaviour in static linking. Or you want to ensure that "Archive" is only static libraries of consistent internal type and use something else for the generic statically-linkable heterogeneous piles of stuff.

At the time I wrote this I did not consider a unix.o dependency a big deal, as this was all ELF-specific code and ELF was considered UNIX-specific. If that's changed, obviously refactoring to something more platform-independent is a good choice.

hainest:
@wrwilliams 

>Assuming that we're talking about ELF archives exclusively, I believe they're required to be single-format. However, given the number of things that could legitimately be called an "archive" these days (various fat binaries for GPU code etc) I suspect that the notion of "is this archive compatible to link in" is the wrong concept: what you want is "here are functions/symbols we seek and names of places to look for them" in keeping with linker behaviour in static linking. Or you want to ensure that "Archive" is only static libraries of consistent internal type and use something else for the generic statically-linkable heterogeneous piles of stuff.

I completely agree, but that seems like a much bigger change that fixing the memory leaks here. I'll pass this along to the student currently working on GPU stuff as that will need to be addressed at some point through his work.

>At the time I wrote this I did not consider a unix.o dependency a big deal, as this was all ELF-specific code and ELF was considered UNIX-specific. If that's changed, obviously refactoring to something more platform-independent is a good choice.

I don't think it would be a problem. I was just feeling too lazy to change the build system, so I took the easier route since the `SymbolReader` was just being used to get the address width -- the same as `BinaryEdit::openFile` -- and the latter was possibly being used anyway.

hainest:
@mxz297 I added back the omission of incompatible archives for the static case. Could you double-check? Thanks!

wrwilliams:
@hainest 

> I completely agree, but that seems like a much bigger change that fixing the memory leaks here. I'll pass this along to the student currently working on GPU stuff as that will need to be addressed at some point through his work.

Indeed. My concern here was that if `Archive` has had semantic drift since I opened this PR back in the Before Times(tm), preserving the old behaviour precisely is largely counterproductive.

I believe the algorithmic change you propose is correct for heterogeneous archives, with the following extra conditions:

1. You definitely want a compatibility check rather than a simple address width check: architecture, endianness, ABI version at a minimum. (And when you're considering GPU stuff, that becomes a check against a list of architectures of course...)
2. You want to hand back an ordered list of candidate BinaryEdits rather than a single matching one, with function resolution proceeding in order if it's not explicitly specified otherwise.
3. Somehow the incompatible stuff needs to be filtered out of the BinaryEdit that's handed back, if there is any.

AddressWidth should die in a fire already and become an architecture trait instead...

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR879

No regressions.

hainest:
@bbiiggppiigg This is a long-term item, but we'll eventually want to handle the items in Bill's [comment](https://github.com/dyninst/dyninst/pull/879#issuecomment-704112371).

mxz297 (approved):

PR #880, opened 2020-10-04
Remove unused lineDict class

hainest:
This was originally part of #317.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR880

No breakages on all three platforms.

PR #881, opened 2020-10-04
Remove CFGFactor::destroy_all

hainest:
This removes the possiblity of object reuse and its associated undefined
behavior. This was originally part of #317.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR881

PR #882, opened 2020-10-05
Require standards-compliant c++11 ABI

hainest:
This has technically been the case for the last two releases, but now we strictly enforce it.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR882

No new regressions.

PR #883, opened 2020-10-06
CodeSource destructor cleanup

hainest:
This ensures that CodeSource is responsible for cleanup up its own objects rather than relying on the derived class to do it. This is particularly important since this class is part of the pubic API.


mxz297:
@hainest I would recommend moving the implementation of the destructor to `parseAPI/src/CodeSource.C`. In this way, if we need to modify the destructor in the future, we don't introduce changes to public header files, which causes long cascading compilation...

hainest:
@mxz297 I wasn't sure if I should leave it that way since the default dtor was inlined, but I agree it's better to put it in the source file.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR883

No new regressions.

PR #891, opened 2020-10-16
code cleanup integer funcs to * and / by constants

kupsch:
* remove functions to multiply and divide by 1000 and 1000000 using
  adds, subtracts and shifts.  Not used, not needed.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR891

Looks good. You can merge when you like.

PR #895, opened 2020-10-21
Fix memory leaks in emitElf

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR895&status=OK

No new regressions.

PR #900, opened 2020-10-27
Add initial support for analyzing AMDGPU binaries

bbiiggppiigg:
This pull request contains the initial effort to support analyzing AMD GPU binaries.

Right now decoding is only supported for the kernels using the VEGA instruction set, and is not even complete because we yet to get the documents from AMD, and all the code is tested only with gfx-803.

When we want to add the support for another AMDGPU ISA, the structure of the code will likely undergo huge changes.


dataflowAPI/rose/semantics/Registers.C
```diff
@@ -643,6 +644,31 @@ RegisterDictionary::print(std::ostream &o) const {
 //    }
 //    return regs;
 //}
+//
+/** AMDGPU registers.
+ * There are a total of 104 scalar general purpose registers each 32 bits wide. Each of these registers can be addressed as its 32-bit or 64-bit form */
+const RegisterDictionary *
+RegisterDictionary::dictionary_amdgpu() {
```
hainest, line 651: Leaky singletons are my least favorite singleton. Could we change this to

    RegisterDictionary const& RegisterDictionary::dictionary_amdgpu() {
        static std::once_flag initialized;
        static RegisterDictionary regs;

        std::call_once(...)

        return regs;
    }

or

    RegisterDictionary const* RegisterDictionary::dictionary_amdgpu() {
        static std::once_flag initialized;
        static RegisterDictionary regs;

        std::call_once(...)

        return &regs;
    }

if you need to preserve the pointer semantics of the return value? Both versions are safe in terms of lifetimes and threading.

dataflowAPI/src/SymbolicExpansion.C
```diff
@@ -112,3 +113,17 @@ bool SymbolicExpansion::expandAarch64(SgAsmInstruction *rose_insn, BaseSemantics
...
 }
 
+bool SymbolicExpansion::expandAmdgpu(SgAsmInstruction *rose_insn, BaseSemantics::RiscOperatorsPtr ops, const std::string &insn_dump) {
+    SgAsmAmdgpuInstruction *insn = static_cast<SgAsmAmdgpuInstruction *>(rose_insn);
+
+    BaseSemantics::DispatcherPtr cpu = DispatcherAMDGPU::instance(ops, 64);
+
+    try {
+        cpu->processInstruction(insn);
+    } catch (rose::BinaryAnalysis::InstructionSemantics2::BaseSemantics::Exception &e) {
```
hainest, line 123: Do we really want to just swallow this exception?

instructionAPI/src/ArchSpecificFormatters.C
```diff
@@ -149,6 +149,61 @@ std::string ArmFormatter::formatBinaryFunc(std::string left, std::string func, s
...
+    binaryFuncModifier["<<"] = "lsl";
+}
+
+std::string AmdgpuFormatter::formatImmediate(std::string evalString) {
+    return "0x" + evalString;
+}
+
+std::string AmdgpuFormatter::formatRegister(std::string regName) {
+    std::string ret = regName;
+    std::transform(ret.begin(), ret.end(), ret.begin(), ::toupper);
```
hainest, line 164: As of C++11, having the destination iterator be part of the input range is undefined behavior. Changing this to a simple range-based loop is the easy fix.

    for(auto &c : ret) ::toupper(c);

Also, there's no need to make a copy of the parameter since it's passed by value.

instructionAPI/src/Ternary.C
```diff
@@ -0,0 +1,131 @@
...
+        
+        /*std::string name = m_Reg.name();
+        std::string::size_type substr = name.rfind(':');
+        if(substr != std::string::npos)
+        {
+            name = name.substr(substr + 1, name.length());
+        }*/
+
+        /* we have moved to AT&T syntax (lowercase registers) */
+        std::transform(name.begin(), name.end(), name.begin(), ::toupper);
```
hainest, line 97: Same as above.

symtabAPI/src/Object-elf.C
```diff
@@ -677,6 +677,15 @@ bool Object::loaded_elf(Offset &txtaddr, Offset &dataddr,
 
             if (txtaddr == 0)
                 txtaddr = scn.sh_addr();
+            // skip header for amdgpu
+            
+            if (getArch() == Dyninst::Arch_amdgpu){
```
hainest, line 682: Empty block. Can be removed?

symtabAPI/src/debug.C
```diff
@@ -35,7 +35,7 @@
 #include "debug.h"
 
 int sym_debug_create = 0;
-int sym_debug_parsing = 0;
+int sym_debug_parsing = 1;
```
hainest, line 38: Let's not enable this by default. It's quite noisy.

instructionAPI/src/Ternary.C
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    void TernaryAST::apply(Visitor* v)
+    {
+        //v->visit(this); // TODO need to support this in visitor
+    }
+    bool TernaryAST::bind(Expression* e, const Result& val)
+    {
+        return false; // false
+        if(Expression::bind(e, val)) {
```
hainest, line 117: Unreachable code.

parseAPI/src/IA_amdgpu.C
```diff
@@ -0,0 +1,150 @@
...
+}
+
+bool IA_amdgpu::isLinkerStub() const
+{
+  // Disabling this code because it ends with an
+    // incorrect CFG.
+    return false;
+}
+
+#if 0
```
hainest, line 139: If this code isn't needed, then let's remove it.

dataflowAPI/rose/SgAsmAmdgpuInstruction.h
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+/* the generated cast function */
+/*! \brief Casts pointer from base class to derived class */
+    friend SgAsmAmdgpuInstruction *isSgAsmAmdgpuInstruction( SgNode * s );
+
+/*! \brief Casts pointer from base class to derived class (for const pointers) */
+     friend const SgAsmAmdgpuInstruction *isSgAsmAmdgpuInstruction(const SgNode *s);
+
+public:
```
hainest, line 29: Only the first `public` is needed.

hainest (changes requested):
There is lots of commented-out code. If it's not needed, then let's go ahead and remove it. If it's useful, then let's add a comment describing it.

A couple of notes on adding new classes:

1. If a class is not intended to be inherited from, then let's mark it `final` and remove the virtual specifiers from the member functions.

2. Base members that are overridden should be marked `override` to help the compiler enforce interface checking.

3. Thanks for using Doxygen comments!

dataflowAPI/rose/semantics/Registers.C
bbiiggppiigg: This part I'm just following what was done with armv8.
I can probably fix the code to use the second pattern.
Can you explain to me what is problem with this code ? 
Like I'm not that familiar with C++, so  what would be the difference between the original version and the version you've suggested.

dataflowAPI/src/SymbolicExpansion.C
bbiiggppiigg: This is also what have been done for the rest of the architectures.(PPC64, AArch64,X86_64) and so on, where the fprintfs are commented out.
I guess the proper way would be to use some sort of debug printf instead of directly writing to stderr ?

instructionAPI/src/Ternary.C
bbiiggppiigg: This part I marked it as TODO as I'm not sure how should we bind an Ternary AST yet.

parseAPI/src/IA_amdgpu.C
bbiiggppiigg: removed

dataflowAPI/rose/SgAsmAmdgpuInstruction.h
bbiiggppiigg: Duplicates removed.

symtabAPI/src/debug.C
bbiiggppiigg: Restore to default = 0.

symtabAPI/src/Object-elf.C
bbiiggppiigg: Empty block removed.

instructionAPI/src/ArchSpecificFormatters.C
bbiiggppiigg: Fixed

instructionAPI/src/Ternary.C
bbiiggppiigg: Fixed

common/h/dyn_regs.h
```diff
@@ -74,6 +75,8 @@ namespace Dyninst
         typedef std::map<signed int, std::string> NameMap;
         static boost::shared_ptr<MachRegister::NameMap> names();
         void init_names();
+
+        void getAMDGPUROSERegister(int &c, int &n, int &p);
```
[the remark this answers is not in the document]
bbiiggppiigg: Fixed

dataflowAPI/src/SymbolicExpansion.C
mxz297: For this one, Dyninst should just swallow this exception. rose has more restrict requirements regarding instruction opcode, operand types, orders, and so on, and rose may also throw exception for unimplemented instructions. For all such error cases, this code will return an null AST ptr and a false return value. So, the user of Dyninst can detect this error and do corresponding error handling.

common/src/dyn_regs.C
```diff
@@ -34,6 +34,8 @@
 #include "external/rose/rose-compat.h"
 #include "external/rose/powerpcInstructionEnum.h"
 #include "external/rose/armv8InstructionEnum.h"
+#include "external/rose/amdgpuInstructionEnum.h"
+#include <boost/format.hpp>
```
mxz297, line 38: @bbiiggppiigg Do we really need to use `<boost/format.hpp>`? @hainest Is there a std c++ equivalent?

mxz297:
@bbiiggppiigg I find that there is some code about the instruction semantics. I did not know that you have started this work. So, what's the status of this work for instruction semantics? Specifically, can we get symbolic expression through dataflowAPI for certain AMD GPU instructions?

bbiiggppiigg:
That is part of my effort trying to reason about function calls using indirect branching working, but it is still work in progress.
My target is to make the backward-slicing thing works. 

I primarily focused on the following instructions
s_setpc_b64, s_swappc_b64 : the instructions that updates the PC
s_getpc_b64 which stores the pc into some scalar register
s_add_u32/ s_addc_u32 : the instructions that are used to calculate the branch target

I'm not sure what do you expect to get using dataflowAPI, like can you give me an example of what you would like to see ?

mxz297:
DataflowAPI has its interface for converting instructions to SSA-like assignments, slicing, and converting assignment to symbolic expressions. Resolving indirect control flow using all of mentioned interface. But, each of them can be used independently and can be used for other things.

mxz297:
@bbiiggppiigg My goal was to only add preliminary support for AMD GPU analysis in terms of decoding instructions and parsing direct control flow. We have tested it with several AMD GPU binaries.

I would recommend that you split your work on dataflowAPI from the other components. Given that dataflowAPI support for AMD GPU has not been shown to work on any AMD GPU binary, I don't think we want to include that in a release.

dataflowAPI/rose/semantics/Registers.C
hainest: I'm definitely not faulting you for using the existing code as a template here, even though the method is sub-optimal. I just want to make sure we are improving the code as we update it. The leaky single pattern (i.e., never freeing the instance) is only useful when there is a problem in ordering of destruction of static objects. That's not an issue here, so we can use better patterns that don't cause fault positives when run under a memory analysis tool (e.g., valgrind).

From the C++ side of things, C++11 added thread safety to function-local static variables. This means that `regs` is only constructed once (when the function is first called) and that initialization is thread-safe. This wasn't the case when the previous code was written (i.e., C++98). This new guarantee allows us to use a local variable with static storage duration instead of storing it on the heap. The call to `call_once` is still necessary because we are doing some non-trivial work that should only be done once. Does that help at all? It's one of the more technical changes made to the language, so I'm not surprised you are unfamiliar with it.

dataflowAPI/src/SymbolicExpansion.C
hainest: @mxz297 Sounds good. I just wanted to make sure this wasn't an accident.

common/src/dyn_regs.C
hainest: I did a quick grep and couldn't find any usage of it. I think we can remove it.

> Is there a std c++ equivalent?

Yes, but it requires C++20.

hainest:
I strongly suggest taking a look at the classes that were added here to improve type safety using the guidelines I mentioned (copied here for convenience).

A couple of notes on adding new classes:

1. If a class is not intended to be inherited from, then let's mark it final and remove the virtual specifiers from the member functions.

2. Base members that are overridden should be marked override to help the compiler enforce interface checking.

bbiiggppiigg:
> @bbiiggppiigg My goal was to only add preliminary support for AMD GPU analysis in terms of decoding instructions and parsing direct control flow. We have tested it with several AMD GPU binaries.
> 
> I would recommend that you split your work on dataflowAPI from the other components. Given that dataflowAPI support for AMD GPU has not been shown to work on any AMD GPU binary, I don't think we want to include that in a release.

The problem here is that, even for running the funcDisassemble example alone, when an instruction is an indirect branch, it will trigger the IndirectAnalyzer in parseAPI, 

https://github.com/bbiiggppiigg/dyninst/blob/5c6f0bd2c3d3e4a26d702e7eaa566b1934487896/parseAPI/src/IndirectAnalyzer.C#L78-L83

which will in turn calls to the AssignmentConverter in dataflowAPI, and the IndirectAnalyzer  expect to see at least one assignment inserted into the assignment vector, so I see no clean way of separating out all dataflowAPI work.

mxz297:
@bbiiggppiigg You can choose to not resolve indirect branch for AMD GPU binaries. The IndirectAnalyzer can simply directly return an empty set of edges for AMD GPU binaries.

dataflowAPI/rose/semantics/Registers.C
bbiiggppiigg: So this part of the code is removed as an effort to separate out the work on dataflowAPI. I'll mark it as resolved here.

bbiiggppiigg:
@mxz297 the changes in dataflowAPI have now been removed.

hainest:
@bbiiggppiigg Is there anything remaining to be updated before testing?

bbiiggppiigg:
I'm not aware of anything remaining to be updated.

bbiiggppiigg:
@hainest In case that we're gonna support multiple AMDGPU architectures in the future, I'm considering whether I should rename the Arch_amdgpu to something like Arch_amdgpu_vega so I can later add something like Arch_amdgpu_rdna. Do you think it is necessary to change this before merging, or is it fine if these Arch specifiers are only used internally

hainest:
This support is still very much experimental, so we make no guarantees about code stability. That said, it would probably be a good idea to split the architectures. We'll likely need to do that with CUDA in the future, as well. It would be nice to have a hierarchy of architectures (e.g., "amdgpu" would encompass vega, navi, etc.), but I don't think we can support that without major changes.

@mxz297 Any thoughts on this?

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR900

Looks good. We can merge now or wait until you have split the AMD GPU architectures.

hainest:
ping.

mxz297:
I think it is a good idea to rename `Arch_amdgpu` to `Arch_amdgpu_vega`. Otherwise, when we later add more ADM GPU architectures, we will break API by renaming the architecture enums.

hainest:
Ok. I'll hold off on merging this until you've made those changes.

bbiiggppiigg:
@hainest I've made changes according to the suggestions.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR900&since_date=2020-11-24

Looks good.

PR #911, opened 2020-10-31
Revert required minimum version of elfutils to 0.173

hainest:


hainest:
We require elfutils >= 0.178 because libdw isn't thread-safe until then.

PR #917, opened 2020-11-11
Improve single thread's backward slicing speed by 40%

Jokeren:
1. Change pass-by-value arguments to pass-by-reference.
2. Pre-allocate vector/map entries to avoid redundant copies.
3. Use hash map for slicing cache.

### DIFF
--- a/dataflowAPI/src/slicing.C
+++ b/dataflowAPI/src/slicing.C
@@ -30,6 +30,7 @@
 #include <set>
 #include <vector>
 #include <map>
+#include <unordered_map>
 #include <utility>
 #include "dataflowAPI/h/Absloc.h"
 #include "dataflowAPI/h/AbslocInterface.h"
@@ -159,11 +160,11 @@ Slicer::sliceInternal(
 
     // this is the unified cache aka the cache that will hold 
     // the merged set of 'defs'.
-    map<Address,DefCache> cache;
+    unordered_map<Address,DefCache> cache;
 
     // this is the single cache aka the cache that holds
     // only the 'defs' from a single instruction. 
-    map<Address, DefCache> singleCache; 
+    unordered_map<Address, DefCache> singleCache; 
     
     ret = Graph::createGraph();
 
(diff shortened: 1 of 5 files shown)


hainest:
@Jokeren You specifically mention pre-allocating the standard containers, but I don't see where that's happening. Is there a missing commit?

dataflowAPI/src/slicing.C
```diff
@@ -776,42 +776,45 @@ Slicer::getPredecessors(
...
+      //Slightly more complicated than the forward case; check
+      //a predicate for each active abstract region to see whether
+      //we should continue
+      bool cont = false;
+      SliceFrame::ActiveMap::const_iterator ait = cand.active.begin();
+      for( ; ait != cand.active.end(); ++ait) {
+        bool add = p.addPredecessor((*ait).first);
+        if(add) {
+          if (nf == NULL) {
+            newCands.emplace_back(cand.loc, cand.con);
```
Jokeren, line 790: @hainest, it's not missing. I should be clearer. Here we allocate an entry in `newCands` and grab the entry's pointer.

In contrast, previously we insert elements into `nf.active.insert` and then copy all map elements to `newCands`. With the new code, the copy process is eliminated.

dataflowAPI/src/slicing.C
```diff
@@ -747,13 +747,13 @@ void Slicer::handlePredecessorEdge(ParseAPI::Edge* e,
       return;
     }
 
-    nf = cand;
+    newCands.emplace_back(cand);
```
Jokeren, line 750: @hainest Likewise, we no longer copy `cand` to `nf`.

dataflowAPI/src/slicing.C
hainest: Ah, ok. I was thinking of a different type of pre-allocation. This makes sense.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR917

No new regressions.

PR #925, opened 2020-11-25
Clang fixes

tstellar:
Here are a few fixes to get dyninst building with clang.  The concurrent.h fix is especially important, because this is a public header and the error is preventing other projects that interface with dyninst from building with clang.

### DIFF
--- a/common/h/concurrent.h
+++ b/common/h/concurrent.h
@@ -145,8 +145,8 @@ public:
 
     using base::clear;
 
-    using base::iterator;
-    using base::const_iterator;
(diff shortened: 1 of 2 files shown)


hainest:
These fixes (and more) have already been established in the [clang branch](https://github.com/dyninst/dyninst/tree/clang). They haven't been merged into the master branch because initial testing showed regressions when building with Clang. I'm working on those currently, and you can track their progress in the Clang Support [project](https://github.com/dyninst/dyninst/projects/4).

PR #927, opened 2020-12-02
Make assert expressions side effect free

### LINKED ISSUE #926: Remove side effects from assert expressions
**Intention**
Fix assert expressions to be pure functions with no side effects that are visible outside the assert.

**Describe the bug**
Several assert expressions in Dyninst have necessary side effects for the correct operation.  Since the assert is completely removed if NDEBUG is defined, this required expression is not present.

**To Reproduce**

**Expected behavior**
The system should function correctly if NDEBUG is defined, and assert's are removed from the code.

all

kupsch:
Fixes 15 asserts that had side effects.  I believe the remainder are side effect free.

Fixes #926


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR927

No new regressions.

PR #928, opened 2020-12-02
Fix a bug in MachRegister::name function

bbiiggppiigg:
As title, the previous implementation of adding support for the name of amdgpu register vector is buggy, and I fix it so now it displays the correct register names. 

I thought I only modify a single file since my last commit that should have been merged in #900 , but I don't know why all my previous commits show up in this pull request

### DIFF
--- a/common/src/dyn_regs.C
+++ b/common/src/dyn_regs.C
@@ -153,43 +153,46 @@ std::string MachRegister::name() const {
 
     std::string ret;
     if (iter != names()->end()) {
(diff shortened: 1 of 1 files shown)


hainest:
Could you rebase this onto the HEAD of master? That will remove the extraneous commits.

bbiiggppiigg:
@hainest I rebased the branch as you have suggested.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR928

No new regressions.

PR #931, opened 2020-12-13
Add direct tail call case for getCallee

tylergu:
When an edge is direct branch or conditional taken branch, and it is inter-procedural, we consider it as a direct tail call and return the target of this edge as the callee.


mxz297 (approved):

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR931&status=OK

No new regressions.

PR #932, opened 2020-12-15
Require gcc >= 6.0

hainest:
We've quietly required the ABI and API changes in libstdc++ for some time, but we make them explicit now.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR932

Issue #935, opened 2020-12-15
several Symbol methods should be const


kupsch:
**Intention**
Several methods of Symbol should be const methods:
- getVersions
- getVersionFileName
- getVersionNum
- getVersionHidden
- getReferringSymbol

**Describe the bug**
You can't call the above methods if you have a const Symbol.

**To Reproduce**

**Expected behavior**
You should be able to call the methods with a const Symbol as they do not modify Symbol.

PR #936, opened 2020-12-15
make Symbol methods const

kupsch:
The following methods should be const methods of the class Symbol:

- getVersions
- getVersionFileName
- getVersionNum
- getVersionHidden
- getReferringSymbol

Fixes: #935


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR936

No new regressions.

PR #939, opened 2020-12-16
Fix wrong return value in DwarfHandle::init_dbg

### LINKED ISSUE #937: Possible bug in AMD GPU architecture handling in DwarfHandle::init_dbg
@bbiiggppiigg  @mxz297 

https://github.com/dyninst/dyninst/blob/master/dwarf/src/dwarfHandle.C#L199 was introduced by #900

1. We are returning an enum/int value in a function that returns a `bool`. In gcc, this fires `-Wint-in-bool-context`.
2. The first assert cannot fire because of the `return`
3. We never assign a value to `arch`. I'm assuming that's because this is only preliminary architecture support for AMD and `DwarfDyninst::DwarfFrameParser::create` doesn't know what to do with it.
4. "probabily" -> "probably" on line 213

Sorry I didn't catch these during the initial review. That was a big PR, and I didn't fully grok what was going on in this part.

bbiiggppiigg:
As title, fixing a bug in DwarfHandle::init_dbg where the arch is not set and wrong value is returned for Arch_amdgpu_vega and Arch_amdgpu_rdna

Fixes #937

### DIFF
--- a/dwarf/src/dwarfHandle.C
+++ b/dwarf/src/dwarfHandle.C
@@ -200,10 +200,11 @@ bool DwarfHandle::init_dbg()
             unsigned int ef_amdgpu_mach = 0x000000ff & file->e_flags();
 			switch(ef_amdgpu_mach){
 				case 0x33: case 0x34: case 0x35: case 0x36: case 0x37: case 0x38:
-					return Dyninst::Arch_amdgpu_rdna;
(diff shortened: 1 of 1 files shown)


hainest (approved):

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR939

LGTM.

PR #948, opened 2021-01-19
Fixing another bug related to amdgpu register names

bbiiggppiigg:
So in one of the previous commits, 
the header file  common/h/amdgpu_vega_sys_reg.h is polluted with wrong mapping 
So incorrect register name will be reported for amdgpu and registers with higher register id might not be identified correctly.

modification made to dyn_regs.C/name function is to account for cases when the register id mapping doesn't start from 0, in which case we want to make sure we use the relative offset from the base registers.


common/src/dyn_regs.C
```diff
@@ -722,7 +729,7 @@ void MachRegister::getAMDGPUROSERegister(int &reg_class, int &reg_idx, int &offs
             offset = 0;
         }
         case amdgpu_vega::SGPR_VEC2:{
-            //std::cout << "class = vec 2 " << std::endl;
+            std::cout << "class = vec 2 " << std::endl;
```
hainest, line 732: If this is useful debugging information, then let's protect it with a check for[ `dyn_debug_inst`](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/debug.C#L301). Otherwise, let's remove it.

hainest (changes requested):

common/src/dyn_regs.C
bbiiggppiigg: The line is removed as suggested.
As for the dyn_debug_inst,I have a question in general about how to enable these debug messages.
Like I see no difference when I run my application with and without DYNINST_DEBUG_INST=1.
Is there some kind of handle that I have to set before I can see the debug messages?

hainest: As long as it's either exported into the global environment or set on invocation, it should work. There don't seem to be many places that `inst_printf` gets called in the codebase, though.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR948

LGTM.

hainest (approved):

PR #950, opened 2021-01-29
add missing libiberty include dir

tomhaber:
On some platforms, the libiberty headers are not in the usual place. The user can override the location in CMake, but this path is ignored by the CMakeLists.txt

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -112,6 +112,10 @@ dyninst_library(common)
 if(TARGET LibIberty)
   add_dependencies(common LibIberty)
   target_link_libraries(common PRIVATE LibIberty)
+
+  if(${ENABLE_STATIC_LIBS})
+	  target_link_libraries(common_static PRIVATE LibIberty)
+  endif()
 endif()
 
 if(TARGET TBB)


hainest:
Woops. I can't believe I didn't notice that. We are trying to move toward target-based CMake builds. Could you move this change to common/CMakeLists.txt:115?

tomhaber:
Now that I look at it again, I noticed you are using cmake with targets.

After doing some more digging, I found the actual problem. `dyninst_library` is not adding the dependencies to the _static target and neither is common/CMakeLists.txt for LibIberty

hainest:
>Now that I look at it again, I noticed you are using cmake with targets.

It's a mish-mash of CMake-2.x and CMake-3.4+. I'm trying to migrate everything to use build targets, but it's going to be a long process.

>dyninst_library is not adding the dependencies to the _static target 

Which dependencies? Only libcommon needs libiberty. Everything else will get it transitively from there.

>neither is common/CMakeLists.txt for LibIberty

Yup. That's why I requested we move it there.

tomhaber:
Sorry, I wasn't clear.

common indeed only depends on libiberty, but I noticed that `dyninst_library` also creates a static version of the same library depending on flags. Now it seems that the same function can also add libraries to the target and some libraries are using this functionality. I was wondering if those dependencies shouldn't be added to the static version as well.

If that is the intended behavior, maybe I should change it to use `dyninst_library` for adding the LibIberty dependency.
That seems nicer to me than adding
```
target_link_libraries(common_static PRIVATE LibIberty)
```
with additional tests to see if the static should be enabled.

hainest:
>That seems nicer to me than adding

Let's do the ugly thing now because it is more inline with using CMake target-based builds. Eventually, `dyninst_library` will be thrown into the core of the Sun.

tomhaber:
I've altered the patch.

Personally, I kind of like the idea of `dyninst_library` helper function: it helps avoid a lot of repetition. I'd probably change it to actually take parameters and use CMake 3.x

Are you set on using `add_dependencies` because that is still missing? It's completely redundant as far as I know.

hainest:
>Are you set on using add_dependencies because that is still missing? It's completely redundant as far as I know.

No, that shouldn't have been added in the last PR. It is redundant. We don't need to fix it here, though. I have a long-term plan to go through and scrub the build system at some point.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR950

LGTM.

@tomhaber If things are working on your side, I'll merge.

tomhaber:
yep, all good. I tested on two machines and a fedora container.

hainest:
Great. Thanks so much!

PR #956, opened 2021-01-31
Fix -Wreturn-type warning

hainest:
This was introduced by 38370b0ee48 in 2016


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR956

No new regressions.

PR #957, opened 2021-02-01
Remove BoundsFactCalculator::ThunkBound

hainest:
This was removed by f87d17a in 2017


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR957

No regressions.

PR #961, opened 2021-02-01
Remove unused 'name' parameter from Collections::addGlobalVariable

hainest:
This was removed by a8def5c64b5 in 2017

@wrwilliams Why does `Collections::addGlobalVariable` use the name from the `Type`'s constructor, but [`BPatch_typeCollection`](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/BPatch_collections.h#L101) uses the user-provided one?

### DIFF
--- a/symtabAPI/h/Collections.h
+++ b/symtabAPI/h/Collections.h
@@ -123,9 +123,9 @@ public:
     void addType(Type* t, dyn_mutex::unique_lock& g) {
       addType(t->reshare(), g);
     }
-    void addGlobalVariable(std::string &name, boost::shared_ptr<Type> type);
-    void addGlobalVariable(std::string &name, Type* t) {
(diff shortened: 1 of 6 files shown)


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR961

wrwilliams:
@hainest I have no memory of this place.

The only thing I can think is that the type info (ie `.debug_info`) contained an always-equal-or-better name than the "user-passed" (in this case `.symtab`?) one, or I simply screwed up when refactoring this. But it's been four years already and the environment in which I made those changes may no longer be accurate/valid.

hainest:
@wrwilliams No worries. I was just curious. Given that we've not had any bug reports or testing failures around this, I'm going to tentatively assume it's fine.

PR #962, opened 2021-02-01
Remove unused parameter 'elf' from Object::parse_all_relocations

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR961

No regressions.

PR #963, opened 2021-02-01
Allow usage of SIGILL for signal trampolines

mxz297:
Add a new environment variable `DYNINST_SIGNAL_TRAMPOLINE_SIGILL` to control whether we use SIGILL as the signal for trampolines.

If `DYNINST_SIGNAL_TRAMPOLINE_SIGILL` is set, we use SIGILL as signal trampolines and the mutator will generate illegal instructions in the mutatee.

In the case of binary rewriting, `DYNINST_SIGNAL_TRAMPOLINE_SIGILL` should be consistently set or unset when rewriting the binary and running the rewritten binaries.

### DIFF
--- a/dyninstAPI/src/addressSpace.C
+++ b/dyninstAPI/src/addressSpace.C
@@ -84,6 +84,7 @@ AddressSpace::AddressSpace () :
     new_instp_cb(NULL),
     heapInitialized_(false),
     useTraps_(true),
+    sigILLTrampoline_(false),
     trampGuardBase_(NULL),
     up_ptr_(NULL),
     costAddr_(0),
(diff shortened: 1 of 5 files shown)


mxz297:
@hainest I don't have a good way to test this. I remembered that you had a use case where you have to debug through trap trampolines. Can you check whether PR fixes your problem?

hainest:
I do. I will test it out when I'm done testing some PRs. Thanks!

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR963

Turning on DYNINST_SIGNAL_TRAMPOLINE_SIGILL, causes regressions- mostly in proccontrol tests. Note that I only used the new flag on cayenne.

mxz297:
@hainest It is expected that turning on `DYNINST_SIGNAL_TRAMPOLINE_SIGILL` will cause proccontrol to behave differently. Proccontrol will basically see SIGILL as fatal errors and terminate the mutatee. 

The purpose of `DYNINST_SIGNAL_TRAMPOLINE_SIGILL` is to enable debugging rewritten binaries. For statically rewritten binaries, proccontrol is out of the scope. Proccontrol is only active during dynamic instrumentation. In addition, it is even more difficult to debug dynamic instrumentation as the mutator uses ptrace to attach to the mutatee, which prevents GDB to attach to the mutatee (a process can only be attached with ptrace by one process).

So, for this PR, I think the criteria is to see what happens with the rewriter tests under GDB and whether it would help with debugging.

mxz297:
Of course, this PR also needs to be tested when not turning on `DYNINST_SIGNAL_TRAMPOLINE_SIGILL`.

hainest:
>It is expected that turning on DYNINST_SIGNAL_TRAMPOLINE_SIGILL will cause proccontrol to behave differently. Proccontrol will basically see SIGILL as fatal errors and terminate the mutatee.

That makes sense. Can we add this to the documentation for this flag?

>Of course, this PR also needs to be tested when not turning on DYNINST_SIGNAL_TRAMPOLINE_SIGILL.

It was. Only the run on cayenne was done with the flag turned on. Running without the flag shows no regressions.

PR #964, opened 2021-02-02
Fix misleading indentation compiler warning in codeRangeTree::remove

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR964

No new regressions.

PR #966, opened 2021-02-02
Fix possible buffer overflow in parseThat::runHunt_binaryEdit

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR966

No new regressions.

PR #969, opened 2021-02-02
Fix cast qualifiers in Object-elf::read_val_of_type

hainest:
The final `const` was always ignored.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR969

No new regressions.

PR #970, opened 2021-02-02
Fix possibly uninitialized local variable in InstructionDecoder_aarch64

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR970

No new regressions.

PR #971, opened 2021-02-02
Remove usage of designated initializer in dwarfHandle.C

hainest:
Designated initializers aren't part of C++ until C++20; use aggregate initialization, instead. This also fixes the missing initializer for Dwfl_Callbacks::debuginfo_path.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR971

No new regressions.

PR #972, opened 2021-02-02
Fix pointer arithmetic on 'void*' in codeGen::insert

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR972

No new regressions.

PR #973, opened 2021-02-03
Fix constructor member intializer list reordering

hainest:
C++ requires that members be initialized in the order declared in the class.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR973

No new regressions.

PR #976, opened 2021-02-03
Fix shifts of negative values

hainest:
This is undefined behavior in C++.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR976

No new regressions.

PR #979, opened 2021-02-03
Add Initial support for Analyzing Indirect Control Flow on Amdgpu Vega

bbiiggppiigg:
This series of commits added the some bare minimum functionalities that are required to resolve one type of indirect control flow that can be observed in kernels of amdgpu, namely functional call performed by the following instruction sequence

getpc
add
addc
swappc


hainest:
@bbiiggppiigg Could you rebase this onto the head of master?

dataflowAPI/src/RoseInsnFactory.C
```diff
@@ -402,8 +406,191 @@ bool RoseInsnArmv8Factory::handleSpecialCases(entryID, SgAsmInstruction *, SgAsm
...
+                auto [oper0,oper1] = expandSgprPair(operands[0]);
+                operands.resize(3);
+                operands[0] = oper0;
+                operands[1] = oper1;
+                operands[2] = Operand(InstructionAPI::Immediate::makeImmediate(Result(u64,_addr+4)),false,false); 
+
+                break;
+            }
+ 
+        case e_lea: {
```
mxz297, line 493: @bbiiggppiigg Why the AMD GPU architecture code needs to deal with x86 CPU opcodes?

mxz297:
@bbiiggppiigg Did you ran this PR against the Dyninst test suite on any platform?

bbiiggppiigg:
@mxz297 I wasn't able to get the test suite working.  I'll try to do it.

hainest:
I ran this through my standard testing battery last night. There are several regressions on x86:

https://bottle.cs.wisc.edu/search?dyninst_branch=PR979

I'll re-run the updates you just posted.

bbiiggppiigg (commented):
This should be fixed by commit 8c93e13

bbiiggppiigg:
@hainest I've reverted the outputting of those messages back to use parsing_printf. so those messages should no longer appear.

hainest:
Great, thanks! Let me know when you're ready, and I'll re-run the test suite battery.

mxz297:
@hainest Thanks for running the test suite.

@bbiiggppiigg This PR will require more testing. Getting no regression is the first step as your change may have side affects on jump table analysis on CPU. After you get the test regression fixed, we will need to setup some way to compare ParseAPI output on more binaries before this PR can be merged.

bbiiggppiigg:
@mxz297 What you've said totally make sense. 
I'll inspect my commits and see if I can find the problem.

@hainest I think I've removed most debugging messages so they shouldn't break your log parser, but since I haven't find out the cause of regression, I expect the regression to persist after your re-run

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR979

The logging output has been fixed, so that my build script's parser is working again. ARM sees no issues, but x86 and Power are having regressions.

bbiiggppiigg:
I've tracked down the commit that causes the regression on x86, and but I'm still analyzing the cause

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR979&since_date=2021-02-07

There are still regressions in test1_33 on Power. The failure on ARM is unrelated to this change.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR979&since_date=2021-02-23

No new regressions now.

hainest:
@bbiiggppiigg What's the status of this PR? Do you need to do more testing before merging?

bbiiggppiigg:
The intended functionality of this PR should be working correctly if there isn't any new regression, and for whatever that was working before on amdgpu, this PR shouldn't introduce new bug.  
I'm aware of some potential buffer overflows that needs to be fixed, but that should probably be another PR of its own.

hainest:
@bbiiggppiigg Are you comfortable adding this to the 11.0 release, or would you rather wait until the buffer overflows are fixed?

bbiiggppiigg:
@hainest I think it would be better if these buffer overflows are fixed before they become a part of the stable release. Saying that, the buffer overflows should have already existed before this PR. 
So maybe we can merge this PR, and I'll try to submit a separate PR for fixing buffer overflows by end of this week?

mxz297:
@bbiiggppiigg We need to test this PR a little bit more before merging it.

Here is a test program that uses Dyninst to print CFG: https://github.com/HPCToolkit/hpctoolkit-tests/blob/master/dyninst/openmp-loops.cpp

I would like you to use this program to compare the output of current master and current master with your PR. Ideally, the output should be the same. 

You should be able to run this comparison across all system libraries. Let me know if you need more helps to set up the test.

bbiiggppiigg:
@mxz297 
I've tested on amd2 and there is not difference in output for all binaries under /bin/*
Are there other binaries that I should test for ?

mxz297:
@bbiiggppiigg please also test with binaries under /lib or /lib64. So we need to test both executables and libraries.

In addition to amd2, can you test on at least zeroah (ARM) and a power machine? Do you have access to a Power machine? 

AMD2 has ubuntu 18.04. @hainest Do you other x86-64 candidates? I think we should at least try a redhat system.

bbiiggppiigg:
@mxz297 
On amd2 the output is the same for /lib/x86_64-linux-gnu/* /bin/* /lib32/* and /libx32/*
For power machine I'm planning to test on llnl@rice and I'm setting it up now, but I don't think I have any access to an arm machine, I"ll ask Tim for access to it.

bbiiggppiigg:
@mxz297 @hainest 
Tested on zeroah, llnl and amd2 and Rebased on master

mxz297:
@bbiiggppiigg Thanks!

@hainest You can merge this PR after a full test suite run.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR979&since_date=2021-04-04

LGTM.

PR #980, opened 2021-02-04
Remove ignored cast qualifier in SnippetGenerator::findParameter

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR980

No new regressions.

PR #981, opened 2021-02-04
Use std::locale when writing a timeStamp to a stream

hainest:
The current implementation causes several -Wstringop-truncation warnings. Yet, the whole thing can be replaced by using locales.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR981

No new regressions.

PR #982, opened 2021-02-04
Fix C-string truncations in parseThat/config.C

hainest:
For some reason, these didn't show up when fixing those in #967.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR982

No new regressions.

PR #987, opened 2021-02-04
Fix unhandled switch case in Region::regionType2Str

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR987

No new regressions.

PR #988, opened 2021-02-04
Correctly declare Aggregate::operator<<

hainest:
The previous declaration relied on header inclusion ordering to be correct. The following example program failed to build:

    #include "Function.h"
     int main(){}


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR988

No new regressions.

PR #989, opened 2021-02-05
Remove dead code from ia32_decode

hainest:
The STUB_OUT_TO_AVOID_RACE was added by 43dd5eb6eb in 2018. It's never set, so this entire calculation is never used.


mxz297 (approved):

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR989&since_date=2021-02-07

No new regressions.

PR #974, opened 2021-02-03
Fix inclusion of boost deprecated headers

hainest:

PR #993, opened 2021-02-05
Bump minimum Boost version to 1.67

hainest:
This is needed because we no longer support the headers deprecated in that version. See also #974

### DIFF
--- a/cmake/Boost.cmake
+++ b/cmake/Boost.cmake
@@ -50,8 +50,8 @@ if(Boost_FOUND)
   return()
 endif()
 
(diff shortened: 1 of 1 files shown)


hainest:
@kupsch Could you verify this works in your environment. I don't have a test environment where I let Dyninst download its dependencies.

kupsch:
This builds correctly.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR993

No new regressions.

PR #997, opened 2021-02-07
Fix breakage introduced by PR990

hainest:
This function call has side effects that were not properly tested for before committing PR990.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR997

Issue is fixed and no new regressions.

PR #999, opened 2021-02-07
Remove assert in block_instance::callee(std::string const&)

hainest:
This was introduce by 14b6941fc in 2020 (PR #875). This fix allows both types of lookup by name, but preserves the "return NULL if not found" semantics of block_instance::callee().


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR999

No new regressions.

PR #1000, opened 2021-02-08
Aarch64: Fix restoring special purpose registers order

bn222:
The special purpose registers are pushed onto the stack by
EmitterAARCH64SaveRegs::saveSPRegisters. This patch reverses the order
used for popping the stored registers from the stack. Without
this patch, some of the state registers get swapped.

### DIFF
--- a/dyninstAPI/src/inst-aarch64.C
+++ b/dyninstAPI/src/inst-aarch64.C
@@ -317,11 +317,12 @@ unsigned EmitterAARCH64RestoreRegs::restoreSPRegisters(
     std::vector<registerSlot *> spRegs;
     map<registerSlot *, int> regMap;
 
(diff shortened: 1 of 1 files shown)


hainest:
@mxz297 The ppc implementation also has this behavior. I'm assuming we should fix it there, as well? I couldn't find anything corresponding to this kind of behavior in x86 (none of these functions exists there).

hainest:
@mxz297 When you have a bit of time, could you let me know your thoughts on the ppc behavior. I think we should change them both.

mxz297:
@hainest This PR looks good to me. I agree with you that ppc currently has the same behavior and we should reverse the order for restoring SPR. Basically, the restoring registers should be done in a reverse order as saving registers.

hainest:
No testing regressions.

PR #1003, opened 2021-02-11
Clean up PLTFunction class from dyninstAPI/Parsing.C

hainest:
This is purely an adaptor to account for Symtab::Function(Symtab::Symbol*) being protected.

@mxz297 I don't think this will affect #1001 as the usage there is polymorphic to `Symtab::Function`. But just in case, can you rebase your branch on this change?


mxz297:
@hainest Given that for #1001, we still needs to explore the static test case, I expect this PR to be merged first. I can rebase #1001 based on the new master branch when #1001 becomes ready.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1003&since_date=2021-02-12%2004:00:00

No new regressions.

PR #1012, opened 2021-03-31
Fix non-deterministic inline function lookup when bad DWARF is generated regarding outlined OpenMP code

### LINKED ISSUE #629: Symtab::getContainingInlinedFunction() is generating non-deterministic results
@mwkrentel has reported non-deterministic results where sometimes an instruction is reported as inlined, but sometimes that instruction is not. I believe the root cause of this issue is in `Symtab::getContainingInlinedFunction()`. 

First of all, I agree with the comments at https://github.com/dyninst/dyninst/blob/7e6b7685bef61ca5465e3f17a5e6b014eb4a29b4/symtabAPI/src/Symtab-lookup.C#L900

stating that the goal of this function is to find the most inlined function given an offset. I thought this function should iterate over all found function ranges and identify the range that is closet to the input offset. 

However, the implementation of this function seems to be doing something else. @mplegendre You wrote this code 7 years ago, so it is likely that you forgot the rationals behind this code. It would be nice if you can take a look at at the implementation of this loop and see:

https://github.com/dyninst/dyninst/blob/7e6b7685bef61ca5465e3f17a5e6b014eb4a29b4/symtabAPI/src/Symtab-lookup.C#L902

The non-deterministic results are caused by `Symtab::getContainingInlinedFunction()` sometimes returning the most inlined function, while sometimes returning the least inlined function...

Of course, `Symtab::getContainingInlinedFunction()` returning non-deterministic results means the range lookup returns function ranges non-deterministically. But I believe the root cause of this issue is that the algorithm in  `Symtab::getContainingInlinedFunction()`  for finding most inlined function is problematic.

mxz297:
This PR fixes #629


mxz297:
@mwkrentel I tested locally that with this PR using your openmp-symtab unit test program. Now the output for quicksilver is the same after 10 runs.

mwkrentel:
@mxz297 
Do you know what inline sequence is produced for 0x40d199 for the qs
binary?  Remember, deterministic is good, but deterministic doesn't
necessarily imply correct.

As I read the patch, if there are multiple inline sequences for one
address, then you prefer the longest sequence, is that correct?
It looks like the comments suggest the old code tried to do that too,
but I guess it didn't do it correctly?

If those conditions occur, is that sufficient to say "bad dwarf"?

I would generally think if there are multiple paths, then probably the
longest one makes the most sense.  Certainly, I can't think of
anything better.

Theoretically, if there are multiple, distinct paths all with the same
length, then it could still be non-deterministic.  But if that
happens, I don't know what correct would be. 
(Sorry, The Machine is deeply embedded in my DNA.)

Thanks !

mxz297:
> @mxz297
> Do you know what inline sequence is produced for 0x40d199 for the qs
> binary? Remember, deterministic is good, but deterministic doesn't
> necessarily imply correct.

Agree. 

The openmp-symtab unit test does not have answers for 0x40d199 because it does not use parseAPI to get the actual instructions. But for adjacent addresses, the inline chain looks like:

```
stmt:  0x40d198  l=113  f='/global/u1/k/krentel/demo/tutorial-examples/examples/gpu/quicksilver/quicksilver/src/GlobalFccGrid.cc'
stmt:  0x40d19c  l=111  f='/global/common/cori_cle7/software/sles15_cgpu/gcc/8.3.0/include/c++/8.3.0/ext/new_allocator.h'
   inline:  l=436  f='/global/common/cori_cle7/software/sles15_cgpu/gcc/8.3.0/include/c++/8.3.0/bits/alloc_traits.h'  p='_ZN9__gnu_cxx13new_allocatorI9MC_VectorE8allocateEmPKv'
   inline:  l=296  f='/global/common/cori_cle7/software/sles15_cgpu/gcc/8.3.0/include/c++/8.3.0/bits/stl_vector.h'  p='_ZNSt16allocator_traitsISaI9MC_VectorEE8allocateERS1_m'
   inline:  l=1398  f='/global/common/cori_cle7/software/sles15_cgpu/gcc/8.3.0/include/c++/8.3.0/bits/stl_vector.h'  p='_ZNSt12_Vector_baseI9MC_VectorSaIS0_EE11_M_allocateEm'
   inline:  l=74  f='/global/common/cori_cle7/software/sles15_cgpu/gcc/8.3.0/include/c++/8.3.0/bits/vector.tcc'  p='_ZNSt6vectorI9MC_VectorSaIS0_EE20_M_allocate_and_copyISt13move_iteratorIPS0_EEES5_mT_S7_'
```
> 
> As I read the patch, if there are multiple inline sequences for one
> address, then you prefer the longest sequence, is that correct?

Yes, you are correct.

> It looks like the comments suggest the old code tried to do that too,
> but I guess it didn't do it correctly?

You are correct. The old implementation assumes that there is only one inline chain. So it starts with the first found function that contains the queried address, and then tries to find the inner most inlined function. Obviously, if there are multiple disjoined inlined call chains, depending who is the first found function, we may get different answers.


> If those conditions occur, is that sufficient to say "bad dwarf"?

Theoretically no. Practically, I have only seen these conditions occurs due to bad DWARF.


> I would generally think if there are multiple paths, then probably the
> longest one makes the most sense. Certainly, I can't think of
> anything better.

Agree.

> 
> Theoretically, if there are multiple, distinct paths all with the same
> length, then it could still be non-deterministic. But if that
> happens, I don't know what correct would be.
> (Sorry, The Machine is deeply embedded in my DNA.)

Agree... I don't know what should be the tie-breaker criteria.

> 
> Thanks !

mwkrentel:
I compared the new version with the debug output from hpcstruct. 
In every case that I looked at where there was a difference, the new
version with the patch had a longer inline sequence.

Similar results for libparseAPI.so.

In both cases, there were a LOT of instructions with different inline
sequences, many with just one inline step, and many with several extra
steps.  For some binaries, this is a very significant bug.

LGTM.  I approve.

mxz297:
@hainest Ready to test.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1012

LGTM.

PR #1014, opened 2021-04-05
Fixing the issue of having insufficient space allocated for unsupported instructions that might result in buffer overflow

bbiiggppiigg:
As title, previously, array spaces are not allocated for unsupported instructions with opcode greater than the opcode of the highest supported instructions, which results in buffer overrun ( which is detected on power machines when testing ).
One of the commits in   #979  temporarily mitigate this issue by simply inserting a lot of spaces, but this PR should fix this problem in a more structured way .


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1014

No regressions. LGTM.

PR #1020, opened 2021-04-20
Set -B and -S when configuring dyninstAPI_RT

stanfordcox:
Set the build directory (-B) and source directory (-S) when
configuring dyninstAPI_RT

This simple change configures out of source tree correctly for dyninstAPI_RT on
 (Fedora release 33)
 cmake version 3.19.7

(without the change: Build files have been written to: /work/scox/dyninst/src/dyninstAPI_RT)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,7 +121,8 @@ if(BUILD_RTLIB)
     -DBUILD_RTLIB_32=${BUILD_RTLIB_32}
     -DPLATFORM=${PLATFORM}
     -G ${CMAKE_GENERATOR}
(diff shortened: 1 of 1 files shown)


hainest:
@stanfordcox Technically, -B and -S are unsupported options. The ultimate fix here is to stop treating the runtime libraries as separate compilation paths for Dyninst's build system. However, this fix will work until I get time to retool the build system.

I'll test this out and merge. Thanks!

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1020

LGTM.

PR #1026, opened 2021-05-10
Do not treat symbols in any text sections as data

mxz297:
This PR aims to fix a hpcstruct segfault reported by a user (https://github.com/HPCToolkit/hpctoolkit/issues/403).

ParseAPI will assert when parsing a binary compiled on Lassen. The root cause of the problem is that PLT stubs in the binary are marked as NOTYPE and they are inside .fini section rather than .plt section. This causes ParseAPI to treat these symbols as data rather than code.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1026

LGTM.

PR #1030, opened 2021-05-12
Fix catch block parsing

### LINKED ISSUE #1027: Dyninst-11.0 regression on parsing blocks that catch exceptions
I noticed a regression on Dyninst-11.0 regarding parsingcatch blocks, which is done in SymtabAPI.

When using Dyninst-11.0, the start and end addresses of catch blocks will have an +0x100000000 offset compared to the correct addresses. I have a simple code that prints out catch blocks and with Dyninst-11.0, I got:

```
tryStart = 100402e8b, tryEnd = 100402e90, catchStart = 100402edc 
```

With Dyninst-10.2.1, I got:

```
tryStart = 402e8b, tryEnd = 402e90, catchStart = 402edc
```

We can see Dyninst-11.0 has a weird huge offset...

@sashanicolas I believe this is likely a side effect of your recent DWARF changes as catch blocks are encoded in .gcc_except_table, which uses DWARF encodings.

I have a simple reproducer:

```
#include "Symtab.h"
#include "Symbol.h"

using namespace std;
using namespace Dyninst;
using namespace Dyninst::SymtabAPI;

int main(int argc, char ** argv) {
    Symtab *sym;
    if (!Symtab::openFile(sym, string(argv[1]))) {
        fprintf(stderr, "Cannot open %s\n", argv[1]);
        exit(-1);
    }

    vector<ExceptionBlock*> exs;
    sym->getAllExceptions(exs);
    for (auto b : exs) {        
        printf("tryStart = %lx, tryEnd = %lx, catchStart = %lx\n", b->tryStart(), b->tryEnd(), b->catchStart());
    }
}
```

You can compile this code and run it with any c++ code that has exceptions such as libdyninstAPI.so.

mxz297:
The root cause is that SymtabAPI uses a 8-byte value to read catch block addresses, which can be a 2, 4 or 8-byte value. While we are careful to clip the value when reading it, there is an addition to it later where we did not clip the value according to its size.

Fixes #1027

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -3238,6 +3238,13 @@ static int read_val_of_type(int type, unsigned long *value, const unsigned char 
 
     if (*value) {
         *value += base;
+        // The addition of the base can lead to arithemtic overflow.
+        // We need to adjust the final value based on its size.
+        if (size == 2) {
+            *value = (*value) & 0xFFFF;
+        } else if (size == 4) {
+            *value = (*value) & 0xFFFFFFFF;
+        }
         if (type & DW_EH_PE_indirect) {
             // When there is a indirect catch block,
             // it is often the case that the indirect pointer would point


mxz297:
@hainest Since you are out next week, I would like to get this fixed this week. While Sasha is reviewing my change, can you also get it a test suite run?

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1030

LGTM.

mxz297:
@sashanicolas Do you have concern on merging this PR?

sashanicolas:
Looks good.

sashanicolas (approved):

PR #1037, opened 2021-05-22
Fixing calling `dwarf_getabbrevcode` and refactoring debug logging in common

### LINKED ISSUE #1034: DWARF debug print causes a crash
On a Fortran binary compiled Intel's Fortran compiler, Dyninst crashed at the following line:

https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/dwarfWalker.C#L363

This reveals two issues:

1. the crash happened inside `dwarf_getabbrevcode`, which is a function from libdw.  However, the debug logging environment variable is not set, so the logging arguments should have not been evaluated. 

`dwarf_printf` implementation here (https://github.com/dyninst/dyninst/blob/master/common/src/debug_common.h#L42) shows that it will be expanded to `dwarf_printf_int` function. And inside the `dwarf_printf_int` function, the environment variable is checked. In this case, all arguments for `dwarf_printf` will be evaluated and passed to `dwarf_printf_int`.

`dwarf_printf` should first check the environment variable before calling `dwarf_printf_int` to avoid evaluating debug logging code when debug logging is not enable. 

2. It is unclear why `dwarf_getabbrevcode` would crash. It looks like that `dwarf_getabbrevcode` has a bad pointer (with value -1). It is unclear it's bad dwarf or bad dwarf parsing.

The user cannot share the binary that caused this problem. I get remote access to the problematic binary and will investigate this issue.

mxz297:
Fix #1034

### DIFF
--- a/common/src/debug_common.C
+++ b/common/src/debug_common.C
@@ -38,6 +38,7 @@ int common_debug_dwarf = 0;
 int common_debug_addrtranslate = 0;
 int common_debug_lineinfo = 0;
 int common_debug_parsing = 0;
+int common_debug_initialized = 0;
 
 #if defined(_MSC_VER)
 #pragma warning(push)
@@ -45,9 +46,8 @@ int common_debug_parsing = 0;
 #endif
 
 bool init_debug_common() {
-    static bool initialized = false;
-    if (initialized) return true;
-    initialized = true;
+    if (common_debug_initialized) return true;
+    common_debug_initialized = 1;
 
     if (getenv("COMMON_DEBUG_DWARF") ||
         getenv("DYNINST_DEBUG_DWARF")) {
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -360,20 +360,11 @@ bool DwarfWalker::parse_int(Dwarf_Die e, bool parseSib, bool dissociate_context)
         setMangledName(false);
 
         Dwarf * desc = dwarf_cu_getdwarf(e.cu);
-        if (!common_debug_initialized || common_debug_dwarf) {
-            int tag = dwarf_tag(&e);
-            unsigned int abbrev_code = 0;
-            if (e.abbrev != DWARF_END_ABBREV) {
-                abbrev_code = dwarf_getabbrevcode(e.abbrev);
-            } else {
-                dwarf_printf("\t abbrev code is DWARF_END_ABBREV\n");
-            }
-            dwarf_printf("(0x%lx) Parsing entry with context size %d, func %s, encl %p, (%s), mod:%s, abbr: %d, tag: %x\n",
+        dwarf_printf("(0x%lx) Parsing entry with context size %d, func %s, encl %p, (%s), mod:%s, tag: %x\n",
                 id(), stack_size(),
                 curFunc()?curFunc()->getName().c_str():"(N/A)",
                 curEnclosure().get(), (dbg()!=desc)?"sup":"not sup",
-                mod()->fullName().c_str(), abbrev_code, tag);
-        }
+                mod()->fullName().c_str(), dwarf_tag(&e));
 
         bool ret = false;
         switch(dwarf_tag(&e)) {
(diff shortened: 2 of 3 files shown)


mxz297:
@hainest Ready for testing

common/src/debug_common.C
```diff
@@ -38,16 +38,16 @@ int common_debug_dwarf = 0;
 int common_debug_addrtranslate = 0;
 int common_debug_lineinfo = 0;
 int common_debug_parsing = 0;
+int common_debug_initialized = 0;
```
hainest, line 41: This is no longer threadsafe. Is that a problem for parallel DWARF parsing?

symtabAPI/src/dwarfWalker.C
```diff
@@ -360,11 +360,20 @@ bool DwarfWalker::parse_int(Dwarf_Die e, bool parseSib, bool dissociate_context)
         setMangledName(false);
 
         Dwarf * desc = dwarf_cu_getdwarf(e.cu);
-        dwarf_printf("(0x%lx) Parsing entry with context size %d, func %s, encl %p, (%s), mod:%s, abbr: %d, tag: %x\n",
+        if (!common_debug_initialized || common_debug_dwarf) {
```
hainest, line 363: Do we really care about the abbreviation code *only if* we are debugging DWARF parsing?

hainest (changes requested):

symtabAPI/src/dwarfWalker.C
mxz297: I was just fixing the code by maintaining its original behavior. This is a better question for @sashanicolas .

common/src/debug_common.C
mxz297: I agree with your assessment, but It was not completely data race free to begin with. The existing debug variables are also global variables. I don't intend to make large changes. 

If I have to guess, I think maybe because we will only set these debug variables from 0 to 1, and never change them from 1 to other values. Such data races are benign: more than one threads may perform initialization, which will check environment variable and set `common_debug_initialized`. Given the initialization is idempotent, we are fine here.

symtabAPI/src/dwarfWalker.C
sashanicolas: we can actually just remove the call to "dwarf_getabbrevcode(e.abbrev)". It was necessary for debug/investigating only. Not needed anymore.

common/src/debug_common.C
hainest: Ah, ok. I wasn't sure what all was going on in the init function. This change is fine, then.

symtabAPI/src/dwarfWalker.C
hainest: @sashanicolas So we can remove the entire block from 363 to 376?

sashanicolas: Just remove the call parameter.
```
diff --git a/symtabAPI/src/dwarfWalker.C b/symtabAPI/src/dwarfWalker.C
index 1c9e583b2..92ae91efc 100644
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -360,11 +360,11 @@ bool DwarfWalker::parse_int(Dwarf_Die e, bool parseSib, bool dissociate_context)
         setMangledName(false);
 
         Dwarf * desc = dwarf_cu_getdwarf(e.cu);
-        dwarf_printf("(0x%lx) Parsing entry with context size %d, func %s, encl %p, (%s), mod:%s, abbr: %d, tag: %x\n",
+        dwarf_printf("(0x%lx) Parsing entry with context size %d, func %s, encl %p, (%s), mod:%s, tag: %x\n",
                 id(), stack_size(),
                 curFunc()?curFunc()->getName().c_str():"(N/A)",
                 curEnclosure().get(), (dbg()!=desc)?"sup":"not sup",
-                mod()->fullName().c_str(), dwarf_getabbrevcode(e.abbrev), dwarf_tag(&e));
+                mod()->fullName().c_str(), dwarf_tag(&e));
 
         bool ret = false;
         switch(dwarf_tag(&e)) {
```

sashanicolas:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1037

hainest:
LGTM. +1

hainest (approved):

Issue #1035, opened 2021-05-19
Data races in ParseAPI and InstructionAPI on Power 9


mxz297:
I discovered several different segfaults when running hpcstruct with 16 threads on all binaries under my local spack directory.

This experiment is done on llnl.cs.rice.edu, which is a power 9 machine running red hat 7.7.

mxz297:
The first one is in ParseAPI during finalization:

```
#0  local_Rb_tree_increment (__x=0x40000) at /tmp/xm13/spack-stage/spack-stage-gcc-8.3.0-vki5t4bqfrtb3yy3rob5wn7f4dcaz4q6/spack-src/libstdc++-v3/src/c++98/tree.cc:65
#1  std::_Rb_tree_increment (__x=0x7fff60000078)
    at /tmp/xm13/spack-stage/spack-stage-gcc-8.3.0-vki5t4bqfrtb3yy3rob5wn7f4dcaz4q6/spack-src/libstdc++-v3/src/c++98/tree.cc:85
#2  0x00007fff90e10908 in operator++ (this=<synthetic pointer>)
    at /tmp/xm13/spack-stage/spack-stage-dyninst-master-xzfserqq35wrkkyu7otaaj4grett2gub/spack-src/parseAPI/h/CFG.h:279
#3  Dyninst::ParseAPI::Parser::finalize (this=0x649d5d70, f=<optimized out>)
    at /tmp/xm13/spack-stage/spack-stage-dyninst-master-xzfserqq35wrkkyu7otaaj4grett2gub/spack-src/parseAPI/src/Parser.C:842
#4  0x00007fff90e29fb4 in Dyninst::ParseAPI::Function::finalize (this=0x7ffedcfd7320)
    at /tmp/xm13/spack-stage/spack-stage-dyninst-master-xzfserqq35wrkkyu7otaaj4grett2gub/spack-src/parseAPI/src/Function.C:222
#5  0x00007fff90e0688c in Dyninst::ParseAPI::Parser::_ZN7Dyninst8ParseAPI6Parser14finalize_funcsERN3tbb17concurrent_vectorIPNS0_8FunctionESaIS5_EEE._omp_fn.5(void) ()
    at /tmp/xm13/spack-stage/spack-stage-dyninst-master-xzfserqq35wrkkyu7otaaj4grett2gub/spack-src/parseAPI/src/Parser.C:1180
#6  0x00007fff9029a2e8 in gomp_thread_start (xdata=<optimized out>)
[...]

PR #1038, opened 2021-05-22
Fix data races and asserts discovered on Power 

mxz297:
Fix all reported segfault and abort mentioned #1035

### DIFF
--- a/dataflowAPI/src/slicing.C
+++ b/dataflowAPI/src/slicing.C
@@ -1704,21 +1704,29 @@ void Slicer::markAsEndNode(Graph::Ptr ret, Direction dir, Element &e) {
 
 void Slicer::fastForward(Location &loc, Address
 			 addr) {
-  while ((loc.current != loc.end) &&
-	 (loc.addr() < addr)) {
-    loc.current++;
-  }
-  assert(loc.current != loc.end);
-  assert(loc.addr() == addr);
+    while ((loc.current != loc.end) && (loc.addr() < addr)) {
+        loc.current++;
+    }
+
+    if (loc.current == loc.end || loc.addr() != addr) {
+        slicing_cerr << "Cannot find addr " << std::hex << addr 
+            << "in block [" << loc.block->start() << "," << loc.block->end() << ")" 
+            << ", function " << loc.func->name() << " at " << loc.func->addr()
+            << std::dec << endl;
+    }
 }
 
 void Slicer::fastBackward(Location &loc, Address addr) {
-    while ((loc.rcurrent != loc.rend) &&
-	 (loc.addr() > addr)) {
-    loc.rcurrent++;
-  }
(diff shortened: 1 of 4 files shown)


mxz297:
@hainest Ready for testing

hainest:
https://bottle.cs.wisc.edu/regressions?id=12853

This reduced the number of crashes when OMP_NUM_THREADS=16 from 156 to 29. I've left the binaries on llnl.cs.rice.edu under /home/th37/paradyn/seW2KFlA.

mxz297:
@hainest I was able to reproduce the aborts. The asserts in the code indeed represent errors, but they are not necessarily fatal. So I converted them to debug logging. Can you retest?

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1038

Looks good on Power9 with 16 OMP threads. I'll re-test on other platforms just to be sure.

hainest:
No regressions on the other platforms.

@mxz297 I'm curious as to why this was only happening on PPC with N>=16 threads, and nowhere else. What sort of conditions would lead a function to be changed from RETURN to NORETURN? Might that decision be a race condition?

mxz297:
@hainest We may see it on arm but arm static rewriting is not enabled anyway. This problem happens when parsing libc.a. The problem is that .a files are not final binaries; they are a collection of .o files. So, calls to other functions from other .o files are not filled. On Power, it will look like:

```
 60:   01 00 00 48     bl      60 <__libc_init_first+0x60>
```

Note that this becomes a function call targeting itself. It will leads to create additional functions and other complication. And yes, in the end, it is a race condition and will lead to inconsistent parsing results. But I don't think I want to tune code parsing for this special case. 

On x86-64, there are also incomplete call instructions:

```
 40:   e8 00 00 00 00          callq  45 <__libc_init_first+0x45>
```

but note that the difference is that on x86, all PC-relative is post-instruction. So this becomes a call to its next instruction, which is like a thunk on 32-bit code for retrieving PC values, and I believe Dyninst has special code for this case.

You may also be interested in why my script does not find this case: my script use `file` command to check if the output includes `ELF`. When you run `file` on a .a file, it says `current ar archive`. So no `ELF`, and .a files are not included...

For Dyninst testing, you may want to change my script to include .a files. Note that hpcstruct does not analyze .a files.

hainest:
That makes sense. I was thinking about it from the perspective of parsing an executable that is linked against a static library, rather than parsing the static library, itself. I agree we don't want special handling for this.

Which script are you referring to?

mxz297:
> That makes sense. I was thinking about it from the perspective of parsing an executable that is linked against a static library, rather than parsing the static library, itself. I agree we don't want special handling for this.
> 
> Which script are you referring to?

The script that I used to crawl binaries under a directory for testing hpcstruct.

PR #1042, opened 2021-05-27
fix Position Independent Executable (PIE) handling

### LINKED ISSUE #605: Handling of PIE binaries
On recent Linux distributions, the system compiler will by default generate position independent executables (PIEs). Dyninst currently has at least two issues when dealing with PIE:

1. A few interface checks whether a binary is a shared library. Such check will wrongly return that PIEs are shared libraries. The boundary between shared library and PIE is indeed blur. We need to come up with a reliable way to distinguish PIE from shared libraries.

2. Static PIE was not legal until recently. To generate static PIE, flag "-static-pie" should be used rather than two flags "-static" and "-fie". Dyninst currently cannot correctly identify static PIE as static binary. Static pie contains necessary dynamic information for relocation, so Dyninst treats static PIE as dynamically linked. This will be the first step towards correctly instrumenting static PIE.

kupsch:
All functions that determine if an object is a shared library and/or an
executable are now based on a single set of functions in Object-elf.
This make all the functions in the various Dyninst libraries and classes
consistent and also makes Dyninst correctly handle PIE executables.
These functions in Object-elf are based on functions found in elfutils's
elfclassify program (with small improvements to better determine if a
shared library is also an executable).   The new functions in Object-elf
include:

    * isOnlyExecutable()
    * isExecutable()
    * isSharedLibrary()
    * isOnlySharedLibrary()
    * isLoadable()
    * isDebugOnly()
    * isLinuxKernelModule()

Symtab exposes some of these functions through its public API.  isExec()
now correctly handles PIE executables, and two new functions were added
to determine if the underlying file is a shared library, an executable,
or both.  These function are:

    * isExec()            (isExecutable() and not isSharedLibrary())
    * isExecutable()
    * isSharedLibrary()

Fixes #605

### DIFF
--- a/dyninstAPI/src/image.h
+++ b/dyninstAPI/src/image.h
@@ -129,7 +129,7 @@ class fileDescriptor {
     // Some platforms have split code and data. If yours is not one of them,
     // hand in the same address for code and data.
     fileDescriptor(string file, Address code, Address data, 
-                   bool isShared=false, Address dynamic=0) :
+			Address dynamic=0) :
 #if defined(os_windows)
 		procHandle_(INVALID_HANDLE_VALUE),
 		fileHandle_(INVALID_HANDLE_VALUE),
@@ -139,7 +139,6 @@ class fileDescriptor {
         code_(code),
         data_(data),
 		dynamic_(dynamic),
-        shared_(isShared),
         pid_(0),
         length_(0),
         rawPtr_(NULL) {}
@@ -147,7 +146,7 @@ class fileDescriptor {
     // ctor for non-files
     fileDescriptor(string file, Address code, Address data, 
                    Address length, void* rawPtr,
-                   bool isShared=false, Address dynamic=0) :
+		    Address dynamic=0) :
 #if defined(os_windows)
 		procHandle_(INVALID_HANDLE_VALUE),
 		fileHandle_(INVALID_HANDLE_VALUE),
@@ -157,7 +156,6 @@ class fileDescriptor {
         code_(code),
         data_(data),
 		dynamic_(dynamic),
-        shared_(isShared),
         pid_(0),
         length_(length),
         rawPtr_(rawPtr) {}
@@ -184,7 +182,6 @@ class fileDescriptor {
      const string &member() const { return member_; }
      Address code() const { return code_; }
      Address data() const { return data_; }
-     bool isSharedObject() const { return shared_; }
      int pid() const { return pid_; }
 //     Address loadAddr() const { return loadAddr_; }
      Address dynamic() const { return dynamic_; }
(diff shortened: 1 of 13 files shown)


kupsch:
This passes on both coriander and zeroah.  This should be tested with the current testsuite and the current testsuite with commit d2d3aba2 reverted so PIE executables are tested (assuming the compiler/linker default to generating PIE).

dyninstAPI/src/image.h
```diff
@@ -361,8 +356,9 @@ class image : public codeRange {
    ParseAPI::CodeObject *codeObject() const { return obj_; }
 
    bool isDyninstRTLib() const { return is_libdyninstRT; }
-   bool isAOut() const { return is_a_out; }
-   bool isSharedObj() const { 
+   bool isExecutable() const { return getObject()->isExec(); }
+   bool isSharedLibrary() const { return getObject()->isSharedLibrary(); }
+   bool isSharedObject() const {
```
hainest, line 361: What's the difference between `isSharedLibrary` and `isSharedObject`?

symtabAPI/h/Symtab.h
```diff
@@ -235,6 +235,8 @@ class SYMTAB_EXPORT Symtab : public LookupInterface,
 
    /*****Query Functions*****/
    bool isExec() const;
+   bool isExecutable() const;
+   bool isSharedLibrary() const;
```
hainest, line 239: Hmm... This is an API break. I know we discussed this before, but what was our rationale for removing these?

symtabAPI/src/Symtab.C
```diff
@@ -373,9 +373,20 @@ SYMTAB_EXPORT Symtab::Symtab() :
 
 SYMTAB_EXPORT bool Symtab::isExec() const 
 {
+    assert( is_a_out == obj_private->isOnlyExecutable() );
```
hainest, line 376: If this is a class invariant, we should enforce it in the constructor.

hainest (changes requested):

hainest:
@kupsch Could you submit a PR to revert d2d3aba2 in the testsuite?

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1042

Runs without PIE binaries in the test suite look good.

dyninstAPI/src/image.h
kupsch: isSharedObject is elf type if DYN.  It is only used to determine a startup_printf message.   renamed old function (was isSharedObj).

isSharedLibrary besides the elf type being DYN has to meet all the other requirement of being a shared library as determined by the Object class.

symtabAPI/h/Symtab.h
kupsch: These are added methods.  Adding two non-virtual functions shouldn't be an API or ABI break.

symtabAPI/src/Symtab.C
kupsch: I will remove.  Present from early validation testing of the old and new code and I forgot to remove this.

symtabAPI/h/Symtab.h
hainest: Wow. I must be going color blind because I thought those were red...

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1042&testsuite_branch=PR188

And runs with PIE enabled are good. Nice!

PR #837, opened 2020-09-07
Remove BPatch_addressSpace::findFunctionByAddr

hainest:
This was deprecated by 7f20129 in 2010. Users should instead use either BPatch_addressSpace::findFunctionByEntry or BPatch_addressSpace::findFunctionsByAddr.

Closes #820

PR #1047, opened 2021-06-08
BPatch_frame::findFunction - find function by address, not by entry

hainest:
This was broken by #837


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1047

No regressions.

sashanicolas (approved):

PR #1055, opened 2021-06-16
Add xsave instruction

hainest:
@tylergu Could you try this out and see if you can now parse this instruction?

### DIFF
--- a/common/h/entryIDs.h
+++ b/common/h/entryIDs.h
@@ -277,6 +277,7 @@ enum entryID : unsigned int {
   e_fxch,
   e_fxrstor,
   e_fxsave,
+  e_xsave,
   e_haddpd,
   e_haddps,
   e_hlt,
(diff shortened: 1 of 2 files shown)


tylergu:
I tried this pull request but it did not solve the problem. Let me write a simple example to test it to make sure.

tylergu:
I wrote a simple hard-coded example to test this, and the `xsave` instruction is correctly parsed. So this pull request is correct.

The problem was that there is another undefined opcode `xrstor` inside the `_dl_runtime_resolve` blocking the parsing. Could you add this opcode too?

Also along the way of writing this test code, I encountered an assertion when I call ParseAPI::CodeObject::parse(target=function_addr, recursive=false). It does not happen when the recursive flag is true. Should I open an issue for this?

tylergu (approved):

hainest:
>I wrote a simple hard-coded example to test this, and the xsave instruction is correctly parsed. So this pull request is correct.

>The problem was that there is another undefined opcode xrstor inside the _dl_runtime_resolve blocking the parsing. Could you add this opcode too?

Ah, ok. I'll make a separate PR for that.

>Also along the way of writing this test code, I encountered an assertion when I call ParseAPI::CodeObject::parse(target=function_addr, recursive=false). It does not happen when the recursive flag is true. Should I open an issue for this?

Yes, and please include any binaries, if possible.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1055

No regressions.

Issue #1045, opened 2021-06-04
aarch64 creates dynsym entry for internal function


stanfordcox:
The routine mutatee has an internal function incr, which has no callers, that the mutator adds a call to.  So the mutated executable ends up with:
Relocation section [31] '.rela.dyn' at offset 0x305d0 contains 7 entries:
   ...
   0x0000000000500070  AARCH64_JUMP_SLOT 000000000000000000      +0 _Z4incrv
Symbol table [27] '.dynsym' contains 18 entries:
   ...
   17: 0000000000000000     36 FUNC    GLOBAL DEFAULT    UNDEF _Z4incrv
Symbol table [41] '.symtab' contains 161 entries:
  ...
  145: 0000000000400894     36 FUNC    GLOBAL DEFAULT       13 _Z4incrv

So _Z4incrv (incr) is in the symbol table but also erroneously appears in the dynamic symbol table

Above is static.tar.gz.   As a double check, converting the test so that incr is defined in a shared object works as expected (static-shared.tar.gz attached)

So the problem is when dyninst creates a call to a static function when in static mode (stapdyn uses dynamic mode).  The dyninst testsuite, as far as I can tell, does not seem to test this.

This is on:
kernel-4.18.0-305.7.el8.kpq1.aarch64
glibc-2.28-160.el8.aarch64
gcc-8.5.0-1.el8.aarch64
dyninst-11.0.0-1.el8.aarch64
[static-shared.tar.gz](https://github.com/dyninst/dyninst/files/6601051/static-shared.tar.gz)
[static.tar.gz](https://github.com/dyninst/dyninst/files/6601061/static.tar.gz)

hainest:
@sashanicolas Could you take a look at this? I know our handling of static binaries on ARM is incomplete.

sashanicolas:
@stanfordcox I still haven't had time to look into this. Static linked binaries are still not fully supported on aarch64. 
If you can, please add here whatever else you found about this issue.

stanfordcox:
I have only tried this on the testcase but it works okay there:
/usr/bin/diff -U1 /tmp/inst-aarch64.C /work/scox/dyninst/src/dyninstAPI/src/
--- /work/scox/dyninst/src/dyninstAPI/src/inst-aarch64.C	2021-03-31 10:37:32.394213499 -0400
+++ /tmp/inst-aarch64.C	2021-06-11 17:43:49.058470680 -0400
@@ -664,4 +664,8 @@
[...]

PR #1056, opened 2021-06-17
Load callee's address when the callee and caller are in the same module

stanfordcox:
If the callee and caller are in the same module then the callee's
address can be loaded directly without using a relocation.

### DIFF
--- a/dyninstAPI/src/inst-aarch64.C
+++ b/dyninstAPI/src/inst-aarch64.C
@@ -660,8 +660,9 @@ Register EmitterAARCH64::emitCall(opCode op,
     assert(scratch != REG_NULL && "cannot get a scratch register");
     gen.markRegDefined(scratch);
 
-    if (gen.addrSpace()->edit() != NULL) {
-        // gen.as.edit() checks if we are in rewriter mode
+    if (gen.addrSpace()->edit() != NULL
+	&& gen.func()->obj() != callee->obj()) {
+        // In rewriter mode with a callee and caller not in the same module
         Address dest = getInterModuleFuncAddr(callee, gen);
 
         // emit ADR instruction
@@ -670,7 +671,6 @@ Register EmitterAARCH64::emitCall(opCode op,
         instruction insn;
         insn.clear();
         INSN_SET(insn, 31, 31, 0);
-        //INSN_SET(insn, 29, 30, disp & 0x3);
         INSN_SET(insn, 28, 28, 1);
         INSN_SET(insn, 5, 23, disp >> 2);
         INSN_SET(insn, 0, 4, scratch);


hainest:
@sashanicolas Could you take a look at this?

hainest:
@stanfordcox Is this fixing a bug you encountered?

stanfordcox:
That fixes the bug.  The problem was the caller and callee were in the same program but the same machinery was being used as for the different module case, use relocation to put address in memory slot and load slot, and in the process the callee was added to .dynsym when it was already in .symtab

hainest:
@stanfordcox @sashanicolas 

This causes regressions in rewriting on arm: https://bottle.cs.wisc.edu/search?dyninst_branch=PR1056

stanfordcox:
Hmm, interesting.   Before the patch pretty much every rewrite test failed with a linker error mentioned in #1045
I no longer have the machine but results with kernel-4.18.0-318.el8 and gcc-8.5.0-2.el8 were:

grep -h 'CRASH\|FAIL' /work/scox/dyn/results/dyn/aarch64-el9-1100.sum
snip_ref_shlib_var         g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)
test1_36                   g++    none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
snip_ref_shlib_var         g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)
test1_36                   g++    none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
snip_ref_shlib_var         g++    none 64  rewriter NA      dynamic nonPIC  CRASHED (Running Test)
test1_36                   g++    none 64  rewriter NA      dynamic nonPIC  FAILED (Running Test)
snip_ref_shlib_var         g++    none 64  rewriter NA      dynamic nonPIC  CRASHED (Running Test)
test1_36                   g++    none 64  rewriter NA      dynamic nonPIC  FAILED (Running Test)
test_callback_2            g++    none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
test_reloc                 g++    none 64  rewriter NA      dynamic nonPIC  CRASHED (Group Teardown)
test_stack_3               g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)
test_stack_4               g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)
test_thread_1              g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)
test_thread_5              g++    none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
test_thread_8              g++    none 64  create   NA      dynamic nonPIC  CRASHED (Running Test)

stanfordcox:
ah, all the failures were PIC tests.  I just run barebones runTests so no pic.  Offhand I cannot think why only pic would be an issue.

hainest:
@stanfordcox @sashanicolas I'm not sure why only PIC would be failing here. @mxz297 Do you have any thoughts?

P.S. Is this fix for #1045?

stanfordcox:
Grabbing an aarch64 machine over here so I can take a look.  Lesson learned: run the full testsuite!

PR #1059, opened 2021-06-22
Add DWARF4 base type entry encodings to symtabAPI::typeScalar

hainest:
These properties are needed for doing ABI analysis in Smeagle.

ping @vsoch

### DIFF
--- a/symtabAPI/h/Type.h
+++ b/symtabAPI/h/Type.h
@@ -436,9 +436,9 @@ typeFunction& Type::asFunctionType() { return dynamic_cast<typeFunction&>(*this)
 
 class SYMTAB_EXPORT typeScalar : public Type {
  private:
-   bool isSigned_;
+   bool isSigned_{false};
  public:
-   typeScalar();
+   typeScalar() = default;
    typeScalar(typeId_t ID, unsigned int size, std::string name = "", bool isSigned = false);
    typeScalar(unsigned int size, std::string name = "", bool isSigned = false);
    static typeScalar *create(std::string &name, int size, Symtab *obj = NULL);
--- a/symtabAPI/h/Type.h
+++ b/symtabAPI/h/Type.h
@@ -435,24 +435,53 @@ class SYMTAB_EXPORT typeFunction : public Type {
 typeFunction& Type::asFunctionType() { return dynamic_cast<typeFunction&>(*this); }
 
 class SYMTAB_EXPORT typeScalar : public Type {
- private:
-   bool is_signed{false};
- public:
-   typeScalar() = default;
+public:
+  struct properties_t {
+	  // Summary properties
+	  // NB: See DwarfWalker::parseBaseType for how these are computed
+	  bool is_integral;
+	  bool is_floating_point;
+	  bool is_string;
 
-   typeScalar(typeId_t ID, unsigned int size, std::string name="", bool isSigned=false) :
-       Type(name, ID, dataScalar), is_signed(isSigned)
-   {
-      size_ = size;
-   }
+	  // Detailed properties
+	  bool is_address;
+	  bool is_boolean;
+	  bool is_complex_float;
+	  bool is_float;
+	  bool is_imaginary_float;
+	  bool is_decimal_float;
+	  bool is_signed;
+	  bool is_signed_char;
+	  bool is_unsigned;
+	  bool is_unsigned_char;
+	  bool is_UTF;
+  };
 
-   typeScalar(unsigned int size, std::string name="", bool isSigned=false) :
-       typeScalar(this->getUniqueTypeId(), size, name, isSigned){}
+private:
+  properties_t props{};
 
-   static typeScalar *create(std::string &name, int size, Symtab *obj = NULL);
-   bool isSigned();
-   bool isCompatible(boost::shared_ptr<Type> x) { return isCompatible(x.get()); };
-   bool isCompatible(Type *otype);
+public:
+  typeScalar() = default;
+
+  typeScalar(typeId_t ID, unsigned int size, std::string name, properties_t p)
+      : Type(name, ID, dataScalar), props{p} {
+    size_ = size;
+  }
+
+  typeScalar(typeId_t ID, unsigned int size, std::string name = "", bool isSigned = false)
+      : Type(name, ID, dataScalar) {
+	props.is_signed = isSigned;
+    size_ = size;
+  }
+
+  typeScalar(unsigned int size, std::string name = "", bool isSigned = false)
+      : typeScalar(this->getUniqueTypeId(), size, name, isSigned) {}
+
+  static typeScalar *create(std::string &name, int size, Symtab *obj = NULL);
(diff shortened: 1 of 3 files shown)


symtabAPI/h/Type.h
```diff
@@ -435,16 +435,53 @@ class SYMTAB_EXPORT typeFunction : public Type {
...
+public:
+  struct properties_t {
+	  // Summary properties
+	  // NB: See DwarfWalker::parseBaseType for how these are computed
+	  bool is_integral;
+	  bool is_floating_point;
+	  bool is_string;
+
+	  // Detailed properties
+	  bool is_address;
```
vsoch, line 447: What about some kind of boolean for is_null, would that be useful?

hainest: I was looking at this, and I think we can add it here. DWARF treats 'void' separately from the base_type attributes handled here, but I think we can lump them together. I was going to chat with @sashanicolas first before adding it.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1059

No new regressions. I want to modify an existing test to explicitly check for the properties here (insofar as C will let us) before merging.

hainest:
@sashanicolas Sorry to press, but we need these changes before we can move forward with Smeagle. Could you take a look as soon as you get a chance?

vsoch:
Thank you @sashanicolas ! +1 what @hainest said. Smeeeeagle needs to parse the preccccious (but he's really using dyninst under the hood :laughing: )

sashanicolas (approved):
Looks good.

hainest:
@sashanicolas Thanks! To address @vsoch's question about `void` types. DWARF treats `void` as `DW_TAG_unspecified_type` instead of `DW_TAG_base_type`. I just checked with gcc-10.2 using `void foo(void *x, int *y);` and it gave me

  [    4b]      formal_parameter     abbrev: 3
                 name                 (string) "x"
                 decl_file            (data1) foo.c (1)
                 decl_line            (data1) 1
                 decl_column          (data1) 16
                 type                 (ref4) [    64]
                 location             (exprloc) 
                  [ 0] reg5
   [    57]      formal_parameter     abbrev: 3
                 name                 (string) "y"
                 decl_file            (data1) foo.c (1)
                 decl_line            (data1) 1
                 decl_column          (data1) 24
                 type                 (ref4) [    66]
                 location             (exprloc) 
                  [ 0] reg4
   [    64]    pointer_type         abbrev: 4
               byte_size            (data1) 8
   [    66]    pointer_type         abbrev: 5
               byte_size            (data1) 8
               type                 (ref4) [    6c]
   [    6c]    base_type            abbrev: 6
               byte_size            (data1) 4
               encoding             (data1) signed (5)
               name                 (string) "int"

so gcc doesn't use DW_TAG_unspecified_type. It just doesn't give a name for `void*` which makes `parseReferences` bail out early (https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/dwarfWalker.C#L1418).

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1059&testsuite_branch=PR196

No new regressions.

vsoch:
Woohoo! Will this be in a release soon (how does the release schedule work for Dyninst?)

hainest:
@vsoch We don't have a fixed release schedule, but we try to do at least two point releases per year. We do patch releases as necessary. We just did one of those a couple weeks ago, so it might be a while before we do another one (unless there's a pressing reason). For now, I would recommend updating Dyninst in the dev container to the current HEAD with these changes.

vsoch:
Gotcha. Autamus (our current development base) only builds from releases so what I'll do is create the same container manually from dyninst master, and then we can use that (and update it) as long as we need. I have it building now, and I can add the Dockerfile and updates to the repo the next time I need to make changes!

vsoch:
https://github.com/orgs/autamus/packages/container/dyninst/3440486?tag=06-25-master

vsoch:
If anyone wants / needs the Dockefile, here you go! It installs dyninst via spack, and I've found this much easier than trying to get it to work locally.

```
# Build stage with Spack pre-installed and ready to be used
# docker build -f Dockerfile.dyninst -t ghcr.io/autamus/dyninst:05-25-master .
FROM spack/ubuntu-bionic:latest as builder

RUN git clone https://github.com/autamus/registry && \
    cd registry && \
    mv spack/d/dyninst /opt/spack/var/spack/repos/builtin/packages/dyninst

# Install OS packages needed to build the software
RUN apt-get -yqq update && apt-get -yqq upgrade \
 && apt-get -yqq install python3-boto3 \
 && rm -rf /var/lib/apt/lists/*

# What we want to install and how we want to install it
# is specified in a manifest file (spack.yaml)
RUN mkdir /opt/spack-environment \
&&  (echo "spack:" \
&&   echo "  specs:" \
&&   echo "  - dyninst@master" \
&&   echo "  view: /opt/view" \
&&   echo "  packages:" \
&&   echo "    all:" \
&&   echo "      target:" \
&&   echo "      - x86_64" \
&&   echo "  config:" \
&&   echo "    compiler:" \
&&   echo "      target: x86_64" \
&&   echo "    install_missing_compilers: true" \
&&   echo "    install_tree: /opt/software" \
&&   echo "  mirrors:" \
&&   echo "    autamus: s3://autamus-cache" \
&&   echo "  concretization: together") > /opt/spack-environment/spack.yaml

# Install the software, remove unnecessary deps
RUN cd /opt/spack-environment && spack env activate . && curl http://s3.amazonaws.com/autamus-cache/build_cache/_pgp/FFEB24B0A9D81F6D5597F9900B59588C86C41BE7.pub > key.pub && spack gpg trust key.pub && spack install --fail-fast --monitor --monitor-save-local --monitor-tags autamus; stat=$?; cd ~/.spack/reports/monitor/; tar -czvf monitor.tar.gz *; curl -F "upload=@monitor.tar.gz" http://localhost:4500/upload; exit $stat && spack gc -y; 

# Strip all the binaries
RUN find -L /opt/view/* -type f -exec readlink -f '{}' \; | \
    xargs file -i | \
    grep 'charset=binary' | \
    grep 'x-executable\|x-archive\|x-sharedlib' | \
    awk -F: '{print $1}' | xargs strip -s

# Modifications to the environment that are necessary to run
RUN cd /opt/spack-environment && \
    spack env activate --sh -d . >> /etc/profile.d/z10_spack_environment.sh

# Bare OS image to run the installed executables
FROM ubuntu:18.04

ENV PATH=/opt/view/bin:/opt/spack/bin:$PATH

COPY --from=builder /opt/spack-environment /opt/spack-environment
COPY --from=builder /opt/software /opt/software
COPY --from=builder /opt/view /opt/view
COPY --from=builder /etc/profile.d/z10_spack_environment.sh /etc/profile.d/z10_spack_environment.sh

ENTRYPOINT ["/bin/bash", "--rcfile", "/etc/profile", "-l", "-c"]
```

PR #1069, opened 2021-06-28
Fix insertion operators in BPatch and Symtab

hainest:


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1069

No regressions.

PR #1070, opened 2021-06-30
Add x86_64 xrstor instruction

hainest:
@tylergu So sorry this took so long! Could you test this in SPI?


tylergu:
Thanks! I tested it in my test program and it correctly parses the instruction and the entire function.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1070

No regressions.

PR #1073, opened 2021-07-08
Convert TRAMP_*_OFFSET macros to functions for aarch64

hainest:
The macro for the TRAMP_*_OFFSET_32 cases expanded into `({assert(0); 0;})` which is not legal C++. It was allowed by gcc by extension. Converting them to functions preserves the behavior without code changes. They are made `inline` to prevent ODR violations.

ping @kupsch


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1073

No new regressions. @sashanicolas Does this look OK to you?

sashanicolas (approved):

PR #1078, opened 2021-08-01
BUG FIX: Fixed dependent using declaration

jrmadsen:
- dependent using declaration resolved to type w/o typename
- Causes compilation error with clang 11
- Instantiation: Dyninst::dyn_c_hash_map<unsigned long, std::vector<Dyninst::SymtabAPI::Variable *>>

### DIFF
--- a/common/h/concurrent.h
+++ b/common/h/concurrent.h
@@ -1,28 +1,28 @@
 /*
  * See the dyninst/COPYRIGHT file for copyright information.
- * 
+ *
  * We provide the Paradyn Tools (below described as "Paradyn")
  * on an AS IS basis, and do not warrant its validity or performance.
  * We reserve the right to update, modify, or discontinue this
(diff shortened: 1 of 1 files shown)


jrmadsen:
It looks like #1021 also fixes this specific issue. However, it appears that PR has some additional edits in order to make dyninst _itself_ compilable with clang. I am not sure what the hold up is on that PR is (looks like it been open for a couple months) but this specific edit is critical -- it prevents a downstream tool which is _using_ a GCC-compiled version of dyninst from being compiled with clang.

hainest:
I've been holding off merging #1021 because building Dyninst with clang results in incorrect parsing of binaries. I've not had time to investigate this, and I've not, until now, had reports of needing to use Dyninst with clang. I'll go ahead and merge #1021 since this is now a broader use issue.

hainest:
Subsumed by #1021

PR #1090, opened 2021-08-06
fix broken cast of a char literal to pointer

### LINKED ISSUE #1089: -Warray-bounds warning in SHA1Final
**Intention**
No diagnostics when compiling with -O3 optimization level.

**Describe the bug**
In `SHA1Final` there is a call to `SHA1Update(context, (unsigned char*)'\200', 1)`, and again with the literal `'\0'`. The intent here isn't to use the memory at addresses 128 and 0, but a buffer containing those values.  The character literal should be replaced with a buffer containing the values (a string literal can not be used since the code is not const-sane).

kupsch:
The code should have been a string literal not a char literal.

Fixes #1089


hainest:
This has clearly never been executed, so running the test suite is of no utility.

PR #1101, opened 2021-09-12
Cmake overhaul

jrmadsen:
## CMake Changes Overview

- Increase cmake min version to 3.13.0
- Warning if in-source build
- Rename files in cmake folder to start with Dyninst
- Moved TPL cmake files to cmake/tpls
- Replace `add_definitions` with `add_compile_definitions`
- `DyninstConfig.cmake` creates interface library
- `Dyninst::Dyninst` handles components
- CPack support
- dynamicAPI_RT uses `add_subdirectory`
- Use `CMAKE_<LANG>_VISIBILITY_PRESET`
- Use `CMAKE_VISIBILITY_INLINES_HIDDEN`
- All Find<PACKAGE>.cmake create interface library
- Fixed Thread_DB
- Significant overhaul of `dyninst_library` function
- Made significant use of interface libraries
- Removed a bunch of unnecessary cmake verbosity
- Better support for building static libraries
- bumped downloaded ElfUtils from 0.178 to 0.182 (0.178 fails to build)
- Explicit control over internal source code builds, e.g. `-DBUILD_BOOST=ON`, `-DBUILD_TBB=ON`, etc.
    - Dyninst will no longer automatically do internal builds without these set
- `DYNINST_OPTION_PREFIX` cmake option
    - converts dyninst options, e.g. `BUILD_BOOST` -> `DYNINST_BUILD_BOOST`
    - useful to prevent option naming conflicts when dyninst is a subproject
    - when activated, non-dyninst-prefixed option is set via local variable

## Minor Source Code Changes

- Fixed uninitialized variable in DispatcherARM64::getBitfieldMask
    - maybe-uninitialized warning
- anonymous namespace for StackAnalysis variables in stackanalysis.C and function.C
    - duplicate symbols in static library builds
- fix potential buffer overflow in BPatch::processCreate
    - warning about buffer overflow
- fix Symbol* symt_current_func (Object-elf.C) + extern Function* symt_current_func (parseStab.C)
    - warning about ODR violation

### Without Components

`Dyninst::Dyninst` links all component libraries (including Boost and TBB):

```cmake
    find_package(Dyninst REQUIRED)
    target_link_libraries(<target> INTERFACE Dyninst::Dyninst)
```

### With Components

`Dyninst::Dyninst` links to specified component libraries (and their dependents):

```cmake
    find_package(Dyninst REQUIRED COMPONENTS dyninstAPI parseAPI instructionAPI symtabAPI)
    target_link_libraries(<target> ... Dyninst::Dyninst)
```

### Dyninst as Subproject

Within another CMake project (e.g. git submodule):

```cmake
    set(DYNINST_OPTION_PREFIX       ON)
    set(DYNINST_BUILD_DOCS          OFF)
    set(DYNINST_QUIET_CONFIG        OFF CACHE BOOL "Suppress dyninst cmake messages")
    set(DYNINST_BUILD_PARSE_THAT    OFF CACHE BOOL "Build dyninst parseThat executable")
    set(DYNINST_BUILD_SHARED_LIBS   OFF CACHE BOOL "Build shared dyninst libraries")
    set(DYNINST_BUILD_STATIC_LIBS   ON  CACHE BOOL "Build static dyninst libraries")
    set(DYNINST_ENABLE_LTO          ON  CACHE BOOL "Enable LTO for dyninst libraries")
    
    add_subdirectory(external/dyninst)

    add_library(Dyninst::Dyninst INTERFACE IMPORTED)

    foreach(_LIB common dyninstAPI parseAPI instructionAPI symtabAPI stackwalk Boost TBB)
        target_link_libraries(Dyninst::Dyninst INTERFACE Dyninst::${_LIB})
    endforeach()

    target_link_libraries(<target> ... Dyninst::Dyninst)
```


jrmadsen:
@hainest @kupsch @bolo 

BTW, with the updates to how external packages are installed, there is a clean was to create a package dyninst into a simple installer script:

```console
cmake -B build-local -DBUILD_{BOOST,TBB,ELFUTILS,LIBIBERTY}=ON -DCMAKE_INSTALL_PREFIX=install-local .
cmake --build build-local --target package --parallel 8
cd build-local
./Dyninst-11.0.1-Linux.sh --skip-license --prefix=/opt/dyninst --exclude-subdir
```

All the external dependencies are installed into `${CMAKE_PREFIX_PATH}/lib/dyninst-tpls` and RPATHed via `$ORIGIN:$ORIGIN/lib/dyninst-tpls/lib` (on Linux).

I have some pretty extensive CI running right now validating everything and when I am confident everything is good to go, I will let you know and post the github actions script so you can see all the variants I've tested.

P.S. I know it looks like I changed a lot but _a lot_ of the changes were whitespace, e.g. indenting a bunch of nested if-else blocks. If you want to set up a zoom or something to go over the changes, let me know.

jrmadsen:
Should I bump the version to `11.1.0` or `11.0.2`? This would help cmake projects ensure that the Dyninst package found has the new cmake semantics

jrmadsen:
Ok, this is ready to go. I have a variety of tests in my dependent package (which is currently not public) but it essentially tests  combination of these various configurations:

- static libs + LTO
- shared libs
- Ubuntu 20.04 with gcc 7, 8, 9, 10
- Ubuntu 18.04 with gcc 7, 8
- Building/installing dyninst with:
    - All third-party library pre-installed
    - All third-party libraries built by dyninst
    - Pre-installed Boost and TBB + elfutils and libiberty built by dyninst
    - Packaging dyninst into a `Dyninst-11.0.1-Linux.sh` script and installing to via `./Dyninst-11.0.1-Linux.sh --skip-license --prefix=/opt/dyninst --exclude-subdir`

The framework using dyninst does validates successfully running both runtime instrumentation and doing a binary-rewrite and running the instrumented executable. 

Here is an overview of the GitHub Actions matrix:

Screen Shot 2021-09-15 at 4 56 48 PM

Here is the full GitHub Actions YAML file for reference:

```yaml
name: linux-ci

on:
  push:
    branches: [ main, develop ]
  pull_request:
    branches: [ main, develop ]

env:
  BUILD_TYPE: Release
  ELFUTILS_DOWNLOAD_VERSION: 0.183

jobs:
  ubuntu-focal:
    runs-on: ubuntu-20.04
    strategy:
      matrix:
        compiler: ['g++']
        mpi: [ '', 'libmpich-dev mpich', 'libopenmpi-dev openmpi-bin libfabric-dev' ]

    steps:
    - uses: actions/checkout@v2

    - name: Install Packages
      run:
        sudo apt-get update &&
        sudo apt-get install -y build-essential python3-pip libtbb-dev libboost-{atomic,system,thread,date-time,filesystem,timer}-dev ${{ matrix.compiler }} ${{ matrix.mpi }} &&
        python3 -m pip install --upgrade pip &&
        python3 -m pip install 'cmake==3.15.3'

    - name: Configure Env
      run:
        echo "CC=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')" >> $GITHUB_ENV &&
        echo "CXX=${{ matrix.compiler }}" >> $GITHUB_ENV &&
        echo "/opt/hosttrace/bin:${HOME}/.local/bin" >> $GITHUB_PATH &&
        echo "LD_LIBRARY_PATH=/opt/hosttrace/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV

    - name: Configure CMake
      run:
        cmake --version &&
        if [ -z "${{ matrix.mpi }}" ]; then USE_MPI=OFF; else USE_MPI=ON; fi &&
        cmake -B ${{ github.workspace }}/build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=/opt/hosttrace
          -DHOSTTRACE_USE_MPI=${USE_MPI}
          -DHOSTTRACE_USE_ROCTRACER=OFF
          -DHOSTTRACE_BUILD_DYNINST=ON
          -DDYNINST_BUILD_ELFUTILS=ON
          -DDYNINST_BUILD_LIBIBERTY=ON
          -DDYNINST_BUILD_SHARED_LIBS=ON
          -DDYNINST_BUILD_STATIC_LIBS=OFF

    - name: Build
      run:
        cmake --build ${{ github.workspace }}/build --target all --parallel 2 -- VERBOSE=1

    - name: Install
      run:
        cmake --build ${{ github.workspace }}/build --target install --parallel 2

    - name: Test
      working-directory: ${{ github.workspace }}/build
      run:
        ctest -V --output-log ${{ github.workspace }}/build/hosttrace-ctest.log

    - name: Test Install
      run:
        hosttrace --help &&
        hosttrace -- sleep 1 &&
        hosttrace -o sleep.inst -- sleep &&
        ./sleep.inst 1 &&
        rm ./sleep.inst

    - name: Artifacts
      uses: actions/upload-artifact@v2
      with:
        name: ctest-log
        path: |
          ${{ github.workspace }}/build/hosttrace-ctest.log

  ubuntu-bionic:
    runs-on: ubuntu-18.04
    strategy:
      matrix:
        compiler: ['g++-7', 'g++-8']
        mpi: [ '', 'libmpich-dev mpich', 'libopenmpi-dev openmpi-bin libfabric-dev' ]

    steps:
    - uses: actions/checkout@v2

    - name: Install Packages
      run:
        sudo apt-get update &&
        sudo apt-get install -y build-essential python3-pip ${{ matrix.compiler }} ${{ matrix.mpi }} &&
        python3 -m pip install --upgrade pip &&
        python3 -m pip install 'cmake==3.15.3'

    - name: Configure Env
      run:
        echo "CC=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')" >> $GITHUB_ENV &&
        echo "CXX=${{ matrix.compiler }}" >> $GITHUB_ENV &&
        echo "/opt/hosttrace/bin:${HOME}/.local/bin" >> $GITHUB_PATH &&
        echo "LD_LIBRARY_PATH=/opt/hosttrace/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV

    - name: Configure CMake
      run:
        cmake --version &&
        if [ -z "${{ matrix.mpi }}" ]; then USE_MPI=OFF; else USE_MPI=ON; fi &&
        cmake -B ${{ github.workspace }}/build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=/opt/hosttrace
          -DHOSTTRACE_USE_MPI=${USE_MPI}
          -DHOSTTRACE_USE_ROCTRACER=OFF
          -DHOSTTRACE_BUILD_DYNINST=ON
          -DDYNINST_BUILD_TBB=ON
          -DDYNINST_BUILD_BOOST=ON
          -DDYNINST_BUILD_ELFUTILS=ON
          -DDYNINST_BUILD_LIBIBERTY=ON

    - name: Build
      run:
        cmake --build ${{ github.workspace }}/build --target all --parallel 2 -- VERBOSE=1

    - name: Install
      run:
        cmake --build ${{ github.workspace }}/build --target install --parallel 2

    - name: Test
      working-directory: ${{ github.workspace }}/build
      run:
        ctest -V --output-log ${{ github.workspace }}/build/hosttrace-ctest.log

    - name: Test Install
      run:
        hosttrace --help &&
        hosttrace -- sleep 1 &&
        hosttrace -o sleep.inst -- sleep &&
        ./sleep.inst 1 &&
        rm ./sleep.inst

    - name: Artifacts
      uses: actions/upload-artifact@v2
      with:
        name: ctest-log
        path: |
          ${{ github.workspace }}/build/hosttrace-ctest.log

  ubuntu-focal-external:
    runs-on: ubuntu-20.04
    strategy:
      matrix:
        compiler: ['g++-7', 'g++-8', 'g++-9', 'g++-10']

    steps:
    - uses: actions/checkout@v2

    - name: Install Packages
      run:
        sudo apt-get update &&
        sudo apt-get install -y build-essential python3-pip libboost-{atomic,system,thread,date-time,filesystem,timer}-dev libtbb-dev libiberty-dev ${{ matrix.compiler }} &&
        sudo python3 -m pip install --upgrade pip &&
        python3 -m pip install 'cmake==3.15.3'

    - name: Configure Env
      run:
        echo "CC=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')" >> $GITHUB_ENV &&
        echo "CXX=${{ matrix.compiler }}" >> $GITHUB_ENV &&
        echo "CMAKE_PREFIX_PATH=/opt/opt/dyninst:/opt/elfutils:${CMAKE_PREFIX_PATH}" >> $GITHUB_ENV &&
        echo "/opt/hosttrace/bin:/opt/dyninst/bin:/opt/elfutils/bin:${HOME}/.local/bin" >> $GITHUB_PATH &&
        echo "LD_LIBRARY_PATH=/opt/hosttrace/lib:/opt/dyninst/lib:/opt/elfutils/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV

    - name: Install ElfUtils
      run:
        pushd external &&
        wget https://sourceware.org/elfutils/ftp/${ELFUTILS_DOWNLOAD_VERSION}/elfutils-${ELFUTILS_DOWNLOAD_VERSION}.tar.bz2 &&
        tar xjf elfutils-${ELFUTILS_DOWNLOAD_VERSION}.tar.bz2 &&
        pushd elfutils-${ELFUTILS_DOWNLOAD_VERSION} &&
        CFLAGS="-g -O2" ./configure --enable-install-elfh --prefix=/opt/elfutils --disable-libdebuginfod --disable-debuginfod &&
        make -j2 &&
        make install -j2 &&
        popd &&
        rm -rf elfutils*

    - name: Install Dyninst
      run:
        cmake --version &&
        git submodule update --init external/dyninst &&
        cd external/dyninst &&
        cmake -B build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=/opt/dyninst &&
        cmake --build build --target all --parallel 2 &&
        cmake --build build --target install --parallel 2 &&
        rm -rf build

    - name: Configure CMake
      run:
        cmake --version &&
        cmake -B ${{ github.workspace }}/build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=/opt/hosttrace
          -DHOSTTRACE_USE_MPI=OFF
          -DHOSTTRACE_USE_ROCTRACER=OFF

    - name: Build
      run:
        cmake --build ${{ github.workspace }}/build --target all --parallel 2 -- VERBOSE=1

    - name: Install
      run:
        cmake --build ${{ github.workspace }}/build --target install --parallel 2

    - name: Test
      working-directory: ${{ github.workspace }}/build
      run:
        ldd ./hosttrace &&
        ./hosttrace --help &&
        ctest -V --output-log ${{ github.workspace }}/build/hosttrace-ctest.log

    - name: Test Install
      run:
        ldd $(which hosttrace) &&
        hosttrace --help &&
        hosttrace -- sleep 1 &&
        hosttrace -o sleep.inst -- sleep &&
        ./sleep.inst 1 &&
        rm ./sleep.inst

    - name: Artifacts
      uses: actions/upload-artifact@v2
      with:
        name: ctest-log
        path: |
          ${{ github.workspace }}/build/hosttrace-ctest.log

  ubuntu-focal-dyninst-package:
    runs-on: ubuntu-20.04
    strategy:
      matrix:
        compiler: ['g++']

    steps:
    - uses: actions/checkout@v2

    - name: Install Packages
      run:
        sudo apt-get update &&
        sudo apt-get install -y build-essential python3-pip ${{ matrix.compiler }} &&
        sudo python3 -m pip install --upgrade pip &&
        python3 -m pip install 'cmake==3.15.3'

    - name: Configure Env
      run:
        echo "CC=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')" >> $GITHUB_ENV &&
        echo "CXX=${{ matrix.compiler }}" >> $GITHUB_ENV &&
        echo "CMAKE_PREFIX_PATH=/opt/opt/dyninst:${CMAKE_PREFIX_PATH}" >> $GITHUB_ENV &&
        echo "/opt/hosttrace/bin:/opt/dyninst/bin:${HOME}/.local/bin" >> $GITHUB_PATH &&
        echo "LD_LIBRARY_PATH=/opt/hosttrace/lib:/opt/dyninst/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV

    - name: Install Dyninst
      run:
        cmake --version &&
        git submodule update --init external/dyninst &&
        cd external/dyninst &&
        cmake -B build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/external/dyninst-install
          -DBUILD_TBB=ON
          -DBUILD_BOOST=ON
          -DBUILD_ELFUTILS=ON
          -DBUILD_LIBIBERTY=ON &&
        cmake --build build --target package --parallel 4 &&
        mkdir /opt/dyninst &&
        ./build/Dyninst-*-Linux.sh --prefix=/opt/dyninst --exclude-subdir --skip-license &&
        rm -rf build

    - name: Configure CMake
      run:
        cmake --version &&
        cmake -B ${{ github.workspace }}/build
          -DCMAKE_C_COMPILER=$(echo '${{ matrix.compiler }}' | sed 's/+/c/g')
          -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
          -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
          -DCMAKE_INSTALL_PREFIX=/opt/hosttrace
          -DHOSTTRACE_USE_MPI=OFF
          -DHOSTTRACE_USE_ROCTRACER=OFF

    - name: Build
      run:
        cmake --build ${{ github.workspace }}/build --target all --parallel 2 -- VERBOSE=1

    - name: Install
      run:
        cmake --build ${{ github.workspace }}/build --target install --parallel 2

    - name: Test
      working-directory: ${{ github.workspace }}/build
      run:
        ldd ./hosttrace &&
        ./hosttrace --help &&
        ctest -V --output-log ${{ github.workspace }}/build/hosttrace-ctest.log

    - name: Test Install
      run:
        ldd $(which hosttrace) &&
        hosttrace --help &&
        hosttrace -- sleep 1 &&
        hosttrace -o sleep.inst -- sleep &&
        ./sleep.inst 1 &&
        rm ./sleep.inst

    - name: Artifacts
      uses: actions/upload-artifact@v2
      with:
        name: ctest-log
        path: |
          ${{ github.workspace }}/build/hosttrace-ctest.log
```

hainest:
>Should I bump the version to 11.1.0 or 11.0.2?

No, we are planning a 12.0 release likely by the end of the month. Optimally, I would like to have these changes ready for that release.

>Ok, this is ready to go

Ok. I will go through this in detail. My initial notes are

1. Please try to minimize whitespace changes when making code changes. If you are using a formatter, just put those pure whitespace changes in a separate commit. I've been using cmake-format on newly-added files, but I haven't generated a .cmake-format.yaml file. If you have one that you like, please feel free to add it to this PR.
2. b84d1fa5c831f30f1911ea26e334967c87637281 is way too big. The non-CMake code changes should be in their own PR, and the changes to dyninst_library should be their own commit. Speaking of, OpenMP shouldn't be a dependency for every library.
3. What is the purpose of DYNINST_OPTION_PREFIX when DYNINST_ADD_OPTION allows for using the previous forms?
4. Why the renames of the CMake files from "Foo.cmake" to "DyninstFoo.cmake"?
5. As I mentioned in the corresponding issue, we don't want to support Windows anymore, so there's no need to make a bunch of changes to that functionality here. We can just remove it (preferably as its own commit).
6. I want to get rid of our custom FindBoost.cmake, so let's just remove it here instead of patching it. Indeed, I want to get rid of as many custom find modules as possible. I've not checked on the shipping modules for CMake in a while, but it would be good to get rid of the elfutils/libdwarf and TBB ones, as well, if possible.
7.  I agree we should fix relative path names in source lists, but we should fix all of them.
8. "bumped ElfUtils from 0.178 to 0.182" We try not to move minimum requirements without a good reason. What's in 0.182 that's needed here?

I see you put a large amount of effort into accommodating building the third-party dependencies from source. As I noted in the corresponding issue, we want to move away from that model. Effectively none of our users are using that functionality, and the maintenance time we've put into it is just not worthwhile. We are also going to be offering containerized versions of Dyninst releases going forward, so this has even less utility. Let's go ahead and just remove it here.

jrmadsen:
>  1. Please try to minimize whitespace changes when making code changes. If you are using a formatter, just put those pure whitespace changes in a separate commit. I've been using cmake-format on newly-added files, but I haven't generated a .cmake-format.yaml file. If you have one that you like, please feel free to add it to this PR.

I completely empathize with the pure whitespace changes being an issue. It's why I always have a CI jobs which requires clang-format to pass for C/C++ codes and black to pass for Python codes before merging a PR in every repo I maintain. But, to be honest, I couldn't even open a file and hit save in my IDE without ending up with 20 lines of whitespace cleanup anywhere -- in the Dyninst source code or the cmake -- so asking me to spend an hour or two rewriting my git history so that the whitespace changes are separate when there hasn't been any enforcement of formatting in the past seems like a pretty huge ask. I'm not saying I won't but I want you to understand where I'm coming from.

In case you didn't know, you can tell the github viewer to "Hide Whitespace Changes":

Screen Shot 2021-09-15 at 11 30 01 PM

>  2. [b84d1fa](https://github.com/dyninst/dyninst/commit/b84d1fa5c831f30f1911ea26e334967c87637281) is way too big.

I'll see what I can do, this seems like it would be part of the aforementioned whitespace-changes-only commit.

> The non-CMake code changes should be in their own PR

I figured that would be the case but I had to commit it because it was the only way I could get the static library builds to work.

> and the changes to dyninst_library should be their own commit. 

I'm not sure about this -- a lot of the changes are intertwined here.

> Speaking of, OpenMP shouldn't be a dependency for every library.

I'm not necessarily disagreeing here but why exactly does this matter? If you don't use `#pragma omp` or include `<omp.h>`, as far as I can tell, the only thing that adding OpenMP as a dependency does is link in the pthread library and add in 4 (seemingly) inconsequential symbols related to offload support:

```console
36a37,40
> 0000000000002088 R __offload_funcs_end
> 0000000000002088 R __offload_func_table
> 0000000000002088 R __offload_vars_end
> 0000000000002088 R __offload_var_table
```

> 3. What is the purpose of DYNINST_OPTION_PREFIX when DYNINST_ADD_OPTION allows for using the previous forms?

Parent project has cache variable BUILD_SHARED_LIBS but wants to build dyninst with PIC and static libraries. Without `DYNINST_OPTION_PREFIX` there is no way to set `BUILD_SHARED_LIBS=ON` in the cache without it propagating down to Dyninst.

> 4. Why the renames of the CMake files from "Foo.cmake" to "DyninstFoo.cmake"?

It is considered a good practice for all project-specific cmake files to be prefixed with their project name. This allows parent projects to add Dyninst's module path (if needed) to their cmake module path and include them without ambiguity and without hardcoding the path to that module file. I used to not think it was a big deal until I encountered it -- I suspect that the reason that `shared.cmake` was using:

```cmake
include (${DYNINST_ROOT}/cmake/platform.cmake)
include (${DYNINST_ROOT}/cmake/cap_arch_def.cmake)
include (${DYNINST_ROOT}/cmake/visibility.cmake)
include (${DYNINST_ROOT}/cmake/warnings.cmake)
include (${DYNINST_ROOT}/cmake/options.cmake)
include (${DYNINST_ROOT}/cmake/optimization.cmake)
include (${DYNINST_ROOT}/cmake/endian.cmake)
```

instead of:

```cmake
include(platform)
include(cap_arch_def)
include(visibility)
include(warnings)
include(options)
include(optimization)
include(endian)
```

was because someone encountered some ambiguity in the past and that was the patch. Either that or someone either really likes typing or didn't know how include + the module path works.

> 5. As I mentioned in the corresponding issue, we don't want to support Windows anymore, so there's no need to make a bunch of changes to that functionality here. We can just remove it (preferably as its own commit).

Oh good. I was going to mention that I did zero testing whether anything broke on Windows and didn't plan to do any.

> 6. I want to get rid of our custom FindBoost.cmake, so let's just remove it here instead of patching it. Indeed, I want to get rid of as many custom find modules as possible. 

I think it's worth keeping in this `FindBoost.cmake` even though cmake distributes one. This is one of the rare exceptions where I hold this opinion because the cmake distributed one does not have the `Boost_FIND_QUIETLY` I added:

```cmake
    if(NOT Boost_VERSION VERSION_LESS 107100 AND NOT Boost_FIND_QUIETLY)
      message(WARNING "New Boost version may have incorrect or missing dependencies and imported targets")
    endif()
```

and since it is a `WARNING` instead of an `AUTHOR_WARNING`, you cannot just suppress it as a developer warning and your cmake configure gets very noisy.

> I've not checked on the shipping modules for CMake in a while, but it would be good to get rid of the elfutils/libdwarf and TBB ones, as well, if possible.

None of these 3 are distributed by cmake.
 
> 8. "bumped ElfUtils from 0.178 to 0.182" We try not to move minimum requirements without a good reason. What's in 0.182 that's needed here?

The original message is wrong. I bumped the download version to 0.182 because 0.178 fails to compile because it uses `-Werror` and there is a warning with more recent GCC compilers about a function which gets called with something along the lines of `int*` and the prototype is `int v[6]`.
 
> I see you put a large amount of effort into accommodating building the third-party dependencies from source. As I noted in the corresponding issue, we want to move away from that model. Effectively none of our users are using that functionality, and the maintenance time we've put into it is just not worthwhile. 

Well, we want to use this feature (by we, I mean my group at AMD) because it avoids having to use spack or something else to get all the dependencies. And even when one does have sudo capabilities, even recent versions such as Ubuntu 20.04 don't have recent enough versions of all the dependencies in apt to just apt-get all of them. 

> We are also going to be offering containerized versions of Dyninst releases going forward, so this has even less utility. 

I absolutely love containers, use them all the time. But tools involving instrumentation (e.g. profilers) aren't ideal for container distribution (which I believe are the primary users of dyninst) because most developers, when they build their container that (later) they wish to profile, are forced to either always install the profiler in the image (leading to significantly larger images -- just take a look at the sizes of the E4S images with TAU, Caliper, HPCToolkit, etc. installed) or they are forced to rebuild their image when they need the profiler. 

Plus, with all the updates to how 3rd-party libraries are installed (which relocate all them to `<prefix>/lib/dyninst-tpls` to avoid overwriting existing installs) and the CPack support, it would be possible for dyninst to actually be distributed via Linux package managers -- making installing dyninst in a container as easy as `apt-get install dyninst`.

hainest:
We should find some time to chat over your preferred teleconferencing system. There are some changes here which add burden to Dyninst's maintenance that I'm not sure we want to take on. I need to better understand how you are using Dyninst and why it is so fundamentally different from all of our other users. My schedule is probably more open than yours, so do you want to try to find some time to chat?

jrmadsen:
I want to pull in the changes from 12.1 but all the improvements to the build system are a mess to merge. I need to know whether these changes will be made a priority such that we can put a freeze on any non-trivial cmake changes until they are merged

hainest:
@jrmadsen I _really_ want to get this in. I also want to break the world because the current CMake is a horror show- including the patches I've put in over the years. I've tried to pick this up several times over the last few months, but it always gets pushed back. One of our core developers left in September, so there are only three of us now and we are each working on at least one external research project. This has left Dyninst maintenance in flux. As I noted before, we don't want to merge this as-is. We are immensely appreciative that you took the time to effectively re-write the whole thing, but we need to understand and document the changes systematically rather than as a monolith. That will just take time. As I noted previously, we are getting out of the business of automatically providing builds of dependencies. We've put too much time into a feature that has sprouted bugs. Because no one noticed them until we found them, we know that our consumers aren't using this 'feature'. There's no reason for us to spend our limited time maintaining it. Moreover, this will only get worse as we are planning to add more dependencies over time (top on the list is Capstone for instruction decoding). I do agree that it would be great to be able to distribute Dyninst through OS package managers like apt. In fact, the Red Hat folks already package their own build as an rpm.

jrmadsen:
>  As I noted before, we don't want to merge this as-is. We are immensely appreciative that you took the time to effectively re-write the whole thing, but we need to understand and document the changes systematically rather than as a monolith. That will just take time. 

Please see #1263, #1264, and #1265 -- I split up the chances as much as humanly possible. #1263 is just formatting. #1264 just renames some files and moves some things around. #1265 is where all the edits occur. Once #1263 is merged, I can rebase #1264. Once #1264 is merged, I can rebase #1265 and all the edits will be much clearer.

> As I noted previously, we are getting out of the business of automatically providing builds of dependencies. We've put too much time into a feature that has sprouted bugs. 

If y'all want to get away from it, that's fine. All I ask is that this happens _after_ all these are merged. Once we are synchronized, it will be much easier for me to maintain a fork.

> Because no one noticed them until we found them, we know that our consumers aren't using this 'feature'. There's no reason for us to spend our limited time maintaining it. 

People didn't use that feature because you had a severely flawed model that didn't allow you to run cmake more than once without building the third-party libraries completely. All the problems are fixed and it is extremely portable now, at least on Linux -- I've built my edits on _numerous_ Linux flavors (Ubuntu, CentOS, OpenSUSE) and several different versions of those OSes. 

And we are using this feature extensively.

> Moreover, this will only get worse as we are planning to add more dependencies over time (top on the list is Capstone for instruction decoding). 

All these problems occur because these particular SW packages do not use CMake and therefore require building via `externalproject_add`. Capstone [uses cmake](https://github.com/capstone-engine/capstone/blob/master/CMakeLists.txt) which makes it generally very easy to support as a subproject via a gitsubmodule + `add_subdirectory`.

> I do agree that it would be great to be able to distribute Dyninst through OS package managers like apt. In fact, the Red Hat folks already package their own build as an rpm.

Please see the new workflow which validates the deb installer (along with a self-extracting tarball script): 

https://github.com/jrmadsen/dyninst/blob/40dfb56eae006e683eba8a1b4f4d41c2e31d5527/.github/workflows/test-packaging.yaml#L64

Build the RPM would simply be something like: `cpack -G RPM -D CPACK_PACKAGING_INSTALL_PREFIX=/opt/dyninst`

PR #1105, opened 2021-09-18
[DO NOT MERGE] Formatting

jrmadsen:
@hainest 

I provided the infrastructure to enforce cmake and source code formatting (via clang-format, v9) during every PR.

If you want to adopt this, please cherry-pick 6a339b6 after #1104 because I'd prefer for it to not look like I've written the entire dyninst source code in `git blame`.


hainest:
Everything here is either merged or in a new PR.

PR #1106, opened 2021-10-01
Adding Dockerfile to make Dyninst easier/faster to develop with

vsoch:
Hi Dyninst! I thought it would be useful to have a Dockerfile alongside the repository that makes it easier to develop Dyninst. Instructions and detail are added to the README in this pull request, and let me know if you have any questions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM ghcr.io/autamus/cmake:3.21.2
+
+# docker build -t dyninst .
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y build-essential python3 python3-dev python3-pip gringo git
+
+WORKDIR /opt
+RUN git clone --depth 1 https://github.com/spack/spack && \
+    pip3 install --upgrade pip && \
+    pip3 install botocore boto3
+
+RUN apt-get install -y software-properties-common && \
+    add-apt-repository 'deb http://mirrors.kernel.org/ubuntu hirsute main universe' && \
+    apt-get update && \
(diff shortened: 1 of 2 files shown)


vsoch:
Just a quick note - I'm working on figuring out a good development workflow after I have the container (this is my first time using spack in this way) so I'll update the PR when I document it!

vsoch:
Okay - Dockerfile and README are updated with the full development workflow that seems to work! This could be extended to point to a quick start guide to show the user a basic CMakeLists.txt (or other setup) but probably for a seasoned Dyninst developer, the most likely user of this, that isn't necessary.

hainest:
We have plans to eventually distribute Dyninst through Docker, but we don't want to intermix that effort here. We have a dockerhub repo (https://hub.docker.com/orgs/dyninst/repositories), but it's currently empty. It looks like the Dockerfile you have here would be the Ubuntu-21.04 version. I think that's as good a place as any to start. We'd like to support the LTS Ubuntus since 18.04 both natively and with spack-- this isn't always feasible because of our dependency minimum versions. Does dockerhub have a PR mechanism?

vsoch:
Docker Hub does not - but I would recommend an automated deployment from a repository on GitHub to a registry. I’d also suggest using a different registry since Docker hub has rate limiting and has toyed with image purging. Would you like me to PR to the other repository with a CI recipe to employ to GitHub packages? And you could push to Docker Hub too if you like, given some brave soul has offered their credential in he environment.

hainest:
Will be replaced by https://github.com/researchapps/dyninst-container

PR #1107, opened 2021-10-05
Adding Support for AMDGPU CDNA2 Architectures based on XML ISA spec.

bbiiggppiigg:
This branch is used for ongoing support for the AMDGPU CDNA2 architecture.
So far only instruction names are decoded, support for decoding operands is still under development.


hainest:
@bbiiggppiigg Do you feel this should go into the 12.0 release?

bbiiggppiigg:
> @bbiiggppiigg Do you feel this should go into the 12.0 release?

No, I don't think so.
The ISA spec is still subjected to change, and based on their change my implementation might change as well.

common/h/dyn_regs.h
```diff
@@ -1638,6 +1640,149 @@ namespace Dyninst
...
+      const signed int BITS_48      = 0x00000C00;
+      const signed int BITS_64      = 0x00000D00;
+      const signed int BITS_128     = 0x00000E00;
+      const signed int BITS_256     = 0x00000F00;
+      const signed int BITS_512     = 0x00001000;
+
+
+
+
+      /*const signed int BIT     = 0x00001000;
```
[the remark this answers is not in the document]
bbiiggppiigg: should be done now

common/h/dyn_regs.h
```diff
@@ -1638,6 +1640,149 @@ namespace Dyninst
...
```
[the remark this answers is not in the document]
bbiiggppiigg: should be done now

common/src/dyn_regs.C
```diff
@@ -195,6 +216,48 @@ std::string MachRegister::name() const {
                     break;
             }
             return ret;
+        }else if(getArchitecture() == Arch_amdgpu_cdna2){
+            signed int category = reg & 0x00ff0000;
+            signed int base_val = reg & 0x000000ff;
+            switch(category){
+                case amdgpu_cdna2::SGPR_VEC2:
+                    base_val -=  (amdgpu_cdna2::sgpr_vec2_0&0xff);
+                    ret = std::string("SGPR["+std::to_string(base_val)+"-"+std::to_string(base_val+1)+"]");
```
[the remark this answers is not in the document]
bbiiggppiigg: This will be done when the functionality is properly implemented

bbiiggppiigg:
@hainest 
Can you review these commits for me?

hainest:
@bbiiggppiigg Sorry, I must have missed the email notice. If you click the "re-request review" button, it will send me a Github notice (I've already clicked it). I will try to get to this soon. There is a lot of code to go through here, but I'll assume you put the opcode tables together correctly.

common/h/dyn_regs.h
```diff
@@ -124,1891 +128,2114 @@ namespace Dyninst
...
+        const signed int BITS_48      = 0x00000C00;
+        const signed int BITS_64      = 0x00000D00;
+        const signed int BITS_128     = 0x00000E00;
+        const signed int BITS_256     = 0x00000F00;
+        const signed int BITS_512     = 0x00001000;
+
+#include "amdgpu_vega_sys_regs.h"
     }
 
+    namespace amdgpu_cdna{
```
hainest: Let's make all of the added constants `constexpr int`.

common/src/dyn_regs.C
```diff
@@ -160,41 +182,83 @@ std::string MachRegister::name() const {
...
+                    ret = iter->second;
                     break;
+            }
+            return ret;
+        }else if(getArchitecture() == Arch_amdgpu_cdna2){
+            signed int category = reg & 0x00ff0000;
+            signed int base_val = reg & 0x000000ff;
+            switch(category){
+                /*case amdgpu_cdna2::SGPR_VEC2:
+                  base_val -=  (amdgpu_cdna2::sgpr_vec2_0&0xff);
```
hainest, line 225: Let's remove this commented-out code, if it's not needed.

hainest (changes requested):
There are many places where a check for a cdna2 variable happens, but there isn't a corresponding check for the cdna variable. Are those implementations intended to asymmetrical?

bbiiggppiigg:
So far I have manual support for Vega, and XML for CDNA2. 
CDNA was never really supported in anyway, and I should and will remove those entries.

common/h/dyn_regs.h
bbiiggppiigg: I'm not sure why this is out dated, Maybe the file is edited in the main repo? 
Right now the highlight is on a namespace line so I'm not sure what needs to be constexpr int.

common/src/dyn_regs.C
bbiiggppiigg: Yeah Ill clean these up.

common/h/dyn_regs.h
hainest: It looks like I added the comment to one of the older commits in the branch and the file was subsequently updated, so it labelled it as outdated.

>Right now the highlight is on a namespace line so I'm not sure what needs to be constexpr int.

That comment was to change the declaration of the namespace members. In a previous commit, they were declared `static const signed int`, but now they're just `const signed int`. Making them `static` wasn't useful, so I was recommending to change them to `constexpr`. You still could, but it's not strictly necessary.

hainest:
> So far I have manual support for Vega, and XML for CDNA2. CDNA was never really supported in anyway, and I should and will remove those entries.

Ah, ok. Removing the CDNA is probably the best solution.

bbiiggppiigg:
I'll create an issue to track the TODO item ( I think it is the one about architecture name, right ?)
I'm not sure what you mean by preserve history of the branch.

hainest:
There were several that I saw:

1. dwarf/src/dwarfHandle.C:200
2. In the commit message of 51d3f8dfc5
3. common/h/dyn_regs.h:1699,1700
4. common/src/dyn_regs.C:502

A whole bunch in the AMDGPU directories: ` grep -Rn TODO instructionAPI/src/AMDGPU`

>I'm not sure what you mean by preserve history of the branch.

Github has three methods of merging a PR:

1. squash-and-merge: all of your commits are packed together into a single commit
2. rebase-and-merge: applies each of your commits one at a time to master. I do not recommend this as you need to make sure every commit can compile in case they are bisected into.
3. merge-commit: create a merge commit that retains all of the commits of the PR. It's not my favorite because it diverges the history, but it can preserve all of your work and is bisectable.

If you don't want to keep a record of every individual change on the bbiiggppiigg/support_cdna2 branch, I recommend squash-and-merge.

bbiiggppiigg:
I think a squash and merge should be good enough.
Even though I tried to split the commits into meaningful small batches, I don't think I've done it successfully.

PR #1113, opened 2021-10-19
[WIP] Remove support for ELF stabs

hainest:


hainest:
>There are 4 mentions of "stabs" that should also be removed in the two files symtabAPI/doc/[14]*.tex

Weird. I thought I included the docs in my search... I'll get those.

>Their are some other uses of "stab" as a variable name (to mean function/block/region) and in DynC_API/tests/*/foo.txt (maybe this file can be removed) that can be part or a future cleanup.

I talked to Xiaozhu about those and confirmed they are just bad local variable names and have nothing to do with stabs.

>This branch also fails to build on coriander

>The declaration of Object::loaded_elf does not match the definition. In symtabAPI/src/Object-elf.{h,C}

I discovered that when I was testing this PR. There is a cascade of errors after fixing that, so I might have broken something. I'm working on it.

PR #1136, opened 2021-11-01
Unify meaning of 'cap_32_64' macro

hainest:
Currently, this is used in several different ways due to the historical
original meaning drifting over the years. It's meaning is now "this host
64-bit platform supports modifying 32-bit binaries". Currently, this is
only x86 and PPC32. Although the latter is going away soon due to
obsolescence.

Requires https://github.com/dyninst/testsuite/pull/202


kupsch (approved):
This looks correct.

mxz297 (approved):

PR #1137, opened 2021-11-02
Remove support for Cray CNL

### LINKED ISSUE #1133: Remove cnl platform detection
It's not a valid platform: https://github.com/dyninst/dyninst/blob/master/cmake/platform_unix.cmake#L3

hainest:
The Compute Node Linux (CNL) was for the old Cray XT/XE platforms that
are obsolete now. This should have been removed by e5484368f in
2021. Support for this platform was removed from the test suite by
7e4ab7c12 in 2012.

Closes #1133


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1137&dyninst_build_status=OK

No regressions. Also builds fine on modern Cray systems.

PR #1139, opened 2021-11-02
Implement ppc64 in MachRegister::getROSERegister

hainest:
Previously, ROSE didn't support ppc64, so the ppc32 registers were used.

@mxz297 The only piece I'm not sure about is `baseID = reg & 0x0000FFFF;`. This was being used for ppc64 previously, so I'm assuming it's ok. Could you verify?

### DIFF
--- a/common/src/dyn_regs.C
+++ b/common/src/dyn_regs.C
@@ -1069,47 +1069,75 @@ void MachRegister::getROSERegister(int &c, int &n, int &p)
             }
             break;
         case Arch_ppc32:
-        case Arch_ppc64: // 64-bit not supported in ROSE
+        {
+            baseID = reg & 0x0000FFFF;
+            n = baseID;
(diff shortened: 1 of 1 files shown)


mxz297 (approved):
I think `baseID = reg & 0x0000FFFF` is fine as the value layout for registers on ppc64 is the same for ppc32.

hainest:
@mxz297 That's what I thought, but I wanted to make sure. I've not seen any regressions on the power machines I've tested thus far. Should we try testing this further- you mentioned trying to parse a bunch of ppc binaries to exercise the Rose code.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1139

The regressions are unrelated to this change. They are due to OS/distribution upgrades that we haven't yet caught up to. I will submit issues/PRs for those separately.

PR #1140, opened 2021-11-02
dataflowAPI/ABI: correctly differentiate between ppc32 and ppc64

hainest:


hainest:
@mxz297 This change caused a very large number of test regressions. Am I missing something here?

mxz297:
@hainest I believe this is likely caused by the internal issues of mixing ppc32 and ppc64 logic. For instruction decoding for power (https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/InstructionDecoder-power.C), if you search ppc32, you can still see many ppc32 registers are used. I know we do conversion from ppc32 register to ppc64 register at various places, but i believe this is the root cause of the problems.

hainest:
@mxz297 Sorry, I meant to post the backtrace from this yesterday. Let me open an issue.

PR #1143, opened 2021-11-05
fix statement-like macros

kupsch:
wrap macro body in 'do {...} while (0)' so they can be used as if they were a
statement with a terminating semicolon.

### DIFF
--- a/dwarf/src/dwarfResult.C
+++ b/dwarf/src/dwarfResult.C
@@ -43,7 +43,7 @@ using namespace Dyninst;
 using namespace DwarfDyninst;
 using namespace std;
 
-#define CHECK_OPER(n) if (operands.size() < n) { error = true; break; }
+#define CHECK_OPER(n) do { if (operands.size() < (n)) { error = true; return; } } while (0)
 
 void SymbolicDwarfResult::pushReg(MachRegister reg) {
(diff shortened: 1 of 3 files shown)


hainest:
It should be illegal to abuse the preprocessor like that. Should we go ahead and add a terminating semicolon? If these macros are used without one, then we'll get a compilation error.

kupsch:
They weren't used with a semicolon.  This is just to prevent this problem in the future. With this change, there will be a compilation error if used without a terminating semicolon. They can only be used in the code where a statement is expected. Previously some of these expanded to a block which didn't require one (and you'd end up with a syntax error if a semicolon was used in the some cases like true clause of an `if` statement that has an `else` statement.

hainest:
That is somehow a worse usage than what I expected.

hainest (approved):

hainest:
Did you run these through the usual testing battery?

kupsch:
I have only built them on x86.

hainest:
Ok. I'll give them a quick twirl on power and arm- just to be sure.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1143

LGTM. Don't worry about that failure on ARM. It's a proccontrol intermittent issue.

PR #1145, opened 2021-11-06
Remove ppc32 from builds

hainest:
This removes the unsupported 32-bit PowerPC architecture from the build. This does *not* remove the ppc32 ISA and associated support code. There is too much intertwining of ppc32 and ppc64 to do that in a single pass (it's a mess). I'll work on it in the future.


kupsch (approved):

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1145

The regressions on Summit are unrelated.

PR #1147, opened 2021-11-09
Improve docs for lookup functions in CodeObject

### LINKED ISSUE #1102: ParseAPI equivalent to SymtabAPI getContainingFunction
Would like to find the function containing an address.  ParseAPI has the following function described in the manual:

bool getContainingFunction(Offset offset,
Function *&ret)

This method returns the function, if any, that contains the provided offset. Returns true on
success and false on failure. The error value is set to No_Such_Function. Note that this method
does not parse, and therefore relies on the symbol table for information. As a result it may return
incorrect information if the symbol table is wrong or if functions are either non-contiguous or
overlapping. For more precision, use the ParseAPI library.

However would like to take the advice of the SymtabAPI manual and using ParseAPI instead of SymtabAPI to get this information.  Looking through the ParseAPI manual the following function looks to be the closest in functionality:

int findFuncs(CodeRegion * cr,
Address addr,
std::set<Function*> & funcs)

Finds all functions spanning addr in the code region, adding each to funcs. The number of results
of this stabbing query are returned.

The question is how to get an appropriate CodeRegion for ParseAPI findfuncs?  Or is there a better alternative to map address back to the containing function than findfuncs? When I attempted to compile code using findFuncs passing in an CodeRegions that is unitialized gcc flagged that as an error:

  CXX      stap-analysis.o
analysis.cxx: In function ‘int liveness(const char*, Dwarf_Addr, location_context)’:
analysis.cxx:63:51: error: ‘cr’ may be used uninitialized [-Werror=maybe-uninitialized]
   63 |         bool found = func_to_analyze.co->findFuncs(cr, addr, f);
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

hainest:
Add wording to improve clarity about when a CodeRegion is strictly needed when using the lookup functions.

Closes #1102 

@wcohen @mxz297

From poking around in the code, it looks like the only place that the ParseData interface is exposed is through member functions in CodeObject. I think that's the best place to document this.

The interface for ParseData is

    1. Function * findFunc(CodeRegion *, Address);
    2. Block * findBlock(CodeRegion *, Address);
    3. int findFuncs(CodeRegion *, Address, set<Function*> &);
    4. int findFuncs(CodeRegion *, Address, Address, set<Function*> &);
    5. int findBlocks(CodeRegion *, Address, set<Block*> &);
    6. ParseFrame * findFrame(CodeRegion *, Address);
    7. ParseFrame::Status frameStatus(CodeRegion *, Address addr);
    8. void setFrameStatus(CodeRegion*,Address,ParseFrame::Status);

CodeObject provides

    Function * findFuncByEntry(CodeRegion * cr, Address entry);
			-> maps to 1

    int findFuncsByBlock(CodeRegion *cr, Block* b, std::set<Function*> &funcs);
			-> Does not map to ParseData. CodeRegion is REQUIRED.

    int findFuncs(CodeRegion * cr, Address addr, std::set<Function*> & funcs);
			-> maps to 3

    int findFuncs(CodeRegion * cr, Address start, Address end, std::set<Function*> & funcs);
			-> maps to 4

    int findCurrentFuncs(CodeRegion * cr, Address addr, std::set<Function*> & funcs);
			-> maps to 4

    Block * findBlockByEntry(CodeRegion * cr, Address entry);
			-> maps to 2

    int findBlocks(CodeRegion * cr, Address addr, std::set<Block*> & blocks);
			-> maps to 5

    int findCurrentBlocks(CodeRegion * cr, Address addr, std::set<Block*> & blocks);
			-> maps to 5

    Block * findNextBlock(CodeRegion * cr, Address addr);
			-> maps to 5

### DIFF
--- a/parseAPI/doc/API/CodeObject.tex
+++ b/parseAPI/doc/API/CodeObject.tex
@@ -92,6 +92,15 @@ void parseGaps(CodeRegion *cr,
 \end{apient}
 \apidesc{Speculatively parse the indicated region of the binary using the specified technique to find likely function entry points, enabled on the x86 and x86-64 platforms.}
 
+\fbox{\begin{minipage}[t]{1\columnwidth}%
+\begin{center}{\textbf{A note on using the lookup functions}}\end{center}\\
+When parsing binary objects such as .o files and static libraries which may have multiple
+\texttt{CodeRegion} objects that overlap in the address space, the \texttt{CodeRegion} argument
+\textit{must} be passed. For executable binaries and shared libraries that are fully linked, there
+is no ambiguity, and {\scshape null} can be passed. The only exception is \texttt{findFuncsByBlock}
+which always requires a valid \texttt{CodeRegion}.%
+\end{minipage}}
+
 \begin{apient}
 Function * findFuncByEntry(CodeRegion * cr,
                            Address entry)


mxz297:
@hainest LGTM.

wcohen:
The documentation addition makes things clearer.  Thanks.

PR #1148, opened 2021-11-10
dynProcess interface cleanup

hainest:
These member functions were removed long ago. `getDeadCode` was just an `assert(0)` and called only from the hybrid analysis code. That was a trap waiting for an unsuspecting user.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1148

No regressions.

PR #1152, opened 2021-11-11
Revise and add some new PowerPC instructions

hainest:
Also add comment for how to implement operands in InstructionDecoder-power.C


hainest:
Replaced by #1803

PR #1154, opened 2021-11-17
DwarfWalker: clean up interfaces for findDieName and findName

hainest:
findDieName always returned 'true', so any checks for its return value
were meaningless. It was also not clear that 'findName' would modify its
argument. The new interface removes that surprise by making the name the
return value.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1154

No regressions.

kupsch (approved):
Looks good.

PR #1159, opened 2021-11-20
SymtabAPI::Type: add support for C++ r-value references

### LINKED ISSUE #1049: SymtabAPI::Type should support C++11 r-value references
These are specified by `DW_TAG_rvalue_reference_type` and should be separate from [`dataReference`](https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Type.h#L80) and [`typeRef`](https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Type.h#L598).

hainest:
Fixes #1049

@kupsch @bigtrak @mxz297 Technically, this doesn't break ABI even though the size of the `typeRef` class changed because Symtab always refers to this type through a pointer. It does have the capacity to break user code if someone had a function taking a `typeRef` by value. I think making a note about this in the release notes is good enough. Thoughts?

### DIFF
--- a/symtabAPI/h/Type.h
+++ b/symtabAPI/h/Type.h
@@ -638,11 +638,16 @@ class SYMTAB_EXPORT typeTypedef: public derivedType {
 };
 
 class SYMTAB_EXPORT typeRef : public derivedType {
+ private:
+	bool is_rvalue_{false};
  protected:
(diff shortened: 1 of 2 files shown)


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1159

No new regressions.

PR #1160, opened 2021-11-22
Refactor dwarfWalker::findValue

hainest:
This function always returns 'true' and few callsites checked the bool
out variable. Using an optional here makes everything simpler. Logging
has been substantially improved for better tracing. The function was
renamed to better indicate its behavior.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1160

No new regressions.

kupsch (approved):

PR #1161, opened 2021-11-23
Increase minimum elfutils version to 0.186

hainest:
This is needed to parse NVIDIA's extended line map information.
HPCToolkit makes extensive use of this functionality.


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1161

No regressions.

PR #1164, opened 2021-11-24
Improve DWARF enum parsing in Symtab::Type

### LINKED ISSUE #1079: Allow access to underlying type in SymtabAPI::typeEnum
This should also add support for `enum class` types.

hainest:
This contains substantial updates to the correctness, capability, and interfaces for SymtabAPI::typeEnum. These changes necessarily break the ABI.

Closes #1079 

@mxz297 @kupsch @bigtrak I carefully broke these changes down into the smallest commits possible, made sure each would build and execute, and wrote as much supporting documentation in the commit messages as possible. Some of these changes fundamentally alter how users interact with this type. This is necessary in order to ensure class invariants for the type information in the derivedType interface. I think it would be a good long-term goal to clean up the interfaces for the rest of the types in Type.h, but we can do that as we make other changes to those classes.

### DIFF
--- a/symtabAPI/src/Type.C
+++ b/symtabAPI/src/Type.C
@@ -297,16 +297,10 @@ bool Type::isCompatible(Type * /*oType*/)
  * ENUM
  */
 typeEnum::typeEnum(int ID, std::string name)
-    : Type(name, ID, dataEnum)
-{
-	size_ = sizeof(int);
-}
+    : derivedType(name, ID, sizeof(int), dataEnum){}
 
 typeEnum::typeEnum(std::string name)
-   : Type(name, ::getUniqueTypeId(), dataEnum)
-{
-   size_ = sizeof(int);
-}
+   : derivedType(name, ::getUniqueTypeId(), sizeof(int), dataEnum) {}
 
 typeEnum *typeEnum::create(std::string &name, dyn_c_vector< std::pair<std::string, int> *> &constants, Symtab *obj)
 {
--- a/symtabAPI/src/Type.C
+++ b/symtabAPI/src/Type.C
@@ -302,6 +302,11 @@ typeEnum::typeEnum(int ID, std::string name)
 typeEnum::typeEnum(std::string name)
    : derivedType(name, ::getUniqueTypeId(), sizeof(int), dataEnum) {}
 
+typeEnum::typeEnum(boost::shared_ptr<Type> underlying_type, std::string name, typeId_t ID) :
+		derivedType(name, ID, underlying_type->getSize(), dataEnum) {
+	baseType_ = underlying_type;
+}
+
 typeEnum *typeEnum::create(std::string &name, dyn_c_vector< std::pair<std::string, int> *> &constants, Symtab *obj)
 {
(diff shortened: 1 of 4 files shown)


symtabAPI/src/Type.C
```diff
@@ -296,42 +296,12 @@ bool Type::isCompatible(Type * /*oType*/)
...
-   	typ->addConstant(constants[i]->first, constants[i]->second);
-    
-    if(obj)
-       obj->addType(typ);
-    //obj->addType(typ); TODO: declare a static container if obj is NULL and add to it.
-    //Symtab::noObjTypes->push_back(typ); ??
-    return typ;	
+typeEnum::typeEnum(boost::shared_ptr<Type> underlying_type, std::string name, typeId_t ID) :
+		derivedType(name, ID, underlying_type->getSize(), dataEnum) {
+	baseType_ = underlying_type;
```
kupsch, line 301: why isn't underlying_type passed to derivedType's constructor, and that constructor initializes baseType_?

kupsch (approved):
Other than the comment about the typeEnum setting a value in the derivedType class instead of passing the value to the constructor, this looks good.  Not changing the constructor would be OK.

symtabAPI/src/Type.C
hainest: I forgot to make a note about that. It's because there is no constructor for `derivedType` that can do that (https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Type.h#L365).

hainest:
@kupsch 

This slices:

    typeEnum tenum{underlying_type, name};
    BPatch_type *newType = new BPatch_type(boost::make_shared<Type>(std::move(tenum)));

I didn't think about it when I wrote it. For some reason, I always think about `make_shared` being covariant, but it's not. I forced-pushed the correct code which is substantially uglier.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1164

No regressions.

mxz297:
@hainest As we are aiming to significantly improve the parsing of DWARF, I feel it is very important to create additional DWARF parsing tests. Of course, this will delay the development to certain degree, but I think it is quite important for us to do this early. The tests can be just stand-alone unit tests for bookkeeping as adding tests to the test suite is just difficult....

hainest:
@mxz297 I absolutely agree, and I've already looked at this. The first hurdle is that the test mutatee is written in C. However, there are some mutatees that have checks for C++-specific macros (e.g., `__cplusplus`), so I *think* we can add C++ code in that way. For the mutator, we can do a runtime check for the [language](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/mapped_module.h#L75), but I haven't tested that out to see if it works. It really depends on if we can get the test suite to build the mutatee with a C++ compiler.

The main DWARF tests are in [test_type_info](https://github.com/dyninst/testsuite/blob/master/src/symtab/test_type_info.C). We should also add parameter parsing to include the same type inspection done in test_type_info. There are currently no such tests. Ideally, we would create a separate test for that and refactor the common type tests, but no one knows how to do that right now.

PR #1170, opened 2021-12-10
Removed incorrect/unnecessary/error-prone use of realloc

### LINKED ISSUE #1169: Strange/incorrect symtabAPI upgradePlaceholder code
https://github.com/dyninst/dyninst/blob/94252b8ce87293da4b10d0cee550e7ad0507a3ae/symtabAPI/src/Type-mem.h#L47

This function is causing me problems and this assert statement doesn't make sense. Assuming from the variable name `placeholder` and the usage of realloc, which explictly states:

> It must be previously allocated by malloc(), calloc() or realloc() and not yet freed with a call to free or realloc. Otherwise, the results are undefined. 

Furthermore, if you [run the example realloc code here](https://en.cppreference.com/w/c/memory/realloc), you will see that realloc _will_ change the pointer. So I am assuming that `placeholder` is already a pre-allocated memory area of `Type::max_size` so you really only need:

```cpp
template <class T>
T*
upgradePlaceholder(Type* placeholder, T* new_type)
{
    assert(sizeof(T) <= Type::max_size);
    memset(static_cast<void*>(placeholder), 0, Type::max_size);
    T* ret = new(placeholder) T{};
    assert(static_cast<void*>(placeholder) == static_cast<void*>(ret));
    *ret = *new_type;
    return ret;
}
```

In other words, since realloc requires an existing memory allocation and you are asserting that `sizeof(T) <= Type::max_size`, you don't need to realloc, you just need to do a placement new into the existing allocation.

jrmadsen:
- Closes #1169

### DIFF
--- a/symtabAPI/src/Type-mem.h
+++ b/symtabAPI/src/Type-mem.h
@@ -41,16 +41,12 @@ using namespace SymtabAPI;
 template<class T>
 T *upgradePlaceholder(Type *placeholder, T *new_type)
 {
-  assert(sizeof(T) <= Type::max_size);
-
-  void* mem = realloc((void *) placeholder, sizeof(T));
-  assert(mem == (void *) placeholder);
-  memset(mem, 0, sizeof(T));
-
-  T *ret = new(mem) T();
-  assert(mem == (void *) ret);
-  *ret = *new_type;
-  return ret;
+    assert(sizeof(T) <= Type::max_size);
+    memset(static_cast<void*>(placeholder), 0, Type::max_size);
+    T* ret = new(placeholder) T{};
+    assert(static_cast<void*>(placeholder) == static_cast<void*>(ret));
+    *ret = *new_type;
+    return ret;
 }
 
 template<class T>


hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1170

No regressions.

kupsch (changes requested):
This could be cleaned up a bit:

1. the memset is unnecessary, but does no harm other than efficiency
2. the old placeholder should be destructed `placeholder->~T()`, but OK if destructors do nothing
3. the `*ret = *new_type` could be done in the new expression `new (placeholder) T{*new_type}`
4. the second assert will not fail unless there is a weird custom allocator associated with the class and could be removed

I'd be OK with the code as is, if there are reasons not to do the above.

hainest:
@kupsch  If this code were being used for sane purposes, I would definitely make the changes you propose. This code is being used for insane purposes. Both `placeholder` and `new_type` are actually pointers to data controlled by a `boost::shared_ptr`. All of this bit-fiddling in `upgradePlaceholder` is being done outside the purview of the control block, so the pointer may not even be valid once it gets here. This whole thing is a ball of muddy undefined behavior.

kupsch:
Do any of the derived Type classes have destructors?  If not, then I'm OK with the change as is. Calling the destructor might still be useful as future proofing if someone ever add a virtual destructor to Type before this code is fixed properly.

If there is brokenness, I think my suggestions neither add nor remove brokenness except for potentially the destructor call being an improvement.

jrmadsen:
> the second assert will not fail unless there is a weird custom allocator associated with the class and could be removed

Agreed, but it was in the original so I left it.

> the *ret = *new_type could be done in the new expression new (placeholder) T{*new_type}

Yes, I am aware but it generates these warnings:

```console
dyninst/symtabAPI/h/Type.h:529:21: warning: implicitly-declared ‘Dyninst::SymtabAPI::fieldListType::fieldListType(const Dyninst::SymtabAPI::fieldListType&)’ is deprecated [-Wdeprecated-copy-dtor]
  529 | class SYMTAB_EXPORT typeStruct : public fieldListType {
      |                     ^~~~~~~~~~
In file included from dyninst/symtabAPI/src/dwarfWalker.h:17,
                 from dyninst/symtabAPI/src/dwarfWalker.C:32:
dyninst/symtabAPI/h/Type.h:321:4: note: because ‘Dyninst::SymtabAPI::fieldListType’ has user-provided ‘virtual Dyninst::SymtabAPI::fieldListType::~fieldListType()’
  321 |    ~fieldListType();
```

and I didn't really want to go down the rabbit-hole of fixing all these type classes.

> the old placeholder should be destructed placeholder->~Type(), but OK if destructors do nothing

I didn't delve into the usage or y'all's object model until now but if all `T` types are derived from `Type` and every call of `addOrUpdateType` involves making a new shared pointer... why does `upgradePlaceholder` even exist at all? This sort of code looks like what is needed with static polymorphism, not dynamic polymorphism. 

Unless I am missing some use case that doesn't look like this:

```cpp
   auto baseType = Type::make_shared<typeScalar>( type_id(), size, curName(), p);
   auto baseTy = tc()->addOrUpdateType( baseType );
```

Then you just need:

```cpp
	if (a->second->getDataClass() == dataUnknownType)
	{
                a->second = type;
		// upgradePlaceholder(a->second.get(), type.get());
	}
```

hainest:
>why does upgradePlaceholder even exist at all?

It shouldn't. Someone was trying to be clever to save space and uphold a severely broken interface. That entire function is 100% undefined behavior and has a possible corner case of a race condition. It needs to go away. However, I need to go through the rest of the DWARF parsing code to make sure we can swap the shared_ptr in calls to `addOrUpdateType` (there is a *lot* of calling `get()` on a shared pointer for some unknown reason).

This code fixes your issue. It doesn't seem to break anything any worse than before on our end, so let's merge. I'll make an issue to follow up with removing `upgradePlaceholder`.

Issue #1173, opened 2021-12-18
Need support for parsing vtable information


mplegendre:
For smeagle, we should start capturing vtables information about classes. A vtable reordering could cause significant ABI issues.

Ideally, we would capture this via SymtabAPI.  DWARF has information about vtables.  Or one might be able to locate the vtable for a class in symbol info, and read the vtable out of RO-data.

I have a prototype PR for this that uses DWARF, but there may be issues with DWARF.  I'll post that PR and discuss the issues there.

PR #1174, opened 2021-12-18
Initial support for parsing vtables out of DWARF information

mplegendre:
I took a stab at adding vtable support to SymtabAPI via the DW_AT_vtable_elem_location attribute in DWARF for #1173 .  

This is only a prototype.  It parses the attribute and adds vtable information to SymtabAPI internal data structures.  I haven't finalized a public interface, nor added tests or documentation.

BUT, I don't think the DW_AT_vtable_elem_location fields are accurate.  With GCC compilers, I'm missing information about destructors.  With clang compilers, the info is all over the place, with different functions occupying the same slots in the same class's vtable.  Based on inspection with dwarfdump, I think this code is reading DW_AT_vtable_elem_location information correctly and it's the compilers that are wrong.
Or, I've got a bug/misunderstanding.  

So I'm not immediately sure how to proceed.  @woodard , any insigned?

### DIFF
--- a/symtabAPI/h/Type.h
+++ b/symtabAPI/h/Type.h
@@ -63,6 +63,7 @@ class fieldListType;
 class rangedType;
 class derivedType;
 class TypeMemManager;
+class DwarfWalker;
 
 //TODO?? class BPatch(to be  ??)function;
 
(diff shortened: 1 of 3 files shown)


vsoch:
@woodard @mplegendre do you have a simple script (or library example in spack) that will generate dwarf with this attribute? I'm testing in Go so I can look at how it works and the standard libs that I use (tcl and libtest) don't have it.

vsoch:
I'm using these as examples - will try different compilers next in case that's the issue. https://www.geeksforgeeks.org/virtual-function-cpp/

vsoch:
Also one thing I think could be an extra check - DW_AT_virtuality ([here in Go](https://cs.opensource.google/go/go/+/master:src/debug/dwarf/const.go;l=72;bpv=1;bpt=1?q=Attr&ss=go%2Fgo)) and [here for docs](https://developer.ibm.com/articles/au-dwarf-debug-format/) should be set if it's virtual.

PR #1175, opened 2021-12-23
glibc r_debug extensions to support multiple namespaces

### LINKED ISSUE #1155: glibc rendezvous structure uninitialized in glibc 2.34+
@stanfordcox Reports

>seeing this with the "rawhide", latest but not released yet kernel and glibc: mutated:
/home/scox/dyninst-11.0.1/src/dyninstAPI_RT/src/RTlinux.c:412: r_debugCheck: Assertion `_r_debug.r_map' failed.
make: *** [Makefile:22: test] Aborted (core dumped)

>So the symptom is a static mutated executable fails as above.   I see what is going on: something changed in the recent runtime environment and now _r_debug, the dynamic linker rendevous struct, cannot be accessed.  I'll ping the glibc folks and figure out a resolution.

>possibly related to glibc commit a93d9e03a  Extend struct r_debug to support multiple namespaces [BZ #15971]

---

Bugzilla [15971](https://sourceware.org/bugzilla/show_bug.cgi?id=15971) says the extended namespace changes won't go in until 2.35 (scheduled for ~Feb 2022). Yet, this is showing up in 2.34 (2.33 is confirmed unaffected). The full commit is (https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=a93d9e03a31ec14405cb3a09aa95413b67067380;hp=885762aa31d75de8b9fea4c0e2e372b582d4c548).

stanfordcox:
When the symbol table is created in
emitElf<ElfTypes>::createSymbolTables, also update the elf _DYNAMIC
symbol to reflect the new address of the .dynamic section.

I did this with regard to issue 1155 (which has a simpler solution).  But it seems reasonable to keep _DYNAMIC up to date.  PT_DYNAMIC in phdr is kept up to date by fixPhdrs.

Fixes #1155

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -342,7 +342,7 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
             else {
                 if (vers) {
                     // There should only be one version string by this time
-                    //If the verison name already exists then add the same version number to the version symbol table
+                    //If the version name already exists then add the same version number to the version symbol table
                     //Else give a new number and add it to the mapping.
                     if (versionNames.find((*vers)[0]) == versionNames.end()) {
                         mpos += sprintf(mpos, "  new version name: %s\n", (*vers)[0].c_str());
@@ -694,7 +694,7 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
         }
         // Change offsets of sections based on the newly added sections
         if (movePHdrsFirst) {
-            /* This special case is specific to FreeBSD but there is no hurt in
+            /* This special case is specific to FreeBSD but there is no harm in
              * handling it on other platforms.
              *
              * This is necessary because the INTERP header must be located within in
@@ -1105,7 +1105,7 @@ void emitElf<ElfTypes>::fixPhdrs(unsigned &extraAlignSize) {
         return;
 
     //We made a new section to contain the program headers--keeps
-    // libelf from overwriting the program headers data when outputing
+    // libelf from overwriting the program headers data when outputting
     // sections.  Fill in the new section's data with what we just wrote.
     Elf_Data *data = elf_newdata(phdrs_scn);
     size_t total_size = (size_t) newEhdr->e_phnum * (size_t) newEhdr->e_phentsize;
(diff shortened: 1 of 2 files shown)


symtabAPI/src/emitElf.C
```diff
@@ -1990,6 +1990,24 @@ bool emitElf<ElfTypes>::createSymbolTables(set<Symbol *> &allSymbols) {
     if (!obj->getAllNewRegions(newSecs))
         log_elferror(err_func_, "No new sections to add");
 
+    for (unsigned long nsi = 0; nsi < newSecs.size(); nsi++) {
+	int sec_addr, prev_size;
+	if (newSecs[nsi]->getDiskOffset())
+	  sec_addr = newSecs[nsi]->getDiskOffset() + library_adjust;
+	else
+	  sec_addr += prev_size;
```
hainest, line 1998: `prev_size` has no defined value here. It's also being updated on line 1999 in every iteration, so there is no lagging value.

symtabAPI/src/emitElf.C
```diff
@@ -1990,6 +1990,24 @@ bool emitElf<ElfTypes>::createSymbolTables(set<Symbol *> &allSymbols) {
...
+	prev_size = newSecs[nsi]->getDiskSize();
+	if (".dynamic" == newSecs[nsi]->getRegionName()) {
```
hainest, line 2000: Hard-coded strings make me nervous. Are '.dynamic' and '_DYNAMIC' guaranteed by the ELF standard? Has this been tested on binaries created by clang?

hainest (changes requested):
If at all possible, it would be great to split the whitespace/spelling changes into a separate commit from the code changes. It's ok if you don't have time, though.

symtabAPI/src/emitElf.C
stanfordcox: #1176 actually fixes the static executable problem so how about we punt on this since it is not essential to fix the problem at hand.

stanfordcox: @hainest  I found the reference in the elf spec:
[elf spec]( https://refspecs.linuxfoundation.org/elf/elf.pdf)
If an object file participates in dynamic linking, its program header table will have an element of type
PT_DYNAMIC. This ‘‘segment’’ contains the .dynamic section. A special symbol, _DYNAMIC, labels the
section, which contains an array of the following structures.

hainest: All good, then. I just wanted to double-check.

symtabAPI/src/emitElf.C
hainest: Have you had a chance to look at this?

stanfordcox:
Added a commit:
Quiesce uninitialized variable warnings.

hainest (approved):

PR #1176, opened 2021-12-23
Remove interposed definition of _r_debug

stanfordcox:
Previously there was a dynamic relocation against _r_debug in the loader which picked up the interposed definition,
but glibc now uses a direct internal hidden symbol reference and thus no longer updates the interposed object.

### DIFF
--- a/dyninstAPI_RT/src/RTlinux.c
+++ b/dyninstAPI_RT/src/RTlinux.c
@@ -404,7 +404,6 @@ void dyninstTrapHandler(int sig, siginfo_t *sg, ucontext_t *context)
 #if defined(cap_binary_rewriter)
 
 extern struct r_debug _r_debug;
-DLLEXPORT struct r_debug _r_debug __attribute__ ((weak));
 
 /* Verify that the r_debug variable is visible */
 void r_debugCheck() { assert(_r_debug.r_map); }


hainest:
@stanfordcox When did this change in glibc occur? We support at least back to glibc-2.17, so I want to make sure we don't accidentally break that.

stanfordcox:
rawhide has glibc-2.34.9000-34.fc36 which is almost but not quite glibc-2.35.   Fedora has glibc-2.34-7.fc35 so the glibc-2.34-NN are the last versions that will work with the _r_debug weak definition.

hainest:
@bigtrak @kupsch Leaving these here for later discussion:

https://sourceware.org/pipermail/libc-alpha/2021-September/130820.html
https://stackoverflow.com/questions/55643058/finding-the-rendezvous-struct-r-debug-structure-of-a-process

hainest:
Testing looks good on glibc 2.17 and 2.31: https://bottle.cs.wisc.edu/search?dyninst_branch=PR1176 

@stanfordcox @kupsch @bigtrak Any objections to merging?

stanfordcox:
@hainest Same here, I put dyninst 12.0.1 plus this patch into Fedora 36 (rawhide) and all seems fine.

hainest:
@stanfordcox Great! Merging.

PR #1180, opened 2022-01-11
Adding first stage of automated docker build

vsoch:
This is the first of two PRs to add automation to Dyninst, including:
 - a container base that can be used for Dyninst development or app development using dyninst
 - the same container base that provides a much quicker way to test (3.5 minutes vs 45-1hour to build from scratch)

This first PR will add the automated build for the base container, which we can leave to also build on PR, but would be okay to remove and just have build on push to main and scheduled nightly (what I recommend). The second PR after this will be to add another workflow to use this base container to run tests.

The base container uses ubuntu 20.04 and a set of dependencies chosen by Tim.  Currently this does not upload test results to Tims server, and this is only because I want to talk to him about it first and then test the workflow before PR here.

Finally, in order for this to work an org owner needs to enable packages (e.g., check the boxes for public and internal) and then the PR, when merged, should be able to deploy. For those not familar, GitHub packages == an OCI registry, and that means it's like Docker Hub in being able to push containers (and other artifacts).

While my preference is to store the Dockerfile in the root of the repository, since we will ultimately have two Dockerfile plus the README and assets that go into the containers (e.g., build.sh) it is much better organized to have a subfolder and then to tweak the build command. E.g.,:

```bash
cd docker/

# build a container with tag "dyninst" with context one level up with the Dockerfile in the present working directory
$ docker build -t dyninst -f Dockerfile ../
```
cc @hainest, master of the C++ universe and diabolical binary analysis leader!

### DIFF
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -43,20 +43,27 @@ jobs:
           sudo rm -rf /usr/share/dotnet
           sudo rm -rf /opt/ghc
 
-      - name: Set Organization Name for Packages Registry
+      - name: Set Organization and Branch Name
         if: (github.event_name != 'schedule')
         env:
           org: ${{ github.event.repository.owner.login }}
-        run: echo "org=${org}" >> ${GITHUB_ENV}
+        run: |
+            branch=$(echo ${GITHUB_REF##*/})
+            echo "branch=${branch}" >> ${GITHUB_ENV}
+            echo "org=${org}" >> ${GITHUB_ENV}
 
-      - name: Set Organization Name for Packages Registry
+      - name: Set Organization and Branch Name
         if: (github.event_name == 'schedule')
-        run: echo "org=dyninst" >> ${GITHUB_ENV}
+        run: |
+            echo "org=dyninst" >> ${GITHUB_ENV}
+            echo "branch=master" >> ${GITHUB_ENV}
 
       - name: Pull Layers for "Cache"
         run: docker pull ghcr.io/${org}/dyninst-ubuntu-${{ matrix.ubuntu }}:latest || echo "No cache available"
 
       - name: Build Dyninst Base Container
+        env:
+          branch: ${{ env.branch }}
         run: |
            cd docker/
            docker build --build-arg ubuntu_version=${{ matrix.ubuntu }} \
@@ -64,6 +71,7 @@ jobs:
                         --build-arg ELFUTILS_VERSION=${{ matrix.elfutils }} \
                         --build-arg LIBIBERTY_VERSION=${{ matrix.libiberty }} \
                         --build-arg INTELTBB_VERSION=${{ matrix.inteltbb }} \
+                        --build-arg DYNINST_BRANCH=${branch} \
                         -f Dockerfile \
                         -t ghcr.io/${org}/dyninst-ubuntu-${{ matrix.ubuntu }}:latest ../
 
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -1,0 +1,79 @@
+name: Build Deploy Containers
+
+on:
+
+  # Always have a base image ready to go - this is a nightly build
+  schedule:
+    - cron: 0 3 * * *
+
+  # On pull request we test updates to images
+  pull_request: []
+ 
+  # On push to main we build and deploy images
+  push: 
+    branches:
+      - main
+ 
+jobs:
+  build:
+    permissions:
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+
+        # Note: this can be extended in two ways
+        # 1. Add more configurations to the matrix, either like this or flattening
+        # 2. Generate matrix programatically and pipe in (recommended)
+        # perl is 5.30.0 provided by 20.04 container base
+        ubuntu: ["20.04"]
+        boost: ["1.73.0"]
+        elfutils: ["0.186"]
+        libiberty: ["2.33.1"] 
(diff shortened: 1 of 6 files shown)


vsoch:
And note that I will need approval to run this job. :point_up:

docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+# Build the Dockerfile in this directory, context one level up
+# docker build -t dyninst -f Dockerfile ../
+
+LABEL maintainer="@hainest,@vsoch"
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=America/Los_Angeles
+
+# We can use build args to populate the specific versions of dependencies (with defaults here)
+ARG BOOST_VERSION=1.73.0
```
hainest, line 13: Is there a way to read these from base-containers.yaml so that we don't have two places to change the dependency version?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Refactor dwarfWalker::findConst ([1160](https://github.com/dyninst/dyninst/issues/1160))
 - Add readable name for Symtab::typeRef ([1157](https://github.com/dyninst/dyninst/issues/1157))
 - DwarfWalker: clean up interfaces for findDieName and findName ([1154](https://github.com/dyninst/dyninst/issues/1154))
+- Added automated docker build for development and testing
```
hainest, line 17: As an aside, we usually only update the CHANGELOG on release, but this is ok.

.github/workflows/base-containers.yaml
```diff
@@ -0,0 +1,79 @@
+name: Build Deploy Containers
+
+on:
+
+  # Always have a base image ready to go - this is a nightly build
+  schedule:
+    - cron: 0 3 * * *
```
hainest, line 7: Does Github have a preferred time to run nightly builds, or is 3AM just a nice time to do it?

docker/Dockerfile
vsoch: This already happens - it's a build argument so it comes from the matrix (what you see there is the default if the build arg is not provided).

CHANGELOG.md
vsoch: Noted -   I update mine as I go because it's easier to remember!

.github/workflows/base-containers.yaml
vsoch: No you can choose whatever time you like!

vsoch:
@hainest going to need another workflow approval! I forgot to cd into docker for the build.

.github/workflows/base-containers.yaml
vsoch: And I actually looked at this one... https://researchapps.github.io/cron/... just because !

docker/Dockerfile
hainest: Ah, ok. I just noticed that when I was looking at the output from the previous run. Good to know!

vsoch:
Update: we still need to tell spack the Dyninst branch being used (I haven't tested this locally - I derive from the action but it may just work to use master)

hainest:
Testing PRs against the HEAD of master is what I do now, so that works. I don't see a button to re-run.

.github/workflows/base-containers.yaml
hainest: Of course you did! :D  That is some interesting data. The correlation with UTC times makes sense, though.

vsoch:
@hainest this built for me locally! So one more workflow approval should do it, and apologies I didn't test it perfectly before submit - I was hoping it would be a relatively simple transfer!

hainest:
No worries at all. Thanks so much for doing the hard work on this!

vsoch:
That's very weird - this worked locally, I must not be on a feature branch. Hmm let me try something else. Can you add me as a contributor to dyninst so it stops asking for your approval to run tests?

vsoch:
@hainest build is successful! If you want to first double check the org has packages enabled (internal and public) then we can merge here, and then we need the container to build and I'll open a follow up PR with the testing.

hainest:
I don't see any packages listed: https://github.com/dyninst/dyninst/packages

https://ghcr.io/dyninst also shows nothing.

vsoch:
@hainest we have to merge the PR before the package is deployed, otherwise any online shmuck could do a PR and deploy a package :)

vsoch:
So - for next steps, you'll want to check the Org settings that packages are enabled (internal and public) for Dyninst, and then when we merge the build will happen that does the deploy. If packages are set up it should push and the package will show up, otherwise we'll get an error and I'll help debug.

vsoch:
E.g., if you look at the workflow, the GHCR Login and Deploy are sort of "greyed out" meaning they don't run:

image

And that happens because of this if in the workflow file:

```yaml
if: (github.event_name != 'pull_request')
```
e.g., "Don't login and deploy on a pull request!" That's also why people that aren't known contributors need the workflow approved every time - a malicious entity could remove that line in a PR and then deploy during the PR.

hainest:
https://docs.github.com/en/packages/quickstart says we need more yamls to make a release package. Do we need that, and if so, do we do that before or after merging here?

vsoch:
To make the package deploy on release we just need to add the [release trigger](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) and then tag the image with some release version and push that one too. We don't need more YAMLs, we will just add to this one. 

I would recommend not adding release yet until we've gotten testing in, which is higher priority. I'll be glad to update the YAML here with release when that is done.

hainest:
Sounds good. Merging!

PR #1181, opened 2022-01-12
Adding dyninst release trigger

vsoch:
To release a tagged (versioned) container on publication of a release we just need to add the release -> published trigger, derive the tag from the GITHUB_REF environment variable, and then tag and push the container.

cc @hainest 

### DIFF
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -6,9 +6,10 @@ on:
   schedule:
     - cron: 0 3 * * *
 
-  # On pull request we test updates to images
-  pull_request: []
- 
+  # Publish packages on release
+  release:
+    types: [published] 
+    
   # On push to main we build and deploy images
   push: 
     branches:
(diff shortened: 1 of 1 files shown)


hainest:
Can we convert this to a pull_request trigger? Know that I understand more of this, I don't think we really need a release trigger. For this container, nightly builds are more than sufficient.

vsoch:
So - the release trigger is to explicitly create a dyninst container tagged with a release, which will only happen on release and is important if developers want to pin a release to use for their project. The nightly build is a different use case - a "the current version of master I don't care about the version" use case.

And I wouldn't recommend adding a PR trigger - that's going to mean every PR is running this hour plus build and it's going to slow things down. If a merge into main breaks something you'll likely know from the tests (and there is no need to rebuild the base container).

hainest:
Oh, derp! I just realized this is for the base container. Agreed! We don't want to build this on every PR. Never mind me! I'll go ahead and merge as-is.

PR #1183, opened 2022-01-12
start of work to add automated testing to dyninst!

vsoch:
The package build was successful, so here is the promised next step - a workflow that will use the automated build to run tests on the new code.

I likely am going to sleep after this and will pick up in the morning

### DIFF
--- a/.github/workflows/test-dyninst.yaml
+++ b/.github/workflows/test-dyninst.yaml
@@ -1,0 +1,39 @@
+name: Test Dyninst
+
+on:
+  pull_request: [] 
+  push: 
+    branches:
+      - master
+ 
+jobs:
+  build:
+    permissions:
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+
+        # Testing containers named by ubuntu version for now
+        ubuntu: ["20.04"]
+
+    runs-on: ubuntu-latest
+    name: Build
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Make Space For Build
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+
+      - name: Set Organization Name for Packages Registry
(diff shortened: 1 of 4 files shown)


vsoch:
That worked pretty well! So in review - when we build the container again (and don't actually add new code) the entire thing takes 3.5 minutes (e.g., the testing repository). When we actually add the code from here and need to rebuild dyninst, the entire thing takes 20 minutes. It's not "immediate and quick" by any means, but having the automation will be better than nothing!


image

cc @hainest and @mplegendre (just if you are interested since I mentioned this yesterday). 

@hainest if you want to add the token for your results server as a secret I can add a step to do that to the workflow. I assume we'd want to upload when tests pass on a merge of new results? Or were you thinking of doing it for every PR?

hainest:
Twenty minutes is more than an order of magnitude speedup, so I think it's amazing! I'm still not sure why it only runs a subset of the tests. I'll see if I can poke around and figure that out.

>if you want to add the token for your results server as a secret I can add a step to do that to the workflow.

Yes, we'll want to do that, but first the dashboard needs to be modified to provide useful results. Currently, it's just a one-sided upload. To make it useful here, it should accept an upload and then return the processed results.

>I assume we'd want to upload when tests pass on a merge of new results? Or were you thinking of doing it for every PR?

I do testing on every PR, so I think that's the way to continue here.

vsoch:
>  I'll see if I can poke around and figure that out.

Okay sounds good! You should be able to shell into the container used by the tester and just reproduce what the testing container does here to poke around - let  me know if you want any help (ping me on Slack).

hainest:
I think I'll go ahead and merge this as-is so we can have some build/test running on PR. I'll take a look at the container to try and figure out why the tests aren't running fully. Objections?

vsoch:
No objections! Please ping me if you need/want any help, or when you are ready to discuss next steps for the upload to the test interface.

PR #1184, opened 2022-01-12
Source file parsing fix for DWARF 5

mxz297:
rocm-4.5 starts to emit AMD GPU binaries with DWARF 5. I noticed missing source line information for these binaries. The root cause is that dyninst assumes that source file table entry 0 represents "unknown file". This no longer seems to be true in DWARF 5, where entry 0 can represents a real application source file.

Fortunately, the fix seems simple: we simply just parse the entry and add it the string table.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -3388,12 +3388,9 @@ void Object::parseLineInfoForCU(Dwarf_Die cuDIE, LineInformation* li_for_module)
         return s_name;
     };
 
-    // dwarf_line_srcfileno == 0 means unknown; 1...n means files[0...n-1]
-    // so we ensure that we're adding a block of unknown, 1...n to the string table
-    // and that offset + dwarf_line_srcfileno points to the correct string
     using namespace boost::filesystem;
     strings->emplace_back("<Unknown file>","");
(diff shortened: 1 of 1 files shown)


kupsch (approved):

hainest:
Do you have a test case we can validate this change against a binary built with DWARF4? The test suite doesn't currently have the DWARF version as a testing dimension.

mxz297:
@hainest Locally I have an AMD GPU binary 

> Do you have a test case we can validate this change against a binary built with DWARF4? The test suite doesn't currently have the DWARF version as a testing dimension.

Do you mean DWARF4 or 5? Locally I have an AMD GPU binary as an reproducer, which is DWARF5. I also ran the PR against several binaries with DWARF 3 and 4 and results look reasonable. I did not have any rigorous testing for this.

hainest:
>I also ran the PR against several binaries with DWARF 3 and 4 and results look reasonable.

That works for me. I just wanted to double-check that reading the extra source file entry wouldn't accidentally break some assumption about DWARF4.

hainest:
https://bottle.cs.wisc.edu/search?dyninst_branch=PR1184

LGTM.

hainest (approved):

PR #1185, opened 2022-01-16
pin to spack release so we have consistent builds

vsoch:
This is a test build for the container base that uses a release of spack (and not develop). What looks to have happened is that some change to speak dealing with patches broke the install of boost (and thus our entire build) and we can't ave that happen.

I'll first test the last release, but we may need to test a little before that.

image

### DIFF
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -9,7 +9,9 @@ on:
   # Publish packages on release
   release:
     types: [published] 
-    
+
+  pull_request: []
+      
   # On push to main we build and deploy images
   push: 
     branches:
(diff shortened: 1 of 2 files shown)


vsoch:
Okay there is something whackado with spack and its infrastructure:
```
==> Error: Failed to install zlib due to InvalidIMDSEndpointError: Invalid endpoint EC2 Instance Metadata endpoint: http://169.254.169.254/latest/api/token
```
The good news is that with the approach I took we don't need the base container to rebuild every night. I posted in the spack slack and will come back here and try again when I know more. It could be that the original error will just resolve itself.

vsoch:
@hainest I'm testing an even earlier release, but this still isn't logical because this was working 2 days ago (and all days before that). 

I'm sorry this isn't reliable - that's kind of the nature of spack, and this kind of thing has happened to me before. We thankfully have the container deployed when it was successful, and long term we probably need to build a container base without spack. Happy to help with that!

vsoch:
Testing removing the cache.

vsoch:
Testing now with spack debug to see issue - but I think it's probably this (added yesterday) https://github.com/spack/spack/commit/415d662ec06ad17e69ff95c72367af7065fdfc78

vsoch:
This has been fixed in upstream spack. phew.

PR #1198, opened 2022-02-23
Fix compiler warnings in amdgpu cdna2 code

hainest:
@mxz297 @bbiiggppiigg

We should try to find a better way of doing code reviews. I missed all of this when looking over the PR. It would be great if we could do them iteratively with smaller amounts of code changes. Understandably, this can be hard to do for active development of new features.

### DIFF
--- a/common/src/dyn_regs.C
+++ b/common/src/dyn_regs.C
@@ -218,15 +218,7 @@ std::string MachRegister::name() const {
             }
             return ret;
         }else if(getArchitecture() == Arch_amdgpu_cdna2){
-            signed int category = reg & 0x00ff0000;
-            signed int base_val = reg & 0x000000ff;
-            switch(category){
-                default:
-                    ret = iter->second;
-                    break;
-            }
-            return ret;
-
+        	return iter->second;
         }else{
             return iter->second;
         }
--- a/instructionAPI/src/AMDGPU/cdna2/InstructionDecoder-amdgpu-cdna2.C
+++ b/instructionAPI/src/AMDGPU/cdna2/InstructionDecoder-amdgpu-cdna2.C
@@ -91,7 +91,7 @@ namespace Dyninst {
 		// decoding opcodes
 		// ****************
 
-		MachRegister InstructionDecoder_amdgpu_cdna2::makeAmdgpuRegID(MachRegister base, unsigned int encoding , unsigned int len) {
+		MachRegister InstructionDecoder_amdgpu_cdna2::makeAmdgpuRegID(MachRegister base, unsigned int encoding , unsigned int) {
 			MachRegister realBase = base;
 			/*if (base == amdgpu_cdna2::s0){
 				switch(len){
--- a/instructionAPI/src/AMDGPU/cdna2/decodeOperands.C
+++ b/instructionAPI/src/AMDGPU/cdna2/decodeOperands.C
@@ -318,7 +318,7 @@ case 2 : return makeRegisterExpression(amdgpu_cdna2::p0,0,opr_size);
 default: return makeRegisterExpression(amdgpu_cdna2::invalid);
 }
 }
-Expression::Ptr InstructionDecoder_amdgpu_cdna2::decodeOPR_PC(uint64_t input,uint32_t opr_size ){
+Expression::Ptr InstructionDecoder_amdgpu_cdna2::decodeOPR_PC(uint64_t input,uint32_t){
 switch(input){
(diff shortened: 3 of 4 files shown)


mxz297:
@hainest is this PR intended to fix the compilation error cased by the internal header file include (https://github.com/dyninst/dyninst/blob/master/common/h/dyn_regs.h#L1817)?

hainest:
@mxz297 Yes. I got not warnings/errors when building this PR.

mxz297:
@hainest I am not certain this PR will fix the error. Note that the compilation error I reported did not happen during compilation of Dyninst. It happened when I use Dyninst. 

In the context of using Dyninst, all public header files are copied into the install directory without the internal header file. The internal header file, however, is available when we compile Dyninst.

Let me double check this PR before you merge it.

mxz297:
@hainest I just confirmed that this PR does not fix the problem I reported. Here is a simple reproducer:

```
[xm13@ufront dyninstapi]$ cat test.cpp 
#include "dyn_regs.h"

int main() { return 0;}
[xm13@ufront dyninstapi]$ gcc -I/home/xm13/dyninstapi/install/include -I/home/xm13/spack/opt/spack/linux-rhel8-zen/gcc-8.5.0/boost-1.78.0-fsrlkuqa56xw6b6zdiauu6gc34s3yzgf/include test.cpp
In file included from test.cpp:1:
/home/xm13/dyninstapi/install/include/dyn_regs.h:1817:10: fatal error: AMDGPU/cdna2/amdgpu_cdna2_sys_regs.h: No such file or directory
 #include "AMDGPU/cdna2/amdgpu_cdna2_sys_regs.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

hainest:
>Note that the compilation error I reported did not happen during compilation of Dyninst. It happened when I use Dyninst.

Ah, I did not catch that part. It's weird that the test suite didn't have any build errors in the docker image. I am seeing them now when building on our internal machines. I need to figure that out so we can catch these issues earlier without me having to run the test suite manually.

bbiiggppiigg:
All changes looked reasonable to me.

instructionAPI/src/AMDGPU/cdna2/InstructionDecoder-amdgpu-cdna2.h
```diff
@@ -31,7 +31,6 @@
 #include "InstructionDecoderImpl.h"
 #include <iostream>
 #include "Immediate.h"
-#include "dyn_regs.h"
```
bbiiggppiigg, line 34: Now I see that this header is also included in InstructionDecoderImpl.h, so it is safe to remove. However, as the same file is included in InstructionDecoder-aarch64.h [](https://github.com/dyninst/dyninst/blob/a85835269188c24a5c27a50bbd06751c1badc3d7/instructionAPI/src/InstructionDecoder-aarch64.h#L34), won't this cause the same build error? Or this problem only exist for amdgpu cdna2?

bbiiggppiigg (approved):
LGTM

instructionAPI/src/AMDGPU/cdna2/InstructionDecoder-amdgpu-cdna2.h
hainest: It won't cause a build error in either place because the macro guards will stop that. I removed it because it wasn't being used. We should remove it from the other files, as well. One of my future tasks is to run IWYU on the whole codebase.

bbiiggppiigg: I think it was included because some of the functions take MachRegister  as parameter, which is defined in dyn_regs.h

bbiiggppiigg: The build error I was referring to, is the comment associated with the first commit in this PR

hainest: Ah, you are right. That makes it worse, not better. Like I said, we can leave that mess until later.

>The build error I was referring to, is the comment associated with the first commit in this PR

Yes. The error there was the absolute path in https://github.com/dyninst/dyninst/blob/a85835269188c24a5c27a50bbd06751c1badc3d7/common/h/dyn_regs.h#L1817 which didn't exist until until #1201.

PR #1200, opened 2022-02-24
Clean up duplicate cdna2 sys_reg headers

bbiiggppiigg:


bbiiggppiigg:
@hainest 

This pull request actually introduce a problem similar to that reported by Xiaozhu in #1198 
Where amdgpu_cdna2_op_table.h is not found for an application using dyninst, but dyninst itself compiles normally.

Do you know what is the right way to make cmake install these headers in the path that they belong?

PR #1201, opened 2022-02-24
Add install rule to install headers under common/h/AMDGPU

bbiiggppiigg:
With this commit, the common/h/AMDGPU headers will be installed to the correct path and should fix the issue that was reported in #1198

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -136,3 +136,10 @@ target_link_libraries(common PUBLIC ${TBB_LIBRARIES})
 if(USE_OpenMP)
   set_target_properties(common PROPERTIES COMPILE_FLAGS ${OpenMP_CXX_FLAGS} LINK_FLAGS ${OpenMP_CXX_FLAGS})
 endif()
+
+# Install AMDGPU headers under subdirectory of common/h/AMDGPU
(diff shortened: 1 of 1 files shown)


hainest:
Yes, I was working on a solution to this. That's why I merged #1200 into a new branch instead of master. I was hoping to be able to take care of this generically, but the `dyninst_library` function doesn't work that way. For now, this will have to do.

Could you move this into common/CMakeLists.txt?

bbiiggppiigg:
@hainest 
I've moved the install rule from top level CMakeLIsts to that under common

hainest:
Could you remove the whitespace changes in CMakeLists.txt?

bbiiggppiigg:
you mean the one in the first commit?

hainest:
The current commit changes both CMakeLists.txt and common/CMakeLists.txt. The changes in CMakeLists.txt are just whitespace, so let's remove those.

hainest:
Oh, nevermind. Github was being slow in refreshing your changes. Nevermind!

hainest (approved):

bbiiggppiigg:
I just reset the last 2 commits and made a new one out of it.

hainest:
That worked. Thanks!

PR #1204, opened 2022-02-25
Tidy up classes in AMDGPU

hainest:
These are just a few items picked up from static analysis. Some of them should have been compiler warnings (e.g., uninitialized class members).


bbiiggppiigg (approved):
All seems reasonable to me.

PR #1207, opened 2022-03-03
Make a couple constant 64-bit ULL to avoid getting truncated during shifts

wcohen:
Noticed constants in BinaryFunction.h and InstructionDecoder-aarch64.C were not properly sized as 64-bit value.  Shift of these constants would be done as 32-bit values and could be truncated if the shift is large. This was detected by a coverity scan.

### DIFF
--- a/instructionAPI/h/BinaryFunction.h
+++ b/instructionAPI/h/BinaryFunction.h
@@ -263,7 +263,7 @@ namespace Dyninst
 						}
 						
 						Result one(u64, 1);
-						Result rot = arg1 & (Result(u64, (1<<arg2val) - 1));
+						Result rot = arg1 & (Result(u64, (1ULL<<arg2val) - 1));
 						
 						rightLogicalShiftResult lhsRightShift;
 						Result arg2mod = Result(arg2.type, arg2val);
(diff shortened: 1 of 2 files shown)


hainest:
Thanks, Will. Static analysis is definitely something we want to incorporate into our CI workflow.

hainest (changes requested):
The `LL` suffix isn't guaranteed to be 64 bits. I would recommend changing these to explicit variables.

`constexpr auto mask = std::uint64_t{0xFFu};`

kupsch:
The ULL suffix is an `unsigned long long`, and the C++ standard does not directly specify a minimum size for this. It does state the existence of the `<climits>` header and refers to the C standard for definitions of the constants. The C standard says that it must be at least 64-bits; https://en.cppreference.com/w/cpp/language/types and Stroustrup's book also say at least 64-bits.

hainest:
@kupsch That's interesting. 3.9.1.2 [basic.fundamental] in C++11 (n3242) does not have that requirement, but 3.9.1.3 in C++14 (n3797) does. I'm assuming it got backported. I am fiercely prejudiced against size literals, so I'll let Will decide if he wants to update to the more explicit form I suggested.

wcohen:
I lean towards the ULL where the constants current exist rather than define additional variables with constants to be more concise.  However, I am checking with other Red Hat people that are involved with the standards and compilers to see whether there are reasons that the variable declaration would be preferred.

wcohen:
It looks like the ULL should be fine to address the issues with the possible truncations of the shifted values and would go with that rather than making new variables based on talking with one of the engineers at Red Hat:

ULL suffix means it is unsigned long long constant.
C++ requires that unsigned long long type is at least 64-bits, it could be
wider than that but I'm not aware of any compiler/architecture where that would
true right now.

PR #1208, opened 2022-03-03
Fix potential buffer overrun in AMDGPU decoders

bbiiggppiigg:
As title, added asserts to check for potential buffer overrun


hainest:
Everything is compiling fine. Have you tried parsing some amd binaries with this new version?

At some point, we should chat about refactoring `amdgpu_cdna2_insn_entry`. It's not needed for the 12.1.0 release, but it will help in the long run.

hainest (approved):

PR #1209, opened 2022-03-04
Docker: use external-tests instead of testsuite in base image

hainest:
We want to use -Werror to catch any warnings introduced by code changes. However, the test suite isn't ready to be built that way, so we use the external-tests instead.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -109,9 +109,6 @@ RUN printf "\n\
 WORKDIR /code
 COPY . /code
 
-# Add test code to base container so we can build tests here
-RUN git clone https://github.com/dyninst/testsuite /opt/testsuite
-
 # Install Dyninst to its own view
 WORKDIR /opt/dyninst-env
 RUN . /opt/spack/share/spack/setup-env.sh && \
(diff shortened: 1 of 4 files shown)


hainest:
@vsoch Aaahhh. I have a bit of a catch-22 here. I need the change in Dockerfile to clone external-tests into /opt, but that change to the base container won't propagate until tomorrow even if I split it out. Everything is working locally, so I'm thinking of merging this now and then testing it again tomorrow after the deployment is done.

vsoch:
Yeah these things happen sometimes :) What I’d do for faster results isbuild and push locally and then you don’t have to wait for tomorrow! With a breaking change like this you either need to do that, update the CI to disable the workflow for a day, or let it break and still wait the day.

hainest:
>What I’d do for faster results isbuild and push locally

What would I be pushing here? Just the PR or the whole container?

vsoch:
The base container.

vsoch:
But is this the base container you need or did you merge after this? This one built two hours ago: https://github.com/dyninst/dyninst/actions/runs/1949289929. We could also add a dispatch event so you can trigger at will.

hainest:
This PR modifies the base container by adding a clone command to grab dyninst/external-tests and then consumes those modifications when running Dockerfile.test. To run this PR successfully, I need to rebuild the base container. If I were to separate out the changes to docker/Dockerfile, is there a way to manually trigger a build+deploy of the base container?

PR #1211, opened 2022-03-07
Docker: use more OS packages

hainest:
@vsoch After using the container a bit, I've had a substantial change of mindset. Using more OS packages is really necessary to get the build times down to something reasonable. The versions aren't as important as I thought they were going to be when we were first writing this. I've tested these changes end-to-end and both Dyninst and the test suite build fine. The only part I was unsure about was the view created for cmake and perl. Now that those are used as external packages, it didn't seem to be necessary to set them up as views. Am I understanding that correctly?


vsoch:
If it builds it's fine with me - please run the build (don't cancel it) to confirm. A heads up that using system (external) packages with spack does lead to weird errors often, but if this is your preference and it builds this is okay with me.

PR #1215, opened 2022-03-08
Correctly handle DWARF subroutines during parallel parsing

### LINKED ISSUE #1214: Dwarf parsing bug for function defined in multiple compilation units
**Intention**
Symtab's API to get function parameters and local variables should return the correct number and types

**Describe the bug**
If.a library or executable has a definition of a function in multiple compilation units within a single library or executable, then the parameters and local variables are duplicated for each occurrence.  This results in number, names and types of parameters being incorrect as the duplicates are interleaved in indeterminate order.  Also the function's local variables are duplicated.  One way that a function can be defined in more than one compilation is for templated classes/functions where the code is emitted in each object file and then the linker picks one copy and make the dwarf location in each function point to this one copy. 

**To Reproduce**
Open a library that uses templated C++ classes or functions and get the parameters for each function.  The symtabAPI library is an instance of this as it use iostreams and STL.

**Expected behavior**
Correct number of parameters and only one copy of local variables.

 - All platforms
 - Present since parallel dwarf parsing was added (10.2)

kupsch:
Fix to only process the DWARF for a function (subprogram) once if it is defined
in multiple compilation units within a library or executable.  This prevents
duplication of parameters, local variables and use of thread-unsafe data
structures from concurrent threads.

Fixes #1214

### DIFF
--- a/symtabAPI/src/dwarfWalker.h
+++ b/symtabAPI/src/dwarfWalker.h
@@ -214,6 +214,10 @@ protected:
     Symtab *symtab_;
     virtual Object * obj() const ;
 
+    // Function object of current subprogram being parsed; used to detect
+    // parseSubprogram recursion
+    FunctionBase *currentSubprogramFunction = nullptr;
+
 }; // class DwarfParseActions 
 
 struct ContextGuard {
@@ -233,7 +237,9 @@ public:
 
     } Error;
 
-    DwarfWalker(Symtab *symtab, Dwarf* dbg);
+    using ParsedFuncs = tbb::concurrent_hash_map<FunctionBase *, bool>;
+
+    DwarfWalker(Symtab *symtab, Dwarf* dbg, std::shared_ptr<ParsedFuncs> pf = nullptr);
 
     DwarfWalker(const DwarfWalker& o) :
             DwarfParseActions(o),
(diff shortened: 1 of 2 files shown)


symtabAPI/src/dwarfWalker.h
```diff
@@ -233,7 +237,9 @@ class DwarfWalker : public DwarfParseActions {
 
     } Error;
 
-    DwarfWalker(Symtab *symtab, Dwarf* dbg);
+    using ParsedFuncs = tbb::concurrent_hash_map<FunctionBase *, bool>;
```
hainest, line 240: I know that the old implementation used a `std::set<FunctionBase*>`, but we have had a great many bugs tied to using pointers as keys in associative containers. Now would be a great time to excise this one. I _think_ the easiest solution is to make a hasher to capture the symbol name, address, and module name to make them unique.

symtabAPI/src/dwarfWalker.h
```diff
@@ -400,8 +411,8 @@ class DwarfWalker : public DwarfParseActions {
             Dwarf_Sword listLength);
 
 
-    // Header-only functions get multiple parsed.
-    std::set<FunctionBase *> parsedFuncs;
+    // Map of Function* to bool (indicates function parsed)
+    std::shared_ptr<ParsedFuncs> parsedFuncs;
```
hainest, line 415: This has to be a `boost::shared_ptr`. Sigh. Maybe one day we'll adopt a C++ standard that's not a decade old. Also, I'm assuming this is shared not because of ownership but for copy semantics (i.e., we don't want to make DwarfWalker move-only by using a unique_ptr here)?

hainest (changes requested):

hainest:
@kupsch @bigtrak Summit and Ray testing are OK.

PR #1217, opened 2022-03-10
Docker: update dependency versions in base container config

hainest:
This should have been part of #1211

### DIFF
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -28,10 +28,11 @@ jobs:
         # 2. Generate matrix programatically and pipe in (recommended)
         # perl is 5.30.0 provided by 20.04 container base
         ubuntu: ["20.04"]
-        boost: ["1.73.0"]
+        boost: ["1.71.0"]
         elfutils: ["0.186"]
         libiberty: ["2.33.1"] 
-        inteltbb: ["2020.2"]
+        inteltbb: ["2020.3"]
+        cmake: ["3.16.3"]
 
     runs-on: ubuntu-latest
     name: Build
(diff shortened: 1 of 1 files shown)


vsoch (commented):
What's the advantage to moving them hard coded into the file? Would the build not be the same?

vsoch:
> @vsoch The motivation here is that the Dockerfile's default versions didn't match the --build-arg being passed in from the matrix in base-containers.yaml. I forgot to synchronize those when I moved to using the versions available in the OS packages. This meant that the spack add boost@${BOOST_VERSION} et al. and the manual entries in packages.yaml didn't match the actual versions from the OS packages.

So my suggestion for this kind of thing is that the build arg defaults (and versions in the CI recipe) are what you would want. That would mean a user (that doesn't change them) gets the default and they should be correct in the CI. It's better practice to expose these things as variables than to hard code and make it harder for a user / development user to change. You are right that it means changing the args in two places, which I think is probably okay. If you do just want to use the defaults as args I guess that's okay to remove them from the recipe, but if you ever add the matrix back (and I'm not around) you'll need to figure out how to do that.

hainest:
>So my suggestion for this kind of thing is that the build arg defaults (and versions in the CI recipe) are what you would want. That would mean a user (that doesn't change them) gets the default and they should be correct in the CI.

Yeah, that was my updated thinking. My only reservation is that a user might say `--build-args BOOST_VERSION=1.78.0`, but the OS version is 1.70.0 (at least in ubuntu-20.04). They would then get a possible build error that would lead to confusion. That's prooobably ok since this container is really intended for our CI use where we have better control of these things.

vsoch:
That's indeed true - although ultimately we want (an educated) developer user to be able to easily change things. Typically people won't change defaults if they are provided.

So that's my suggestion - if you feel really strongly about hard coding it's not functionally going to break anything (and I'll approve) but I do think it makes sense to preserve the variables. Let me know what you think!

hainest:
Let's just change them in the build matrix for now. I think that's the most flexible way forward with the caveat of the user knows what they are doing. Let me change this PR.

vsoch (approved):

vsoch:
Nicely done!

PR #1221, opened 2022-03-11
Docker: use more OS packages for dependencies

hainest:
This should have been part of #1211

### DIFF
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -14,6 +14,10 @@ on:
   push: 
     branches:
       - master
+  
+  pull_request:
+    branches:
+      - master
  
 jobs:
   build:
--- a/.github/workflows/base-containers.yaml
+++ b/.github/workflows/base-containers.yaml
@@ -14,10 +14,6 @@ on:
   push: 
     branches:
       - master
-  
-  pull_request:
-    branches:
-      - master
  
 jobs:
   build:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,7 +52,10 @@ RUN apt-get -qq update && \
       xsltproc \
       cmake \
       libboost1.71-all-dev \
-      libtbb-dev
+      libtbb-dev \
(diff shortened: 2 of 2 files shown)


hainest:
> Does slack external find it work to find those?

No. It won't find libraries- just executables in $PATH. :(

vsoch:
Any reason to printf into the file vs. storing a packages.yaml alongside the container?

hainest:
None other than I didn't know you could do that. If we have both a hand-written packages.yaml and are using `spack external find`, where would the best place be to keep the packages.yaml?

vsoch:
If you are just building the container, you'd keep it in the docker/ folder with COPY to where it needs to be and add it before you setup the environment.

hainest:
@vsoch I was able to get two of the three added with `spack find external`. The last one I'll leave for now. You, @kupsch, @bigtrak, and I should have a chat (preferably in the soon-ish timeframe) to expand what we have now to support multiple OSes and what we want to do about version evolution for dependencies. I'd also like to think about abstracting the base containers into their own layers, if that makes sense for whatever we decide on an overarching architecture.

vsoch:
@hainest could you temporarily add the PR trigger to the workflow to build the new container (and double check it won't deploy) so we can be 100% sure it builds ok? once that is done ping me to take a look, then we can undo that change and merge.

hainest:
Did I do that correctly?

.github/workflows/base-containers.yaml
```diff
@@ -14,6 +14,10 @@ on:
   push: 
     branches:
       - master
+  
+  pull_request:
```
vsoch, line 18: Yep! If you are lazy in the future you can just do:
```yaml
pull_request: []
```

vsoch:
@hainest looks good! Just remove the PR trigger and +1 to merge.

52 minutes - Dyninst you CHONK!

hainest:
52 minutes?! That's a paddlin'.

PR #1222, opened 2022-03-11
Docker: don't build Dyninst through spack for the environment

hainest:


vsoch:
@hainest you'll want to update all the different recipes at once verifying they work otherwise you are going to create chaos.

vsoch:
There are:

 - the base container
 - the test container
 - the release workflow (not merged, but I recommend you integrate and get working before changing)
 - the libabigail builder (also not merged, same)

hainest:
> There are:
>     * the base container
>     * the test container
>     * the release workflow (not merged, but I recommend you integrate and get working before changing)
>     * the libabigail builder (also not merged, same)

Oh, yeah, definitely. I think we should start with the release workflow since the abigail one depends on it.

vsoch:
I guess go nuts then, good luck! I don't have expertise building dyninst outside of spack (I believe my first effort early on I tried and failed) so I can't give much comment in terms of reviewing the PRs.

hainest:
In the current workflow, the base container builds dyninst once using spack and then immediately again using build.sh. This produces two dyninst installations: one in the view (made by spack) and one in /opt/dyninst-env/install/dyninst (made by build.sh). For the base container, this is fine- it's just extra space and cycles.

However, Dockerfile.test only uses build.sh to build dyninst using the updated source tree in /code. When build.sh runs, it sees the source files in /code _and_ the header files in the view (/opt/dyninst-env/.spack-env/view/include/). This is because spack modifies the compiler commands at the shell level to prepend those spack-specific directories. This incompatibility between header files was the cause of the build error in dyninst-branch-builder we recently debugged.

I can think of three possible options for moving forward (definitely propose your own if you think of one!):

1. always use spack to build dyninst
2. always manually build dyninst (as is done in build.sh now)
3. synchronize the two

The first one works fine in the base container- we just need to remove the call to 'build.sh' there[1]. In the test container, things are less straightforward. To build the latest source tree in /code, we need to `spack develop --path /code dyninst && spack add dyninst && spack install dyninst` as is being done in the base container now. However, spack does not allow multiple versions of a package in an environment. The fix here may be as simple as doing 'spack rm dyninst' in build.sh immediately after activating the environment. I'd want to check that the existing spack _install_ isn't interfering with the subsequent `spack develop ...` build. A second item is that we want to programmatically control passing the -Werror family of flags to dyninst if-and-only-if it is being built in the test container. This is so we can detect compiler warnings added by PR code. I have never had much success doing custom builds using `spack develop ...`, but it may be possible.

The second item would move the responsibility of building dyninst into build.sh. This has the nice advantage of being a single point of maintenance and guarantees consistent builds for all versions. The cost is that there is a division between headers/libraries coming from spack (stored in the view) and those coming from dyninst (stored in install/dyninst) which necessitates updating the test, release, and abigail workflows. This was the motivation behind the (now incomplete) changes here.

The third one won't work because spack fiddles with RPATHs.

I'm not sure what would be 'best' (or if there is a 'best'). I think there are some experiments that could shape our thinking here, though.

[1] This has the nice property of separating the responsibilities of the containers. The base container would have all of the dependencies set up according to the base OS and then just a copy of the dyninst source tree. Updating the base container would just be `COPY . /code`. Every other container would then have its own build artifacts- build and tests for the test container and build in /usr for the release container.

vsoch:
I think my original intention was to build with spack.

PR #1224, opened 2022-03-14
DO NOT MERGE: ABI breakage test

hainest:


vsoch:
ooh this should be fun!

vsoch:
:popcorn: :popcorn: :popcorn:

vsoch:
So @hainest you are saying I should merge this, and right away? :laughing:

hainest:
It compiles, so ship it!

vsoch:
ahh so making ABI errors is hard because they don't even build! A new challenge, lol.

hainest:
It's the same issue as before: the headers in the view are older than what's being compiled in the test container. I'll need to fix that first.

hainest:
Oh, neat. It topologically sorts the jobs and runs as many as it can in parallel.

vsoch:
Ah ok - and just to check this is because of some change you've made here but not the base container yet? This typically should not happen!

vsoch:
> Oh, neat. It topologically sorts the jobs and runs as many as it can in parallel.

Yeah! The only bit of potential "could be faster" bit of that is that it needs to download the _entire_ cache per job just to get one file. If you wanted to speed it up you could create a separate artifact per file, and have the job download it (akin to how we name the cache based on the stage, you'd add the filename you want there and then only save that file to it for each).

I figured the time was trivial compared to the compile and left it as is.

hainest:
Yaaay!

PR #1226, opened 2022-03-17
Correctly propagate pc ranges for blocks and local variables

### LINKED ISSUE #1225: The symbol information ignores additional restrictions on range provided by surrounding DW_TAG_lexical_block
**Intention**
I have some tooling note when multiple variables share the same register due to optimization (https://github.com/wcohen/quality_info/blob/master/dyninsttools/cse.C).  It outputs the ranges where there the variables are live and register represents more than one variable. 

**Describe the bug**
I noticed for /usr/lib/debug/lib/modules/5.16.12-200.fc35.x86_64/vmlinux there were some overlaps reported that had ranges that started at 0 or ended at 0xfffffffffffffff.  Like the following for fault_in_readablet:

```
fault_in_readable
[0,ffffffffffffffff) x86_64::rdx __val_gu __val_gu 
[ffffffff812c9e30,ffffffff812c9e50) x86_64::rdi uaddr start 
[ffffffff812c9ea5,ffffffff812c9ead) x86_64::rdi uaddr start 
[ffffffff812c9ec5,ffffffff812c9ec8) x86_64::rdi uaddr start 
```
Looked at the resulting  debuginfo from the vmlinux using:
`llvm-dwarfdump -c --name fault_in_readable /usr/lib/debug/lib/modules/5.16.12-200.fc35.x86_64/vmlinux > fault_in_readable
`

It becomes clear what is happening.  A number of  DW_TAG_lexical_block tags limit the scope of the variables and dyninst is not using the DW_TAG_lexical_block to limit the range of those variables.   One __val_gu defined by:

```
0x02bc0bd0:   DW_TAG_lexical_block
                DW_AT_low_pc	(0xffffffff812c9ea5)
                DW_AT_high_pc	(0xffffffff812c9eb1)
                DW_AT_sibling	(0x02bc0c09)
```
```
0x02bc0bf9:     DW_TAG_variable
                  DW_AT_name	("__val_gu")
                  DW_AT_decl_file	("/usr/src/debug/kernel-5.16.12/linux-5.16.12-200.fc35.x86_64/mm/gup.c")
                  DW_AT_decl_line	(1800)
                  DW_AT_decl_column	(0x07)
                  DW_AT_type	(0x02ba4d2f "unsigned char")
                  DW_AT_location	(DW_OP_reg1 RDX)

```
And the other is defined by
```
0x02bc0c09:   DW_TAG_lexical_block
                DW_AT_low_pc	(0xffffffff812c9e7b)
                DW_AT_high_pc	(0xffffffff812c9e87)

```
```
0x02bc0c2e:     DW_TAG_variable
                  DW_AT_name	("__val_gu")
                  DW_AT_decl_file	("/usr/src/debug/kernel-5.16.12/linux-5.16.12-200.fc35.x86_64/mm/gup.c")
                  DW_AT_decl_line	(1808)
                  DW_AT_decl_column	(0x07)
                  DW_AT_type	(0x02ba4d2f "unsigned char")
                  DW_AT_location	(DW_OP_reg1 RDX)
```

In this particular case the __val_gu is coming from two different instances of the [do_get_user_call](https://elixir.bootlin.com/linux/v5.16.12/C/ident/do_get_user_call) macro.

They don't actually overlap based on the DW_TAG_lexical_block information.  However, dyninst isn't incorporating the DW_TAG_lexical_block information into the information about the variable ranges.

**To Reproduce**
Will work on a smaller reproducer.

**Expected behavior**
For the paricular fault_in_readable function no overlap reported for __val_gu variables. by cse.

 - x86_64 Fedora 35
 - dyninst-11.0.1-3.fc35.x86_64
 - glibc-2.34-28.fc35.x86_64

**Additional context**
The issue still exists in the newly release dyninst-12.1.0.

kupsch:
- fix valid pc address ranges for local variables that are declared in a
  sub-block (brace, try and catch blocks) of their function
 - fix Context class's constructors so they correctly initialize and copy
   all member (use in-class initialize and default constructors)

Fixes #1225

### DIFF
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -381,6 +381,12 @@ bool DwarfWalker::parse_int(Dwarf_Die e, bool parseSib, bool dissociate_context)
             case DW_TAG_lexical_block:
                 ret = parseLexicalBlock();
                 break;
+            case DW_TAG_try_block:
+                ret = parseTryBlock();
+                break;
+            case DW_TAG_catch_block:
+                ret = parseCatchBlock();
+                break;
             case DW_TAG_common_block:
                 ret = parseCommonBlock();
                 break;
@@ -856,6 +862,16 @@ vector<AddressRange> DwarfWalker::getDieRanges(Dwarf * /*dbg*/, Dwarf_Die die, O
 
 bool DwarfWalker::parseLexicalBlock() {
    dwarf_printf("(0x%lx) Parsing lexical block\n", id());
+   return parseRangeTypes(dbg(), entry());
+}
+
+bool DwarfWalker::parseTryBlock() {
+   dwarf_printf("(0x%lx) Parsing try block ranges\n", id());
+   return parseRangeTypes(dbg(), entry());
+}
+
+bool DwarfWalker::parseCatchBlock() {
+   dwarf_printf("(0x%lx) Parsing catch block ranges\n", id());
    return parseRangeTypes(dbg(), entry());
 }
 
--- a/symtabAPI/src/dwarfWalker.h
+++ b/symtabAPI/src/dwarfWalker.h
@@ -98,37 +98,18 @@ public:
     typedef boost::shared_ptr<std::vector<std::pair<Address, Address> > > range_set_ptr;
 private:
     struct Context {
-        FunctionBase *func;
-        boost::shared_ptr<Type> commonBlock;
-        boost::shared_ptr<Type> enumType;
-        boost::shared_ptr<Type> enclosure;
-        bool parseSibling;
-        bool parseChild;
-        Dwarf_Die offset;
-        Dwarf_Die specEntry;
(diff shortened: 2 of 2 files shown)


hainest (commented):
The `Context` constructor also doesn't initialize the `Dwarf_Die` members. They're aggregates for a C interface, so it probably doesn't matter. But it also initializes the members out of order, so that's a barrel of chuckles waiting to happen. I say we use in-class initializers and smack the class with the rule-of-zero hammer.

kupsch:
I made initialized all members in-class and removed the constructors. The Dwarf_Die member were always set before used, but for safety, they are now all value (zero) initialized.  The member were initialized in order (the compiler warns if this is not the case).

hainest (approved):
LGTM. I would suggest a slightly less ambiguous commit title. Something like 'Correctly propagate address ranges for lexical, try, and catch blocks'.

PR #1240, opened 2022-03-30
Remove unneeded #pragma's

kupsch:
- Remove once and interface pragmas
- Replace use of "external/stdint-win.h" and "external/inttypes.h" with
  <stdint.h> and <inttypes.h>
- Remove unused files containing pragmas


proccontrol/src/procset.C
```diff
@@ -45,10 +45,6 @@
 #include <map>
 #include <algorithm>
 
-#if defined(os_windows)
```
hainest, line 48: Did we need to include stdint.h and inttype.h here, or are they not used in this file?

hainest (commented):
common/src/refCounter.h can go away completely as it's never used. I would do that separately.

proccontrol/src/procset.C
kupsch: They are not necessary for this file.

hainest (approved):
LGTM.

PR #1241, opened 2022-03-30
Cleanup (remove) ancient linux kernel support

kupsch:


hainest:
Not to make more work, but while we're here, there is one other old kernel check: https://github.com/dyninst/dyninst/blob/master/common/src/linuxKludges.C#L923

kupsch:
@hainest, updated to also remove old kernel support from linuxKludges.C.

hainest (approved):

hainest:
Thanks!

PR #1246, opened 2022-04-05
Add cmake options for C/C++ language standards

kupsch:
Add cmake options to set C/C++ language standard versions used to build to
facilitate testing:

- DYNINST_C_LANGUAGE_STANDARD cmake option:  C Standard version
- DYNINST_CXX_LANGUAGE_STANDARD cmake option:  C++ Standard version

Make dyninstAPI_RT files build with standard C:
     
- use same language standards as the rest of Dyninst
- define _DEFAULT_SOURCE in some dyninstAPI_RT source files so functions and macros used are defined when using standard C


hainest (commented):
We should probably namespace very generic names like these in DYNINST_.

kupsch:
The options are now DYNINST_C_LANGUAGE_STANDARD and DYNINST_CXX_LANGUAGE_STANDARD.

hainest (approved):

hainest (approved):

PR #1249, opened 2022-04-07
Redo finalization to get correct function boundaries

mxz297:
When multiple functions share code and include a tail call, and when the tail call is first wrongly identified, during finalization, and one of the multiple function will correct the tail call and recompute its function boundary, other functions will still have the wrong function boundaries. This PR fixes this issue.

### DIFF
--- a/parseAPI/src/Parser.C
+++ b/parseAPI/src/Parser.C
@@ -1010,6 +1010,21 @@ Parser::finalize()
         finalize_funcs(discover_funcs);
         clean_bogus_funcs(discover_funcs);
 
+        // We need to redo finalization one more time to get correct
+        // function boundaries. This is intended to handle tail call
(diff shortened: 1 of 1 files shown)


hainest (approved):
This seems like an internal compiler optimization. Is there any reasonable way we can test this repeatably?

mxz297:
@hainest We never have a solution to test interesting binary code constructs. This has been a topic since I joined the group... The best thing that we can do is to add a binary to a repo of interesting binary. In this case, we cannot even do that because it is from JLSE.

hainest:
Yeah, we've talked about making a repo of binaries, but it's never come to fruition. I'll see if I can push that as we build out our containerization efforts.

Everything here looks good, so you can merge when you are ready.

Issue #800, opened 2020-08-17
Verify that the existing documentation examples compile


hainest:
Some of the examples have been moved over to the dyninst/examples repository. Those are working fine, so we don't need to worry about them. However, there are still a few examples in the documentation, itself. Some of them are snippets that aren't complete, but we should make sure they are still reflective of the current APIs.

Issue #711, opened 2020-02-25
A "make doc_examples_check" to verify that examples in documentation compile


wcohen:
The examples are likely the first thing that people will try when learning to use Dyninst. It can be frustrating when they don't work as expected. It would be good to proactively check that the examples in the documentation work to avoid issues like:

The slicing and symbolic evaluation in the dataflowAPI.pdf manual do not compile #699

It would be really nice if there was a "make doc_examples_check" to verify that each of the examples of the documentation can be compiled with the Dyninst being built.  This would be similar to Systemtap's examples being in the testsuite and are run with the "make installcheck".

hainest:
How do you synchronize your compilable code with the documentation text? Do you have some sort of 'include' macro in your LaTeX?

wcohen:
Systemtap isn't using LaTeX and the examples are left as normal systemtap files with *.stp and there is a companion *.meta file that describes the example and is used to generate https://sourceware.org/systemtap/examples/.   I haven't thought through how this would be done in LaTeX.  Searching around looks like one can have separate source files pulled in:

https://en.wikibooks.org/wiki/LaTeX/Source_Code_Listings

So could have something like to have just the source file:

\lstinputlisting{slicing.C}

mxz297:
@hainest We do a lot of different things for the manual...

For ParseAPI, the examples are stored in source files and directly included into latex:

https://github.com/dyninst/dyninst/blob/master/parseAPI/doc/3-Example.tex#L18

For other component libraries, it may directly write examples use source code listing:

https://github.com/dyninst/dyninst/blob/master/symtabAPI/doc/3-Examples.tex#L12

And for BPatch and Proccontrol, the manuals are written in Microsoft Word...

Issue #394, opened 2017-07-25
Documentation examples should use container returns properly


wrwilliams:
We have a nontrivial number of code examples that use the following idiom:
```
vector<Thing> foo = findSomeThings();
doSomething(foo[0]); // assume the first one is an okay representative
```
These should be replaced by safe uses of the following forms:
```
if(!foo.empty()) doSomething(foo[0});
for(f in foo) doSomething(f);
```

hainest:
This is going to get some attention soon, so we should also upgrade code like

	auto it = container.begin();
	for(; it != container.end(); ++it) {
		some_type *x = *it;
		...
	}

to use range-based for loops

	for(auto *x : container) {
		...
	}

since we now require full C++11 support.

PR #1253, opened 2022-04-12
Remove unused build options

hainest:
We should revisit using CPack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,64 +176,6 @@ if(BUILD_TARBALLS)
     endif()
   endif()
 endif()
-if(BUILD_DOCS)
-  add_custom_target(doc)
-  set(LATEX_DOCS dynC_API dataflowAPI instructionAPI parseAPI patchAPI symtabAPI stackwalk)
-  set(WORD_DOCS proccontrol dyninstAPI)
-  add_custom_target(proccontrol-doc
-    DEPENDS ${CMAKE_SOURCE_DIR}/proccontrol/doc/proccontrol.pdf)
-  add_custom_target(dyninstAPI-doc
-    DEPENDS ${CMAKE_SOURCE_DIR}/dyninstAPI/doc/dyninstAPI.pdf)
-  foreach(COMPONENT ${WORD_DOCS})
-    add_dependencies(doc ${COMPONENT}-doc)
-    set_target_properties(${COMPONENT}-doc PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD 1)
-    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${COMPONENT}/doc/${COMPONENT}.pdf
-      DESTINATION ${INSTALL_DOC_DIR}
-      RENAME ${COMPONENT}-${VERSION_STRING}.pdf
-      OPTIONAL
-      )
-  endforeach()
-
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -163,78 +163,6 @@ endif()
 set (VERSION_STRING "${DYNINST_MAJOR_VERSION}.${DYNINST_MINOR_VERSION}.${DYNINST_PATCH_VERSION}")
 set (DYNINST_NAME "DyninstAPI-${VERSION_STRING}")
 
-if(BUILD_TARBALLS)
```
vsoch, line 166: This giant big beautiful block of red deletion is very satisfying, +1 destroy it.

cmake/options.cmake
```diff
@@ -7,11 +7,8 @@ option (LIGHTWEIGHT_SYMTAB "Use lightweight symtab interface for ParseAPI, ProcC
 # Use ParseAPI analysis in Stackwalker?
 option (SW_ANALYSIS_STEPPER "Use ParseAPI-based analysis stepper in Stackwalker" ON)
 
-option (BUILD_TARBALLS "Build Dyninst package tarballs. Requires git archive, tar, gzip." OFF)
 option (BUILD_RTLIB_32 "Build 32-bit runtime library on mixed 32/64 systems" OFF)
 
-option(BUILD_DOCS "Build manuals from LaTeX sources" ON)
```
vsoch, line 13: Would be nice to someday have web docs because reading pdfs is sooooo 2009.

kupsch (commented):
Removing BUILD_TARBALLS is good as it is unused.

The BUILD_DOCS currently is broken; `make docs` sort of works the first time, but is broken for subsequent invocations.  It would be nice to have a make target to build the latex docs that had proper dependencies so it would rebuild correctly. If this is removed, there needs to be instructions on how build the docs added to the repo and perhaps a script to create the pdfs.

vsoch:
I'm pretty good at that, I could probably help in a week or so, have too many "after work personal projects" right now but probably that won't be the case after I finish some up! I've done readthedocs with rst a gazillion times.

hainest:
@vsoch I wasn't hinting- I promise. :smiley: If you have the cycles and will, I certainly won't turn away the help, though.

vsoch:
yeah I can add to my list of "stuff to do" I'll get to it sooner than later. :)

kupsch:
@hainest I understand that the latex documentation is only part of the documentation, but it wasn't clear to me how someone that forked the repo would know how to build the latex documentation; it would require them to know the existence of the google doc. If we are moving from latex soon then this is OK and the merge can proceed.  If not, then putting the instructions in README.md (or similar) that is versioned in the repo would be useful.

hainest:
@kupsch Oh, sorry, I didn't get that part of what you were saying. As far as I know, no one but us has ever tried to build the documentation. Even the Red Hat folks use the generated PDFs. I'm pretty sure the BUILD_DOCS was a holdover from the autotools->CMake conversion. I would _really_ like to see us move to readthedocs soon. Creating the documentation PDFs is by far the biggest pain of doing a release. Since Vanessa offered a bit of her time, I think the two of us can put our heads together to do a bulk export. There are still doc bugs that need to be fixed (#800, #711, #394), but we can do that afterward.

PR #1256, opened 2022-04-13
start of work to refactor the docs to use readthedocs

vsoch:
This is the first pass to format the previous latex into rst. I have done the conversion and only started to go through ensuring that content is preserved (meaning I did not miss anything) and all the code blocks are formatted (after the automated conversion they indeed are not!) I will want a few more hours to finish this up, and then we need to discuss deployment. E.g., I recommend readthedocs so you can automate deployment and keep versioned docs. We can also deploy to github pages (I can make a workflow) but I do not have a good suggestion for versioning things that way. I also have not added in an ability to still render the pdfs if that is desired, which I think should be possibly. Finally, it would be nice if some of these docs could render from docstrings - I know how to do this for Python so I wonder if Cpp is that much different. It is probably terrible because it is cpp, but what can you do?

For everyone's sanity, we should probably pause adding new content to docs via the PR here so it can be merged here first and then done. In the meantime while I'm finishing this up, let's discuss:
 
 - deployment 
 - renderings to do aside from rst on readthedocs
 - other assets in the doc/ folders that I might have missed
 - anything else?

FYI these ones are super cute. Behold, the dyninst dragon! If Dyninst doesn't scare you away he sure might :laughing: 

image

No seriously he's super cute I'm losing it. cc @hainest 

image

Looking at it now, I probably need to lighten up the purple - it's a little too dark for the text background.
I should have a github pages preview to show y'all tomorrow after I finish up detail work. For now, dinosaur awwwaaay!


vsoch:
okay still haven't pushed changes, but a current version is here: https://researchapps.github.io/dyninst-docs/ there are maybe 600 lines of issues to go through (haven't done that yet). @hainest is this a direction you want to go in / should I continue working on? I don't want to do all this time intensive work and then not have it be what you were looking for.

hainest:
Holy moly. This is great! You are making us all look bad with your crazy wizardry! :smile_cat:

The first thing to check is how well pandoc was able to convert the latex. Spending some time comparing the rendered version you have and the PDFs is probably(?) the best way. I don't expect you to do that, but I won't have time until Friday. My evening time is consumed by prep for my biochem test that's tomorrow.

Being able to automatically deploy docs on release would be amazing. I'm assuming we can set up a key between github and readthedocs to fire that off in an action of some sort?

Can we directly import rst into readthedocs, or does it need to be translated into Sphynx?

We definitely don't have to address it in this PR, but I would _love_ to be able to extract the code examples from the docs and make sure they are built as part of our PR testing. That's a big chunk of work that I don't expect you to do, but it would be amaze-balls.

>renderings to do aside from rst on readthedocs
I'm not sure. Aside from what readthedocs does to convert rst into html, what other rendering do they do? We have some figures in the docs that should be preserved. I'm pretty sure they are sitting in the doc subdirectories, so I should be able to account for them all.

vsoch:
okay I think I'm done! https://researchapps.github.io/dyninst-docs/ 

> The first thing to check is how well pandoc was able to convert the latex. Spending some time comparing the rendered version you have and the PDFs is probably(?) the best way. I don't expect you to do that, but I won't have time until Friday. My evening time is consumed by prep for my biochem test that's tomorrow.

No worries - study away! I did some early comparison today and the major difference is that we can't easily "embed" a pdf into the rst directly, so the previous pdf embeds are now links. We would ideally re-generate these images in png/svg and then be able to include them. It's not a "must do now" thing but eventually would be nice. Also, I think readthedocs will generate pdfs for us out of the box, so we don't need to think about it.

> Being able to automatically deploy docs on release would be amazing. I'm assuming we can set up a key between github and readthedocs to fire that off in an action of some sort?

It's so easy - you literally just log in, and then setup / connect the repository. Someone with permission to dyninst will need to (so you are a good contender). Then they will build on pushes to main (and you can configure for releases).

> Can we directly import rst into readthedocs, or does it need to be translated into Sphynx?

Oh so I already did the full rst conversion, and yes readthedocs can render that into docs via sphinx.

> We definitely don't have to address it in this PR, but I would love to be able to extract the code examples from the docs and make sure they are built as part of our PR testing. That's a big chunk of work that I don't expect you to do, but it would be amaze-balls.

That's definitely possible, and I could probably point you to examples but I'm not offering to do it because, well, C++ :laughing: 

> I'm not sure. Aside from what readthedocs does to convert rst into html, what other rendering do they do? We have some figures in the docs that should be preserved. I'm pretty sure they are sitting in the doc subdirectories, so I should be able to account for them all.

yep and they are there, but currently linked as pdf so we'd want to convert them to an actual image and then they will show up alongside.

vsoch:
Note that many repos are facing this weird bug - GitHub is on it. https://github.com/actions/checkout/issues/760

vsoch:
Just pushed fix for action so it should build now https://github.com/actions/checkout/releases/tag/v2.4.1

hainest:
@me to check:

- [ ] common
- [ ] dataflow
- [ ] dwarf?
- [ ] dynC
- [ ] dyninstAPI
- [ ] dyninstAPI_RT
- [ ] elf?
- [ ] external?
- [ ] instrutionAPI
- [ ] parseAPI
- [ ] parseThat? (should really be in dyninst/tools repo)
- [ ] patchAPI
- [ ] proccontrol
- [ ] stackwalk
- [ ] symlite?
- [ ] symtab

vsoch:
@hainest did you get a chance to look at this? As changes are made to the docs in other PRs it's going to make it impossible t keep up to date here, and I'm not offering to do this conversion again (it was a lot of work).

hainest:
We definitely don't want to redo the conversion - and a million thanks to you for tackling it! I can't believe it's already been two weeks. ugh. Let me dedicate some time to this today and tomorrow. We _will_ get this merged.

vsoch:
phew ok! sounds good!

hainest:
@hainest Items to add:

Debugging instrumented code: https://docs.google.com/document/d/1xS1dxbX9-iieyIDg8_c0XP1bnBxqHsbjtyIaCgwHp1U/edit

gap parsing: https://docs.google.com/document/d/1cC47hxNvep0K-MuITbfkWDgrdaXhvI584kJSOzV6y9k/edit

cuda profiling tool: https://docs.google.com/document/d/1h12Uq-cQyNSRuajZQo9bhcpFFPZVmL1g-ztCRifze5s/edit

static-time code: https://docs.google.com/document/d/1_1jHuZo1J7O_Raza0xdCHmQfqCPYylr9kLh-JL7JFKs/edit

PR #1262, opened 2022-05-03
add case in DwarfEncToReg for Arch_intelGen9

jmellorcrummey:
Add a missing case to MachineRegister::DwarfEncToReg for Arch_intelGen9 that returns Dyninst::InvalidReg instead of reaching the default case with an assert. 

This change was necessary to swallow DWARF variable information for Intel GPUs, which HPCToolkit ignores.

### DIFF
--- a/common/src/dyn_regs.C
+++ b/common/src/dyn_regs.C
@@ -1834,10 +1834,12 @@ MachRegister MachRegister::DwarfEncToReg(int encoding, Dyninst::Architecture arc
             break;
         case Arch_amdgpu_vega:
         case Arch_amdgpu_cdna2:
-            // ignore CUDA register encodings for now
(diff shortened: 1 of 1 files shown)


jmellorcrummey:
@jmellorcrummey taking over for @mxz297

hainest:
@jmellorcrummey We don't have a way to test GPU code right now. If this patch is working on your end, we'll merge.

jmellorcrummey:
It works on my end. Without it, we can't process DWARF for intel GPU binaries: the default assertion is triggered.

hainest:
@jmellorcrummey Excellent.

@kupsch This couldn't be detected by '-Wswitch` because of the default. I think removing the default is a good idea.

PR #1269, opened 2022-05-05
[WIP] Machinery for updating and enforcing source formatting

hainest:
Originally part of #1105


jrmadsen:
@hainest This might be useful for the git blame issue I mentioned: 

https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

Particularly the creation of the .git-blame-ignore-revs file

hainest:
@jrmadsen Oh, that's nice. We'll definitely want to do that. Before running clang-format, we'll need to sprinkle some `// clang-format off` around, for example, in the whitespace for the instruction decode tables which has meaning. I'll chat with the other devs and see if they have preferences about layout. At this point, I'd be happy with just _consistent_ formatting...

Does this need to be done before the other PRs?

jrmadsen:
> Does this need to be done before the other PRs?

Not at all. The few changes to the source have been relatively minimal and aren't difficult to merge.

jrmadsen:
Just a quick note though, `DyninstUtilities.cmake` was a 100% new file but `DyninstLibrary.cmake` was originally `shared.cmake`.

PR #1273, opened 2022-05-10
Tweak build.sh + Dockerfile.test

### LINKED ISSUE #1272: Testing suite container bug when files are renamed
**Describe the bug**

When files get moved within the repository, the containers used for testing copy the PR source code onto an existing source code tree. Thus, you will be unable to catch any invalid `#include "..."` statements in the source code or `include(...)` in cmake.

**To Reproduce**

See #1271 

**Expected behavior**

You need to either remove the existing source code in the container before copying or checkout the revision instead.

jrmadsen:
- remove old code before copying code into testing container
- build with verbosity in build.sh (this helps when build errors occur)
- Closes #1272


hainest:
This was the fix I was going to do. Thanks for getting it early.

PR #1276, opened 2022-05-27
Add callback for parsing unknown instructions

hainest:


hainest:
TODO:

1. Is `InstructionDecoder::m_buf` the real address of the instruction in the binary?
2. Can we make an N-byte NOP from `e_nop`, size is determined by `InstructionDecoder`?

hainest:
@mwkrentel I removed the second parameter and (transparently?) modified the implementation. Could you verify this is still working for you, modulo the avx512 issue?

kupsch (changes requested):
Build fails:

```
CMakeFiles/instructionAPI.dir/src/InstructionDecoder.C.o: In function `Dyninst::InstructionAPI::InstructionDecoder::decode()':
/usr/include/c++/7/bits/stl_algobase.h:202: undefined reference to `Dyninst::InstructionAPI::InstructionDecoder::maxInstructionLength'
```

hainest:
> Build fails:
> 
> ```
> CMakeFiles/instructionAPI.dir/src/InstructionDecoder.C.o: In function `Dyninst::InstructionAPI::InstructionDecoder::decode()':
> /usr/include/c++/7/bits/stl_algobase.h:202: undefined reference to `Dyninst::InstructionAPI::InstructionDecoder::maxInstructionLength'
> ```

What compiler? It's working for me with gcc-9 and the CI builds passed.

kupsch:
Same result with gcc 7.5 and 11.3.

kupsch:
This is for a DEBUG build. RELEASE builds are OK.

kupsch:
If std::min is not inlined, then it fails as std::min parameters are passed by const reference which requires that maxInstructionLength is resolvable, but it is not exported. If I pass `maxInstructionLength + 0` then it succeeds as a temporary is passed and doesn't need to be resolved by the linker.

hainest:
I found the same, but I can't get the same behavior out of a godbolt example: https://godbolt.org/z/xfnP47bve. I guess I'll add a +0. Eww.

kupsch (approved):

mwkrentel:
@hainest Yes, this is working for me.
Actually, it works quite well.
What a strange issue with the maxInstructionLength.

PR #1283, opened 2022-07-06
Added support for gfx908 based on the XML-ISA-DROP for MI100

bbiiggppiigg:
+ Fixed a bug in the instruction decoder that returns incorrect value on the last instruction.
+ Added support for gfx908 based on the XML-ISA-DROP


bbiiggppiigg:
The abi break detected seems to be caused by newly inserted enum values, which should be reasonable in this case as we are adding support for a new architecture.

common/h/dyn_regs.h
```diff
@@ -60,6 +60,7 @@ namespace Dyninst
         Arch_amdgpu_vega  =  0x84000000,
         Arch_cuda         =  0x88000000,
         Arch_amdgpu_cdna2 =  0x94000000, //future support for cdna2
+        Arch_amdgpu_gfx908 =  0x98000000, //future support for gfx908
```
hainest, line 63: Is gfx908 the "official" AMD name? I ask because it varies substantially from the previous ones.

hainest (commented):
The changes outside of adding the gfx908 architecture should really go in a separate PR so that we can more explicitly call them out. Future you will appreciate it when you are debugging. :)

bigtrak:
> 
> > @@ -60,6 +60,7 @@ namespace Dyninst
>          Arch_amdgpu_vega  =  0x84000000,
>          Arch_cuda         =  0x88000000,
>          Arch_amdgpu_cdna2 =  0x94000000, //future support for cdna2
> +        Arch_amdgpu_gfx908 =  0x98000000, //future support for gfx908
> 
> Is gfx908 the "official" AMD name? I ask because it varies
> substantially from the previous ones.

My crib notes say ...

	architecture	AMD code	GPU		alt names
	------------	--------	-----------	---------
	vega10          gfx900          mi-25
	vega20          gfx906          mi-50 mi-60
	cdna 1          gfx908          mi-100          arcturus
	cdna 2          gfx90a          mi-200  2xx     aldebaran
	[ Hsuan-Heng may have access to the next 2, I'm using public info 
	  that is guess-work ]
	cdna 3          gfx90c ???      mi-300          ???
		-- that's probably a graphics card
	cdna 3          gfx940 ????     mi-300          ???
		-- that's probably mi-300

I would suggest staying with the arch names, as it helps
identify the family of the GPU.   I think if there were
both Arch_amdgpu_cdna1 and arch_amdgpu_gfx908 identical
values in the enum, we would be free to change things
in the future.

I believe the best way to proceed is treating amdgpu
support as a "production prototype" for now, and accept
that it is not finalized nor ABI stable until sometime
in the future.   We are really pushing out the AMD stuff
before it is ready for AMD's benefit, and that is the cost
of doing it that way.


Bolo

common/h/dyn_regs.h
bbiiggppiigg: Yes.
In later patches vega and cdna2 will be renamed to similar style.

bbiiggppiigg:
@bigtrak 

So every architecture might have [variants](https://llvm.org/docs/AMDGPUUsage.html#instructions).

Each variant might [map](https://llvm.org/docs/AMDGPUUsage.html#processors) to more than one product (mi-xxx). 

I think in one of the past meetings we brought up this issue with naming before, and the AMD folks suggested us to use the gfx-xxx names for better clarity.

kupsch:
That is correct.  AMD recommended using the processor column from 
https://llvm.org/docs/AMDGPUUsage.html#processors

woodard:
The thing with the libabigail ABI checks is, you can just increment the library age and revision and it should be fine. That would allow backward compatible binaries but still allow new code to make use of those enums.

bigtrak:
> The thing with the libabigail ABI checks is, you can just increment
> the library age and revision and it should be fine. That would allow
> backward compatible binaries but still allow new code to make use of
> those enums.

Ahh, so you are saying we are doing our release cycles backwards!?

Not really, but sorta close.

We only change library versions for the next major release that
has an ABI or API break.

	However, we are free to create ABI breaks anytime after making
	a release, as we no longer guarantee ABI compatability in
	our dev train.

So, when a release is made, we should bump the ABI as soon as we
have an abi break in our dev, and then leave it there until the next
release, at which point the cycle repeats itself.

	No reason to bump multiple times in a dev cycle, as that is
	for the outside world, not for the developers who build from
	scratch on every build.

Currently we do the bump at the next release.

How do you handle it at Redhat?

Bolo -- Josef T. Burger

woodard:
So how we handle it at RH varies from project to project. Like the DOE and the labs now that tooling like libabigail is getting near production ready it is being used more. 

Exactly when you bump the library version is a good question and probably needs to be thought of in the context of the CI automation. Here is what I propose. We should probably run this by Vanessa to make sure that this works with her automation or at least can be made to work with her automation.

1. Start with the most recently released version of libabgail. 
2. create a "gold" ABI from the release. 
3. This gold ABI is copied to the the "current" ABI.
4. when a PR comes in it, it is compared to the current ABI.
5. if it matches the current ABI then the test passes.
6. if it doesn't match the current ABI then you apply a PR that updates the library version, and updates the current ABI for the development branch according to the library versioning rules https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html before you apply the PR which broke the ABI. Then the test should pass.
7. As subsequent PRs come in which break ABI, then just verify that that the library version changes are correct in comparison to the release gold ABI. Then you regenerate the current ABI.
8. When you release, you move the current ABI to the gold ABI
9. One of the first things that you take AFTER the release is update libabigail to the current released version and regenerate the "gold" release ABI to pick up the any new bug fixes that we have. You are then back at step 2.

common/h/dyn_regs.h
hainest: We should let the Rice folks know about this soon. If they've made a release of HPCToolkit that uses the vega/cdna2 names from Dyninst, that will complicate how we handle this.

instructionAPI/src/debug_decode.C
```diff
@@ -0,0 +1,77 @@
+/*
```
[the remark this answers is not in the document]
hainest: Agreed.

hainest:
@bigtrak @woodard We've been down this road before. It would be a huge breakage to use the libtool conventions because Dyninst's SONAME includes both the major and minor version. Currently, we only merge ABI-breaking[1] commits when we are ready to do a major version release. It's a pain, but that's what we have to do. I'm not strictly against changing this, but we must be prepared to break the world to do it.

[1] We do not consider the _addition_ of names to be an ABI break.

From Stan's email

```
> Dyninst currently follows these two conventions for every release. No code changes are made to the libraries without also changing the real name, and no ABI changes are made without updating the major version (i.e., the version number in the soname). We follow something very close to the kde rules to determine breakages (https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C++#The_Do.27s_and_Don.27ts).
>
> All of this said, I would really like to be able to break ABI on minor versions (but not patch versions)- 12.1.0 could be a break, but 12.0.1 never is. That would violate (1). Does RedHat strictly follow this convention? In essence, are you linking against the real name or the soname?

Most libraries DT_SONAME is of the form:
% eu-readelf -d /usr/lib64/libssh.so.4.8.7
SONAME            Library soname: [libssh.so.4]

(libNAME.so.MAJOR)

So a change to libssh.so.4.9.0 would not require a user of the library to relink.

I notice dyninst uses:
% SONAME            Library soname: [libdyninstAPI.so.11.1]

and so stap links against:
  NEEDED            Shared library: [libdyninstAPI.so.11.1]


which means that everytime either the major or minor version changes and I add the new version of dyninst to the release then I relink systemtap so it corresponds to the correct version.

I notice that is set in dyninst/src/cmake/shared.cmake

set (SOVERSION "${DYNINST_MAJOR_VERSION}.${DYNINST_MINOR_VERSION}")
set (LIBVERSION "${SOVERSION}.${DYNINST_PATCH_VERSION}")
set (DYNINST_VERSION "${LIBVERSION}")

Typically abi breakage would only be with the major version, minor version would be non abi breakage changes , and the final .N is the release number.  The last component, release #, may get bumped for example, create a new release just because everything is build with a new compiler or a new version of boost was installed.

There are, however, libraries that use major.minor
% eu-readelf -d /usr/lib64/libruby.so.3.0 | grep SONAME
(standard input):  SONAME            Library soname: [libruby.so.3.0]

So dyninst has company in that area.
```

bbiiggppiigg:
>

hainest:
Any motion on this since the last discussion? I'd like to do a point release (probably 12.3) so we can get this out for testing with our partners before the 13.0 release.

bbiiggppiigg:
@hainest 
1. You still think that I should separate out the change to the debugging interface?  I'm not sure what is the right way to pull them out of this PR. Or perhaps I just undo them and redo them in another PR?
2. Based on the lastest ISA drop, the decoding for one of the instructions is different then before. I'm waiting on the AMD side to confirm whether the change is intentional or not.

hainest:
>     1. You still think that I should separate out the change to the debugging interface?  I'm not sure what is the right way to pull them out of this PR. Or perhaps I just undo them and redo them in another PR?

I think having those in a separate PR would be best.

>     2. Based on the lastest ISA drop, the decoding for one of the instructions is different then before. I'm waiting on the AMD side to confirm whether the change is intentional or not.

Ok. Were you able to get in contact with the Rice folks about a migration strategy for renaming the existing architectures (e.g., cdna2 -> gfx90a)?

bbiiggppiigg:
@hainest I think this PR is ready for review again.

hainest:
@bigtrak I don't see any more failures than usual: https://bottle.cs.wisc.edu/search?dyninst_branch=PR1283

bigtrak:
> @kupsch @bigtrak
> There are at least two 'debug' libraries: one in libcommon and one
> in libsymtabapi. I've never understood why the ones in libsymtabapi
> are not in libcommon as symtab links against it. libinstructionapi
> also links against libcommon.
>
> What are your thoughts on the sprawl of debug interfaces here? Should
> we move these over to libcommon? I don't have a strong opinion, but
> I think it would be good to have them in one place (optimally in a
> libdebug, but that's for the future).


Leave it for now.  We have bigger fish to fry than
creating the perfect monster.    

One thing that is great about the current setup is that if you
are needing to add some debugging to symtab .. you can do it
locally without affecting the rest of dyninst.   Add something,
it spreads into symtab as you need, just that part of symtab
recompiles, and then off to continue debugging with a new
debug option .. and not recompiling the rest of dyninst.


Bolo

kupsch:
Tony said the official arch names to product names are on this web page:

https://llvm.org/docs/AMDGPUUsage.html#processors

Jim

bbiiggppiigg:
I'm working on some more stuff for gfx908 so please don't merge yet

bigtrak:
This pull request Resolves #1380.

PR #1284, opened 2022-07-10
fix races with parallel analysis of cubins

jmellorcrummey:
- Boost multi_index_set is not thread safe. Add an OpenMP critical section to guard against parallel inserts. 
    - Is this enough? A more comprehensive fix would be to add a reader/writer lock and also synchronize reads of this structure.
- dwarf_next_lines uses calls code that uses glibc's tsearch, which is not thread safe. A better fix would be to replace or synchronize the use of tsearch in elfutils. 

NOTE: in practice, these fixes led to safe parallel analysis using 16 threads of a 240MB LLNL export-controlled cubin that can't be used externally for testing.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -3646,9 +3646,15 @@ LineInformation* Object::parseLineInfoForObject(StringTablePtr strings)
 
     int status;
 
-    while ((status = dwarf_next_lines(dbg, off = next_off, &next_off, &cu,
-			     &files, &fileCount, &lineBuffer, &lineCount)) == 0)
-    {
+    while (1) {
+
+#pragma omp critical (next_lines)
+{
+    status = dwarf_next_lines(dbg, off = next_off, &next_off, &cu,
+                              &files, &fileCount, &lineBuffer, &lineCount);
+}
+    if (status != 0) break;
+      
 
     boost::unique_lock<dyn_mutex> l(strings->lock);
     size_t offset = strings->size();
--- a/symtabAPI/src/LineInformation.C
+++ b/symtabAPI/src/LineInformation.C
@@ -60,8 +60,12 @@ bool LineInformation::addLine( unsigned int lineSource,
                                         lowInclusiveAddr, highExclusiveAddr);
     Statement::Ptr insert_me(the_stmt);
     insert_me->setStrings_(strings_);
-   return insert( insert_me).second;
-
+   bool result;
+#pragma omp critical (addLine)
+{
+   result = insert( insert_me).second;
+}
+   return result;
 } /* end setLineToAddressRangeMapping() */
 bool LineInformation::addLine( std::string lineSource,
                                unsigned int lineNo,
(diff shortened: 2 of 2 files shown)


jmellorcrummey:
@hainest I am giving a tutorial at LANL, Sandia, and LLNL this Wednesday. We need this commit to update our install at the labs to avoid races in hpcstruct when analyzing NVIDIA CUBINs in parallel. Will it be possible to accept this? As you can see, it just adds critical sections, which are harmless - they fix bugs for OpenMP users but are a no-op otherwise.

symtabAPI/src/Object-elf.C
```diff
@@ -3646,9 +3646,15 @@ LineInformation* Object::parseLineInfoForObject(StringTablePtr strings)
 
     int status;
 
-    while ((status = dwarf_next_lines(dbg, off = next_off, &next_off, &cu,
-			     &files, &fileCount, &lineBuffer, &lineCount)) == 0)
-    {
+    while (1) {
```
hainest, line 3649: Explicit infinite loops like this are technically UB in C++. Most compilers just "do the right thing". I'll make a note to revisit.

hainest (commented):
I'm running tests now. They should be done by end-of-day. If they are good, I'll merge so you can have this for your presentation on Wed.

symtabAPI/src/Object-elf.C
jmellorcrummey: If you look at the code, all I wanted to do was to enclose the call to dwarf_next_lines in the critical section. Rather than having the test in the loop header, I had the test to break from the loop a few lines below the loop header.

The loop could be for(;;) rather than while (1) if you prefer.

hainest: >If you look at the code, all I wanted to do was to enclose the call to dwarf_next_lines in the critical section. Rather than having the test in the loop header, I had the test to break from the loop a few lines below the loop header.

Yup. I was originally going to recommend moving the synchronized call to `dwarf_next_line` into a lambda and then put the test back into the `while`. Something like

    auto next_line = [&](){
      return dwarf_next_lines(dbg, off = next_off, &next_off, &cu,
			       &files, &fileCount, &lineBuffer, &lineCount) == 0;
    };

    while(next_line()) {
      ...

but I couldn't prove to myself that was thread-safe wrt the variables used in the dwarf call without looking at the context of the dwarf call in greater detail. I didn't want to hold up this change because you need it soon.

>The loop could be for(;;) rather than while (1) if you prefer.

Those fall under the same rule (intro.progress).

I double-checked the standard and the current form is actually thread-safe so long as OpenMP is enabled because of

```perform a synchronization operation or an atomic operation```

I agree that we should still move to reader/writer semantics so we don't get bad reads or ABBA issues. I'll make a Github issue for us to revisit this.

jmellorcrummey: The loop has side effects and an exit condition. I don't think it meets the definition of an infinite loop. I thought the way I rewrote

while (status = dwarf_next_lines(...)) {
...
}

was much better than the following:

as
#pragma omp critical (next_lines)
status = dwarf_next_lines(...)
while (status) {
...
#pragma omp critical (next_lines)
status = dwarf_next_lines(...)
}

hainest:
No new regressions. Merging.

PR #1292, opened 2022-08-09
tbb version file moved

rscohn2:
The FindTBB.cmake in the repo does not work with recent releases of TBB because tbb/version.h no longer contains the macros defining the version. If you remove FindTBB, then you will use the file provided by tbb.  They have been providing the file for a few years so it should work for most users.


hainest:
We are currently working to retool our CMake. Part of that includes increasing the minimum supported version of TBB to one that comes with a CMake installation. I _think_ that's around 2020.1-ish. Do you happen to know the actual version?

rscohn2:
> We are currently working to retool our CMake. Part of that includes increasing the minimum supported version of TBB to one that comes with a CMake installation. I _think_ that's around 2020.1-ish. Do you happen to know the actual version?

I see the TBBConfig.cmake in 2017, update 7 https://github.com/oneapi-src/oneTBB/releases/tag/2017_U7

rscohn2:
Closing because this is covered in CMake retooling.

PR #1293, opened 2022-08-20
Make SymtabCodeSource take a const param

Myles1:
This just makes it slightly easier to pass in an `std::string` to the constructor, allowing `my_str.c_str()` directly, without needing to do a const cast or copy the string.


hainest:
Thanks! There are quite a few const correctness issues in Symtab that have been on our TODO list for a while. They are ABI breaks, so we have to wait until a major release to publish them.

hainest:
No test regressions when rebased on master.

PR #1308, opened 2022-10-21
Fix possible null pointer access in BPatch_module::findFunctionByAddress

hainest:
'isValid' checks that 'mod' is not null. The code here only runs if mod _is_ null. Oddly, this only showed up as a compile warning when ENABLE_STATIC_BUILDS=ON.


kupsch (approved):
You could remove the ': ' from the end of the message.  Otherwise looks good.

hainest:
> You could remove the ': ' from the end of the message. Otherwise looks good.

Fixed.

hainest:
No new regressions. https://bottle.cs.wisc.edu/search?dyninst_branch=PR1308

PR #1309, opened 2022-10-24
Do not build dyninstAPI_RT as separate CMake project

hainest:


hainest:
This is a necessary step to get target-based TBB working and a generally good cleanup. From the comments, the initial reason the runtime was built separately was to allow using different compilers for the RT and the rest of Dyninst. That seems to no longer be relevant (and certainly not tested).

kupsch (approved):
looks good

hainest:
No new regressions: https://bottle.cs.wisc.edu/search?dyninst_branch=PR1309

PR #1311, opened 2022-10-24
Fix incorrect format string in Result.h for u48 / s48 / u64 / s64

### LINKED ISSUE #1310: InstructionAPI Result type outputs wrong format string on certain types
**Intention**
Disassemble a binary and check that the output assembly is correct

**Describe the bug**

https://github.com/dyninst/dyninst/blob/eea540a94564371d3170d01ab16c62ccbc8f9b9d/instructionAPI/h/Result.h#L441-L522

Result::format  is supposed to return its value in hex type, but for u48 / s48 / u64 / s64, the format function returns a formatted string in dec instead of hex.

And as ArchSpecificFormatters always prepend "0x" to the format string of Immediate, value 48 will be incorrectly outputted as 0x48 when it is of type s64.
https://github.com/dyninst/dyninst/blob/eea540a94564371d3170d01ab16c62ccbc8f9b9d/instructionAPI/src/ArchSpecificFormatters.C#L98-L99

https://github.com/dyninst/dyninst/blob/eea540a94564371d3170d01ab16c62ccbc8f9b9d/instructionAPI/src/ArchSpecificFormatters.C#L158-L159

https://github.com/dyninst/dyninst/blob/eea540a94564371d3170d01ab16c62ccbc8f9b9d/instructionAPI/src/ArchSpecificFormatters.C#L214-L216

**To Reproduce**
Create a Result object with type mentioned above and call format on it.

bbiiggppiigg:
Fixes  #1310

### DIFF
--- a/instructionAPI/h/Result.h
+++ b/instructionAPI/h/Result.h
@@ -473,9 +473,10 @@ namespace Dyninst
                                 snprintf(hex, 20, "%x", (unsigned int)val.s32val);
                                 break;
                             case u64:
-                                return std::to_string(val.u64val);
+                                snprintf(hex, 20, "%lx", val.u64val);
+                                break;
                             case s64:
-                                return std::to_string(val.s64val);
+                                snprintf(hex, 20, "%lx", (uint64_t) val.s64val);
                                 break;
                             case sp_float:
                                 snprintf(hex, 20, "%f", (double)val.floatval);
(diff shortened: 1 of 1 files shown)


hainest:
Ha! Looks like I broke this aeons ago: 7e6b7685b. How did I miss the hex conversion? It should definitely be fixed. I do hate the `printf` family, but it's the best solution we have unless/until we use `fmt`.

hainest (approved):

PR #1313, opened 2022-10-25
Manually add hex prefix when formatting an Operand

bbiiggppiigg:
I believe this is a misuse of snprintf format string.
Instead of outputting the numeric value in hex, it would output the value as dec followed by a literal x which makes no sense at all.

### DIFF
--- a/instructionAPI/src/Operand.C
+++ b/instructionAPI/src/Operand.C
@@ -134,7 +134,7 @@ namespace Dyninst
                 Result res = op_value->eval();
                 if (res.defined) {
                     char hex[20];
-                    snprintf(hex, 20, "%lux", res.convert<uintmax_t>());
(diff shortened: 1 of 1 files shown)


kupsch (commented):
Slight improvement would be to use "%#lx" instead of "0x%lx".

hainest:
> Slight improvement would be to use "%#lx" instead of "0x%lx".

I was going to recommend that, but it looks like it doesn't append the prefix when the value is zero. A quick grep found that other usages manually prepend the "0x". I don't know if it that is to handle the zero case or just from copy/paste.

kupsch:
If you want "0x0" then leave the code as is.  If just "0" is acceptable then use the '#' modifier.  I could see either case being acceptable.

PR #1314, opened 2022-11-02
Allow assignment conversion without framepointer

kupsch:
* When performing assignment conversion of an instruction and with stack frame analysis, the call would assert if the frame pointer (or stack pointer) was TOP.  Change to treat this as unknown (just as a bottom value is).


ajain365 (approved):

hainest (approved):

hainest:
Is there any way to have an explicit, reproducible test for this? Given that the test suite hasn't been failing, I'm assuming this wasn't encountered in those tests.

PR #1315, opened 2022-11-07
ParseAPI: improve tail call recognition

jmellorcrummey:
Don't consider a branch that is a self-loop as a potential tail call. 

Note: This commit avoids having function finalization never reach a fixed point for the bad parse of function PMPI_Accumulate in  libmpi_nvidia.so.12.0.0. However, this does nothing to fix the underlying parsing problems that led to the anomaly:

  ` 0x116d954 <PMPI_Accumulate+106900>:  jmp    0x116d954 <PMPI_Accumulate+106900>
`
which according to objdump should have been:

 ```
  0x116d951 <PMPI_Accumulate+106897>:  callq  0x2a270 <MPIR_Err_create_code@plt>
  0x116d956 <PMPI_Accumulate+106902>:  mov    %eax,%edi
```
Note that 0x116d954 is in the middle of the instruction at 0x116d951 reported by objdump.
@kupsch

### DIFF
--- a/parseAPI/src/Parser.C
+++ b/parseAPI/src/Parser.C
@@ -920,9 +920,11 @@ Parser::finalize(Function *f)
             b->copy_targets(targets);
             for (auto e : targets) {
                 if (!e->interproc() && (e->type() == INDIRECT || e->type() == DIRECT)) {
+		  if (b->last() != e->trg()->start()) { // if not an instruction that branches to itself
                     e->_type._interproc = true;
                     parsing_printf("from %lx to %lx, marked as tail call (jump at entry), re-finalize\n", b->last(), e->trg()->start());
                     return false;
+		  }
                 }
             }
         }


jmellorcrummey:
@hainest I don't understand why 13 libabigail checks would fail from adding an if statement in a function. I think that this is a sign that your CI is off the rails rather than my patch.

woodard:
looking at the patch, I have to agree with John here. It looks like there is an unrelated warning causing problems.

kupsch:
The problem here is that the .so version is incremented after release (@hainest can explain the rationale), and libabigail flags .so version number differences.  @woodard is there a way to disable this check?

hainest:
> @hainest I don't understand why 13 libabigail checks would fail from adding an if statement in a function. I think that this is a sign that your CI is off the rails rather than my patch.

Not off the rails- just broken. You can safely ignore the libabigail tests.

kupsch (approved):
This will fix this particular instance, but there are likely other ways to cause the fixed pointed analysis loop that are still under investigation.

hainest:
LGTM.  https://bottle.cs.wisc.edu/search?dyninst_branch=PR1315

hainest (approved):

PR #1323, opened 2022-11-14
fix Instruction class to allow valid assignment

### LINKED ISSUE #1312: InstructionAPI Instruction assignment operator won't update the formatter member
**Intention**
Decode Instructions and output their format using the formatter corresponding to the architecture it decodes from.
Check [disassemble example](https://github.com/dyninst/examples/blob/master/disassemble/disassemble.cpp#L29)

**Describe the bug**
Since the formatter is of reference type, it can't be updated a default Instruction is constructed.
Which means the update in

https://github.com/dyninst/dyninst/blob/eea540a94564371d3170d01ab16c62ccbc8f9b9d/instructionAPI/src/Instruction.C#L215

will take no effect. 
(This line will be skipped when single stepping through Instruction::operator=() in gdb)

**To Reproduce**
Run the disassemble example on non-x86-64 platforms and check what formatter is actually used when calling instr.format()

**Expected behavior**
The formatter that corresponds to the architecture of which the Instruction is decoded from should be used.

**Additional context**
1. While it is possible to change the disassemble example and move the Instruction instr declaration into the loop and combined it with the assignment, I'm not sure if we should make a restriction that Instruction object should always be initialized with the correct architecture.
2. What would be the right fix for this? Changing the member to a pointer or an object?

kupsch:
* Allow the Instruction class to allow correct assignment semantics by changing a member from a reference to a pointer

* The formatter member is a reference (to a singleton) that can not be changed after it is initialized.  Assignment to this member wrongly assigns to the previously bound referent singleton and wrongly slices the rhs to the ArchSpecificFormatter base class.  The base class has no data members, and since the assignment does not change the type of the object (its vtable), it appears to have no effect.  Also a default constructed Instuction object sets the formatter to the x86 formatter even if the host was not x86.

@bbiiggppiigg please test that this fixes your use case. Test suite shows no regressions on x86_64 and arm.

fixes #1312

### DIFF
--- a/instructionAPI/h/Instruction.h
+++ b/instructionAPI/h/Instruction.h
@@ -317,7 +317,8 @@ namespace Dyninst
       Architecture arch_decoded_from;
       mutable std::list<CFT> m_Successors;
       static int numInsnsAllocated;
-      ArchSpecificFormatter& formatter;
+      // formatter is a non-owning pointer to a singleton object
+      ArchSpecificFormatter* formatter;
     };
   }
 }
--- a/instructionAPI/src/Instruction.C
+++ b/instructionAPI/src/Instruction.C
@@ -85,7 +85,7 @@ namespace Dyninst
                 size_t size, const unsigned char* raw,
                 Dyninst::Architecture arch)
             : m_InsnOp(what), m_Valid(true), arch_decoded_from(arch),
-            formatter(ArchSpecificFormatter::getFormatter(arch))
+            formatter(&ArchSpecificFormatter::getFormatter(arch))
         {
             copyRaw(size, raw);
 
@@ -130,7 +130,7 @@ namespace Dyninst
         }
 
         INSTRUCTION_EXPORT Instruction::Instruction() :
-            m_Valid(false), m_size(0), arch_decoded_from(Arch_none), formatter(ArchSpecificFormatter::getFormatter(Arch_x86_64))
+            m_Valid(false), m_size(0), arch_decoded_from(Arch_none), formatter(nullptr)
         {
 #if defined(DEBUG_INSN_ALLOCATIONS)
             numInsnsAllocated++;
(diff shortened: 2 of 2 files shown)


hainest (commented):
My only concern here is that the following is no longer valid

```
Instruction i{};
auto &f = i.getFormatter();
```

That said, `getFormatter` is not in the documentation, so we might be able to get away with that. Buuuuut

```
Instruction i{};
auto f = i.format(0xdeadbeef);
```
is also now invalid and _is_ in the docs. The class design here is upside down, but we're stuck with it. That said, I'm not sure if allowing this case would be worth the penalty of a possible ABI break by switching to reference_wrapper. Thoughts?

kupsch:
This behavior of calling `i.format(addr)` using a default constructed `Instruction` happens to work on x86_64, but will not work on any other architectures.  I think we have a couple of choices if `arch_decoded_from == Arch_none` is true when format is called:

1. abort
2. throw an exception
3. return a string an error string "ERROR_NO_ARCHITECTURE_SET_FOR_INSTRUCTION"

If this functionality is used then the user should use the non-default constructor and specify the architecture.  If you are using the instruction decoder interface as recommended then this is done correctly.

Also `ArchSpecificFormatter::getFormatter(Architecture)` returns the x86Formatter as a default for unknown architectures (like Arch_none) which is probably not correct.

bbiiggppiigg:
@kupsch 
I've tested and this PR did fix my use case.

As for the case when architecture is not specified, I think option 3 seems reasonable as 
1. It's  probably not good to break existing user code, whether because of seg fault on nullptr or un caught exception.
2. By returning the error string as the format string of the Instruction the users will know how they should fix their code.

bbiiggppiigg (commented):
@kupsch 
I've tested and this PR did fix my use case.

As for the case when architecture is not specified, I think option 3 seems reasonable as 
1. It's  probably not good to break existing user code, whether because of seg fault on nullptr or un caught exception.
2. By returning the error string as the format string of the Instruction the users will know how they should fix their code.

bbiiggppiigg (approved):

PR #1326, opened 2022-11-16
Fix possible memory leak in Elf_X::findDebugFile

hainest:
This will conflict with #1325, so I'll rebase onto whichever gets merged first.


kupsch (changes requested):
filename is only valid (and must be freed) if fd >= 0, so the proposed change could result in undefined behavior since filename is undefined (null or random address). The unmodified code is correct without changes.

hainest:
Oh! I didn't catch that `fd` was the returned value. I thought it had come from above the `debuginfod_find_debuginfo` call. Derp. Thanks for double-checking.

PR #1332, opened 2022-12-09
use default allocator for tbb::concurrent_hash_map

kupsch:
Fixes warning produced by gcc 12: a non-default allocator for tbb::concurrent_hash_map causes a new/delete mismatch warning (that appears to be a false positive).  Warning produced since tbb headers are not marked as a system headers yet. 

In the future the allocator should be changed to use the tbb scalable allocator.

Related to #1331

### DIFF
--- a/common/h/concurrent.h
+++ b/common/h/concurrent.h
@@ -94,12 +94,10 @@ namespace concurrent {
 
 template<typename K, typename V>
 class dyn_c_hash_map : protected tbb::concurrent_hash_map<K, V,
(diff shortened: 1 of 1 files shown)


jrmadsen:
This would probably be silenced by adding SYSTEM to the target_include_directories for the TBB headers

kupsch:
Thanks, we already have that in the pipeline and it does silence the warning. We are also removing the default as it did nothing and for a future change to the scalable allocator.

hainest (approved):

PR #1339, opened 2022-12-14
Use OpenMP target

hainest:
This also provides a dummy target so we don't have to do any additional checking when USE_OpenMP=OFF. We only use OpenMP_CXX, so I didn't create a target for the other languages (C,Fortran).


kupsch (approved):

hainest:
No new regressions: https://bottle.cs.wisc.edu//search?dyninst_branch=PR1339

PR #1345, opened 2022-12-22
CMake modernization - platform detection

### LINKED ISSUE #707: Canonicalize platform support
There is a lot of code floating around in both Dyninst and the test suite for antiquated platforms like AIX, Sun, and BlueGene (see #617 for this one). FreeBSD seems to be supported, but it's never tested. We've also been getting Github issues for Mac and Windows. What about 32-bit platforms (linux-32 is dead)? Do we want to support building with MSVC (it hasn't been tested in years), or just require MinGW/[LSW](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and gcc/clang there?

1. What platforms do we officially support?
2. Which can we test on?
3. How should they be prioritized?

hainest:
Closes #707

This definitely needs a careful examination. I am fairly confident I didn't break anything and it tests fine, but a tiny typo could easily cause no issues.

Eventually, we'll want to clean up the internal macro checks to simplify platform detection, but that's beyond the scope of this work.


hainest:
No new regressions: https://bottle.cs.wisc.edu/search?dyninst_branch=PR1345

kupsch (changes requested):
looks good except one remaining use of PLATFORM that should be DYNINST_OS_i386, and two places that could be changed to DYNINST_OS_UNIX.  After fixing at least the first one, I approve.

PR #1346, opened 2022-12-23
CMake modernization - update options

hainest:


cmake/DyninstOptions.cmake
```diff
@@ -13,27 +13,19 @@ option(ENABLE_LTO "Enable Link-Time Optimization" OFF)
...
-    CACHE BOOL "Do not download/build any third-party dependencies from source")
+option(STERILE_BUILD "DEPRECATED -- Do not use" OFF)
 
-set(ADD_VALGRIND_ANNOTATIONS
-    OFF
-    CACHE BOOL "Enable annotations for Valgrind analysis")
+option(ADD_VALGRIND_ANNOTATIONS "Enable annotations for Valgrind analysis" OFF)
 
-set(ENABLE_STATIC_LIBS
-    NO
```
kupsch, line 25: OFF instead of NO for consistency?

kupsch (commented):
Looks good.  Consider using consistent boolean value OFF/ON.

cmake/DyninstOptions.cmake
hainest: Oops, sorry. I didn't mean to request a review on this. I agree we should use OFF instead of NO. I just copy/pasted.

PR #1350, opened 2023-01-09
Remove symlite

hainest:
The current structure of depending on symtab/symLite makes separating them in the build system cumbersome. There are several reasons to remove it:

- parseAPI requires symtab headers- even when LIGHTWEIGHT_SYMTAB=ON
- It wasn't buildable until this year
- its headers are exported, but '#include' files not in the install tree (i.e., it was always broken)
- More than 75% of the testsuite requires symtab

Given that it didn't build for a long time and its headers have never been usable, I'd say no one was using it.


bigtrak:
The last time we discussed keeping or discarding Symlite Bart wanted 
to have it remain in Dyninst.   I think you need to discuss this removal with Bart.

kupsch (approved):

hainest:
@bigtrak At that time, we didn't have a particularly good reason to remove it the code wasn't causing any trouble. With the new build system, that's no longer the case. We _could_ just leave it out of the CMake updates, but then it would just sit here and rot. I've found enough problems to fix in these CMake changes, I'd rather not spend much time on something that very likely isn't being used. All of this would be for the 13.0 release, so we have time to discuss with Bart once he gets back.

vsoch:
Off with its head!

PR #1351, opened 2023-01-10
Refactor common/src/Types.h

hainest:
The platform macro guards (e.g., `os_windows`) are causing issues in the CMake refactor. The simplest solution is to fix the root cause here. Moreover, the custom fixed-size types are unnecessary since we require C++11.

### DIFF
--- a/common/src/arch-x86.h
+++ b/common/src/arch-x86.h
@@ -1159,13 +1159,13 @@ inline bool is_disp16(long disp) {
 }
 
 inline bool is_disp32(long disp) {
-  return (disp <= I32_MAX && disp >= I32_MIN);
+  return (disp <= INT32_MAX && disp >= INT32_MIN);
 }
 inline bool is_disp32(Address a1, Address a2) {
   return is_disp32(a2 - (a1 + JUMP_REL32_SZ));
 }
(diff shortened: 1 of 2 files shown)


common/src/arch-x86.h
```diff
@@ -1159,13 +1159,13 @@ inline bool is_disp16(long disp) {
...
 inline bool is_disp32(long disp) {
-  return (disp <= I32_MAX && disp >= I32_MIN);
+  return (disp <= INT32_MAX && disp >= INT32_MIN);
 }
 inline bool is_disp32(Address a1, Address a2) {
   return is_disp32(a2 - (a1 + JUMP_REL32_SZ));
 }
 inline bool is_addr32(Address addr) {
-    return (addr < UI32_MAX);
+    return (addr < INT32_MAX);
```
kupsch, line 1168: `UINT32_MAX` to allow 32 bits instead of 31 bits.

kupsch (changes requested):
Looks good except the one comment to change INT32_MAX to UINT32_MAX

common/src/arch-x86.h
hainest: Good catch!

kupsch (approved):

PR #1353, opened 2023-01-25
Remove special global ctor/dtor search in ppc for static binaries

hainest:
Neither of the other platforms searches through .init to try to find a global constructor/destructor. Moreover, the gmon_start function has been removed in the latest glibc's.


kupsch (approved):
Looks good.  One minor additional cleanup would be to change the variable name `ctorFuncs` (as they are not ctors) to `dtorFuncs` in `findGlobalDestructorFunc`, or use the variable name `funcs` in both functions.

hainest:
> Looks good. One minor additional cleanup would be to change the variable name `ctorFuncs` (as they are not ctors) to `dtorFuncs` in `findGlobalDestructorFunc`, or use the variable name `funcs` in both functions.

Oooh, copy/paste, how would we write code without you? I'll fix those in the other update for these files.

hainest:
No regressions (the stupid dashboard stopped accepting uploads, so no links).

PR #1355, opened 2023-01-26
Update global ctor/dtor instrumentation for static executables

hainest:
glibc-2.34 removed __libc_csu_init(fini) which was used as instrumentation points for Dyninst's global constructor(destructor). We now insert the global Dyninst constructor into the beginning of 'main' and the global destructor into the beginning of 'exit'.


hainest:
This is just for x86. I'll update ARM and ppc once I test this.

kupsch (approved):

hainest:
Testing fine all all platforms.

kupsch (approved):

PR #1358, opened 2023-01-31
Remove AObject::pickLanguage

hainest:
This was only used by the STABs parsing code. It should have been removed by 5e142eff. Language detection is handled by `void Object::getModuleLanguageInfo` and requires DWARF (it does nothing on non-ELF binaries).


kupsch (approved):

hainest:
Building and testing with no issues.

PR #1359, opened 2023-02-03
Correctly check form value in DwarfWalker::findString

hainest:
dwarf_whatform will only return 0 if the attribute object has no form value.


hainest:
Superceded by #1360

PR #1360, opened 2023-02-03
Replace DwarfWalker::findString with DwarfWalker::find_call_file

hainest:
The original function was only ever used to find DW_AT_call_file. During the transition to libdw, the added check for that attribute made the rest of the string processing useless. Moreover, the code update also introduced a bug in checking the return value of dwarf_whatform which rendered the function completely useless as a generic stringform parser.

### DIFF
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -529,13 +529,13 @@ bool DwarfWalker::parseCallsite()
     if (!has_line)
         return true;
 
-    std::string inline_file;
-    bool result = findString(DW_AT_call_file, inline_file);
-    if (!result)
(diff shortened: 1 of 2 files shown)


symtabAPI/src/dwarfWalker.C
```diff
@@ -529,21 +529,21 @@ bool DwarfWalker::parseCallsite()
     if (!has_line)
         return true;
 
-    std::string inline_file;
-    bool result = findString(DW_AT_call_file, inline_file);
-    if (!result)
+    using opt_string = boost::optional<std::string>;
```
kupsch, line 532: delete this line and use auto on the next line?

kupsch (approved):
Looks good.  Might want to use auto to simply (see below).

PR #1361, opened 2023-02-03
Fixed nullptr issues in dyninstAPI/src/mapped_object.C

jrmadsen:
- checking `std::string::c_str() == 0` is pointless because `std::string` cannot be constructed from a `nullptr`
  - C++11 standard, sect. 21.4.2.9
  - Replaced with `std::string::empty()`
- add `nullptr` checks in `find.*Vector.*` functions
- Fixes issues noted here: https://github.com/AMDResearch/omnitrace/pull/235#issuecomment-1414264630


hainest:
No testing regressions.

Issue #1363, opened 2023-02-07
Relocating AVX instruction with PC relative addresses cause an assertion failure


Maetveis:
**Intention**

Instrumenting a library containing functions with avx instructions using pc relative addressing.

**Describe the bug**

This assertion fires when the edit is finalized (from `BPatch_binaryEdit::writeFile`):
```plaintext
dyninst/dyninstAPI/src/Relocation/Transformers/Movement-adhoc.C:341:
bool Dyninst::Relocation::adhocMovementTransformer::isPCRelData(Dyninst::Relocation::Widget::Ptr, Dyninst::InstructionAPI::Instruction, Dyninst::Address&):
Assertion `res.defined' failed.
```

**To Reproduce**

```console
❯ gcc -x c - EOF
void f() {
    asm("vpand 0x0(%rip),%xmm3,%xmm3");
}

int main() {}
EOF
❯ parseThat -i 1 ./a.out
```

**Expected behavior**

No assertion is fired, the instrumentation completes without errors.

**System
 - Platform x86_64, Ubuntu 20.04
 - Dyninst built from source using commit: 3937542bdc773b6f10740e0746e9c4fb7f8e02d6
 - libc version
```console
❯ dpkg -s libc6 | grep Version
Version: 2.31-0ubuntu9.9
```

PR #1364, opened 2023-02-08
Fix sema type for vex2 encoded vpand

Maetveis:
The vex encoded form of VPAND takes three operands dest,src1,src2.

Fixes: #1363


hainest:
Thanks for the fix. We are planning to transition our instruction decoding to a third-party library so things like this will be less common.

hainest:
No testing regressions.

PR #1371, opened 2023-02-15
Remove use of couts in the AMDGPU instruction decoders

bbiiggppiigg:
As title, remove the use of cout in the AMDGPU instruction decoders.
Debug information should be controlled by the debug_printfs, which is still missing for instructionAPI.


hainest (commented):
Is there a difference between the debugging statements removed and the ones commented out?

bbiiggppiigg:
> 

The ones commented out are the ones that I'll probably rewrite using the debug_printf for instructionAPI, once they are supported.

hainest (approved):

Issue #1372, opened 2023-02-16
x86 code not interpreted correctly on arm and powerpc


bigtrak:
The Rice group noticed that Dyninst wasn't correctly interpreting CFGs when 
an intel binary was analyzed on arm.

Briefly, due to differences in the signed-ed-ness of chars between platforms and
architectures, 8 bit immediate values were not being sign extended.  When this 
occurred with a branch instruction, it resulted in a stop in CFG parsing, as
branch destinations no longer made sense.     This issues also affects constant
values as well as branch offsets.

Here is the text of the original query:
------------------------------------------------
Are ParseAPI and InstructionAPI expected to be compatible across architectu=
res? That is, if I have a binary compiled for arch X, can I expect to use D=
yninst to generate the complete and correct CFG on arch Y?

We (the HPCToolkit folks) recently found a possible bug in Dyninst: ParseAP=
I running on ARM generates an incorrect CFG for a simple x86-64 binary, but=
 ParseAPI running on x86-64 produces the correct CFG. This caused our hpcst=
ruct tool to generate different results for the same binary, depending on t=
he host architecture. ([More detailed debugging by Mark Krentel in our issu=
e tracker][1])

We couldn't find documentation stating whether Dyninst officially supports =
cross-architecture binary analysis, so we decided to ask here first before =
submitting a proper bug report on GitHub.

[1]: https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/734

bigtrak:
Closed by PR#1373

PR #1373, opened 2023-02-16
Insure intel 8 bit immediate values are interpreted correctly on other systems.

bigtrak:
Intel 8 bit immediate values are signed.   They were being interpreted
as unsigned on power and arm due to char differences.   This PR
ensures 8 bit values related to x86 are interpreted correctly on other
systems.   This resolves issue #1372 which the Rice group had with
incorrect CFG interpretation cross-arch.   As well as other problems
such as normal immediate values being interpreted incorrectly.

### DIFF
--- a/common/src/arch-x86.C
+++ b/common/src/arch-x86.C
@@ -10490,7 +10490,7 @@ static unsigned int ia32_decode_modrm(const unsigned int addrSzAttr, const unsig
    addr++;
 
    /* Get displacements we're going to use */
-   const char* disp8 = (const char*)addr;
+   const signed char* disp8 = (const signed char*)addr;
    const short* disp16 = (const short*)addr;
    const int* disp32 = (const int*)addr;
 
@@ -10675,7 +10675,7 @@ static unsigned int ia32_decode_modrm(const unsigned int addrSzAttr, const unsig
       }
 
       /* Update displacement pointers  */
-      disp8 = (const char*)addr;
+      disp8 = (const signed char*)addr;
       disp16 = (const short*)addr;
       disp32 = (const int*)addr;
 
@@ -10953,7 +10953,7 @@ unsigned int ia32_decode_operands (const ia32_prefixes& pref,
                nib += wordSzB * addrSzAttr;
                if(mac)
                {
-                  int offset = 0;
+                  long offset = 0;
                   switch(addrSzAttr)
                   {
                      case 1: // 16-bit offset
@@ -11744,7 +11744,7 @@ int displacement(const unsigned char *instr, unsigned type) {
       disp = *(const int *)(instr+2);
    } else if (type & IS_JUMP) {
       if (type & REL_B) {
-         disp = *(const char *)(instr+1);
+         disp = *(const signed char *)(instr+1);
       } else if (type & REL_W) {
          disp = *(const short *)(instr+1); // skip opcode
       } else if (type & REL_D) {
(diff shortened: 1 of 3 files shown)


kupsch:
Please update the commit message to use the git convention that the first line is the title (try to keep it less than 70 characters) followed by a blank line and then a longer commit message.  This way when you do 'git log --oneline' it displays just the title (really up to the first blank line).  Currently this outputs your first paragraph.

You can update it using the following:

```
git commit --amend
git push -f
```

kupsch:
There are additional char/size problems that I found that can go in this commit.  You can edit file, `git add arch-x86.C`, `git commit --amend`,  and then `git push -f`.  Here is the diff:

```
  --- a/common/src/arch-x86.C
  +++ b/common/src/arch-x86.C
  @@ -10490,7 +10490,7 @@ static unsigned int ia32_decode_modrm(const unsigned int   addrSzAttr, const unsig
      addr++;
   
      /* Get displacements we're going to use */
  -   const char* disp8 = (const char*)addr;
  +   const signed char* disp8 = (const signed char*)addr;
      const short* disp16 = (const short*)addr;
      const int* disp32 = (const int*)addr;
   
  @@ -10675,7 +10675,7 @@ static unsigned int ia32_decode_modrm(const unsigned int   addrSzAttr, const unsig
         }
   
         /* Update displacement pointers  */
  -      disp8 = (const char*)addr;
  +      disp8 = (const signed char*)addr;
         disp16 = (const short*)addr;
         disp32 = (const int*)addr;
   
  @@ -10953,7 +10953,7 @@ unsigned int ia32_decode_operands (const ia32_prefixes&   pref,
                  nib += wordSzB * addrSzAttr;
                  if(mac)
                  {
  -                  int offset = 0;
  +                  long offset = 0;
                     switch(addrSzAttr)
                     {
                        case 1: // 16-bit offset
  @@ -11744,7 +11744,7 @@ int displacement(const unsigned char *instr, unsigned ty  pe) {
         disp = *(const int *)(instr+2);
      } else if (type & IS_JUMP) {
         if (type & REL_B) {
  -         disp = *(const char *)(instr+1);
  +         disp = *(const signed char *)(instr+1);
         } else if (type & REL_W) {
            disp = *(const short *)(instr+1); // skip opcode
         } else if (type & REL_D) {
  @@ -11752,7 +11752,7 @@ int displacement(const unsigned char *instr, unsigned ty  pe) {
         }
      } else if (type & IS_JCC) {
         if (type & REL_B) {
  -         disp = *(const char *)(instr+1);
  +         disp = *(const signed char *)(instr+1);
         } else if (type & REL_W) {
            disp = *(const short *)(instr+2); // skip two byte opcode
         } else if (type & REL_D) {
```

bigtrak:
> There are additional char/size problems that I found that can go
> in this commit.  You can edit file, `git add arch-x86.C`, `git commit
> --amend`,  and then `git push -f`.  Here is the diff:

Great.

Thank you, Jim.


What I'd really like to know is why those few lines didn't
show up in testing -- at all.   Wonderful.

Did you have  a test case where that showed up that I can
use as a repeater?   I checked the very instructions
that were affected and they were decoded correctly with
my changes.


My long-term plan on things like was that we should open
a ticket and make sure the md code uses a proper typedef
for the immediate values as a long term code cleanup issue,
instead of having all these hard-wired types in the code.

Then work on fixing the code where ad-hoc types are used.

They archs already have the types. and the code doesn't
use them uniformly.


Bolo

bigtrak:
> When the code requires a fixed size & signedness using the types
> defined in <cstdint> is the correct thing to do.  There is generally
> no good reason for a class to create its own set of aliases for these
> types; from a glance I know what a uint16_t is but have to look to see
> what a dword_t is. In my upcoming PR, I have changes many primitive C++
> int types to use the stdint types when I was modifying code near them,
> but I sure that this is only a small fraction of those that should
> be changed.

More along the lines of "this is an x86 8 bit immediate
value" typedef, which can be implemented either with the
current types or newer types such as you suggest.

If we had internal types, it would be clear what is being
accessed and why it is being accessed.   Especially if thre
are multiple stored values of the same width, depending upon
which instruction is used, such as signed/unsigned values.

Bolo

kupsch (approved):

PR #1374, opened 2023-02-20
CMake modernization - update compiler flags

hainest:
The options for setting the linker and stdlib are there to let us explore that during our testing. I don't anticipate users will try them.

The gymnastics of getting compiler flag overrides to work is heavily commented. Let me know if there is anything that's not clear on that.


cmake/DyninstOptimization.cmake
```diff
@@ -79,12 +80,17 @@ else()
...
-  set(CMAKE_CXX_FLAGS_${bt} "${_${bt}} ${DYNINST_CXX_FLAGS}")
-  unset(_${bt})
+# Set the CMake globals, preserving user-specified values in CMAKE_<LANG>_FLAGS
+string(TOUPPER ${CMAKE_BUILD_TYPE} _build_type)
+foreach(l "C" "CXX")
+  if(CMAKE_${l}_FLAGS)
+    set(_p ${CMAKE_${l}_FLAGS})
+  endif()
+  set(CMAKE_${l}_FLAGS_${_build_type} "${_${_build_type}} ${DYNINST_${l}_FLAGS} ${_p}")
+  set(CMAKE_${l}_FLAGS "" CACHE STRING "" FORCE)
```
jrmadsen: This should be a local variable scoped to the Dyninst project. Forcing those as cache variables unnecessarily propagates them outside of the Dyninst build when Dyninst is a submodule.

jrmadsen: Also, there's no need to append the global `CMAKE_<LANG>_FLAGS` to the build type flags. The non-build type suffixed flags are applied regardless of build type and _in addition to the build-type flags_

jrmadsen: Oh disregard the last comment, I just read the comments. The first one is still valid though. But, why resort to this trickery? Won't converting `CMAKE_<LANG>_FLAGS` into a list and add them to the target_compile_options?

jrmadsen: Another alternative, set the compile flags property to the language flags:

https://cmake.org/cmake/help/latest/prop_sf/COMPILE_FLAGS.html

hainest: I tried all of this and more. The underlying problem is that CMake both populates and inserts `CMAKE_<LANG>_FLAGS_<BUILD_TYPE>` unconditionally. After a lot of research, the only way I could find to disable it is to overwrite it. That was satisfactory given that we were previously overwriting `CMAKE_<LANG>_FLAGS` which is worse. However, I didn't consider being consumed as a submodule. Here's the specific example I was using:

```cmake
set(MY_CXX_FLAGS -DMYDEF ${CMAKE_CXX_FLAGS}) 
add_executable(test test.cpp)
target_compile_options(test PRIVATE $<$<CONFIG:RELWITHDEBINFO>:${MY_CXX_FLAGS}>)
```

```terminal
$ cmake .. -DCMAKE_CXX_FLAGS="-DUSERDEF" -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

generates

```terminal
$ /usr/bin/c++ -DUSERDEF -O2 -g -DNDEBUG -DMYDEF -DUSERDEF ...
```

Dyninst can't currently work correctly with NDEBUG, so we don't want that. Of course, we could add `-UNDEBUG` to `MY_CXX_FLAGS`, but then we're chasing CMake's defaults for the rest of time. Maybe we don't have a choice.

I also just realized we're explicitly setting `CMAKE_BUILD_TYPE` in DyninstLibrarySettings.cmake, so I should update that, as well.

jrmadsen: Try this:

```cmake
if(NOT CMAKE_CONFIGURATION_TYPES)
    set(CMAKE_CONFIGURATION_TYPES "Release" "RelWithDebInfo" "MinSizeRel" "Debug")
endif()

foreach(_LANG C CXX)
    foreach(_CONFIG ${CMAKE_CONFIGURATION_TYPES})
        string(TOUPPER CMAKE_${_LANG}_FLAGS_${_CONFIG} _VAR)
        string(REGEX REPLACE "( |^).DNDEBUG( |$)" " " ${_VAR} "${${_VAR}}")
    endforeach()
endforeach()
```

Here's the full `CMakeLists.txt`:

```cmake

cmake_minimum_required(VERSION 3.15 FATAL_ERROR)

project(DyninstTest LANGUAGES C CXX)

if(NOT CMAKE_CONFIGURATION_TYPES)
    set(CMAKE_CONFIGURATION_TYPES "Release" "RelWithDebInfo" "MinSizeRel" "Debug")
endif()

foreach(_LANG C CXX)
    foreach(_CONFIG ${CMAKE_CONFIGURATION_TYPES})
        string(TOUPPER CMAKE_${_LANG}_FLAGS_${_CONFIG} _VAR)
        string(REGEX REPLACE "( |^).DNDEBUG( |$)" " " ${_VAR} "${${_VAR}}")
    endforeach()
endforeach()

if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Build type" FORCE)
endif()

set(MY_CXX_FLAGS -DMYDEF ${CMAKE_CXX_FLAGS})
add_executable(test test.cpp)
target_compile_options(test PRIVATE $<$<CONFIG:RELWITHDEBINFO>:${MY_CXX_FLAGS}>)
```

And the result without `-DNDEBUG`:

```console
$ cmake -B build -DCMAKE_CXX_FLAGS="-DUSERDEF" . -G Ninja && cmake --build build --target all -- -v
-- The C compiler identification is GNU 11.1.0
-- The CXX compiler identification is GNU 11.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/jrmadsen/devel/c++/AMDResearch/omnitrace/build-omnitrace/scratch/dyninst-test/build
[1/2] /usr/bin/g++   -DUSERDEF -O2 -g  -DMYDEF -DUSERDEF -MD -MT CMakeFiles/test.dir/test.cpp.o -MF CMakeFiles/test.dir/test.cpp.o.d -o CMakeFiles/test.dir/test.cpp.o -c ../test.cpp
[2/2] : && /usr/bin/g++ -DUSERDEF -O2 -g  CMakeFiles/test.dir/test.cpp.o -o test   && :
```

hainest: I just added -UNDEBUG.

kupsch (approved):
Looks good.  Consider also changing option ENABLE_LTO to DYNINST_ENABLE_LTO

hainest:
> Looks good. Consider also changing option ENABLE_LTO to DYNINST_ENABLE_LTO

That's been there a while. Do you think anyone's using it?

kupsch:
Unlikely that anyone is using it.  I think that we could just change it.  I'd be more leery of changing other non-DYNINST_ options.

If you want to be safe, you could change it and leave the old option (mark as deprecated), and then warn or error if it is set, and remove it in a future release.

PR #1382, opened 2023-02-22
Rename AMDGPU cdna2 architecture to gfx90a and update its decoder implementation

bbiiggppiigg:
As title.


bbiiggppiigg:
> I would suggest splitting this into two PRs: one to rename and one to update the implementation. I'll leave it to you to decide if that's something you want to do.

Should I close this pull request and open two additional ones in order to achieve the splitting?

hainest:
> > I would suggest splitting this into two PRs: one to rename and one to update the implementation. I'll leave it to you to decide if that's something you want to do.
> 
> Should I close this pull request and open two additional ones in order to achieve the splitting?

You can re-use this one for part of it, or you can make two new ones.

instructionAPI/src/AMDGPU/gfx90a/amdgpu_gfx90a_decoder_impl.C
```diff
@@ -1635,7 +1635,7 @@ void InstructionDecoder_amdgpu_gfx90a::decodeENC_SOP1(){
 	layout.OP = longfield<8,15>(insn_long);
 	layout.SDST = longfield<16,22>(insn_long);
 	layout.SSRC0 = longfield<0,7>(insn_long);
-	assert( layout.OP >= 0 && layout.OP < sizeof(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table) / sizeof(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table[0]) && "Opcode over or underflow");
+	assert( layout.OP >= 0 && layout.OP <= (sizeof(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table) / sizeof(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table[0])-1) && "Opcode over or underflow");
```
kupsch, line 1638: Consider simplifying the array bound check by defining isArrayIndexValid in a common header file (@hainest, suggestion of file to contain this function?), and changing the assert as shown below.

```
#include <cstddef>    // may not be required if size_t is defined by another include file

// define in namespace Dyninst if necessary


// Check if the index (2nd arg) is valid for the array (1st arg)
template <typename ArrayType, std::size_t n, typename IndexType>
constexpr bool isArrayIndexValid(ArrayType (&)[n], const IndexType& i)
{
    return 0 <= i && i <= (n - 1);
}
```

Then you can replace the assert with

```
assert(isArrayIndexValid(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table, layout.OP)
                  && "Opcode over or underflow")
```

Then you can replace all the similar asserts in the fashion.

After this is committed, then a separate PR can be made to change the small number of similar asserts in the other instruction decoders.

PR #1390, opened 2023-02-28
Avoid possible copy in BPatch::createEnum

hainest:
The Type class doesn't have a user-defined copy constructor and its implicit declaration is deprecated in C++20. Rearranging the construction here removes the possibility of copying.


kupsch (approved):
Looks good and warning is gone.

hainest:
No testing issues.

PR #1392, opened 2023-03-01
Restore PPC special-purpose registers in opposite order of saving them

hainest:
This is the same fix as for aarch64 (https://github.com/dyninst/dyninst/pull/1000), but for PowerPC. x86 does not have this issue.


kupsch (approved):

hainest:
No testing regressions.

PR #1397, opened 2023-03-14
enable compiler warnings in dataflowAPI

### LINKED ISSUE #1138: datflowAPI: switch fall throughs
There are some switch statements that fall through in https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/ExpressionConversionVisitor.C#L60. I'm not sure why the compiler warnings are picking these up. Maybe the flags are getting trampled when building the dataflowAPI toolkit?

@kupsch These don't need to be fixed before 12.0. This is just here for a reminder.

### LINKED ISSUE #1396: enable warnings for dataflowAPI source files
**Intention**
Enable compiler warnings in all of the Dyninst source code that isn't a third party library.

**Describe the bug**
The dataflowAPI directory is built as part of the parseAPI library.  Currently all the files in the dataflowAPI is built with all compiler warnings disabled. The dataflowAPI consists of dataflowAPI/src and dataflowAPI/rose (a third party library).

**Expected behavior**
The files in dataflowAPI/src should be built with the same compiler warnings enabled as the rest of Dyninst and should build cleanly.

kupsch:
Enable warnings in source files contained in dataflowAPI/src.  Source files in dataflowAPI/rose continue to have warnings disabled.

Fix the code to eliminate warning produced by these files.  Now builds cleanly with gcc 6-13 and clang 7-15, and C++ versions 11, 14, 17, 20 and 23 with the exception of clang and C++20 and C++23 (due to an error with `operator==` in dataflowAPI/rose that might be fixed by C++ DR).

Also fixed infinite recursion in SgAsmBinaryMultiply::get_type() as it did not call the base class's get_type in a sane way.

Fixes #1396
Fixes #1138

### DIFF
--- a/dataflowAPI/rose/util/PoolAllocator.h
+++ b/dataflowAPI/rose/util/PoolAllocator.h
@@ -108,7 +108,7 @@ private:
         const Chunk *chunk;
         size_t nUsed;
         ChunkInfo(): chunk(NULL), nUsed(0) {}
-        ChunkInfo(const Chunk *chunk, size_t nUsed): chunk(chunk), nUsed(nUsed) {}
+        ChunkInfo(const Chunk *chunk_, size_t nUsed_): chunk(chunk_), nUsed(nUsed_) {}
         bool operator==(const ChunkInfo &other) const {
             return chunk==other.chunk && nUsed==other.nUsed;
         }
@@ -192,7 +192,7 @@ private:
             if (!freeLists_[freeListIdx]) {
                 Chunk *chunk = new Chunk;
                 freeLists_[freeListIdx] = chunk->fill(cellSize_);
-                SAWYER_THREAD_TRAITS::LockGuard lock(chunkMutex_);
+                SAWYER_THREAD_TRAITS::LockGuard chunkLock(chunkMutex_);
                 chunks_.push_back(chunk);
             }
             ASSERT_not_null(freeLists_[freeListIdx]);
--- a/dataflowAPI/rose/util/PoolAllocator.h
+++ b/dataflowAPI/rose/util/PoolAllocator.h
@@ -180,6 +180,8 @@ private:
                 delete *ci;
         }
 
+DYNINST_DIAGNOSTIC_BEGIN_SUPPRESS_UNUSED_VARIABLE
+
         bool isEmpty() const {
             SAWYER_THREAD_TRAITS::LockGuard lock(chunkMutex_);
             return chunks_.empty();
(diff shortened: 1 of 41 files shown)


dataflowAPI/src/stackanalysis.C
```diff
@@ -588,7 +588,7 @@ void StackAnalysis::summarize() {
       Block *block = bIter->first;
       for (auto aIter = (*intervals_)[block].begin();
          aIter != (*intervals_)[block].end(); aIter++) {
-         Address addr = aIter->first;
+         //Address addr = aIter->first;
```
hainest, line 591: Any reason to not just delete these?

dataflowAPI/src/RoseImpl.C
```diff
@@ -838,7 +838,7 @@ SgAsmBinaryMultiply::SgAsmBinaryMultiply(SgAsmExpression *lhs, SgAsmExpression *
 }
 
 SgAsmType *SgAsmBinaryMultiply::get_type() const {
-    return ((SgAsmBinaryExpression *) this)->get_type();
+    return static_cast<const SgAsmBinaryExpression *>(this)->get_type();
```
hainest, line 841: I'm not disagreeing with the change here, but should we actually fix this and use the type-safe `return SgAsmBinaryExpression::get_type();`?

dataflowAPI/rose/util/PoolAllocator.h
```diff
@@ -192,7 +192,7 @@ class PoolAllocatorBase {
             if (!freeLists_[freeListIdx]) {
                 Chunk *chunk = new Chunk;
                 freeLists_[freeListIdx] = chunk->fill(cellSize_);
-                SAWYER_THREAD_TRAITS::LockGuard lock(chunkMutex_);
+                SAWYER_THREAD_TRAITS::LockGuard chunkLock(chunkMutex_);
                 chunks_.push_back(chunk);
```
hainest, line 198: Is the rename lock -> chunkLock intentional?

kupsch: Intentional to eliminate shadowed name

dataflowAPI/src/stackanalysis.C
kupsch: left these for documentation of the unused first and second members

dataflowAPI/src/RoseImpl.C
kupsch: The code as written results in infinite recursion

kupsch: I will rewrite this and SgAsmBinaryAdd::get_type to call `SgAsmBinaryExpression::get_type()`

PR #1402, opened 2023-03-23
Fix UB in dwarf/dwarf_subrange.cpp::get_type

hainest:
The dwarf_* functions return a pointer to their last argument. This causes 'get_type' to return a pointer to a local variable.


kupsch (approved):
I also wrote and tested this exact change and it corrects the problem.

PR #1403, opened 2023-03-30
remove unneeded zero length files

kupsch:
- two old commits (accidentally?) committed zero length files that seem to serve no purpose


hainest (approved):
Some of these are removed in #1391, but it's good to have them as a separate PR here.

PR #1404, opened 2023-03-30
Rename Architecture AMDGPU_CDNA2 to AMDGPU_GFX90A

bbiiggppiigg:
Rename the Architecture Arch_amdgpu_cdan2 to Arch_amdgpu_gf90a, as recommended by the vendor.

This is an effort to split implementation update and architecture renaming into two separate PRs,
as suggested in #1382.


hainest (approved):
LGTM. If the CI builds work, I'm fine to merge.

hainest:
@kupsch You had comments on the original. Did you want any changes here?

bbiiggppiigg:
I believe those are related to updating implementation details, but I'll wait his response.

kupsch:
My only comment was to change assert with the bounds check to use a template function to factor out the range check.  There were 2-3 of these in other instruction parsers, but now there are many more.  I can make this change in all the files after this merged.

Consider simplifying the array bound check by defining isArrayIndexValid in a common header file, and changing the assert as shown below.

```
#include <cstddef>    // may not be required if size_t is defined by another include file

// define in namespace Dyninst if necessary


// Check if the index (2nd arg) is valid for the array (1st arg)
template <typename ArrayType, std::size_t n, typename IndexType>
constexpr bool isArrayIndexValid(ArrayType (&)[n], const IndexType& i)
{
    return 0 <= i && I < n;
}
```

Then you can replace the assert (and all similar)

```
assert( layout.OP >= 0 && layout.OP < sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table) / sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table[0]) && "Opcode over or underflow");
```

with

```
assert(isArrayIndexValid(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table, layout.OP)
                  && "Opcode over or underflow")
```
.

bbiiggppiigg:
> My only comment was to change assert with the bounds check to use a template function to factor out the range check. There were 2-3 of these in other instruction parsers, but now there are many more. I can make this change in all the files after this merged.
> 
> Consider simplifying the array bound check by defining isArrayIndexValid in a common header file, and changing the assert as shown below.
> 
> ```
> #include <cstddef>    // may not be required if size_t is defined by another include file
> 
> // define in namespace Dyninst if necessary
> 
> 
> // Check if the index (2nd arg) is valid for the array (1st arg)
> template <typename ArrayType, std::size_t n, typename IndexType>
> constexpr bool isArrayIndexValid(ArrayType (&)[n], const IndexType& i)
> {
>     return 0 <= i && I < n;
> }
> ```
> 
> Then you can replace the assert (and all similar)
> 
> ```
> assert( layout.OP >= 0 && layout.OP < sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table) / sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table[0]) && "Opcode over or underflow");
> ```
> 
> with
> 
> ```
> assert(isArrayIndexValid(amdgpu_gfx90a_insn_entry::ENC_SOP1_insn_table, layout.OP)
>                   && "Opcode over or underflow")
> ```
> 
> .

I'm planning to make another PR following this one dedicated to update the implementations, where these changes will be integrated.

kupsch:
If the opcode array contains 256 elements and the layout.OP is a uint8_t then this code will produce Wtautological-comparison warnings with clang unless the 2nd conjunctive term is replaced with a `<=` and a `- 1`, so if we want no clang warnings maybe the change needs to be made now.

```
layout.OP <= sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table) / sizeof(amdgpu_gfx90a_insn_entry::ENC_SOPC_insn_table[0]) - 1
```

the template function does not warn for `0 <= i && i < n` for some reason even though they are essentially the same.

bbiiggppiigg:
Cause Tim's comments about the original PR is to separate updates of implementation from name changes.
I don't believe that this PR is going to introduce new compiler warnings ( as this PR is basically a find-and-replace ).

If you guys think it is better to include these changes in this PR I'll make a commit for this.

kupsch:
The expression in the asserts do not produce a warning as the tables contain less than 256 elements.  So the suggested change could be made later.

hainest:
Sounds like those changes can go in the second PR. Merging here. Thanks!

PR #1405, opened 2023-04-03
Fix link bug in insnCodeGen::loadImmIntoReg on aarch64

hainest:
The function template has a separate declaration and "definition" AND is called from translation units where neither is located. This only happens to work correctly when the TUs are linked in a specific order.

### DIFF
--- a/dyninstAPI/src/inst-aarch64.C
+++ b/dyninstAPI/src/inst-aarch64.C
@@ -681,7 +681,7 @@ Register EmitterAARCH64::emitCall(opCode op,
 
         insnCodeGen::generateMemAccess(gen, insnCodeGen::Load, scratch, scratch, 0, 8, insnCodeGen::Offset);
     } else {
-        insnCodeGen::loadImmIntoReg<Address>(gen, scratch, callee->addr());
+        insnCodeGen::loadImmIntoReg(gen, scratch, callee->addr());
     }
 
     instruction branchInsn;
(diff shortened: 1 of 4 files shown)


hainest:
@kupsch This is now building on Zeroah. Now running test suite.

dyninstAPI/src/inst-aarch64.C
```diff
@@ -884,14 +884,14 @@ void emitASload(const BPatch_addrSpec_NP *as, Register dest, int stackShift,
...
-	    insnCodeGen::loadImmIntoReg<long int>(gen, dest, imm);
+	    insnCodeGen::loadImmIntoReg(gen, dest, static_cast<Address>(imm));
 	    return;
 	}
 	else {
 	    restoreGPRtoGPR(gen, ra, dest);
 	}
     } else {
-        insnCodeGen::loadImmIntoReg<long int>(gen, dest, 0);
+        insnCodeGen::loadImmIntoReg(gen, dest, static_cast<Address>(0));
```
kupsch, line 894: unnecessary cast

kupsch (approved):
Looks good, only suggestion is to eliminate the unnecessary cast.

dyninstAPI/src/inst-aarch64.C
hainest: Agreed, but I was trying to make it more obvious what was happening at the callsite since the template parameter _should_ have been an `Address`. I guess I could have used `Address{0}`, but I'm going to say it's still better than what was there.

PR #1406, opened 2023-04-05
fix tbb detection

spotaws:
The FindTBB.cmake module does not find current TBB (after 2021.1.1) because it is trying to find version defines in tbb/version.h. tbb/version.h still exists, but only as a forward to the new oneapi/tbb/version.h, which does have the right defines in it.

This change adjusts the logic in FindTBB.cmake such that:

It looks first for tbb/tbb_stddef.h (location of defines before 2020.1.1), then oneapi/tbb/version.h (current location of defines), and finally, tbb/version.h (home of defines between 2020.1.1 and 2021.1.1). It searches in that specific order (and break()s out of the foreach loop if it finds one) because that _should_ work in all known scenarios.

### DIFF
--- a/cmake/Modules/FindTBB.cmake
+++ b/cmake/Modules/FindTBB.cmake
@@ -153,11 +153,16 @@ find_path(
 #
 if(TBB_INCLUDE_DIRS)
     # Starting in 2020.1.1, tbb_stddef.h is replaced by version.h
+    # As of 2021.1.1, the old tbb/version.h no longer has the right defines.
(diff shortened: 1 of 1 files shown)


hainest:
Thanks for the code contribution. However, we'll soon be requiring consumption of TBB through their CMake configs (TBBConfig.cmake). This is implemented in #1391 (see https://github.com/dyninst/dyninst/pull/1391/files#diff-895c3f35dfb1c1b85992f5183ae7559ea9609301571430a2c9682086e8a682f9 for details) and will be in the next release of Dyninst.

PR #1412, opened 2023-04-20
Add CI workflow to build consumers of Dyninst

hainest:
This was previously https://github.com/dyninst/external-tests/pull/12, but I think it would fit better in this repo.

### DIFF
--- a/.github/workflows/consumers.yaml
+++ b/.github/workflows/consumers.yaml
@@ -1,0 +1,35 @@
+# Build the latest versions of applications that consume Dyninst
+
+name: Build Consumers
+
+on:
+  pull_request:
+    branches:
+      - master
+#  schedule:
+#    - cron: '0 2 3 * *' # Every 3rd of the month
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        consumer: [
+          "hpctoolkit"
+          # must+stackwalker  - python is broken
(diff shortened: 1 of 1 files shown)


kupsch (approved):

.github/workflows/consumers.yaml
```diff
@@ -0,0 +1,32 @@
...
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        consumer: [
+          "hpctoolkit"
+          # must+stackwalker  - python is broken
+          # stat              - takes >6hrs
+          # extrae+dyninst    - not yet tested, may not support dyninst >10.0.0
+          # omnitrace         - Needs updated cmake
```
jrmadsen, line 19: What does `Needs updated cmake` mean? It may be missing some of the latest versions but that is fine for your use case I think. Or you could potentially target `omnitrace@main`.

hainest: It was my understanding that omnitrace and timemory wwere using a custom fork of Dyninst, so testing them here wouldn't necessary reflect changes made in this repo.

jrmadsen: Omnitrace has a custom submodule for when it is built for packaging but it fully supports using an external Dyninst installation and the spack recipe for omnitrace is written such that it always relies on the Dyninst installed via spack. 

Don't worry about timemory.

PR #1416, opened 2023-04-27
Make InstructionAPI::isArrayIndexValid const

hainest:
clang complains:

  'constexpr' non-static member function will not be implicitly 'const'
  in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]


kupsch (approved):
This is OK and the simplest change. Eventually these two identical methods in different classes should be a stand-alone function in a utility library.

@bbiiggppiigg you will need to update your generator program to reflect this change so this problem will not reintroduced if you regenerate these files.

kupsch:
If this is going to stay a member function, it would be better to make it a static member function instead of a const member function.

PR #1421, opened 2023-04-27
Fix x86 misspelled, incorrect, and invalid opcodes

hainest:
These were found when comparing against Capstone. There are a few breakages in here. For example, 1dcab37d renames `e_cmpsd_sse` to `e_cmpsd`, consistent with the rest of the SEE opcodes.


kupsch (approved):

hainest:
@mwkrentel No, these are bugs I found while comparing against Capstone. The vpermpd issue is much more involved and will require changes to the decoder implementation.

PR #1427, opened 2023-05-09
AddressTranslateSysV::adjustForAddrSpaceWrap: Fix C file descriptor leak

hainest:
Found using cppcheck:
    
    common/src/addrtranslate-sysv.C:1046:58: error: Resource leak: fd [resourceLeak]
       if (read(fd, &e_hdr, sizeof(e_hdr)) != sizeof(e_hdr)) return base;
                                                             ^
    common/src/addrtranslate-sysv.C:1048:28: error: Resource leak: fd [resourceLeak]
       if (e_hdr.e_phoff == 0) return base;


kupsch (approved):
This is good.  Consider using `goto done;` with the `done` label before the final `close` at the end of the function, or you could use a `std::unique_ptr` with a deleter that closes the file descriptor and get rid of all the `close` calls.

PR #1431, opened 2023-05-10
Fix uninitialized member variables

hainest:
These were detected by cppcheck's uninitMemberVar*. No attempt was made to be consistent in solving each item. Some are directly initialized in the class declaration, others have modified constructor initializer lists.

@kupsch I'll take a careful look over these once the CI runs to see if there are any enum members that should have better defaults.


hainest:
@kupsch I've verified these and found no needed changes.

PR #1436, opened 2023-05-11
BPatch_addressSpace::deleteSnippet: fix enumeral conversion in conditional

hainest:
This was picked up by gcc with -Wextra, but not by clang.


hainest:
Folded into #1434

PR #1438, opened 2023-06-10
Include CheckCXXSourceCompiles for check_cxx_source_compiles

opoplawski:
Building 12.3.0 with cmake 3.27 fails with:
```
-- Performing Test HAS_C_FLAG_Wwrite_strings - Success
CMake Error at cmake/warnings.cmake:133 (check_cxx_source_compiles):
  Unknown CMake command "check_cxx_source_compiles".
Call Stack (most recent call first):
  cmake/shared.cmake:106 (include)
  CMakeLists.txt:43 (include)
```
This should fix that on the master branch.

### DIFF
--- a/cmake/DyninstWarnings.cmake
+++ b/cmake/DyninstWarnings.cmake
@@ -116,6 +116,7 @@ endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "^(GNU|Clang)$")
   include(CheckCXXCompilerFlag)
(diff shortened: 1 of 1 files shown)


hainest:
Thanks for the fix! I have a container that builds against all CMake we support, but it only goes to 3.24...

opoplawski:
Well, maybe this was premature.  Builds are now working again without this change with cmake 3.27.0-rc2.

PR #1443, opened 2023-07-10
add header files to directly define std:: syms

### LINKED ISSUE #1442: FTBFS git-master on fedora rawhide f39 with gcc 13.1.1
instructionAPI/src/InstructionDecoder-aarch64 and instructionAPI/src/InstructionDecoder-amdgpu-vega fail to build, with same reasons as issue #1378.  The corresponding header files need a

```#include <array>```

kupsch:
- update dyninst header files to directly include the standard header file defining symbols from the standard C++ library that are used by the dyninst header file; in some instances, the code relied on symbols being defined via an unrelated include file indirectly including the necessary header file leading to fragile code

- minor other cleanups:  remove unnecessary header files, remove definitions of names that are defined in a standard header file

Fixes #1442

### DIFF
--- a/common/h/Annotatable.h
+++ b/common/h/Annotatable.h
@@ -35,6 +35,7 @@
 #define DYN_DETAIL_BOOST_NO_INTRINSIC_WCHAR_T 1
 #endif
 #include "dyntypes.h"
+#include <stddef.h>
 #include <vector>
 #include <map>
 #include <typeinfo>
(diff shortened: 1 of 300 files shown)


hainest (approved):
LGTM, assuming the most recent commit builds.

kupsch:
@bbiiggppiigg you will need to update your script to include the changes made by this commit to the AMD GPU files so this change is not reverted in a future commit.

PR #1444, opened 2023-07-11
Fix EM_AMDGPU to use RELA as Region Type

bbiiggppiigg:
According to the llvm page
https://llvm.org/docs/AMDGPUUsage.html#relocation-records AMDGPU backend generates Elf64_Rela records.

This fix added the case for EM_AMDGPU to
static Region::RegionType getRelTypeByElfMachine
such that it uses RELA instead of the default REL type.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -281,6 +281,7 @@ static Region::RegionType getRelTypeByElfMachine(Elf_X *localHdr) {
         case EM_X86_64:
         case EM_IA_64:
         case EM_AARCH64:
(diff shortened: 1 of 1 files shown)


kupsch:
Since there is a hard coded table this update seems reasonable. In general there could be both types in the same ELF, but in practice there is only one for all instances of the architecture. The type of the relocation is specified by the type of the dynamic entry and the type of the section (if present, generally is present but isn't required).

hainest:
@bbiiggppiigg Are there more changes that need to go into this? If not, I'll merge.

hainest (approved):

PR #1451, opened 2023-07-14
Clean up Symtab constructors

hainest:
This fixes several bugs and makes the constructors now behave all the same.


kupsch (approved):
the bool members could also be 0 initialized `m{}` instead of `m{false}` for consistency, but being explicit is OK.

hainest:
> the bool members could also be 0 initialized `m{}` instead of `m{false}` for consistency, but being explicit is OK.

I considered doing that, but thought the explicit might be clearer in the long run.

PR #1452, opened 2023-07-19
Make Symtab::getContainingInlinedFunction lazy parsing threadsafe

### LINKED ISSUE #1446: Locking around Symtab::mod_lookup
@mwkrentel @jmellorcrummey asked about need of locking in `Symtab::findModuleByOffset`. It looks to be needed because `Symtab::mod_lookup` lazily assigns the `mod_lookup_` member which is an `IBSTree<ModRange>`. Constructing an `IBSTree` appears to be inexpensive: IBSTree.h:258 because an `IBSNode<T>` just has a bunch of primitives and an enum. I see no reason why `mod_lookup_` couldn't be initialized in the constructor and its existence be a class invariant.

It also looks like the same could be done with `func_lookup`. We should survey the whole class.

@kupsch @bigtrak

hainest:
This fix does not address the thread safety issues in Symtab::parseFunctionRanges. That function is still thread unsafe and will require separate modifications.

@jmellorcrummey @mwkrentel This should allow you to remove the need to explicitly call `parseFunctionRanges` before calling `findModuleByOffset`. It does not fix #1446. That will be a separate PR.

### DIFF
--- a/symtabAPI/h/Symtab.h
+++ b/symtabAPI/h/Symtab.h
@@ -38,6 +38,7 @@
 #include <string>
 #include <vector>
 #include <set>
+#include <mutex>
 
 #include "Symbol.h"
 #include "Module.h"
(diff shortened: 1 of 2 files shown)


hainest:
@jmellorcrummey @mwkrentel I'm going to merge this because I need it for the next set of fixes for #1446.

PR #1454, opened 2023-07-22
Remove lazy initialization of IBSTrees in Symtab

hainest:
The IBSTree type is threadsafe, but the lazy initialization of them is not. Generally, we don't expect users to use these directly in a multithreaded context (e.g., via `mod_lookup()`). However, it is possible for users to indirectly access them through `getContainingInlinedFunction` or `findModuleByOffset`. At this time, an empty `IBSTree` occupies about 300 bytes of the total ~8.5kB of a `Symtab` object.

This works toward fixing #1446.


hainest:
No testing regressions.

PR #1456, opened 2023-07-24
modernize InstructionAPI::Operand class

kupsch:
- eliminate all but one of the constructors by using default parameters

- remove destructor (unnecessarily virtual, and reset a member shared_ptr), copy ctor, and copy assignment declarations to allow the compiler to generate default versions that are equivalent or better than the hand written versions

- add default member initializers for all members

### DIFF
--- a/instructionAPI/h/Operand.h
+++ b/instructionAPI/h/Operand.h
@@ -60,44 +60,15 @@ namespace Dyninst
     {
     public:
         typedef boost::shared_ptr<Operand> Ptr;
-    Operand() noexcept : m_isRead(false), m_isWritten(false), m_isImplicit(false), m_isTruePredicate(false), m_isFalsePredicate(false) {}
       /// \brief Create an operand from a %Expression and flags describing whether the %ValueComputation
       /// is read, written or both.
       /// \param val Reference-counted pointer to the %Expression that will be contained in the %Operand being constructed
       /// \param read True if this operand is read
       /// \param written True if this operand is written
-      Operand(Expression::Ptr val, bool read, bool written) : 
-          op_value(val), m_isRead(read), m_isWritten(written), m_isImplicit(false), m_isTruePredicate(false), m_isFalsePredicate(false) {}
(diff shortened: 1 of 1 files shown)


instructionAPI/h/Operand.h
```diff
@@ -60,45 +60,16 @@ namespace Dyninst
...
       /// \param read True if this operand is read
       /// \param written True if this operand is written
-      Operand(Expression::Ptr val, bool read, bool written) : 
-          op_value(val), m_isRead(read), m_isWritten(written), m_isImplicit(false), m_isTruePredicate(false), m_isFalsePredicate(false) {}
-      Operand(Expression::Ptr val, bool read, bool written, bool implicit) : 
-          op_value(val), m_isRead(read), m_isWritten(written), m_isImplicit(implicit), m_isTruePredicate(false), m_isFalsePredicate(false) {}
       // An instruction can be true predicated, false predicated, or not predicated at all
-      Operand(Expression::Ptr val, bool read, bool written, bool implicit,
-              bool trueP, bool falseP):
+      Operand(Expression::Ptr val = {}, bool read = false, bool written = false, bool implicit = false,
```
hainest, line 69: I would suggest making this `explict`, but I'm not sure how much that would break user code.

instructionAPI/h/Operand.h
```diff
@@ -60,45 +60,16 @@ namespace Dyninst
...
+              bool trueP = false, bool falseP = false):
```
hainest, line 70: The old default ctor was `noexcept`. If `boost::shared_ptr`'s default constructor is `noexcept`, I would mark this, as well.

instructionAPI/h/Operand.h
kupsch: Done; unlikely to impact user code.

instructionAPI/h/Operand.h
kupsch: `boost::shared_ptr` default ctor is `noexcept`, but the ctor taking a pointer is not, so this constructor is not `noexcept`.

hainest (approved):

PR #1463, opened 2023-07-25
improve intel instruction & operand formatting

### LINKED ISSUE #1459: Fix x86_64 instruction formatting for >2 operands
Formatting an x86_64 instruction with three or more operands has the operands in the wrong order.  Dyninst stores the operands in intel syntax order, but wants to formats them in AT&T syntax order.  Essentially the formatter rotates the operands right one instead of reversing the list.  If the intel order was `A, B, C`, the AT&T syntax should be `C, B, A`, but it currently outputs as `C, A, B`.  One and two operand instructions format correctly as rotating and reversing produce the same result.

### LINKED ISSUE #1460: Intel X86_64 operand formatting is incorrect
Formatting an x86_64 instruction operand can produce an incorrect string representation for indirect operands. The string representation for the operands when formatting a whole instruction is correct.  An indirect operand expression produces a result that starts with `##Operand` and needs to be transformed to `0x0(Operand)`.  This fix-up code is in the Instruction::format method instead of the Operand::format method.

### LINKED ISSUE #1461: formatting x86_64 instruction incorrectly makes kN registers masks in some cases
formatting instruction that use a mask register as non-mask register such as `kmovb %k1, %k2`.  Get formatted as `kmovb %k1{%k2}`. Dyninst when formatting an instruction for disassembly incorrectly treats any formatted operand value that starts with `%k...` as a mask register, and outputs `{%k...}`.  Mask registers are already formatted as `{kN}` so this special treatment should be eliminated.

kupsch:
This fixes three issues when formatting an x86_64 instruction or its operands:

1) Fix the order of operands when formatting the instruction to be the
   AT&T syntax order.  If the instruction had three or more operands.
   Conversion from the internal (Intel) order rotated right by 1 instead
   of reversing the operands.
Fixes #1459

2) Eliminate special treatment of register %kN as the first operand as
   mask registers in Instruction::format as Operand::format already
   formats these as masks (enclosed in braces) and not all uses of mask
   registers are as a mask.
Fixes #1461

3) Fix Operand::format to produce the correct disassembly operand
   string.  The formatting of the internal Expression needs to be done
   for some indirect values.  This was done in the Instruction::format
   instead of directly in Operand::format
Fixes #1460

Methods changed:

* x86Formatter::getInstructionString - fixes 1, 2, 3
* Operand::format - fixes 3
* x86Formatter::formatRegister - cleanup, remove malloc that could leak

### DIFF
--- a/instructionAPI/src/ArchSpecificFormatters.C
+++ b/instructionAPI/src/ArchSpecificFormatters.C
@@ -252,6 +252,8 @@ std::string x86Formatter::getInstructionString(std::vector<std::string> operands
 {
     std::string s;
     bool oneOperandAdded{false};
+
+    // append the operands in reverse order to convert from Intel to AT&T syntax order
     for (auto i = operands.crbegin(); i != operands.crend(); ++i)  {
 	if (oneOperandAdded)  {
 	    s += ',';
@@ -259,6 +261,7 @@ std::string x86Formatter::getInstructionString(std::vector<std::string> operands
 	s += *i;
 	oneOperandAdded = true;
     }
+
     return s;
 }
 
--- a/instructionAPI/src/ArchSpecificFormatters.C
+++ b/instructionAPI/src/ArchSpecificFormatters.C
@@ -4,6 +4,7 @@
 
 #include "ArchSpecificFormatters.h"
 #include <algorithm>
+#include <iterator>
 #include <sstream>
 #include <iostream>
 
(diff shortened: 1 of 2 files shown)


instructionAPI/src/ArchSpecificFormatters.C
```diff
@@ -257,53 +250,16 @@ std::string x86Formatter::formatDeref(std::string addrString)
...
-        } else if(!op.compare(0, 2, "%k"))
-        {
-            kmask_op = "{" + op + "}";
-        } else if(!source_ops.compare(""))
-        {
-            source_ops = op;
-        } else {
-            source_ops += "," + op;
-        }
+    std::string s;
```
hainest, line 253: Could you add a quick comment with an example? On first pass, I missed the 'r' in `crbegin`.

Since we're in here fixing things, we might as well change the param to a `const&`. But I'll let you decide if you want to do that.

kupsch: I added a comment.  What type of example are looking for?

I was going to update the ArchSpecificFormatter interface to use const refs for vectors and strings for all the methods in a separate PR.

hainest: I guess an example wouldn't really be helpful here.

hainest (approved):

PR #1464, opened 2023-07-25
modernize ArchSpecificFormatter classes

kupsch:
- pass string and vector parameters by const reference instead of by value for performance

- add ArchSpecificFormatter::getInstructionString and remove the overridden versions in the non-x86_64 as they were functionally the same, but not identically written

- remove virtual and add override to overridden methods

### DIFF
--- a/instructionAPI/src/ArchSpecificFormatters.C
+++ b/instructionAPI/src/ArchSpecificFormatters.C
@@ -16,7 +16,7 @@ using namespace Dyninst::InstructionAPI;
 
 ///////// Base Formatter
 
-std::string ArchSpecificFormatter::getInstructionString(const std::vector<std::string> &operands)
+std::string ArchSpecificFormatter::getInstructionString(const std::vector<std::string> &operands) const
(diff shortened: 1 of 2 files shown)


instructionAPI/src/ArchSpecificFormatters.C
```diff
@@ -16,7 +16,25 @@ using namespace Dyninst::InstructionAPI;
 
 ///////// Base Formatter
 
-std::string ArchSpecificFormatter::formatBinaryFunc(std::string left, std::string func, std::string right) {
+std::string ArchSpecificFormatter::getInstructionString(const std::vector<std::string> &operands) const
+{
+    std::string s;
```
hainest, line 21: I would strongly encourage the use of [`boost::algorithm::join_if`](https://www.boost.org/doc/libs/1_82_0/doc/html/boost/algorithm/join_if_1_3_3_9_14_3_1_2.html) here.

hainest (approved):

PR #1473, opened 2023-08-02
Fix AMDGPU register formatting

bbiiggppiigg:
Fix missing parameter for makeRegisterExpression
During one of the PRs that targets compiler warning,
the num_elements parameter to makeRegisterExpression is omitted,
resulting in all register operands are displayed separately,
whether they should be displayed as a group or not.

In addition to adding the parameter, certern unused helper functions
and fields of InstructionDecoder-amdgpu-gfx*** has been removed
to avoid shadowing of variable names.
(Note that this change isn't applied to vega as we don't have ISA-XML for it.)

Also, the logic for handling this type of formatting is extracted out of Register.C
into a helper function in ArchitectureSpecificFormatter.C

### DIFF
--- a/instructionAPI/h/ArchSpecificFormatters.h
+++ b/instructionAPI/h/ArchSpecificFormatters.h
@@ -85,7 +85,9 @@ namespace Dyninst {
             std::string formatDeref(const std::string&) const override;
             std::string formatRegister(const std::string&) const override;
             std::string formatBinaryFunc(const std::string&, const std::string&, const std::string&) const override;
-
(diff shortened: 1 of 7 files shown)


kupsch (approved):
Try to be more consistent in formatting in the future (dyninst has no consistent formatting, but new code should be have consistency in the new code). At some point we'll reformat everything for consistency. For instance, keywords and binary logical and numeric operators should have whitespace before and after; space before '{' that introduces a block; no whitespace before ')', no whitespace after '(', ...

hainest:
@bbiiggppiigg If this is complete, feel free to squash-and-merge.

PR #1474, opened 2023-08-02
Fix hang when creating ELF dynamic sections

### LINKED ISSUE #1445: Potential Infinite Loop in emitElf.C
**Describe the bug**
At 
https://github.com/dyninst/dyninst/blob/9696c64a3a1d6f41241d27c3c51820616490efa6/symtabAPI/src/emitElf.C#L2402
variable i is used as the loop index variable to iterate through new_dynamic_entries

But in its inner loop, at 
https://github.com/dyninst/dyninst/blob/9696c64a3a1d6f41241d27c3c51820616490efa6/symtabAPI/src/emitElf.C#L2432
variable i is used as the loop index variable to iterate through header 

**To Reproduce**
A binary with no .rel.dyn and .rela.dyn sections, but its target platform uses rel type relocation records.
https://github.com/dyninst/dyninst/blob/9696c64a3a1d6f41241d27c3c51820616490efa6/symtabAPI/src/emitElf.C#L2526-L2527
This kind of binary will create two additional new_dynamic_entries.
So when header->num_entries < new_dynamic_entries.size(), the loop variable will keep getting resetted. 

**Expected behavior**
The two loops uses two different index variable.

**Additional context**
This potential bug is discovered when attempting to instrument AMDGPU binaries and  is related to #1444 . 
Because an AMDGPU binary was previously incorrectly specified as using a rel relocation record, the loop index variable of the outer loop will keep getting resetted.

bbiiggppiigg:
This is triggered when a binary with no .rel.dyn and .rela.dyn sections, but its target platform uses rel type relocation records. This happened on an AMDGPU binary that was previously incorrectly specified as using a rel relocation record.

Fixes #1445

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -2399,9 +2399,9 @@ void emitElf<ElfTypes>::createDynamicSection(void *dynData_, unsigned size, Elf_
         dynamicSecData[DT_NEEDED].push_back(dynsecData + curpos);
         curpos++;
     }
-    for (unsigned i = 0; i < new_dynamic_entries.size(); i++) {
-        long name = new_dynamic_entries[i].first;
-        long value = new_dynamic_entries[i].second;
+    for (auto const& entry: new_dynamic_entries){
+        long name = entry.first;
+        long value = entry.second;
         dynsecData[curpos].d_tag = name;
         long adjust = 0;
         switch(name)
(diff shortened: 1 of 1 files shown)


hainest (approved):
LGTM. Please update the commit message with the updated title and description before committing.

bbiiggppiigg:
@hainest 
The title of this PR is longer than 50 characters.
What title should I use for the commit ?

hainest:
@bbiiggppiigg Derp I should have checked that. I updated it.

PR #1476, opened 2023-08-09
Fix overflow of fields in instruction layout

bbiiggppiigg:
Fields in instruction layouts can have implicit LSBs. 

Previously this wasn't taken into consideration,
and causes overflow and decoding instruction.

This PR fixes the layout definition to include the LSBs.


instructionAPI/src/AMDGPU/gfx908/amdgpu_gfx908_decoder_impl.h
```diff
@@ -1,5 +1,3 @@
-#include <stdint.h>
```
hainest, line 1: Did you update your generation scripts to include Jim's updates for including standard headers?

bbiiggppiigg: Resolved.

hainest (approved):

PR #1484, opened 2023-08-17
Continue parsing amdgpu binaries after procedure call instructions

### LINKED ISSUE #1470: Dyninst doesn't continue parsing AMDGPU code after a procedure call
This was by earlier design.

A recent review of AMDGPU binaries found that procedure calls are now being
used in AMDGPU code.   

As a result, without continuation parsing, Dyninst can miss parsing a substantial
portion of an AMDGPU binary, as it doesn't continue to parse code following a
procedure call instruction.

bigtrak:
Resolves #1470 

There was an additional area of code examination which 
needed to be changed to better recognize return instructions
in the amdgpu.

The initial issue/pr was correct, it could do better.

In addition, the remainder of inline hand-edited amdgpu instructions
have been moved to include files that can be mechanically generated
by the data dumper.


instructionAPI/src/Instruction.C
```diff
@@ -558,13 +558,24 @@ namespace Dyninst
...
                 case e_call:
                 case e_syscall:
-                case amdgpu_op_s_setpc_b64:
-                case amdgpu_op_s_swappc_b64:
-                case amdgpu_gfx908_op_S_SETPC_B64:
-                case amdgpu_gfx908_op_S_SWAPPC_B64:
-                case amdgpu_gfx90a_op_S_SETPC_B64:
-                case amdgpu_gfx90a_op_S_SWAPPC_B64:
                     return false;
+#if 1
```
hainest, line 562: I'm assuming this was for debugging? Let's remove it.

instructionAPI/src/Instruction.C
```diff
@@ -558,13 +558,24 @@ namespace Dyninst
...
+                 * of Semantic Evaluation feature for amdgpu, and a few
+                 * other odd cases due to instruction classification issues.
+                 *
+                 * This temporary fix allows almost full dyninst CFG
+                 * until we have amdgpu semantic information.
+                 * Once semantic evaluation is received & implemented
+                 * for amdgpu architectures, this will no longer be needed;
+                 * then GPU instructions will be processed as e_ "instructions".
+                 */
+#include "amdgpu_callinsn_table.h"
```
hainest, line 575: I know there are lots of places in Dyninst where we include source files and headers in the middle of blocks, but let's not perpetuate that terrible coding style. If these cases have a special meaning, we should call them out in a function.

bigtrak: Those are machine generated files from the database dumps.
This way, as changes happen and new AMD GPU architectures are added, the machine
generated changes stay in machine generated files, not in regular files
that need to have those changes  hand edited into.

instructionAPI/src/Instruction.C
bigtrak: This code sequence is a temporary until real issues are fixed.
By #if the sequence that will be removed when further work is done,
this is completely identified as to what will  be removed.

kupsch (changes requested):
Header files should only be used to import declarations and definitions, and should not be used to imports bits of code into the middle of functions used as a sort of macro expansion as it causes maintenance and understandability issues.

Having 3 `#include`ed files that contain just case labels allows the labels for the instruction types to get generated without having change InstructionCategories.C, it makes it very difficult for anyone having to understand the `entryToCategory` function as you have to look at 4 different files after first realizing that there are 4 different file.

You can achieve the same goal without having the body of `entryToCategory` in 4 different files. Make one header file `AmdInstructionCategories.h` that contains 1 inlined function `amdEntryToCategory` in an anonymous namespace that covers all the AMD GPU instructions that have special categories and returns their type or `c_NoCategory` if unknown.  Then replace the `return c_NoCategory;` with `return amdEntryToCategory(e);` after the `default:` label.

Make a similar change to `Instruction::allowsFallThrough`, call an inline function as the `return` expression of the `default:` whose definition is contained in a generate header file (same or different as above). The `#if 1` can then go away as it is the current version. Unclear why the `#if 1` is present.

PR #1493, opened 2023-08-19
Make MappedFile::mapped_files threadsafe

hainest:
Currently, a MappedFile is created from a constructor in either the Symtab or Archive classes. In both cases, a user is allowed to invoke them in parallel (although only one thread may enter the constructor), but the cache here is not threadsafe.

### DIFF
--- a/common/src/MappedFile.C
+++ b/common/src/MappedFile.C
@@ -29,23 +29,28 @@
  */
 #include "common/src/MappedFile.h"
 #include "common/src/pathName.h"
+#include "concurrent.h"
 #include <iostream>
 using namespace std;
 
-dyn_hash_map<std::string, MappedFile *> MappedFile::mapped_files;
+namespace {
+  dyn_c_hash_map<std::string, MappedFile *> mapped_files;
+}
 
 MappedFile *MappedFile::createMappedFile(std::string fullpath_)
 {
    //fprintf(stderr, "%s[%d]:  createMappedFile %s\n", FILE__, __LINE__, fullpath_.c_str());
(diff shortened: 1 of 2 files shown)


common/src/MappedFile.C
```diff
@@ -29,23 +29,28 @@
...
-      //fprintf(stderr, "%s[%d]:  mapped file exists for %s\n", FILE__, __LINE__, fullpath_.c_str());
-      MappedFile  *ret = mapped_files[fullpath_];
-      if (ret->can_share) {
+   {
+     decltype(mapped_files)::const_accessor a;
+     if (mapped_files.find(a, fullpath_)) {
+       //fprintf(stderr, "%s[%d]:  mapped file exists for %s\n", FILE__, __LINE__, fullpath_.c_str());
+       MappedFile  *ret = a->second;
+       if (ret->can_share) {
          ret->refCount++;
```
kupsch, line 49: This is not thread safe.  Make refCount atomic.

common/src/MappedFile.C
```diff
@@ -168,17 +177,7 @@ void MappedFile::closeMappedFile(MappedFile *&mf)
```
kupsch, line 177: decrement and then test of `refCount` is not thread safe.  `refCount` needs to be atomic and then and then atomically decrement and return the pre decrement value; and then test the value.

kupsch (changes requested):

hainest:
I'm closing this as making the class threadsafe is nearly impossible given the semantics of `setSharing`.

PR #1500, opened 2023-09-01
Unify naming of Modules

hainest:


hainest:
> I think the `ref_sig8` changes are broken and were broken before.

They were. I've removed that commit for now. We'll want to revisit how we handle .debug_type entries at some point.

> Rename `dwarf_names.hpp` to `dwarf_names.h`; only header not using the `.h` suffix is `dwarf_cu_info.hpp`. Perhaps `dwarf_cu_info.hpp` should be renamed (`dwarf_cu_info.hpp` could be left with a `#error ...` body for a release or two).

Those were added by me in the PR, so I just renamed them.

kupsch (approved):

PR #1502, opened 2023-09-01
Workarounds for GFX908 and GFX90A

bbiiggppiigg:
This commit contains the workaround (until we have updated data dumps) for the following problems 

1. Instructions that use SDWA have wrong length
2. Instructions that use SDWA does not have the _SDWA suffix
3. MUBUF instructions decodes the VADDR field incorrectly
4. MUBUF instructions missing information about offset
5. DS instructions missing information about offset


bbiiggppiigg:
@bigtrak 
This branch should make it easier for you to compare your output to toolchain output.

instructionAPI/h/Instruction.h
```diff
@@ -48,281 +48,288 @@
 
 namespace Dyninst
 {
```
kupsch, line 50: Undo the the reindentation of this file.  Reindenting a file (that wasn't confusingly indented) just added noise to the commit and obfuscates the real changes.  Diffing

``` 
  instructionAPI/h/Instruction.h    | 541 ++++++++++++++++++++--------------------
  instructionAPI/h/Operation_impl.h | 236 +++++++++---------
```

If I ignore whitespace changes the diff shows only 2% of the lines changed are real:

```
  instructionAPI/h/Instruction.h    | 7 +++++++
  instructionAPI/h/Operation_impl.h | 8 ++++++++
```

instructionAPI/src/Immediate.C
```diff
@@ -102,31 +118,31 @@ namespace Dyninst {
         std::string ArmConditionImmediate::format(formatStyle) const {
             unsigned int cond_val = eval().convert<unsigned int>();
             if(m_condLookupMap.count(cond_val) > 0)
```
kupsch, line 120: More whitespace changes. Undo the whitespace changes.

kupsch (changes requested):
In the future use a better branch name.  It should be of the form **USENAME/DESCRIPTIVE_NAME**.  This  makes it easier to determine the branch owner and purpose. The name *workaround* is quite generic and could just as well be *stuff*.

Use a better PR title that more specifically describes the changes like *GFX908/A handcrafted instr property improvements*.

There are several files that are reformatted with whitespace only changes. See inline comments. Undo these and avoid making them in the future.  If they do need to be made, talk to people and make a PR with only the whitespace changes.

PR #1512, opened 2023-09-07
Github CI: execute external tests on PR

hainest:


kupsch (approved):
I assume this is for a future update to external_tests since they don't seem to be ctest-ified.

hainest:
> I assume this is for a future update to external_tests since they don't seem to be ctest-ified.

Only a few have ctests (https://github.com/dyninst/external-tests/blob/master/dyninstAPI/static_ctors_dtors/CMakeLists.txt).

PR #1518, opened 2023-09-08
Add logicalNot operation to Rose for correctness

bbiiggppiigg:
Previously rose use invertOp for logical Not.
As the Dyninst interpretation of Rose  didn't limit
the size of the input to single bit, inverting a non-zero
value can still result in an non-zero value, causing
incorrect result to be populated.

This PR adds a separate logicalNotOp that
returns 0 given non-zero value
returns 1 given 0


bbiiggppiigg:
@kupsch 
Rebased on master after merging 1511.

PR #1522, opened 2023-09-11
 Fix int size of rose operations return values

### LINKED ISSUE #1517: Rose logicalNot implemented as 1's complement
**Describe the bug**

https://github.com/dyninst/dyninst/blob/bec7e21a5008adc1c0445775602ce740943e9b55/dataflowAPI/rose/semantics/BaseSemantics2.h#L1734-L1735

Currently, Rose didn't extinguish between logical and bitwise operation ( and / or / invert ).
And invert is used to implement logicalNot.

https://github.com/dyninst/dyninst/blob/bec7e21a5008adc1c0445775602ce740943e9b55/dataflowAPI/rose/semantics/BaseSemantics2.C#L243-L246

https://github.com/dyninst/dyninst/blob/bec7e21a5008adc1c0445775602ce740943e9b55/dataflowAPI/rose/semantics/BaseSemantics2.C#L225-L233

However, since we didn't limit its input to invert to 1 bit, inverting an non-zero value can still result in a non-zero value, causing incorrect result.

https://github.com/dyninst/dyninst/blob/bec7e21a5008adc1c0445775602ce740943e9b55/parseAPI/src/SymbolicExpression.C#L240-L245

https://github.com/dyninst/dyninst/blob/bec7e21a5008adc1c0445775602ce740943e9b55/parseAPI/src/SymbolicExpression.C#L271-L281

bbiiggppiigg:
- set integer bit size of equalsizeOp to be 1
- set integer bit size of andOp to be max size of operands
- set integer bit size of orOp to be max size of operands

Fix #1517

### DIFF
--- a/parseAPI/src/SymbolicExpression.C
+++ b/parseAPI/src/SymbolicExpression.C
@@ -242,14 +242,20 @@ AST::Ptr SymbolicExpression::SimplifyRoot(AST::Ptr ast, Address addr, bool keepM
                                           if (roseAST->child(0)->getID() == AST::V_ConstantAST && roseAST->child(1)->getID() == AST::V_ConstantAST) {
                                               ConstantAST::Ptr child0 = boost::static_pointer_cast<ConstantAST>(roseAST->child(0));
                                               ConstantAST::Ptr child1 = boost::static_pointer_cast<ConstantAST>(roseAST->child(1));
-                                              return ConstantAST::create(Constant(child0->val().val & child1->val().val, 64));
+                                              auto size0{child0->val().size};
+                                              auto size1{child1->val().size};
+                                              uint64_t newSize{std::max(size0,size1)};
+                                              return ConstantAST::create(Constant(child0->val().val & child1->val().val, newSize));
                                           }
                                           break;
             case ROSEOperation::orOp:
                                           if (roseAST->child(0)->getID() == AST::V_ConstantAST && roseAST->child(1)->getID() == AST::V_ConstantAST) {
                                               ConstantAST::Ptr child0 = boost::static_pointer_cast<ConstantAST>(roseAST->child(0));
                                               ConstantAST::Ptr child1 = boost::static_pointer_cast<ConstantAST>(roseAST->child(1));
-                                              return ConstantAST::create(Constant(child0->val().val | child1->val().val, 64));
+                                              auto size0{child0->val().size};
+                                              auto size1{child1->val().size};
+                                              uint64_t newSize{std::max(size0,size1)};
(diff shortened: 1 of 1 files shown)


parseAPI/src/SymbolicExpression.C
```diff
@@ -242,14 +242,20 @@ AST::Ptr SymbolicExpression::SimplifyRoot(AST::Ptr ast, Address addr, bool keepM
                                           if (roseAST->child(0)->getID() == AST::V_ConstantAST && roseAST->child(1)->getID() == AST::V_ConstantAST) {
                                               ConstantAST::Ptr child0 = boost::static_pointer_cast<ConstantAST>(roseAST->child(0));
                                               ConstantAST::Ptr child1 = boost::static_pointer_cast<ConstantAST>(roseAST->child(1));
-                                              return ConstantAST::create(Constant(child0->val().val & child1->val().val, 64));
+                                              auto size0{child0->val().size};
+                                              auto size1{child1->val().size};
+                                              uint64_t newSize{size0 > size1 ? size0 : size1};
```
kupsch, line 247: use std::max

parseAPI/src/SymbolicExpression.C
```diff
@@ -242,14 +242,20 @@ AST::Ptr SymbolicExpression::SimplifyRoot(AST::Ptr ast, Address addr, bool keepM
...
                                           }
                                           break;
             case ROSEOperation::orOp:
                                           if (roseAST->child(0)->getID() == AST::V_ConstantAST && roseAST->child(1)->getID() == AST::V_ConstantAST) {
                                               ConstantAST::Ptr child0 = boost::static_pointer_cast<ConstantAST>(roseAST->child(0));
                                               ConstantAST::Ptr child1 = boost::static_pointer_cast<ConstantAST>(roseAST->child(1));
-                                              return ConstantAST::create(Constant(child0->val().val | child1->val().val, 64));
+                                              auto size0{child0->val().size};
+                                              auto size1{child1->val().size};
+                                              uint64_t newSize{size0 > size1 ? size0 : size1};
```
kupsch, line 257: use std::max

kupsch (approved):
Looks good, but make the following changes before merging (if you make just these changes I don't need to review again):

1) Use std::max instead writing your own

2) Change the title, commit title and commit body to be clearer

For the title omething like *Fix int size of rose operations return values*.

For the items, use a markdown list and make the change clearer:

```
- set integer bit size of equalsizeOp to be 1
- set integer bit size of andOp to be max size of operands
- set integer bit size of orOp to be max size of operands
```

PR #1528, opened 2023-09-19
GithubCI: Allow consumer builds to be run manually

hainest:


kupsch (approved):
Looks good except there is a typo in commit (and PR) header "conumer" -> consumer

PR #1531, opened 2023-09-20
Hide implementations of complex data structures in SymtabAPI::Symtab

hainest:
This allows us to more easily modify the parallelism around the big tables (e.g., `indexed_modules`) without bumping the ABI around, meaning we don't have to wait until a major release to make those changes. This also removes internal details like `concurrent.h` which puts TBB into the pubic API and the multiple Boost headers needed for `multi_index`.

Although `ModRangeLookup* mod_lookup();` was a public member, it was never documented, and should never be called by a user. I'm not considering this an API-breaking change.

### DIFF
--- a/symtabAPI/h/Symtab.h
+++ b/symtabAPI/h/Symtab.h
@@ -51,8 +51,6 @@
 
 #include "dyninstversion.h"
 
-#include "concurrent.h"
-
 #include "boost/shared_ptr.hpp"
 
 class MappedFile;
--- a/symtabAPI/h/Symtab.h
+++ b/symtabAPI/h/Symtab.h
@@ -532,8 +532,6 @@ class SYMTAB_EXPORT Symtab : public LookupInterface,
 
    // Similar for Variables
    std::vector<Variable *> everyVariable{};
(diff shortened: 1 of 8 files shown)


hainest:
This will have to wait until https://github.com/dyninst/testsuite/pull/233 is merged.

symtabAPI/h/Symtab.h
```diff
@@ -111,6 +91,9 @@ class SYMTAB_EXPORT Symtab : public LookupInterface,
    friend class relocationEntry;
    friend class Object;
 
+   // Hide implementation details that are complex or add large dependencies
+   std::unique_ptr<symtab_impl> impl;
```
kupsch, line 95: `const std::unique_ptr...`?  The impl never changes.

kupsch (approved):
Only suggestion: `impl` could be `const`

symtabAPI/h/Symtab.h
hainest: Since `Symtab` isn't moveable, that is a good idea.

PR #1534, opened 2023-09-21
Refactor Symtab::addSymbolToAggregate

hainest:
This makes it easier to understand what's happening. It's still not clear why the region check is necessary.

### DIFF
--- a/symtabAPI/src/Symtab.C
+++ b/symtabAPI/src/Symtab.C
@@ -689,8 +689,11 @@ bool Symtab::addSymbolToAggregates(const Symbol *sym_tmp)
 
             if( func->getRegion() != sym->getRegion() ) {
                 func = new Function(sym);
+            } else {
+        	// The function has an additional name
+        	// e.g., two symbols aliasing the same code location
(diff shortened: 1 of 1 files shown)


symtabAPI/src/Symtab.C
```diff
@@ -687,16 +689,16 @@ bool Symtab::addSymbolToAggregates(const Symbol *sym_tmp)
 
             if( func->getRegion() != sym->getRegion() ) {
                 func = new Function(sym);
-                boost::unique_lock<dyn_rwlock> l(symbols_rwlock);
-                everyFunction.push_back(func);
-                sorted_everyFunction = false;
             }
             func->addSymbol(sym);
```
kupsch, line 693: clearer as

```
}  else  {
    func->addSymbol(sym);  // function has additional name
}
```

kupsch (approved):

symtabAPI/src/Symtab.C
hainest: added.

PR #1535, opened 2023-09-22
Refactor DwarfWalker::dieRanges

hainest:


symtabAPI/src/dwarfWalker.C
```diff
@@ -819,38 +819,38 @@ bool DwarfWalker::parseRangeTypes(Dwarf * dbg, Dwarf_Die die) {
...
-        offset = dwarf_ranges(&die, offset, &base, &start, &end);
         if(offset < 0) return std::vector<AddressRange>();
-        if(offset == 0) break;
         dwarf_printf("Lexical block from 0x%lx to 0x%lx\n", start, end);
-//        printf("Lexical block from 0x%lx to 0x%lx, offset is %p, %d entries\n", start, end, offset, newRanges.size());
         newRanges.push_back(AddressRange(start, end));
+        offset = dwarf_ranges(&die, offset, &base, &start, &end);
     }
     return newRanges;
 }
```
kupsch, line 839: The change is correct, but I don't think the previous version of the loop is UB. There are observable side effect in `dwarf_print`, and the compiler can't know if the calls to functions not defined in this compilation unit have observable side effects.  I also think the `return` and `break` control flow statement also make this non-UB.  A loop is only UB if there are no observable side-effects and it does not terminate.

kupsch (approved):
The loop UB commit seems unnecessary as I don't think the loop is UB (see comment on change).

symtabAPI/src/dwarfWalker.C
hainest: I went back and looked, and the wording around this changes between drafts. Good to know. I'll revert that commit so it doesn't add confusion.

From n3090 (2010-03-09)

    6.5(5) [stmt.iter]

    A loop that, outside of the for-init-statement in the case of a for statement,
    - makes no calls to library I/O functions, and
    ...
    may be assumed by the implementation to terminate.

From n3242 (2011-02-08)

    1.10(24) [intro.multithread]

    The implementation may assume that any thread will eventually do one of the following:
     - terminate,
     - make a call to a library I/O function,
    ...

From 4659 (2017-03-21)

    4.7.2(3) [intro.progress]

    During the execution of a thread of execution, each of the following is termed an execution step:
      (3.1)— termination of the thread of execution,
      (3.2)— performing an access through a volatile glvalue, or
      (3.3)— completion of a call to a library I/O function, a synchronization operation, or an atomic operation.

PR #1544, opened 2023-09-28
Replace boost::multi_index_container with tbb::concurrent_unordered_set in symtab_impl

hainest:
There are now only two dimensions to each Module, so the multi_index isn't needed. This also replaces the mutex with the intrinsic reader/writer locks in TBB.


symtabAPI/src/indexed_modules.hpp
kupsch, line 1: add copyright and header guard

symtabAPI/src/Symtab.C
```diff
@@ -504,10 +504,10 @@ bool Symtab::fixSymModules(std::vector<Symbol *> &raw_syms)
 	createDefaultModule();
     }
 
-    for (auto i = impl->indexed_modules.begin(); i != impl->indexed_modules.end(); ++i)
+    for (auto *m : this->impl->modules)
```
kupsch, line 507: remove the `this->` for consistentcy

kupsch (approved):
in another PR (or here) change the `.hpp` found in the PR and the other two to `.h` files to be consistent with the rest of dyninst.  Couple of other minor comments below

symtabAPI/src/indexed_modules.hpp
hainest: I always forget these...

symtabAPI/src/Symtab.C
hainest: No idea how that got there. Fixed.

hainest:
> in another PR (or here) change the `.hpp` found in the PR and the other two to `.h` files to be consistent with the rest of dyninst. Couple of other minor comments below

Fixed indexed_modules here. I'll submit a separate PR for the other two.

PR #1546, opened 2023-09-28
Remove DwarfWalker::setModuleFromName

hainest:
It's replaced with the updated Symtab::findModuleByOffset. Finding by name was always unnecessary as we know the offset of the current DIE.

This is a step toward removing the broken Symtab::findModuleByName.


hainest:
I'm an idiot. This shouldn't be checked until #1545 is merged.

kupsch (approved):

PR #1550, opened 2023-09-29
add deprecated annotation support

kupsch:
- Added the macro DYNINST_DEPRECATED(msg). It can be placed before a function, method, type or variable so that on use a deprecated diagnostic is produced.  The macro works using language standard annotation or compiler specific annotations or has no effect if neither is available.

### DIFF
--- a/common/h/compiler_annotations.h
+++ b/common/h/compiler_annotations.h
@@ -94,6 +94,41 @@
 
 /***********************************************************************
  *
+ * DYNINST_DEPRECATED(msg)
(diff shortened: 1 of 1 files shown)


kupsch:
@hainest you can use this in PR #1545

hainest (approved):
Thanks for this. I was considering using the one from boost/config. The 'deprecated' attribute requires the parentheses be omitted if no message is given. I'm fine with forcing the use of a message here.

PR #1554, opened 2023-10-02
fix deprecated annotation warning using clang

kupsch:
The clang compiler allows the use of some attributes introduced in a later language standard than use to compile the source as a non-standard extension.  Clang's __has_cpp_attribute and __has_c_attribute returns true for these attributes.  If one of these attributes is used, clang warns of a non-standard language feature usage.  So for clang, only use a standard attribute if the feature test returns true and the language standard version is valid.

- fix clang's [[deprecated]] (only if C++ >= 14 and C >= 23)

- use now known C-23 __STDC_VERSION__ value 202311L

### DIFF
--- a/common/h/compiler_annotations.h
+++ b/common/h/compiler_annotations.h
@@ -53,7 +53,7 @@
  */
 
 #if defined(__cplusplus) && defined(__has_cpp_attribute)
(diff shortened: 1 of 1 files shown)


hainest (approved):
#1551 builds cleanly now

PR #1559, opened 2023-10-05
warning fix:  improve logical-op handling

kupsch:
- use DYNINST_DIAGNOSTIC_BEGIN_SUPPRESS_LOGICAL_OP and END macros to suppress logical-op warning

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -2366,15 +2366,10 @@ bool shouldSaveReg(registerSlot *reg, baseTramp *inst, bool saveFlags)
    if (saveFlags) {
       // Saving flags takes up EAX/RAX, and so if they're live they must
       // be saved even if we don't explicitly use them
-#if !defined(__GNUC__) || __GNUC__ >= 10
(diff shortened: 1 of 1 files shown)


hainest (approved):

hainest:
I just noticed that the suppressions are only checking for __GNUC__. Does that get defined for clang or is it only __GNU__?

kupsch:
It expands to nothing on compiler that do not produce the incorrect warning (gcc >= 9 and all clang versions).  On gcc <= 8 it expands to _Pragma's that push the current set of warning on stack disables the one warning,  The END macro pops the stack to restore the warning.

hainest:
> It expands to nothing on compiler that do not produce the incorrect warning (gcc >= 9 and all clang versions). On gcc <= 8 it expands to _Pragma's that push the current set of warning on a stack and disables the warning. The END macro pops the stack to restore the warning.

Ah, so this was just to refactor to use the macro?

kupsch:
Yep. Instead of the hack that break if the EAX and RAX constants were ever not equal.

PR #1560, opened 2023-10-06
GithubCI: add testing on Fedora-39

hainest:
This uses gcc-13. Clang won't work because the only available libomp is for clang-17, but the highest installable version of clang is 15.


kupsch (approved):
Add the missing new-line, otherwise looks good.

PR #1571, opened 2023-10-10
Add Symtab::getContainingModule(Offset)

hainest:
Returns the module with PC ranges that contain a given offset (really address). By contrast, findModuleByOffset(Offset) finds a module starting at the given offset.

This is the correct replacement for `Symtab::findModuleByOffset(std::set<Module *>&, Offset)` removed in #1545.


kupsch (approved):
You might add to the documentation that the default module is never returned even though it covers the entire range of the symtab object.

hainest:
> You might add to the documentation that the default module is never returned even though it covers the entire range of the symtab object.

The default module will be returned if and only if it is the only module. I'll update the docs to explicitly mention that.

PR #1575, opened 2023-10-12
Remove DWARFisms from Symtab::Module

hainest:
There is no need to store the CU DIE from which a Module instance is
derived. The address of the CU can be used to reconstitute the
entry in the .debug_info section using dwarf_addrdie.

Because Module.h is part of the public API for Dyninst, this also
removes the transitive dependency on libdw.

### DIFF
--- a/symtabAPI/h/Module.h
+++ b/symtabAPI/h/Module.h
@@ -83,9 +83,6 @@ namespace Dyninst { namespace SymtabAPI {
     bool isShared() const;
     ~Module();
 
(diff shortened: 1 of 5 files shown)


kupsch (approved):
`dyn_c_queue` is no longer used, you could delete it from *concurrent.h*.

hainest:
> `dyn_c_queue` is no longer used, you could delete it from _concurrent.h_.

Annoyingly, concurrent.h is in the public header space. Someone could be using it.

PR #1585, opened 2023-10-15
Move entryIDs into separate data files

hainest:
This will make it much easier to update them as we migrate to Capstone.

This is a very poor solution, but is the only means of splitting the architectures apart while maintaining the existing entryID enum usage. I tried several different means of making this better, but the only one that worked well requires C++20's `using enum`.

@bbiiggppiigg How much effort is it to modify your generator scripts to accommodate the new locations and file formats?


bbiiggppiigg:
@hainest 
I generate the file by script and copy them into Dyninst manually.
So I assume there isn't any additional effort if we are just putting them in a different location.

bbiiggppiigg (approved):

kupsch (approved):
OK, but two things

1. `common/h/mnemonics/x86` is missing a NL as its last character
2. should the names of the file end in `.h` (perhaps `_entryIds.h` or `_entryId_enums.h`) to indicate that they contain C/C++? This would help with editors and other tools that determine the file type by suffix.

hainest:
> OK, but two things
> 
>     1. `common/h/mnemonics/x86` is missing a NL as its last character
Fixed.

>     2. should the names of the file end in `.h` (perhaps `_entryIds.h` or `_entryId_enums.h`) to indicate that they contain C/C++? This would help with editors and other tools that determine the file type by suffix.
I didn't give them file extensions since they are supposed to be just data files. They also don't contain any valid syntax. EclipseCDT can read them as enum values because it does a full parse when it indexes files, but I guess I was biased by that. Is your editor able to see them like that?

kupsch:
Vim's syntax highlighting for enum values is the same as text, so hard to tell. Line counting tools like cloc ignore these files without extensions using the default settings.

bbiiggppiigg:
I'm not sure but a file extension like .include.h  or .include.data 
or maybe change the directory name from mnemonic to something like entryID.include 
might give people a better understanding of what these files are or where there are used?

kupsch:
Looks good.

bbiiggppiigg:
Looks good to me.

PR #1590, opened 2023-10-26
Refactor common/dyn_regs.h

hainest:
This will make the integration to Capstone easier.

It wasn't my original intention to make such sweeping changes; they just came out organically from what I wanted to do for Capstone automatic updates. There are a few big items I want to discuss before merging:

1) I do not like how some architectures split system registers from the rest, cf. aarch64_sys_regs.h. We should do it for all or none.

2) I'll clean up the MachRegister `name` weirdness in a separate PR.

3) This stuff belongs in instructionAPI, not common. Putting it into the InstructionAPI namespace would require leaving a few namespace directives as breadcrumbs in common, but I think it would still be a good move. Definitely not required and I'm fine leaving this to the far future.

4) @bbiiggppiigg How much does this perturb your automation process for AMD GPU stuff?

5) The business of dyninstAPI/RegisterConversion* seems wholly over-complicated and unnecessary. MachRegister already keeps a container of registers for _all_ architectures, has an index mapping to the DWARF registers via `DwarfEncToReg` which is a unique pair of (index,arch), and it stores the name so we don't need to do that twice. MachRegister is also sizeof(signed int), so it's not expensive to copy around. All of this is internal, so no user code will be affected.

NOTE: This won't pass CI until I push the requisite changes to the test suite, examples, and external-tests repos. I wanted to discuss the design here first.


kupsch (approved):

bbiiggppiigg:
@hainest 

For 3, which stuff do you mean by "this stuff" ?
For 4, if I'm not mistaken, you didn't touch the content of the include file, just change their path? In that case the automation is not really affected.
For 5, I didn't touch that stuff personally so I don't know how it is used,  but I think @rochauha worked on something related to RegisterConversion.

hainest:
> For 3, which stuff do you mean by "this stuff" ?

Everything to do with dyn_regs.h. The concepts of registers only make sense in the presence of an instruction, so they should go in that toolkit.

>For 4, if I'm not mistaken, you didn't touch the content of the include file, just change their path? In that case the automation is not really affected.

Yup, I just moved them to a different directory like with the entryID updates I did last week.

>For 5, I didn't touch that stuff personally so I don't know how it is used, but I think @rochauha worked on something related to RegisterConversion.

The general items I was talking about are from circa 2008. I'll check with him to make sure he's not working on that code.

bbiiggppiigg:
I feel like registers should belong to its architecture namespace. 
And putting it in common might avoid some unnecessary dependency on instructionAPI?

I don't really know. 
And I can't think of use case of registers without associating them with a instruction anyway.

hainest:
> I feel like registers should belong to its architecture namespace.

Yeah, there needs to be a 'machine' representation, but that's a huge change.

>And putting it in common might avoid some unnecessary dependency on instructionAPI?

I see it as _removing_ an unnecessary dependency on common. :)
 
> I don't really know. And I can't think of use case of registers without associating them with a instruction anyway.

Maybe ABI? But even then, it's not entirely separate.

---

Do you have any objections to the changes in e24db465, 7c06190e, 88df8321, or da16ce4b2?

bbiiggppiigg:
@hainest 
So if the first 3 commits are just swapping the bits and type then I have no objections.
There seems to be some extra / in the 4th commit though.
For example:
#include "registers/AMDGPU//amdgpu_gfx908_regs.h"
in
[instructionAPI/src/ArchSpecificFormatters.C](https://github.com/dyninst/dyninst/commit/da16ce4b27c8d0055b8bdc28ace2a32cd01077e0#diff-af390766c98939c9a05e502feda5a4165585b4cd2d61c7d92933b59a2a6b4029)

hainest:
> @hainest So if the first 3 commits are just swapping the bits and type then I have no objections.

Yup. I just reordered them to match what the other files have.

> There seems to be some extra / in the 4th commit though. For example: #include "registers/AMDGPU//amdgpu_gfx908_regs.h" in [instructionAPI/src/ArchSpecificFormatters.C](https://github.com/dyninst/dyninst/commit/da16ce4b27c8d0055b8bdc28ace2a32cd01077e0#diff-af390766c98939c9a05e502feda5a4165585b4cd2d61c7d92933b59a2a6b4029)

Good catch! I'll fix that.

PR #1612, opened 2023-10-30
Improve name handling in MachRegister

hainest:


common/src/registers/MachRegister.C
```diff
@@ -100,16 +100,8 @@ namespace Dyninst {
...
-
-    std::string ret;
-    if(iter != names()->end()) {
-      return iter->second;
-    }
-    common_parsing_printf(" can't find register with index %x\n", (unsigned int)reg);
-    return std::string("<INVALID_REG>");
+  std::string const& MachRegister::name() const {
+    return names.find(reg)->second;
   }
```
kupsch, line 105: @hainest, doesn't this need to check for reg not being in the map?  Couldn't someone use `MachRegister::MachRegister(bogusRegId)`, and bogusRegId would not be in `names`?

kupsch (approved):

PR #1621, opened 2023-11-05
Fix astOperatorNode semantics

hainest:
The lack of consistent checking of empty pointers was fixed by #1609, but those changes broke the semantics of the class. This reverts those changes and fixes the pointer checks more directly.

### DIFF
--- a/dyninstAPI/src/ast.C
+++ b/dyninstAPI/src/ast.C
@@ -1592,7 +1592,6 @@ bool AstOperatorNode::generateCode_phase2(codeGen &gen, bool noCost,
                break;
             }
             case operandType::RegOffset: {
-               assert(loperand);
                assert(loperand->operand());
 
                // load the address reg + addr into dest
--- a/dyninstAPI/src/ast.C
+++ b/dyninstAPI/src/ast.C
@@ -319,10 +319,6 @@ AstOperatorNode::AstOperatorNode(opCode opC, AstNodePtr l, AstNodePtr r, AstNode
     roperand(r),
     eoperand(e)
 {
-    if(!loperand) loperand = boost::make_shared<AstNullNode>();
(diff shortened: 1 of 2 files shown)


dyninstAPI/src/ast.h
```diff
@@ -566,6 +572,7 @@ class AstOperatorNode : public AstNode {
 
     bool generateOptimizedAssignment(codeGen &gen, int size, bool noCost);
 
+    AstOperatorNode() {}
```
kupsch, line 575: `= default`?

kupsch (approved):
A couple of comments on changes in-line and 2 more since github won't allow commenting on lines not near a change:

Need add two more checks to ast.C lines 2460 & 2464 as both have unprotected `loperand->print()`'s.

dyninstAPI/src/ast.C
```diff
@@ -1584,6 +1592,7 @@ bool AstOperatorNode::generateCode_phase2(codeGen &gen, bool noCost,
                break;
             }
             case operandType::RegOffset: {
+               assert(loperand);
```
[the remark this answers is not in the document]
hainest: No, it can't. I just missed it.

dyninstAPI/src/ast.h
hainest: It's never used, so I removed it.

hainest:
> A couple of comments on changes in-line and 2 more since github won't allow commenting on lines not near a change:
> 
> Need add two more checks to ast.C lines 2460 & 2464 as both have unprotected `loperand->print()`'s.

I fixed them, but that whole definition can be removed. Its declaration was removed in 2006. I think everything guarded by ASTDEBUG can be removed, given how long that's been broken.

PR #1623, opened 2023-11-07
Deprecate Symtab::getOrCreateModule

hainest:
There are several problems here:

1) Users shouldn't be creating modules

2) When created, the returned module must be "fixed up" before it's
useful. There's no need for that when one could be properly constructed
at the callsite inside Dyninst.

3) It violates the Single Responsibility Principle


hainest:
This trips clang's -Winconsistent-missing-override. Let me go fix that first.

kupsch (approved):

PR #1626, opened 2023-11-08
Construct a Module from the CU's offset not its PC

### LINKED ISSUE #1625: Regression from 12.3.0 to master with line map and inline info
I'm seeing a definite regression from dyninst 12.3.0 to master (nov 7,
git hash 9256b319789d) for both coverage and non-determinism for both
the line map info and the inline sequences.

This is on x86_64, Ubuntu 22.04, gcc 11.4.0, spack-build dyninst,
elfutils, etc.  I'm using a unit test that calls `Symtab::openFile()`
and `parse()` and then iterates over the parseapi functions and basic
blocks (in order) and dumps the statements (instructions) and prints
the line map info and inline sequences for every instruction.

For example:
```
func:  0x10b130  std::endl<char, std::char_traits<char> >.isra.0

block:
stmt:  0x10b130  l=684  f='/usr/include/c++/11/ostream'
stmt:  0x10b132  l=684  f='/usr/include/c++/11/ostream'
stmt:  0x10b133  l=684  f='/usr/include/c++/11/ostream'
stmt:  0x10b137  l=685  f='/usr/include/c++/11/ostream'
...

block:
stmt:  0x10b14b  l=877  f='/usr/include/c++/11/bits/locale_facets.h'
inline:  l=450  f='/usr/include/c++/11/bits/basic_ios.h'  p='_ZNKSt5ctypeIcE5widenEc'
inline:  l=685  f='/usr/include/c++/11/ostream'  p='_ZNKSt9basic_iosIcSt11char_traitsIcEE5widenEc'
stmt:  0x10b151  l=877  f='/usr/include/c++/11/bits/locale_facets.h'
inline:  l=450  f='/usr/include/c++/11/bits/basic_ios.h'  p='_ZNKSt5ctypeIcE5widenEc'
inline:  l=685  f='/usr/include/c++/11/ostream'  p='_ZNKSt9basic_iosIcSt11char_traitsIcEE5widenEc'
...
```

I'm seeing at least three kinds of problems.  All of these are running
my unit test with 12 threads on the hpcstruct-bin binary (compiled on ubuntu),
comparing dyninst 12.3.0 and master as of Nov 7, 2023.

(1) Non-determinism in the inline sequences

I ran dyninst 12.3.0 100 times and got the same answer 100 times.

But with dyninst master, I'm seeing things like:

```
$ diff   -u  out-master-33  out-master-72
 block:
 stmt:  0x136cfe  l=0  f=''
 stmt:  0x136d01  l=0  f=''
-inline:  l=36  f='/usr/include/c++/11/bits/stl_algo.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
+inline:  l=36  f='/usr/include/c++/11/bits/stl_algobase.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
 stmt:  0x136d03  l=0  f=''
-inline:  l=36  f='/usr/include/c++/11/bits/stl_algo.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
+inline:  l=36  f='/usr/include/c++/11/bits/stl_algobase.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
 stmt:  0x136d06  l=0  f=''
-inline:  l=36  f='/usr/include/c++/11/bits/stl_algo.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
+inline:  l=36  f='/usr/include/c++/11/bits/stl_algobase.h'  p='_ZN5boost6systemeqERKNS0_15error_conditionES3_'
```

Here, the file names are different: `stl_algo.h` vs `stl_algobase.h`.
This reminds me of the problem identified in

https://github.com/dyninst/dyninst/issues/1469

where there were multiple file name tables and something was being
looked up in the wrong table, except in #1469 the issue was the line
map info and here it's the inline sequence.

This seems to happen mostly when the line map info is unknown (`l=0 f=''`).

----------

(2) Non-determinism and missing line map info.

Again, with dyninst master, I'm seeing:

```
$ diff -u out-master-10  out-master-22

 block:
-stmt:  0x1094c1  l=0  f=''
+stmt:  0x1094c1  l=195  f='/usr/include/c++/11/bits/basic_string.h'
...
-stmt:  0x1094c5  l=0  f=''
+stmt:  0x1094c5  l=239  f='/usr/include/c++/11/bits/basic_string.h'
```

Here, sometimes dyninst finds line map info, sometimes it goes
missing.  Note: this is two runs of the same dyninst (master), not two
versions of dyninst.

----------

(3) Missing line map info compared to 12.3.0.

Here, we're comparing the output from 12.3.0 with master.

```
$ diff -u out-1230-1 out-master-10

-stmt:  0x10404c  l=151  f='/home/krentel/non-det/hpctoolkit/build/src/tool/hpcstruct/../../../../src/tool/hpcstruct/main.cpp'
+stmt:  0x10404c  l=0  f=''
...
-stmt:  0x1040c9  l=148  f='/home/krentel/non-det/hpctoolkit/build/src/tool/hpcstruct/../../../../src/tool/hpcstruct/main.cpp'
-stmt:  0x1040cd  l=148  f='/home/krentel/non-det/hpctoolkit/build/src/tool/hpcstruct/../../../../src/tool/hpcstruct/main.cpp'
-stmt:  0x1040d0  l=148  f='/home/krentel/non-det/hpctoolkit/build/src/tool/hpcstruct/../../../../src/tool/hpcstruct/main.cpp'
+stmt:  0x1040c9  l=0  f=''
+stmt:  0x1040cd  l=0  f=''
+stmt:  0x1040d0  l=0  f=''
```

Here, 12.3.0 finds line map info, but sometimes master does not.  It's
pretty much always this direction.  That is, master is missing the
info, never master finds something that 12.3.0 misses.

Also, the output from master is maybe 8-10% smaller than 12.3.0, so
it's missing a fair amount of something.

----------

Note, the API for asking for line map info has changed from 12.3.0 to
master.  In 12.3.0, I was using:

```
    set <Module *> modSet;
    the_symtab->findModuleByOffset(modSet, addr);

    for (auto mit = modSet.begin(); mit != modSet.end(); ++mit) {
        (*mit)->getSourceLines(svec, addr);
        if (! svec.empty()) {
            break;
        }
    }
```

So, we get a set of modules and try each one.
But in master, I'm using:

```
    mod = the_symtab->getContainingModule(addr);
    if (mod != NULL) {
        mod->getSourceLines(svec, addr);
    }
```

Here, `getContainingModule()` returns just one module and either it
has line map info or not.

I don't know if that's a factor, perhaps some other module has line
map info that getContainingModule() misses.

So, if I'm using the wrong code to ask for the line map info for a given
address, please let me know.

----------

I'll copy the input binary (STRUCT), test program (openmp-inline.cpp)
and sample output files onto ufront in:

ufront:  /home/krentel/Dyninst-non-det/

All of the above was compiled and run on ubuntu 22.04.  You're right
that if I use an input file compiled on RH, then I get the same
problems, but vastly more of them.

ping @hainest @kupsch @bigtrak 

 - What platform (x86, x86_64, Power8, Power9, ARMv8)
 - What commit? (Or branch)
 - Maybe libc version

x86_64, Ubuntu 22.04, dyninst master as of Nov 7, 2023 (9256b319789d).

hainest:
The PC value can be non-unique across CUs. For example, they can all be 0x0 for a PIE binary. The offset of the CU is unique as it's the location inside of the .debug_info table.

Fixes #1625

### DIFF
--- a/dwarf/h/dwarf_cu_info.h
+++ b/dwarf/h/dwarf_cu_info.h
@@ -97,6 +97,12 @@ namespace Dyninst { namespace DwarfDyninst {
    *
    */
   inline Dwarf_Die *find_cu(Dwarf *dbg, Dwarf_Addr addr, Dwarf_Die *result) {
+    // The .debug_info DIE offset is the starting address for modules.
+    if(dwarf_offdie(dbg, addr, result)) {
+      return result;
+    }
+
+    // Search for the given address (assumes .debug_aranges is present)
     Dwarf_Die cuDIE{};
     if (dwarf_addrdie(dbg, addr, &cuDIE)) {
       *result = cuDIE;
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -2193,7 +2193,7 @@ bool Object::fix_global_symbol_modules_static_dwarf() {
 
         dwarf_printf("Locating ranges for module '%s' at offset 0x%zx\n", modname.c_str(), loc);
         std::vector<AddressRange> mod_ranges = DwarfWalker::getDieRanges(cu_die);
-        auto *m = associated_symtab->getContainingModule(loc);
+        auto *m = associated_symtab->findModuleByOffset(loc);
         if(!m) {
           m = new SymtabAPI::Module(lang_Unknown, loc, modname, associated_symtab);
         }
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -2189,17 +2189,13 @@ bool Object::fix_global_symbol_modules_static_dwarf() {
 
         std::string modname = DwarfDyninst::cu_name(cu_die);
 
-        Address tempModLow;
-        Address modLow = 0;
-        if (DwarfWalker::findConstant(DW_AT_low_pc, tempModLow, &cu_die, dbg)) {
-            convertDebugOffset(tempModLow, modLow);
-        }
+        auto const loc = dwarf_dieoffset(&cu_die);
 
-        dwarf_printf("Locating ranges for module '%s' at offset 0x%zx\n", modname.c_str(), modLow);
+        dwarf_printf("Locating ranges for module '%s' at offset 0x%zx\n", modname.c_str(), loc);
         std::vector<AddressRange> mod_ranges = DwarfWalker::getDieRanges(cu_die);
-        auto *m = associated_symtab->getContainingModule(modLow);
+        auto *m = associated_symtab->getContainingModule(loc);
         if(!m) {
-          m = new SymtabAPI::Module(lang_Unknown, modLow, modname, associated_symtab);
+          m = new SymtabAPI::Module(lang_Unknown, loc, modname, associated_symtab);
         }
         dwarf_printf("Adding %zu ranges to '%s'\n", mod_ranges.size(), m->fileName().c_str());
         for (auto r = mod_ranges.begin(); r != mod_ranges.end(); ++r)
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -262,12 +262,12 @@ bool DwarfWalker::parseModule(Dwarf_Die moduleDIE, Module *&fixUnknownMod) {
 
     // Find the Symtab Module corresponding to this CU
     mod() = [this]() {
-      SymtabAPI::Module* m = symtab()->findModuleByOffset(modLow);
+      SymtabAPI::Module* m = symtab()->findModuleByOffset(offset());
       if(m) return m;
       return symtab()->getDefaultModule();
     }();
 
-    dwarf_printf("Mapped to Symtab module %p from '%s' at offset %zx\n", (void*)mod(), mod()->fileName().c_str(), modLow);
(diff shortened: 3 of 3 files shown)


kupsch (approved):

mwkrentel:
As I said in #1625, I think this last commit here solves 99% of the
problems I raised in that issue.  So, I suggest this should be
committed and we declare victory.

I still see some odd problems with binaries compiled on RedHat 8.x
that are a bit different.  But I think that's worth a separate issue.
Plus, I'm not sure if it's the dwarf that's broken or a bug in
dyninst, or what.

hainest:
> As I said in #1625, I think this last commit here solves 99% of the problems I raised in that issue. So, I suggest this should be committed and we declare victory.

I will try to work on the thread safety issue soon. I'm trying hard to get Capstone working before the end of the year, but I know that bug is a real pain for you right now.

> I still see some odd problems with binaries compiled on RedHat 8.x that are a bit different. But I think that's worth a separate issue. Plus, I'm not sure if it's the dwarf that's broken or a bug in dyninst, or what.

Bolo (@bigtrak) has seen some issues, as well. However, I've not seen anything on the multiple RHEL 8.X systems I test on (both x86 and ppc). Definitely let us know if you are having problems there.

PR #1633, opened 2023-11-24
Update x86 DWARF registers maps

hainest:
I will add the missing control registers in a separate PR after this is merged.


common/src/registers/MachRegister.C
```diff
@@ -1093,8 +1110,99 @@ namespace Dyninst {
...
+          case 1059: return Dyninst::x86::k0;
+          case 1060: return Dyninst::x86::k2;
+          case 1061: return Dyninst::x86::k4;
+          case 1062: return Dyninst::x86::k6;
+          case 1063: return Dyninst::x86::oeax;
+          case 1064: return Dyninst::x86::gsbase;
+          case 1065: return Dyninst::x86::tr1;
+          case 1066: return Dyninst::x86::tr3;
+          case 1067: return Dyninst::x86::tr5;
+          case 1068: return Dyninst::x86::tr7;
```
kupsch, line 1205: Is there a reason that the odd  cr, dr, ymm, zmm, and k registers are missing, along with the even tr registers?

common/src/registers/MachRegister.C
```diff
@@ -1117,144 +1225,204 @@ namespace Dyninst {
...
+          case 1096: return Dyninst::x86_64::zmm21;
+          case 1097: return Dyninst::x86_64::zmm23;
+          case 1098: return Dyninst::x86_64::zmm25;
+          case 1099: return Dyninst::x86_64::zmm27;
+          case 1100: return Dyninst::x86_64::zmm29;
+          case 1101: return Dyninst::x86_64::zmm31;
+          case 1102: return Dyninst::x86_64::tr0;
+          case 1103: return Dyninst::x86_64::tr2;
+          case 1104: return Dyninst::x86_64::tr4;
+          case 1105: return Dyninst::x86_64::tr6;
```
kupsch, line 1425: why are the odd tr registers missing?

common/src/registers/MachRegister.C
```diff
@@ -1117,144 +1225,204 @@ namespace Dyninst {
...
+          case 121: return Dyninst::x86_64::k3;
+          case 122: return Dyninst::x86_64::k4;
+          case 123: return Dyninst::x86_64::k5;
+          case 124: return Dyninst::x86_64::k6;
+          case 125: return Dyninst::x86_64::k7;
+          case 126: return Dyninst::InvalidReg;
+          case 127: return Dyninst::InvalidReg;
+          case 128: return Dyninst::InvalidReg;
+          case 129: return Dyninst::InvalidReg;
+
```
kupsch, line 1341: The Oct 2023 ABI doc maps 130-145 to r16-r31 (the new registered defined by APX). These can likely be ignored for now.

kupsch (approved):

common/src/registers/MachRegister.C
hainest: Because the perl script I was using to generate these had a huge bug in it that I didn't notice. Fixed now.

common/src/registers/MachRegister.C
hainest: Same as above.

common/src/registers/MachRegister.C
hainest: Yup. Capstone doesn't support AMX yet because it's based on LLVM 15 which has very poor support for them. The next release of Capstone will be based on at least LLVM 17 which has full support. I'll probably add the TMM registers soon just to have them.

PR #1634, opened 2023-11-24
Synchronize x86 mnemonics with Capstone

hainest:
This only adds the mnemonics. It does not include instruction decoding.


hainest:
This broke the CFG test (test1_33) on x86. I'm guessing one of the branch instructions is wrong.

hainest:
@kupsch The problem was in aliasing cbw with cwde. This cause e_cwde to map to ROSE's x86_cbw. In dataflowAPI/rose/x86_64InstructionSemantics.h, the code for x86_cbw is commented out- because of course it is. I don't feel like fighting this, so I just undid the change and added it to my notes to deal with later.

Tests are passing now.

kupsch (approved):

PR #1636, opened 2023-11-28
Add x86_{i386,64} memory-management and x87 control/status registers

hainest:
These have explicit DWARF mappings in their respective ABI docs.


common/h/registers/x86_64_regs.h
```diff
@@ -266,6 +274,8 @@ namespace Dyninst { namespace x86_64 {
   DEF_REGISTER(       st5,    0x5 | FPDBL |   X87 | Arch_x86_64, "x86_64");
   DEF_REGISTER(       st6,    0x6 | FPDBL |   X87 | Arch_x86_64, "x86_64");
   DEF_REGISTER(       st7,    0x7 | FPDBL |   X87 | Arch_x86_64, "x86_64");
+  DEF_REGISTER(       fcw,    0x8 | W_REG |   X87 | Arch_x86_64, "x86_64");
+  DEF_REGISTER(       fsw,    0x9 | W_REG |   X87 | Arch_x86_64, "x86_64");
```
kupsch, line 278: There needs to be a new category type for `fcw` and `fsw` (and `mxcsr`) such as `FPCTL`, otherwise `getBaseRegister` will convert these to a register with category `<id> | W_REG | ST | Arch_x86_64` which is incorrect.  `getBaseRegister` should return the register unmodified.

common/h/registers/x86_64_regs.h
```diff
@@ -308,6 +318,7 @@ namespace Dyninst { namespace x86_64 {
   DEF_REGISTER(     xmm29,   0x1D |  XMMS |   XMM | Arch_x86_64, "x86_64");
   DEF_REGISTER(     xmm30,   0x1E |  XMMS |   XMM | Arch_x86_64, "x86_64");
   DEF_REGISTER(     xmm31,   0x1F |  XMMS |   XMM | Arch_x86_64, "x86_64");
+  DEF_REGISTER(     mxcsr,   0x20 | D_REG |   XMM | Arch_x86_64, "x86_64");
```
kupsch, line 321: `XMM` category needs to be changed as above to a new category such as `FPCTL`.

kupsch (changes requested):
Looks good except the category for `fcw`, `fsw`, and `mxcsr` needs to a new category.

common/h/registers/x86_64_regs.h
hainest: Ah, that's right. I forgot about that. I actually started out making a new category for them, but then changed my mind at the last second...

common/h/registers/x86_64_regs.h
hainest: Same as above.

common/h/registers/x86_regs.h
```diff
@@ -80,6 +80,7 @@ namespace Dyninst { namespace x86 {
   const signed int YMM    = 0x000B0000;  // YMM0-YMM7 Registers from AVX2/FMA
   const signed int ZMM    = 0x000C0000;  // ZMM0-ZMM7 Registers from AVX-512
   const signed int KMASK  = 0x000D0000;  // K0-K7 opmask Registers from AVX-512
+  const signed int FPCTL  = 0x000E0000;  // 32-bit x87 and SSE control registers
```
kupsch, line 83: both 32 and 64-bit and mxcsr is used by avx so change comment to `x87, SSE and AVX control/status registers`.

common/h/registers/x86_regs.h
```diff
@@ -238,6 +249,7 @@ namespace Dyninst { namespace x86 {
   DEF_REGISTER(      xmm5,   0x05 |  XMMS |   XMM | Arch_x86, "x86");
   DEF_REGISTER(      xmm6,   0x06 |  XMMS |   XMM | Arch_x86, "x86");
   DEF_REGISTER(      xmm7,   0x07 |  XMMS |   XMM | Arch_x86, "x86");
+  DEF_REGISTER(     mxcsr,   0x08 |  FULL | FPCTL | Arch_x86, "x86");
```
kupsch, line 252: mxcsr is 32-bits so `FULL` should be `D_REG`

common/h/registers/x86_64_regs.h
```diff
@@ -77,6 +77,7 @@ namespace Dyninst { namespace x86_64 {
   const signed int YMM    = 0x000B0000;  // YMM0-YMM15 Registers from AVX2/FMA (YMM0-YMM31 for AVX-512)
   const signed int ZMM    = 0x000C0000;  // ZMM0-ZMM31 Registers from AVX-512
   const signed int KMASK  = 0x000D0000;  // K0-K7 opmask Registers from AVX-512
+  const signed int FPCTL  = 0x000E0000;  // 32-bit x87 and SSE control registers
```
kupsch, line 80: both 32 and 64-bit and mxcsr is used by avx so change comment to `x87, SSE and AVX control/status registers`.

kupsch (approved):
Good to go after fixing the x86 `mxcsr` size, and the `FPCTL` comment.

common/h/registers/x86_regs.h
hainest: I had only noticed the mention of it in the prose for SSE. Looking at the tables for AVX, it's definitely there. I guess whoever wrote that didn't feel it necessary to explicitly call it out. Good catch.

common/h/registers/x86_regs.h
hainest: This is for x86_i386, so FULL is 32-bit. It is D_REG in x86_64_regs.h.

common/h/registers/x86_64_regs.h
kupsch: I was looking at the wrong definition of FULL.

PR #1643, opened 2023-12-13
Clean up InstructionAPI::Operation

hainest:
I ultimately decided not to rename the file, even though it really should be.

### DIFF
--- a/instructionAPI/h/Expression.h
+++ b/instructionAPI/h/Expression.h
@@ -122,7 +122,6 @@ namespace Dyninst
     public:
       /// \brief A type definition for a reference counted pointer to a %Expression.
       typedef boost::shared_ptr<Expression> Ptr;
-      friend class Operation_impl;
     protected:      
       Expression(Result_Type t);
       Expression(MachRegister r);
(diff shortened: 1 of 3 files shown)


kupsch (approved):
Looks Good.  If you want to do further cleanup:

1. copy ctor, and assignment operator could be removed (implicit versions are equivalent)
2. the assignment operator return value should not be `const` (fixed by 1)
3. default member inititialization would cleanup and allow the default ctor to be `= default`

hainest:
> Looks Good. If you want to do further cleanup:
> 
>     1. copy ctor, and assignment operator could be removed (implicit versions are equivalent)
> 
>     2. the assignment operator return value should not be `const` (fixed by 1)
> 
>     3. default member inititialization would cleanup and allow the default ctor to be `= default`

I _think_ these are here since once_flag isn't copyable or movable. I would be surprised if Operation is every copied or moved. The flag will go away with Capstone because it's only used to initialize a bunch of data that Capstone gives us. At that time, I'll remove all of these SMFs.

PR #1644, opened 2023-12-13
Fix register check in LivenessAnalyzer::isMMX

hainest:
There is no requirement that the two values for MMX be the same between the two architectures.

### DIFF
--- a/dataflowAPI/src/liveness.C
+++ b/dataflowAPI/src/liveness.C
@@ -625,10 +625,10 @@ void LivenessAnalyzer::clean(Function *func){
 bool LivenessAnalyzer::isMMX(MachRegister machReg){
 	auto const arch = machReg.getArchitecture();
 	if (arch == Arch_x86) {
-		return (machReg.val() & 0x00ff0000) == x86::MMX;
+		return machReg.regClass() == x86::MMX;
 	}
 	if (arch == Arch_x86_64) {
-		return (machReg.val() & 0x00ff0000) == x86_64::MMX;
+		return machReg.regClass() == x86_64::MMX;
 	}
 	return false;
 }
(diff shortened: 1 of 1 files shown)


dataflowAPI/src/liveness.C
```diff
@@ -623,10 +623,13 @@ void LivenessAnalyzer::clean(Function *func){
 }
 
 bool LivenessAnalyzer::isMMX(MachRegister machReg){
-	if ((machReg.val() & Arch_x86) == Arch_x86 || (machReg.val() & Arch_x86_64) == Arch_x86_64){
-		assert( ((machReg.val() & x86::MMX) == x86::MMX) == ((machReg.val() & x86_64::MMX) == x86_64::MMX) );
+	auto const arch = machReg.getArchitecture();
+	if (arch == Arch_x86) {
 		return (machReg.val() & x86::MMX) == x86::MMX;
```
kupsch, line 628: The old test is not correct.  MMX has the value `0x00090000` so any category with those two bits set matches so second bytes of `*9` `*B` `*D` and `*F` matches.  The correct expression should be `(machReg.val() & 0x00ff0000) == x86::MMX`.

kupsch (changes requested):
This change would be correct, but there is an existing bug that should be fixed in both return expressions.

dataflowAPI/src/liveness.C
hainest: Oh my. I thought that check looked weird, but I was just heading in to fix the crazy assert. Good catch.

hainest:
@kupsch I just realized I closed the conversation, but forgot to write a message that I had updated the checks. Could you take a look?

kupsch (approved):

PR #1646, opened 2023-12-14
Use Capstone for x86 instruction decoding

### LINKED ISSUE #1881: fucompp incorrectly decoded
**x86**
`fucompp %st1 reads {ST1, ST0, }, writes {ST1, ST0, }`
Missing write to x86::fsw.

x86_64:
`fucompp %<invalid_reg> reads {<INVALID_REG>, <INVALID_REG>, }, writes {<INVALID_REG>, <INVALID_REG>, }`
Really broken.

Capstone is missing read from st(1):

```
 0  da e9                                            fucompp	
	ID: 743 (fucompp)
	Prefix:0x00 0x00 0x00 0x00 
	Opcode:0xda 0x00 0x00 0x00 
	rex: 0x0
	addr_size: 4
	modrm: 0xe9
	disp: 0x0
	sib: 0x0
	Registers read: st(0)   <----
	Registers modified: fpsw
	FPU_FLAGS: MOD_C0 MOD_C1 MOD_C2 MOD_C3
	Groups: fpu
```

hainest:
This replaces #638.

This is still a work in progress, but I want to get the code out for review. There are several items I want to check on before even considering merging:

- [x] Verify all decode cases from `InstructionDecoder_x86` have been captured
- [x] Check completeness and accuracy of `decode_reg`
- [x] Check completeness and accuracy of `decode_imm`
- [x] Check completeness and accuracy of `decode_mem`
- [x] Handle segment registers in `decode_mem`
- [x] Replace calculations in `Operation::SetUpNonOperandData` (at least for x86)
- [x] Fix error handling in `decode_mem`
- [x] Update container builds to include Capstone source
- [ ] Check that the description of `Instruction::getWriteSet/getReadSet` is correct

Big items as separate PRs
- [ ] Get rid of old decode tables
- [ ] Check consistency of register usages in dataflowAPI, parseAPI, etc. For example, popf{d,q} are not uniformly checked in switches.
- [ ] Try to replace manual register checks with generalized concepts and put them into new methods in the Instruction class

**decoder.C**
- [ ] x86_decoder::decode_operands
  - Don't need to check for endbr32/64 because no operands have been set on the `insn` object (it's also const here).

**Operation.C**
- [x] Rename file
  -  File is no longer needed
- [x] Why is Operation_impl inheriting from boost::lockable_adapter?
  - It shouldn't.
- [x] Remove `Operation(NS_X86::ia32_entry*, ...)`
  - Fixed by #2169
- [x] ~~Why can't `SetUpNonOperandData` be done in the ctor?~~ It's not needed. Removed.
  - this would get rid of mutables and let the accessors be 'const' again
  - Why is it x86 only?
  - read{write}Flags checks are incomplete.
- [x] `Operation(entryID id,...)` should check all architecture values in switch; no default
  - `archDecodedFrom` is no longer used
- [x] ~~Move once_flag ito pImpl~~ It's not needed. Removed.
  - Get rid of SMFs.
- [ ] Use updateMnemonic throughout Dyninst instead of the idiom 'Instruction::getOperation().mnemonic = ...'
  - Probably make them friends and move updateMnemonic to private
  - Then can remove non-const Instruction::getOperation

**instructionAPI/h/InstructionCategories.{h,C}**
- [ ] Update entryToCategory after getting new Capstone mnemonics
- [ ] entryToCategory
  - This is just buggered
- [ ] makeReturnExpression should use u32/u64 depending on `arch_decoded_from`.
  - This makes its use in x86_decoder::decode_operands consistent with the expr created in InstructionDecoder_x86::decodeOperands:1649.

**instruction.C**
- [ ] Remove DEBUG_INSN_ALLOCATIONS
- [ ] makeReturnExpression should be synchronized with InstructionDecoderImpl::makeRegisterExpression

**x86{_64}_regs.h**
- [ ] Remove tr* registers (they only exist for Dyninst's internal decoder)
- [ ] Possibly remove MISC registers

**entryIDs**
- [ ] popf{d,q} are not uniformly checked in switches
- [ ] collapse all e_ret* into just e_ret
  - Capstone will generate aliases for them, but we should not treat them distinctly
  - For example, convertOpcodes.C:159 should go away
- [ ] Remove e_fp_generic
- [ ] Remove e_int80
- [ ] cwde and cwdq should alias cbw, but Dyninst has separate semantics for them (see x86_64InstructionSemantics.h) even though it doesn't have separate decode table entries for them.

**BoundFactCalculator.C**
- [ ] 457: Assumes all SETxx entry ids are contiguous

**x86{_64}_regs.h**
- [ ] Check which instructions read/write IOPL bit of EFLAGS

**instructionAPI/InstructionTest.C**
- [ ] Keep or throw away?
  - throw it away. It has fixed mnemonics that won't necessarily match capstone's.

---
Lower priority items:
- [ ] Handle implicit x87 control/state registers

Fixes #1881 (modulo missing reg read)


kupsch (dismissed):
OK.  Just one incorrect comment.

instructionAPI/src/x86/decoder.C
```diff
@@ -0,0 +1,359 @@
...
+#include "entryIDs.h"
+#include "registers/x86_regs.h"
+#include "registers/x86_64_regs.h"
+#include "registers/MachRegister.h"
+
+#include <cstdint>
+#include <map>
+#include <vector>
+
+namespace di = Dyninst::InstructionAPI;
```
bbiiggppiigg, line 15: I'm not entirely sure.
Is this a good practice to shorten namespace like this?
Or is it better to keep them as Dyninst::instructionAPI::Result_Type?
Or maybe these are not needed cause I assume decoder is in the instructionAPI namespace?

bbiiggppiigg (dismissed):
I left a comment on the use of shortened namespace.

instructionAPI/src/x86/register-xlat.C
```diff
@@ -0,0 +1,531 @@
...
+          case X86_REG_FP0:
+          case X86_REG_FP1:
+          case X86_REG_FP2:
+          case X86_REG_FP3:
+          case X86_REG_FP4:
+          case X86_REG_FP5:
+          case X86_REG_FP6:
+          case X86_REG_FP7:
+
+          /* AMX registers */
```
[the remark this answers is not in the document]
hainest: Fixed.

instructionAPI/src/x86/decoder.C
hainest: It's all good. I just wouldn't do this in the global namespace in a public header.

hainest:
@kupsch @wxrdnx I merged master. You should be able to fast forward, unless you've made conflicting local changes. Let me know if you have any issues.

PR #1650, opened 2023-12-24
Write to stderr in ParseAPI::parsing_printf

hainest:
Using a per-thread file here is cumbersome. Including the thread id in the message and synchronizing vprintf makes this work for multiple threads.

### DIFF
--- a/parseAPI/src/debug_parse.C
+++ b/parseAPI/src/debug_parse.C
@@ -74,7 +74,14 @@ int Dyninst::ParseAPI::parsing_printf_int(const char *format, ...)
       int v{};
       #pragma omp critical
       {
-        v = vfprintf(stderr, msg.c_str(), va);
+#ifdef __clang__
+        #pragma clang diagnostic push
+        #pragma clang diagnostic ignored "-Wformat-nonliteral"
+#endif
+        v = vprintf(msg.c_str(), va);
+#ifdef __clang__
+        #pragma clang diagnostic pop
+#endif
         fflush(stdout);
       }
       return v;
--- a/parseAPI/src/debug_parse.C
+++ b/parseAPI/src/debug_parse.C
@@ -82,7 +82,7 @@ int Dyninst::ParseAPI::parsing_printf_int(const char *format, ...)
 #ifdef __clang__
         #pragma clang diagnostic pop
 #endif
-        fflush(stdout);
+        fflush(stderr);
       }
       return v;
     }();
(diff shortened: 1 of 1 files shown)


parseAPI/src/debug_parse.C
```diff
@@ -59,23 +57,36 @@ int Dyninst::ParseAPI::parsing_printf_int(const char *format, ...)
...
     va_start(va,format);
-    int ret = vfprintf(log_file, format, va);
-    fflush(log_file);
-    va_end(va);
 
+    auto ret = [&]() {
+      int v{};
+      #pragma omp critical
+      {
+#ifdef __clang__
```
kupsch, line 77: To preserve the -Wformat-nonliteral, which is a useful security property, I'd suggest removing the creation of a new format string, by just using `fprintf(stderr, "[thread %u] ", id);` and then the `vfprintf(stderr, format, va);` without the `#pragma`'s.

parseAPI/src/debug_parse.C
```diff
@@ -59,23 +57,36 @@ int Dyninst::ParseAPI::parsing_printf_int(const char *format, ...)
...
+        #pragma clang diagnostic push
+        #pragma clang diagnostic ignored "-Wformat-nonliteral"
+#endif
+        v = vfprintf(stderr, msg.c_str(), va);
+#ifdef __clang__
+        #pragma clang diagnostic pop
+#endif
+        fflush(stdout);
```
kupsch, line 85: `fflush(stderr)` instead of `stdout`.

parseAPI/src/debug_parse.C
hainest: Not sure why I didn't think of that... Changed.

parseAPI/src/debug_parse.C
hainest: Derp. Fixed.

parseAPI/src/debug_parse.C
```diff
@@ -59,23 +57,28 @@ int Dyninst::ParseAPI::parsing_printf_int(const char *format, ...)
...
-    int ret = vfprintf(log_file, format, va);
-    fflush(log_file);
-    va_end(va);
 
+    auto ret = [&]() {
+      int v{};
+      #pragma omp critical
+      {
+        v = fprintf(stderr, "[thread %d] ", id);
+        v &= vfprintf(stderr, format, va);
```
kupsch, line 76: The `&=` doesn't make sense.  fprintf returns the number of characters written or -1 on error.  You should record each return value separately in `r1` and `r2` with a return value of `r1 != -1 && r2 != -1) ? (r1 + r2) : -1`.  Slightly better would be to also only execute the second `vprintf` if `r1 != -1`.  It is also likely that nothing looks at the return value.

kupsch (changes requested):
slight change in error handling.

kupsch (approved):

PR #1653, opened 2023-12-27
[WIP] Move documentation to RST/readthedocs

hainest:
There is a lot to cover here, but I wanted to get an alpha preview up to start a conversation. When reviewing, skip everything before 'dataflowAPI init'- that stuff is all naked imports and me trying out different styles of formatting. What's in dataflowAPI is how I would recommend formatting everything else.

There are still many notes I need to add to the README.


hainest:
@kupsch @bbiiggppiigg @rochauha Could you take a look at the TODO items in the description and provide feedback if/when you have time? The instructions for building are in docs/README.md.

bbiiggppiigg:
So what I should do is build the docs, and then check if the TODO items can be checked off based on the built documentation ?

hainest:
> So what I should do is build the docs, and then check if the TODO items can be checked off based on the built documentation ?

Yup. You don't have to necessarily go through each TODO item. I'm really looking for feedback on the usability of the docs in general.

bbiiggppiigg:
> > So what I should do is build the docs, and then check if the TODO items can be checked off based on the built documentation ?
> 
> Yup. You don't have to necessarily go through each TODO item. I'm really looking for feedback on the usability of the docs in general.

I checked that I can build the html doc, and the navigation seems fine to me.

For the layout, I don't know if it would be better for the developer docs for an API to be directly below the corresponding toolkit API. Even though the separation of developer doc from toolkit API make sense, the developer docs on its own looks kind of arbitrary (Maybe because the file are named kind of arbitrarily). 

Also, I thought the documentation should be centric around C++ classes, but right now it seems to be centric around files, which I'm not sure if it is an issue or not.

I don't know too much about writing RSTs. This commit is too huge that my chrome will hang when I navigate the files changed tab.

hainest:
> > > So what I should do is build the docs, and then check if the TODO items can be checked off based on the built documentation ?
> > 
> > 
> > Yup. You don't have to necessarily go through each TODO item. I'm really looking for feedback on the usability of the docs in general.
> 
> I checked that I can build the html doc, and the navigation seems fine to me.

Roughly how long did it take to build on your machine? I forgot to include a patch to enable parallel builds, so I'll do that.

> For the layout, I don't know if it would be better for the developer docs for an API to be directly below the corresponding toolkit API. Even though the separation of developer doc from toolkit API make sense, the developer docs on its own looks kind of arbitrary (Maybe because the file are named kind of arbitrarily).

I took the hint from other readthedocs pages on separating them into different sections.

> Also, I thought the documentation should be centric around C++ classes, but right now it seems to be centric around files, which I'm not sure if it is an issue or not.

Funny enough, this is something I wanted to change during this migration. I just completely forgot about it as I slogged through the existing "documentation". I definitely agree we should do this.

> 
> I don't know too much about writing RSTs.

Yeah, I need to include a small example with links to the relevant RST docs.

>This PR is too huge that my chrome will hang when I navigate the files changed tab.

No surprises. GitHub's interface is ass.

docs/README.md
```diff
@@ -0,0 +1,26 @@
+# Documentation Instructions
+
+To prepare the local docs for preview:
+
+```bash
+$ python -m venv env
```
kupsch, line 6: - `python` doesn't exist in Ubuntu 22.04; `python3` should work everywhere.
- removing the `$` prompt would allow easier copy and paste into a shell

docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+$ source env/bin/activate
+$ pip install -r requirements.txt
+$ `cd $(find env/lib -type d -name sphinxcontrib) && wget https://raw.githubusercontent.com/wpilibsuite/sphinxext-remoteliteralinclude/main/sphinxext/remoteliteralinclude.py`
```
kupsch, line 9: This happens to work, but is not what is intended.  The command substitution ``` `cmd` ``` creates a new shell and then evaluates its stdout as bash. In this case the only output is to stderr so nothing gets executed. Using just a subshell `(cmd)` is what is intended (to not make the `cd` change the current shell).  Better yet would be to use `wget`'s `--directory-prefix` or `-P` to set the output directory which eliminates the `cd` and hence the need for a subshell:

```shell
wget -P $(find env/lib -type d -name sphinxcontrib) https://raw.githubusercontent.com/wpilibsuite/sphinxext-remoteliteralinclude/main/sphinxext/remoteliteralinclude.py 
```

bbiiggppiigg:
Now I think about it, a page for runtime debug environmental variable can really help.

docs/README.md
hainest: Agreed. Updated.

docs/README.md
hainest: Oh, I didn't know that wget could do that. Guess I should read the man pages more often. Fixed.

hainest:
> Now I think about it, a page for runtime debug environmental variable can really help.

That's a good idea. Added to TODO.

hainest:
@kupsch I'm still looking at getting the latex to build. There seems to be certain permutations of directories that cause it to fail. I'm trying to find the minimum set to see if I can narrow it down to a particular file. Based on what I could find in forums, it seems like having circular references in the latex generated by sphinx doesn't work very well. I have a suspicion that may be what's going on here.

hainest:
@kupsch After more experimenting, it really is the case that the generated tex file is too big to be parsed by pdflatex. I found that using luatex worked.

```
make latex
cd _build/latex/
make LATEXMKOPTS="-lualatex"
```

The navigation pane for the resulting PDF is terrible, but I can't figure out how to make RST create the same structure in PDF as it does for HTML. One problem at a time, I guess.

As an aside, I've had no problems using parallel builds by enabling `parallel_write_safe` in env/lib/python3.10/site-packages/sphinxcontrib/remoteliteralinclude.py. Then you can use `make latex SPHINXOPTS='-jN'`.

hainest:
@kwryankrattiger The reason this PR stalled is because the tex files generated by sphinx when creating a PDF version of the docs are too large for pdflatex to consume- it exhausts the 64GB of memory on my local machine. They _can_ be rendered using luatex, but there are lots of issues that I never figured out how to fix (https://github.com/dyninst/dyninst/pull/1653#issuecomment-2046160142). If you have any experience with this, ideas are very much welcome.

PR #1669, opened 2024-01-11
Fix detection of LEA instruction variants as NOP

batuzovk:
LEA instruction can sometimes be used as NOP variant. There were two issues with detecting it (both fixed in this pull request):

* In Dyninst LEA doesn't dereference its memory argument, thus Instruction::getMemoryReadOperands() always returns empty set causing the first 'if' to always fail.
* We should check for displacement only if the instruction uses the same register for reading and writing.

In practice this resulted in some 'LEA REG1, [REG2]' (with REG1 != REG2) instructions being incorrectly identified as NOPs, and consequently causing additional basic blocks splits due to 'nop block'. The real-life example of this is glibc (e.g. libc.so.6 from Debian or Ubuntu distributions):

```
   4c5c8:	0f 89 8a 00 00 00    	jns    4c658 <__xpg_basename@@GLIBC_2.2.5+0x248>
   4c5ce:	4c 8d 00             	lea    (%rax),%r8
   4c5d1:	48 8b 04 de          	mov    (%rsi,%rbx,8),%rax
   4c5d5:	4c 8d 0a             	lea    (%rdx),%r9
   4c5d8:	48 f7 e1             	mul    %rcx
   4c5db:	4c 8d 18             	lea    (%rax),%r11
   4c5de:	48 8b 44 de 08       	mov    0x8(%rsi,%rbx,8),%rax
   4c5e3:	48 8d 2a             	lea    (%rdx),%rbp
   4c5e6:	eb 49                	jmp    4c631 <__xpg_basename@@GLIBC_2.2.5+0x221> 
```

Instruction at address 4c5ce was detected as NOP and split into a separate basic block.

In theory there is one more LEA variant that is also a NOP instruction, but I haven't seen it in practice yet.

### DIFF
--- a/parseAPI/src/IA_x86.C
+++ b/parseAPI/src/IA_x86.C
@@ -131,24 +131,28 @@ bool isNopInsn(Instruction insn)
         return true;
     if(insn.getOperation().getID() == e_lea)
     {
-        std::set<Expression::Ptr> memReadAddr;
-        insn.getMemoryReadOperands(memReadAddr);
+        // LEA doesn't dereference memory expression
+        std::set<RegisterAST::Ptr> readRegs;
+        insn.getReadSet(readRegs);
         std::set<RegisterAST::Ptr> writtenRegs;
         insn.getWriteSet(writtenRegs);
 
-        if(memReadAddr.size() == 1 && writtenRegs.size() == 1)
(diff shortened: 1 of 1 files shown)


hainest:
Subsumed by #1678.

PR #1670, opened 2024-01-11
Improve support of different ways to implement syscall

batuzovk:
Dyninst supports different architectures, yet in Parser.C system calls are detected by explicitly comparing instruction to e_syscall, which is an x86_64 instruction. The result was a segmentation fault on 32-bit code like this:

```
_start:
    mov     $0, %ebx
    mov     $1, %eax
    int     $0x80
```

This change aims at moving architecture-specific parts of syscall detection to architecture-specific files and using generalized interface in common code.

This code doesn't handle e_sysenter because it is not listed in `man 2 syscall` and I'm not sure what the convention regarding it is.

### DIFF
--- a/instructionAPI/src/InstructionCategories.C
+++ b/instructionAPI/src/InstructionCategories.C
@@ -113,6 +113,7 @@ namespace Dyninst
       case e_sysenter:
 	return c_SysEnterInsn;
       case e_syscall:
+      case e_int:
     return c_SyscallInsn;
           default:
 	return c_NoCategory;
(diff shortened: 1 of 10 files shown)


hainest:
I ran into this problem in working on the Capstone migration. I have a fix over there, but it's the same as here. This PR contains bug fixes, refactorings, and new code, so I'd like to split each of them into their own PR. I can do this, unless you want to (I can list out my recommendations).

batuzovk:
> I ran into this problem in working on the Capstone migration. I have a fix over there, but it's the same as here. This PR contains bug fixes, refactorings, and new code, so I'd like to split each of them into their own PR. I can do this, unless you want to (I can list out my recommendations).

I don't mind either way for as long as it gets fixed in the end. If it would be easier for you to explain, then I'll do it. If it would be easier for you to do it yourself, than I have no objections either.

hainest:
These are now all present as of #1692. Thanks for the bug report and code fixes!

PR #1672, opened 2024-01-11
IndirectAnalyzer: fix potential segfault

ilil01:
When processing a `call` instruction, the `IndirectControlFlowAnalyzer::FindAllThunks` method assumes that both the instruction and its target are located in the same `Region`. This is not always the case &mdash; for example, in 32-bit libc there are instructions in `__libc_freeres_fn` section calling functions from `.text`. In such cases this method provokes a segmentation fault since the result of the `getPtrToInstruction` method is not tested against `nullptr`.


bbiiggppiigg:
If the fix is just adding a if(target) check around the decoder and the for loop, it looks reasonable to me.
However, the indentation seems to be a bit off when viewed in the browser.

bbiiggppiigg (approved):

kupsch:
@ilil01, can you open a [bug report issue](https://github.com/dyninst/dyninst/issues/new?template=bug_report.md) for this problem with the information on how to reproduce? I am unable to reproduce this with a simple test. This will also allow us to make a more robust fix in the future.

ilil01:
> @ilil01, can you open a [bug report issue](https://github.com/dyninst/dyninst/issues/new?template=bug_report.md) for this problem with the information on how to reproduce? I am unable to reproduce this with a simple test. This will also allow us to make a more robust fix in the future.

I've opened issue https://github.com/dyninst/dyninst/issues/1680

kupsch:
This prevents the crash. Fixing issue #1680, will be a more complete fix to find the correct region and instruction.

PR #1678, opened 2024-02-09
Correctly detect NOP variants of x86 lea instruction

hainest:
Extends and subsumes #1669.

Works for all of these cases:

```
  lea rax, [rax]
  lea rax, [rax + rax*1 + 0 ]
  lea rax, [rax + rax*1 + 1 ]
  lea rax, [rax + 0]
  lea rax, [rax + 1]
  lea rax, [rax - 1]
  lea rax, [1 + rax]
  lea rax, [rax*1]
  lea rax, [1*rax]
  lea rax, [rax*1 + 0]
  lea rax, [rax*1 + 1]
  lea rax, [rbx]
  lea rax, [rbx + rbx*1 + 0 ]
  lea rax, [rbx + rbx*1 + 1 ]
  lea rax, [rbx + 0]
  lea rax, [rbx + 1]
  lea rax, [rbx - 1]
  lea rax, [1 + rbx]
  lea rax, [rbx*1]
  lea rax, [1*rbx]
  lea rax, [rbx*1 + 0]
  lea rax, [rbx*1 + 1]
  lea rax, [ 12*7 + 2]
```

No test regressions on x86.

### DIFF
--- a/parseAPI/src/IA_x86.C
+++ b/parseAPI/src/IA_x86.C
@@ -91,66 +91,132 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
     return false;
 }
 
-class nopVisitor : public InstructionAPI::Visitor
-{
-    public:
-        nopVisitor() : foundReg(false), foundImm(false), foundBin(false), isNop(true) {}
-        virtual ~nopVisitor() {}
+class leaVisitor : public Visitor {
+public:
+  Immediate *imm{};
+  BinaryFunction *bf{};
 
-        bool foundReg;
-        bool foundImm;
-        bool foundBin;
-        bool isNop;
-
-        virtual void visit(BinaryFunction*)
-        {
-            if (foundBin) isNop = false;
-            if (!foundImm) isNop = false;
-            if (!foundReg) isNop = false;
-            foundBin = true;
-        }
-        virtual void visit(Immediate *imm)
-        {
-            if (imm != 0) isNop = false;
-            foundImm = true;
-        }
-        virtual void visit(RegisterAST *)
-        {
-            foundReg = true;
-        }
-        virtual void visit(Dereference *)
-        {
-            isNop = false;
-        }
+  void visit(BinaryFunction *bf_) override {
+    bf = bf_;
+  }
+  void visit(Immediate *imm_) override {
+    imm = imm_;
+  }
+  void visit(RegisterAST *) override {
+    // Don't care about registers
+  }
+  void visit(Dereference *) override {
+    // Can't have a deref in a lea calculation
+  }
 };
 
-bool isNopInsn(Instruction insn)
-{
-    // TODO: add LEA no-ops
-    if(insn.getOperation().getID() == e_nop)
-        return true;
-    if(insn.getOperation().getID() == e_lea)
-    {
-        std::set<Expression::Ptr> memReadAddr;
-        insn.getMemoryReadOperands(memReadAddr);
-        std::set<RegisterAST::Ptr> writtenRegs;
-        insn.getWriteSet(writtenRegs);
+bool isNopInsn(Instruction insn) {
+  if (insn.getOperation().getID() == e_nop)
+    return true;
 
-        if(memReadAddr.size() == 1 && writtenRegs.size() == 1)
-        {
-            if(**(memReadAddr.begin()) == **(writtenRegs.begin()))
-            {
-                return true;
-            }
-        }
-        // Check for zero displacement
-        nopVisitor visitor;
+  if (insn.getOperation().getID() == e_lea) {
+    // Get all registers read or written
+    std::set<RegisterAST::Ptr> readRegs;
+    insn.getReadSet(readRegs);
+    std::set<RegisterAST::Ptr> writtenRegs;
+    insn.getWriteSet(writtenRegs);
 
-	// We need to get the src operand
-        insn.getOperand(1).getValue()->apply(&visitor);
-        if (visitor.isNop) return true; 
+    // If more than one register is read or written, it can't be a nop
+    if(readRegs.size() != 1 || writtenRegs.size() != 1) {
+      return false;
     }
-    return false;
+
+    auto src_reg_ast = **(readRegs.begin());
+    auto dest_reg_ast = **(writtenRegs.begin());
+
+    // If source and destination are different, it's not a nop
+    if(!(src_reg_ast == dest_reg_ast)) {
+      return false;
+    }
+
+    auto is_index_identity = [](Immediate *imm) {
+      auto const index = imm->eval().convert<int64_t>();
+      if(index == 1LL) {
+        // gas doesn't allow 'index' to be negative or zero, but
+        // this requirement doesn't appear to be in the Intel SDM.
+        return true;
+      }
+      return false;
+    };
+
+    auto is_scale_identity = [](Immediate *imm) {
+      if(!imm) {
+        // Something went wrong
+        return false;
+      }
+      auto const scale = imm->eval().convert<int64_t>();
+      if(scale != 0LL) {
+        // Scale can be positive or negative
+        // [base*index+0x0] is a legal addressing form
+        return false;
+      }
+      return true;
+    };
+
+    auto is_nop_baseindex = [&is_index_identity](BinaryFunction *bf) {
+      // The AST could be [base*index] or [index*base], so check both
+      std::vector<Expression::Ptr> children;
+      bf->getChildren(children);
+
+      for(auto child : children) {
+        leaVisitor index;
+        child->apply(&index);
+        if(index.imm && is_index_identity(index.imm)) {
+          return true;
+        }
+      }
+      return false;
+    };
+
+    auto is_nop_scale = [&is_scale_identity](BinaryFunction *bf) {
+      // The AST could be [bi+scale] or [scale+bi], so check both
+      std::vector<Expression::Ptr> children;
+      bf->getChildren(children);
+
+      for(auto child : children) {
+        leaVisitor scale;
+        child->apply(&scale);
+        if(scale.imm && is_scale_identity(scale.imm)) {
+          return true;
+        }
+      }
+      return false;
+    };
+
+    // The addressing form has the structure base*index+scale
+    leaVisitor whole_form;
+    insn.getOperand(1).getValue()->apply(&whole_form);
+
+    if(!whole_form.bf) {
+      // No index, no scale. Instruction is 'lea reg, [reg]'.
+      return true;
+    }
+
+    if(whole_form.bf->isMultiply()) {
+      // Form is definitely [base*index], check the index
+      if(!is_nop_baseindex(whole_form.bf)) {
+        // Instruction is 'lea reg, [reg * N]', N != 1
+        return false;
+      }
+    }
+
+    // Form is definitely [base*index+scale]
+    // Arbitrarily check the scale first
+    if(!is_nop_scale(whole_form.bf)) {
+      // Instruction is 'lea reg, [reg * N + M]', M != 0
+      return false;
+    }
+
+    // Form is definitely base*index+0x0, check the index
+    return is_nop_baseindex(whole_form.bf);
+  }
+
+  return false;
 }
 
 bool IA_x86::isNop() const


hainest:
I'm using this to test. There's currently on way to test `IA_x86::isNop`.

```c++
#include "BinaryFunction.h"
#include "Dereference.h"
#include "Expression.h"
#include "Immediate.h"
#include "Instruction.h"
#include "Operand.h"
#include "Register.h"
#include "Visitor.h"

class leaVisitor : public di::Visitor {
public:
  di::Immediate *imm{};
  di::BinaryFunction *bf{};

  void visit(di::BinaryFunction *bf_) override {
    bf = bf_;
  }
  void visit(di::Immediate *imm_) override {
    imm = imm_;
  }
  void visit(di::RegisterAST *) override {
    // Don't care about registers
  }
  void visit(di::Dereference *) override {
    // Can't have a deref in a lea calculation
  }
};

bool isNopInsn(di::Instruction insn) {
  if (insn.getOperation().getID() == e_nop)
    return true;

  if (insn.getOperation().getID() != e_lea) {
    return false;
  }
  // Get all registers read or written
  std::set<di::RegisterAST::Ptr> readRegs;
  insn.getReadSet(readRegs);
  std::set<di::RegisterAST::Ptr> writtenRegs;
  insn.getWriteSet(writtenRegs);

  // If more than one register is read or written, it can't be a nop
  if(readRegs.size() != 1 || writtenRegs.size() != 1) {
    return false;
  }
  
  auto src_reg_ast = **(readRegs.begin());
  auto dest_reg_ast = **(writtenRegs.begin());
  
  // If source and destination are different, it's not a nop
  if(!(src_reg_ast == dest_reg_ast)) {
    return false;
  }

  // At this point, the addressing form has the structure base*index+scale
  // where 'base' is a register but 'index' and 'scale' may not be present.
  leaVisitor whole_form;
  insn.getOperand(1).getValue()->apply(&whole_form);
  if(!whole_form.bf) {
    // No index, no scale. Instruction is 'lea reg, [reg]'.
    return true;
  }
  
  auto is_index_identity = [](di::Immediate *imm) {
    auto const index = imm->eval().convert<int64_t>();
    if(index == 1LL) {
      // gas doesn't allow 'index' to be negative or zero, but
      // this requirement doesn't appear to be in the Intel SDM.
      return true;
    }
    return false;
  };
  
  auto is_scale_identity = [](di::Immediate *imm) {
    if(!imm) {
      // Something went wrong
      return false;
    }
    auto const scale = imm->eval().convert<int64_t>();
    if(scale != 0LL) {
      // Scale can be positive or negative
      // [base*index+0x0] is a legal addressing form
      return false;
    }
    return true;
  };

  auto is_nop_baseindex = [&is_index_identity](di::BinaryFunction *bf) {
    // The AST could be [base*index] or [index*base], so check both
    std::vector<di::Expression::Ptr> children;
    bf->getChildren(children);
    
    for(auto child : children) {
      leaVisitor index;
      child->apply(&index);
      if(index.imm && is_index_identity(index.imm)) {
        return true;
      }
    }
    return false;
  };
  
  auto is_nop_scale = [&is_scale_identity](di::BinaryFunction *bf) {
    // The AST could be [bi+scale] or [scale+bi], so check both
    std::vector<di::Expression::Ptr> children;
    bf->getChildren(children);
    
    for(auto child : children) {
      leaVisitor scale;
      child->apply(&scale);
      if(scale.imm && is_scale_identity(scale.imm)) {
        return true;
      }
    }
    return false;
  };
  
  if(whole_form.bf->isMultiply()) {
    // Form is definitely [base*index], check the index
    if(!is_nop_baseindex(whole_form.bf)) {
      // Instruction is 'lea reg, [reg * N]', N != 1
      return false;
    }
  }

  // Form is definitely [base*index+scale]
  // Arbitrarily check the scale first
  if(!is_nop_scale(whole_form.bf)) {
    // Instruction is 'lea reg, [reg * N + M]', M != 0
    return false;
  }
  
  // Form is definitely base*index+0x0, check the index
  return is_nop_baseindex(whole_form.bf);
}

int main() {
  // clang-format off
  std::array<const unsigned char, 126> buffer = {
    0x48, 0x8d, 0x00,                                 // (0) lea rax, [rax]
    0x48, 0x8d, 0x04, 0x00,                           // (1) lea rax, [rax + rax*1 + 0 ]
    0x48, 0x8d, 0x44, 0x00, 0x01,                     // (2) lea rax, [rax + rax*1 + 1 ]
    0x48, 0x8d, 0x00,                                 // (3) lea rax, [rax + 0]
    0x48, 0x8d, 0x40, 0x01,                           // (4) lea rax, [rax + 1]
    0x48, 0x8d, 0x40, 0xff,                           // (5) lea rax, [rax - 1]
    0x48, 0x8d, 0x40, 0x01,                           // (6) lea rax, [1 + rax]
    0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (7) lea rax, [rax*1]
    0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (8) lea rax, [1*rax]
    0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (9) lea rax, [rax*1 + 0]
    0x48, 0x8d, 0x04, 0x05, 0x01, 0x00, 0x00, 0x00,   // (10) lea rax, [rax*1 + 1]
    0x48, 0x8d, 0x03,                                 // (11) lea rax, [rbx]
    0x48, 0x8d, 0x04, 0x1b,                           // (12) lea rax, [rbx + rbx*1 + 0 ]
    0x48, 0x8d, 0x44, 0x1b, 0x01,                     // (13) lea rax, [rbx + rbx*1 + 1 ]
    0x48, 0x8d, 0x03,                                 // (14) lea rax, [rbx + 0]
    0x48, 0x8d, 0x43, 0x01,                           // (15) lea rax, [rbx + 1]
    0x48, 0x8d, 0x43, 0xff,                           // (16) lea rax, [rbx - 1]
    0x48, 0x8d, 0x43, 0x01,                           // (17) lea rax, [1 + rbx]
    0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (18) lea rax, [rbx*1]
    0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (19) lea rax, [1*rbx]
    0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (20) lea rax, [rbx*1 + 0]
    0x48, 0x8d, 0x04, 0x1d, 0x01, 0x00, 0x00, 0x00,   // (21) lea rax, [rbx*1 + 1]
    0x48, 0x8d, 0x04, 0x25, 0x56, 0x00, 0x00, 0x00,   // (22) lea rax, [ 12*7 + 2]  (lea rax,ds:0x56)
  };
  // clang-format on

  di::InstructionDecoder decoder(buffer.data(), buffer.size(),
                                 Dyninst::Architecture::Arch_x86_64);

  di::Instruction i;
  int cnt=0;
  do {
    i = decoder.decode();
    if (i.isValid()) {
      std::cout << "instruction (" << cnt << "): --"
                << std::boolalpha << isNopInsn(i) << "-- "
                << i.format() << '\n';
      cnt++;
    }
  } while (i.isValid());
}
```

batuzovk:
I would suggest renaming variables/functions to match the naming scheme used in Intel/AMD manuals. Right now code uses the same words manuals use, but uses them for different entities. This is very confusing. AMD manual is more detailed, so I'll be referencing it. Intel manual uses the same names for the same fields.

The part of the instruction in question in encoded in SIB byte. The following formula is given in vol 3 (General-Purpose and System Instructions), chapter 1 (Instruction Encoding), section 1.4 (ModRM and SIB Bytes):

```
effective_address = scale * index + base + offset
```

where:

* **base** is the base register,
* **index** is the index register,
* **scale** is the scale factor index register is multiplied by (immediate),
* **offset** (more often called **displacement** ) is the offset added to the result (immediate).

This corresponds to Intel notation like `[RBX + 4 * RAX + 0x123]` where RBX is the base, RAX is the index, 4 is the scale factor, 0x123 is the displacement. All parts are optional (within reasons, i.e. you can't have index without scale).

The scale factor is encoded in 2 bits and can be 1, 2, 4, or 8. No negative values or other numbers are possible.

batuzovk:
After thinking a bit more about the proposed solution I've spotted one issue with the algorithm. It tries to check different cases of possible AST, but it doesn't provide any outline why all the cases were checked. In fact, for arbitrary AST it is not even the case. For example, if topmost operation is multiplication, and one of its arguments is 1, the other argument isn't even checked. In other words, something like `1 * (REG + 1)` would be identified as identity AST. I think all ASTs that can appear in LEA arguments are covered, but it is hard to check (and nearly impossible to prove).

Clear semantic is the key here, so I suggest the following modification of the visitor.

Given an AST operating on one register (and possibly multiple immediate values) visitor classifies it as one of the following:

* always 0,
* always 1,
* identity, i.e. always has the same value as REG,
* something else.

The classifier is conservative, i.e. it will never classify as `identity` something that is not, but it can classify as `something else` what in fact is identity. The classifier is exact if there are no operations in the AST other than addition and multiplication.

```cpp
#include "BinaryFunction.h"
#include "Dereference.h"
#include "Expression.h"
#include "Immediate.h"
#include "Instruction.h"
#include "Operand.h"
#include "Register.h"
#include "Visitor.h"

class leaVisitor : public di::Visitor {
    private:
        enum ValType {
            VAL_TYPE_UNKNOWN = 0,
            VAL_TYPE_ZERO,
            VAL_TYPE_ONE,
            VAL_TYPE_IDENTITY
        };

        std::unordered_map<di::Expression *, ValType> val_type;
    public:
        bool isIdentity(di::Expression *e) {
            if (val_type.find(e) == val_type.end()) return false;
            return val_type[e] == VAL_TYPE_IDENTITY;
        }

        void visit(di::BinaryFunction *bf_) override {
            if (bf_->isMultiply()) {
                // 1 * <something> has the same type as <something>
                // 0 * <something> is 0
                std::vector<di::Expression::Ptr> children;
                bf_->getChildren(children);
                assert(children.size() == 2);
                if (val_type[children[0].get()] == VAL_TYPE_ONE) {
                    val_type[bf_] = val_type[children[1].get()];
                } else if (val_type[children[1].get()] == VAL_TYPE_ONE) {
                    val_type[bf_] = val_type[children[0].get()];
                } else if (val_type[children[0].get()] == VAL_TYPE_ZERO) {
                    val_type[bf_] = VAL_TYPE_ZERO;
                } else if (val_type[children[1].get()] == VAL_TYPE_ZERO) {
                    val_type[bf_] = VAL_TYPE_ZERO;
                } else {
                    val_type[bf_] = VAL_TYPE_UNKNOWN;
                }
            } else if (bf_->isAdd()) {
                // 0 + <something> has the same type as <something>
                std::vector<di::Expression::Ptr> children;
                bf_->getChildren(children);
                assert(children.size() == 2);
                if (val_type[children[0].get()] == VAL_TYPE_ZERO) {
                    val_type[bf_] = val_type[children[1].get()];
                } else if (val_type[children[1].get()] == VAL_TYPE_ZERO) {
                    val_type[bf_] = val_type[children[0].get()];
                } else {
                    val_type[bf_] = VAL_TYPE_UNKNOWN;
                }
            } else {
                // LEA shouldn't have any other operations
                val_type[bf_] = VAL_TYPE_UNKNOWN;
            }
        }
        void visit(di::Immediate *imm_) override {
            int64_t val = imm_->eval().convert<int64_t>();
            if (val == 0) {
                val_type[imm_] = VAL_TYPE_ZERO;
            } else if (val == 1) {
                val_type[imm_] = VAL_TYPE_ONE;
            } else {
                val_type[imm_] = VAL_TYPE_UNKNOWN;
            }
        }
        void visit(di::RegisterAST *reg_) override {
            // There is only one register used in this AST and it is the one we need
            val_type[reg_] = VAL_TYPE_IDENTITY;
        }
        void visit(di::Dereference *d) override {
            // Can't have a deref in a lea calculation
            val_type[d] = VAL_TYPE_UNKNOWN;
        }
};

bool isNopInsn(di::Instruction insn) {
    if (insn.getOperation().getID() == e_nop)
        return true;

    if (insn.getOperation().getID() != e_lea) {
        return false;
    }
    // Get all registers read or written
    std::set<di::RegisterAST::Ptr> readRegs;
    insn.getReadSet(readRegs);
    std::set<di::RegisterAST::Ptr> writtenRegs;
    insn.getWriteSet(writtenRegs);

    // If more than one register is read or written, it can't be a nop
    if(readRegs.size() != 1 || writtenRegs.size() != 1) {
        return false;
    }

    di::RegisterAST src_reg_ast = **(readRegs.begin());
    di::RegisterAST dest_reg_ast = **(writtenRegs.begin());

    // If source and destination are different, it's not a nop
    if(!(src_reg_ast == dest_reg_ast)) {
        return false;
    }

    leaVisitor whole_form;
    insn.getOperand(1).getValue()->apply(&whole_form);

    return whole_form.isIdentity(insn.getOperand(1).getValue().get());
}

int main() {
    // clang-format off
    std::array<const unsigned char, 126> buffer = {
        0x48, 0x8d, 0x00,                                 // (0) lea rax, [rax]
        0x48, 0x8d, 0x04, 0x00,                           // (1) lea rax, [rax + rax*1 + 0 ]
        0x48, 0x8d, 0x44, 0x00, 0x01,                     // (2) lea rax, [rax + rax*1 + 1 ]
        0x48, 0x8d, 0x00,                                 // (3) lea rax, [rax + 0]
        0x48, 0x8d, 0x40, 0x01,                           // (4) lea rax, [rax + 1]
        0x48, 0x8d, 0x40, 0xff,                           // (5) lea rax, [rax - 1]
        0x48, 0x8d, 0x40, 0x01,                           // (6) lea rax, [1 + rax]
        0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (7) lea rax, [rax*1]
        0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (8) lea rax, [1*rax]
        0x48, 0x8d, 0x04, 0x05, 0x00, 0x00, 0x00, 0x00,   // (9) lea rax, [rax*1 + 0]
        0x48, 0x8d, 0x04, 0x05, 0x01, 0x00, 0x00, 0x00,   // (10) lea rax, [rax*1 + 1]
        0x48, 0x8d, 0x03,                                 // (11) lea rax, [rbx]
        0x48, 0x8d, 0x04, 0x1b,                           // (12) lea rax, [rbx + rbx*1 + 0 ]
        0x48, 0x8d, 0x44, 0x1b, 0x01,                     // (13) lea rax, [rbx + rbx*1 + 1 ]
        0x48, 0x8d, 0x03,                                 // (14) lea rax, [rbx + 0]
        0x48, 0x8d, 0x43, 0x01,                           // (15) lea rax, [rbx + 1]
        0x48, 0x8d, 0x43, 0xff,                           // (16) lea rax, [rbx - 1]
        0x48, 0x8d, 0x43, 0x01,                           // (17) lea rax, [1 + rbx]
        0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (18) lea rax, [rbx*1]
        0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (19) lea rax, [1*rbx]
        0x48, 0x8d, 0x04, 0x1d, 0x00, 0x00, 0x00, 0x00,   // (20) lea rax, [rbx*1 + 0]
        0x48, 0x8d, 0x04, 0x1d, 0x01, 0x00, 0x00, 0x00,   // (21) lea rax, [rbx*1 + 1]
        0x48, 0x8d, 0x04, 0x25, 0x56, 0x00, 0x00, 0x00,   // (22) lea rax, [ 12*7 + 2]  (lea rax,ds:0x56)
    };
    // clang-format on

    di::InstructionDecoder decoder(buffer.data(), buffer.size(),
            Dyninst::Architecture::Arch_x86_64);

    di::Instruction i;
    int cnt=0;
    do {
        i = decoder.decode();
        if (i.isValid()) {
            std::cout << "instruction (" << cnt << "): --"
                      << std::boolalpha << isNopInsn(i) << "-- "
                      << i.format() << '\n';
            cnt++;
        }
    } while (i.isValid());
} 
```

hainest:
> I would suggest renaming variables/functions to match the naming scheme used in Intel/AMD manuals. Right now code uses the same words manuals use, but uses them for different entities. This is very confusing. AMD manual is more detailed, so I'll be referencing it. Intel manual uses the same names for the same fields.
> 
> The part of the instruction in question in encoded in SIB byte. The following formula is given in vol 3 (General-Purpose and System Instructions), chapter 1 (Instruction Encoding), section 1.4 (ModRM and SIB Bytes):

I had tunnel vision and was only searching for 'SIB' which (quite naturally) isn't in the section describing the 'effective address' calculation. Interestingly, that's in Vol 1, 3.7.5 in the version I'm using (Dec 2023) which seems like quite a bit of drift...

Fixed.

hainest:
> After thinking a bit more about the proposed solution I've spotted one issue with the algorithm. It tries to check different cases of possible AST, but it doesn't provide any outline why all the cases were checked.

There are comments about which pieces of the equation are ruled out at each step. They can undoubtedly be expanded upon- particularly now that I have a correct picture of the effective address calculation.

>In fact, for arbitrary AST it is not even the case.

We don't need to handle arbitrary AST here. In fact, I shouldn't have added the extra checks for commutativity. The AST for memory access is generated in [makeModRMExpression](https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/InstructionDecoder-x86.C#L208), and is constructed exactly as `base + (index*scale) + displacement`.

batuzovk:
I think that implementation that doesn't rely on particular form of the AST would be less prone to breaking unexpectedly, but I don't insist.

Meanwhile I've run into an actual example of LEA being used as NOP in 32-bit glibc, and I don't think that current implementation detects it as one (but my variant did):

```
  199716:	c3                   	ret
  199717:	8d b4 26 00 00 00 00 	lea    0x0(%esi,%eiz,1),%esi
  19971e:	66 90                	xchg   %ax,%ax
  199720:	53                   	push   %ebx 
```

`%eiz` here is some kind of pseudo-register that exists purely for text representation of the instruction. 64-bit variant of this instruction will get additional `0x48` prefix (REX I think?)

```
    0x48, 0x8d, 0xb4, 0x26, 0x00, 0x00, 0x00, 0x00,   // (23) lea rsi, [rsi + 1 * riz + 0] 
```

Basically this is `lea rsi, [rsi]` with redundant SIB byte. I'm not sure how to generate it from text representation, but you can specify it in binary form for gas `.byte 0x48, 0x8d, 0xb4, 0x26, 0x00, 0x00, 0x00, 0x00`

kupsch:
Thanks for the new example.  I'm writing a general a new patch that will simplify all opcode expressions to an equivalent register if one exists (without any specific knowledge of an intel addressing modes). I will post a new PR soon.

hainest:
For "8d b4 26 00 00 00 00", Capstone gives the mnemonic

```
8d b4 26 00 00 00 00  lea esi, [rsi + riz]
```

Assembling that mnemonic with gcc/gas and dumping with objdump gives

8d b6 00 00 00 00   lea esi,[rsi+0x0]

which is a different opcode. On the face of it, gcc appears to substitute
riz->0x0, but it produces a different opcode. That's weird. Capstone
generates the same mnemonic as objdump for "8db6000000", so that's good.

Dyninst decodes "8d b4 26 00 00 00 00" as

lea esi, [rsi+0x0]

which would be a nop if we allow aliasing. However, the current check
for equality of register ASTs does not:

```c++
if(!(src_reg_ast == dest_reg_ast)) {
  return false;
}
```

Assembling "48 8d b4 26 00 00 00 00" with gcc and dumping with objdump gives

```
0:  48 8d b4 26 00 00 00  lea    rsi,[rsi+riz*1+0x0]
7:  00
```

Dyninst says it's

```
lea rsi, [rsi+0x0]
```

and Capstone says it's

```console
48 8d b4 26 00 00 00 00    lea  rsi, [rsi + riz]
```

No one can agree on what to do with the pseudoregister and how it affects
the opcode.

1) Do we just want to say that pseudo registers can be ignored
for the purposes of detecting a nop?

2) Do we want to allow aliased registers to be flagged as a nop?

    I don't think we should because the value of the destination and source
    will be different at the end of the instruction.
    
    The SDM table "64-bit Mode LEA Operation with Address and Operand Size
    Attributes" says that if the destination has fewer bits than the address,
    the value is truncated. If the destination has more bits, then the
    address is zero-extended.

PR #1681, opened 2024-02-14
correctly detect LEA instructions as NOP

kupsch:
LEA is a NOP if and only if the two operands refer to the same register after simplifying additive and multiplicative identity operations

Extends and subsumes #1669 & #1678

@hainest and @batuzovk please take a look at this and give me feedback.  This correctly handles all instructions presented so far.


parseAPI/src/IA_x86.C
```diff
@@ -91,68 +94,77 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
...
-        {
-            foundReg = true;
+        void visit(RegisterAST *reg) override {
+            exprStack.push(reg);
         }
-        virtual void visit(Dereference *)
-        {
-            isNop = false;
+        void visit(Dereference *deref) override {
+            exprStack.push(deref);
```
batuzovk, line 129: Don't we need exprStack.pop() here? It is not going to affect LEA instruction, but if someone tries to use the same visitor for other ASTs it might.

batuzovk (commented):
I'm a bit worried about possible implicit upcasting of RegisterAST* to Expression* somewhere, but I'm not C++ guru. If you are sure having both base class and derived one in the same variant will work reliably, then I'm fine with it.

Everything else looks good to me.

parseAPI/src/IA_x86.C
kupsch: I added a reset() method that resets the visitor to the initial state. It is unused as the iterator is never resused, but could be if there is a need to reuse the iterator.  The only other change is whitespace and a comment.

kupsch: Sorry, I misunderstood your comment. You are correct that there should be a pop. I have pushed a new version.

batuzovk (approved):
Looks good to me

hainest:
@kupsch I'm reviewing. Was that force-push just a rebase on master?

kupsch:
It was just a rebase and a comment change.  The code is no different than what batuzovk reviewed.

parseAPI/src/IA_x86.C
```diff
@@ -91,68 +94,81 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
...
+            }  else  {
+                exprStack.push(bf);   // no simplification, use BinaryFunction
+            }
         }
-        virtual void visit(Immediate *imm)
-        {
-            if (imm != 0) isNop = false;
-            foundImm = true;
+        void visit(Immediate *imm) override {
+            auto const value = imm->eval().convert<uint64_t>();
```
hainest, line 122: It won't make any difference to the calculation in `visit(BinaryFunction*)`, but displacements can be signed values. The scalar field of a SIB byte is only positive.

batuzovk:
> The pseudoregister `riz` in `lea rsi, [rsi+riz*1+0x0]` is not considered to be read. This reduces the expression to `rsi+0x0`.

I don't think it is necessary to mention those riz and eiz anywhere at all. Those aren't even pseudo-registers in full sense, they are just crutches for text disassemblers that encountered some weird combination of ModRM and SIB.

From encoding perspective `index` field of SIB is 100b = RSP. RSP can't be an index register, thus this value is treated as a special case according to documentation:

> Register specification is null. The scale*index portion of the indexed register-indirect effective address is set to 0.

Nothing is read, no mentions of any pseudo-registers.

Text disassemblers tend to get confused by it since without index part memory operand can be encoded with ModRM alone, and there is no need to introduce SIB (unless someone was reading from stack or making a very long NOP instruction like in our case). From Dyninst perspective it is just "no index part" of the expression, there is no need to introduce new entities because of it.

> If the source and destination registers are different sizes, then the instruction is not considered a nop. For example, `lea eax, [rax]`.

`lea eax, [rax]` is definitely not a NOP because it sets high 32 bits of RAX to zero. Same goes for `mov eax, eax`, which is a NOP in 32-bit code, but not a NOP in 64-bit code. I don't see any reason to mention this case specifically.

parseAPI/src/IA_x86.C
```diff
@@ -91,68 +94,83 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
...
-        {
-            if(**(memReadAddr.begin()) == **(writtenRegs.begin()))
-            {
-                return true;
-            }
+    if (insn.getOperation().getID() == e_lea) {
+        std::vector<Operand> operands;
+        insn.getOperands(operands);
+
+        if (operands.size() != 2)  {
```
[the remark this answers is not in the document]
kupsch: added

parseAPI/src/IA_x86.C
```diff
@@ -91,68 +94,81 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
...
+        operands[0].getValue()->apply(&op1Visitor);
+        operands[1].getValue()->apply(&op2Visitor);
+
+        auto op1Reg = op1Visitor.getRegister();
+        auto op2Reg = op2Visitor.getRegister();
+
+        // both operands simplify to registers and they are the same
+        auto r = op1Reg && op2Reg && *op1Reg == *op2Reg;
+
+        return r;
```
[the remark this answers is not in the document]
kupsch: done

parseAPI/src/IA_x86.C
```diff
@@ -91,68 +94,81 @@ bool IA_x86::isFrameSetupInsn(Instruction i) const
...
-        virtual void visit(RegisterAST *)
-        {
-            foundReg = true;
+        void visit(RegisterAST *reg) override {
+            exprStack.push(reg);
         }
-        virtual void visit(Dereference *)
-        {
-            isNop = false;
+        void visit(Dereference *deref) override {
```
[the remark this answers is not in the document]
kupsch: Diagnostic added. The presence of deref will result in not a nop.

parseAPI/src/IA_x86.C
kupsch: Changed.

kupsch:
@hainest, I have pushed a new version that addresses your review points.

hainest (approved):
No test regressions.

PR #1683, opened 2024-02-16
Fix logic error in IA_IAPI::isInterruptOrSyscall

hainest:
This was broken by e7db16f5 in 2010.

Originally part of #1670


kupsch (approved):

hainest:
This is causing some versions of glibc to throw `__gnu_cxx::recursive_init_error` when initializing the static in `isSyscall` when enough threads are running. I'll fix that first.

hainest:
No test regressions.

PR #1684, opened 2024-02-18
Refactor IA_IAPI::isSyscall,isInterrupt into arch-specific files

hainest:
This also requires those members be defined for all classes derived from IA_IAPI.

Originally part of #1670


kupsch (approved):
Looks good.  The formatting for additions to arch-specific header files is inconsistent and should be adjusted.

hainest:
> Looks good. The formatting for additions to arch-specific header files is inconsistent and should be adjusted.

Yeah, I noticed that. I finished testing, so I'll fix that.

PR #1686, opened 2024-02-20
Improve syscall and interrupt instruction detection

hainest:
Was originally part of #1670.

### DIFF
--- a/instructionAPI/src/InstructionCategories.C
+++ b/instructionAPI/src/InstructionCategories.C
@@ -113,6 +113,7 @@ namespace Dyninst
       case e_sysenter:
 	return c_SysEnterInsn;
       case e_syscall:
+      case e_int:
     return c_SyscallInsn;
           default:
 	return c_NoCategory;
--- a/parseAPI/src/IA_aarch64.C
+++ b/parseAPI/src/IA_aarch64.C
@@ -275,7 +275,7 @@ bool IA_aarch64::isNopJump() const
 }
 
 bool IA_aarch64::isSyscall() const {
-  return false;
+  return curInsn().getOperation().getID() == aarch64_op_svc;
 }
 
 bool IA_aarch64::isInterrupt() const {
--- a/parseAPI/src/IA_power.C
+++ b/parseAPI/src/IA_power.C
@@ -529,8 +529,10 @@ bool IA_power::isNopJump() const
 }
 
 bool IA_power::isSyscall() const {
-  auto const& i = curInsn();
-  return i.getOperation().getID() == power_op_sc;
+  auto const id = curInsn().getOperation().getID();
+
+  // There is also a vectorized form (scv), but Dyninst can't decode it.
+  return id == power_op_sc;
 }
 
 bool IA_power::isInterrupt() const {
--- a/parseAPI/src/IA_x86.C
+++ b/parseAPI/src/IA_x86.C
@@ -828,15 +828,15 @@ bool IA_x86::isInterrupt() const
     return ((ci.getOperation().getID() == e_int) ||
             (ci.getOperation().getID() == e_int3));
 }
+
 /* Check for system call idioms
  *
  * Idioms checked:
  *
  *  syscall
+ *  int <vector> (e.g., 'int 0x80')
  */
 bool IA_x86::isSyscall() const {
-  static RegisterAST::Ptr gs(new RegisterAST(x86::gs));
-  
   auto const id = curInsn().getOperation().getID();
 
   /*
@@ -847,10 +847,20 @@ bool IA_x86::isSyscall() const {
     return true;
   }
 
+  /* Software interrupts
+   *
+   *  All interrupts that specify a vector go directly through the kernel,
+   *  so they are system calls.
+   *
+   *  Interrupts for traps/breakpoints (notable int3, int1, and into) don't
+   *  directly go through the kernel, so aren't considered system calls.
+   */
+  if(id == e_int) {
+    return true;
+  }
+
   Instruction ci = curInsn();
   return (((ci.getOperation().getID() == e_call) &&
               (ci.getOperation().isRead(gs)) &&
-              (ci.getOperand(0).format(ci.getArch()) == "16")) ||
-          (ci.getOperation().getID() == e_int));
+              (ci.getOperand(0).format(ci.getArch()) == "16")));
 }
-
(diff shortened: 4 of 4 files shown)


hainest:
No test regressions.

hainest:
Using the test code below, this PR works as I would expect in the case of 32-bit code. For 64-bit code, I explicitly discount the elf auxvec case. I'm open to discussion on this.

Bonus fun, Dyninst doesn't correctly decode `into` and `int1` in 64-bit mode (Capstone, xed, and bddisasm do). :shrug:

```
Running in x86_64 mode

call $0x10
is syscall: false
is interrupt: false

call 0x10(%rip)
is syscall: false
is interrupt: false

int $0x80
is syscall: true
is interrupt: true

int 3 
is syscall: false
is interrupt: true
```
---
```
Running in x86 mode

call $0x10
is syscall: true
is interrupt: false

call $0x10
is syscall: true
is interrupt: false

int $0x80
is syscall: true
is interrupt: true

int 3 
is syscall: false
is interrupt: true

into 
is syscall: false
is interrupt: true

int1 
is syscall: false
is interrupt: true
```
---
```c++
#include "BinaryFunction.h"
#include "Dereference.h"
#include "Expression.h"
#include "Immediate.h"
#include "Instruction.h"
#include "Operand.h"
#include "Register.h"
#include "Visitor.h"

#include <iostream>
#include <vector>
#include <array>

bool isSyscall(di::Instruction in);
bool isInterrupt(di::Instruction ci);

struct ins_visitor : di::Visitor {
  di::Result r;
  void visit(di::BinaryFunction*) override {}
  void visit(di::Immediate *imm) override {
    r = imm->eval();
  }
  void visit(di::RegisterAST*) override {}
  void visit(di::Dereference*) override {}
};

di::Result segment_offset(di::Expression::Ptr expr) {
  ins_visitor v;
  expr->apply(&v);
  return v.r;
}

int main() {
  std::array<const unsigned char, 20> buffer = {
    0x65, 0xff, 0x14, 0x25, 0x10, 0x00, 0x00, 0x00,   // call	[q|d]word ptr gs:[0x10]
    0x65, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00,         // 32-bit 'call dword ptr gs:0x10', 64-bit 'call qword ptr gs:[rip + 0x10]'
    0xcd, 0x80,                                       // int 0x80
    0xcc,                                             // int3
    0xce,                                             // into
    0xf1                                              // int1
  };

  constexpr auto arch = Dyninst::Architecture::Arch_x86_64;
  
  auto mode = (arch == Dyninst::Architecture::Arch_x86) ? "x86" : "x86_64";
  std::cout << "Running in " << mode << " mode\n\n";
  di::InstructionDecoder decoder(buffer.data(), buffer.size(), arch);

  di::Instruction i;
  do {
    i = decoder.decode();
    if(i.isValid()) {
      std::cout << i.format() << '\n';
      std::cout << "is syscall: " << std::boolalpha << isSyscall(i) << "\n"
                << "is interrupt: " << std::boolalpha << isInterrupt(i) << "\n\n";
    }
  } while (i.isValid());
}

bool isInterrupt(di::Instruction ci) {
    return ((ci.getOperation().getID() == e_int) ||
            (ci.getOperation().getID() == e_int1) ||
            (ci.getOperation().getID() == e_into) ||
            (ci.getOperation().getID() == e_int3));
}

bool isSyscall(di::Instruction in) {
  auto const id = in.getOperation().getID();
  if(id == e_syscall) {
    return true;
  }

  if(id == e_int) {
    return true;
  }

  if(in.getArch() != Dyninst::Arch_x86) {
    return false;
  }

  if(id == e_call) {
    std::vector<di::Operand> operands;
    in.getOperands(operands);

    if(operands.size() != 1) {
      return false;
    }

    auto result = segment_offset(operands[0].getValue());
    if(result.defined && result.convert<int64_t>() == 0x10) {
      return true;
    }
  }

  return false;
}
```

kupsch:
Since the vistor, in addition to visiting the node that the apply is called on, also does a postfix traversal of the expression tree and therefore is called on all nodes in the epxression tree, the code really is looking for a 0x10 anywhere in the expression tree not just at the top level.  Another issue is that the use of GS segment prefix is not checked, so no prefix and FS prefix also return true as a system call.

Here are test cases that should be false in x86-32.  The first is a call that uses 0x10 , the others are not using the GS segment prefix.
```
    0x38, 0x10, 0x00, 0x00, 0x00,                     // call 0x10 + rip
    0x64, 0xff, 0x14, 0x25, 0x10, 0x00, 0x00, 0x00,   // call	[q|d]word ptr fs:[0x10]
    0x64, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00,         // 32-bit 'call dword ptr fs:0x10', 64-bit 'call qword ptr gs:[rip + 0x10]'
    0x65, 0xff, 0x14, 0x25, 0x10, 0x00, 0x00, 0x00,   // call	[q|d]word ptr [0x10]
    0x65, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00,         // 32-bit 'call dword ptr 0x10', 64-bit 'call qword ptr [rip + 0x10]'
```

hainest:
> Since the vistor, in addition to visiting the node that the apply is called on, also does a postfix traversal of the expression tree and therefore is called on all nodes in the epxression tree, the code really is looking for a 0x10 anywhere in the expression tree not just at the top level. Another issue is that the use of GS segment prefix is not checked, so no prefix and FS prefix also return true as a system call.
> 
> Here are test cases that should be false in x86-32. The first is a call that uses 0x10 , the others are not using the GS segment prefix.
> 
> ```
>     0x38, 0x10, 0x00, 0x00, 0x00,                     // call 0x10 + rip
>     0x64, 0xff, 0x14, 0x25, 0x10, 0x00, 0x00, 0x00,   // call	[q|d]word ptr fs:[0x10]
>     0x64, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00,         // 32-bit 'call dword ptr fs:0x10', 64-bit 'call qword ptr gs:[rip + 0x10]'
>     0x65, 0xff, 0x14, 0x25, 0x10, 0x00, 0x00, 0x00,   // call	[q|d]word ptr [0x10]
>     0x65, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00,         // 32-bit 'call dword ptr 0x10', 64-bit 'call qword ptr [rip + 0x10]'
> ```

Bah, yes, of course. I should have thought of these tests. Given that Dyninst doesn't generate AST for segment registers, I think we'll have to table this until we integrate Capstone. The old code didn't work, anyway, so we aren't losing functionality.

kupsch (approved):

hainest:
No test regressions.

PR #1688, opened 2024-02-26
Add CMake option to control symbol visibility

hainest:
Currently, all symbols are hidden by default and
then explicitly made visible. This is the behavior we want for publishing Dyninst libraries, but unit testing the internal functionality requires access to these symbols. This option allows that.

It is explicitly not described in DyninstOptions.cmake because it is not intended to be used by regular users.

### DIFF
--- a/cmake/DyninstLibrarySettings.cmake
+++ b/cmake/DyninstLibrarySettings.cmake
@@ -15,9 +15,19 @@ endif()
 
 set(BUILD_SHARED_LIBS ON)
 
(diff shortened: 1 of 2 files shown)


kupsch (approved):
Looks good.  Should `DYNINST_EXPORT_ALL` also be added as an advanced option in `cmake/DyninstOptions.cmake`?

hainest:
Blerg. ccmake... I guess so.

PR #1691, opened 2024-02-27
Remove explicit check for system calls in parse_frame_one_iteration

hainest:
165f19cd and 091929d6 changed the parsing semantics of system calls such that software interrupts (e.g., `int` on x86) are always interpreted as system calls. The `hasCFT()` check on line 1824 now covers all of the system call cases because it considers those instructions to have a control flow target (CFT).

Its non-use was verified against /usr/libx32/libc.so.6, /usr/lib/i386-linux-gnu/libc.so.6, /usr/lib/x86_64-linux-gnu/libc.so.6, and /usr/lib32/libc.so.6 on Ubuntu 22.04.


bbiiggppiigg (approved):

hainest:
No idea what that CI Fedora-37 check failure was. No testing regressions.

PR #1692, opened 2024-02-27
Unify system call detection

hainest:
I'll refactor the interrupt detection in a separate PR.

### DIFF
--- a/instructionAPI/src/syscalls.C
+++ b/instructionAPI/src/syscalls.C
@@ -1,3 +1,33 @@
+/*
+ * See the dyninst/COPYRIGHT file for copyright information.
+ *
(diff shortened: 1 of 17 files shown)


instructionAPI/src/syscalls.C
kupsch, line 1: add dyninst copyright/license comment

kupsch (approved):
Looks good except missing comment in syscalls.C.

instructionAPI/src/syscalls.C
hainest: Why do I always forget that?

hainest:
@kupsch I decided to go ahead and remove IA_IAPI::isSyscall here, as well.

kupsch (approved):

PR #1700, opened 2024-03-07
Fix and extend i386 ABI registers

hainest:


hainest:
@kupsch No test regressions.

hainest:
Subsumed by #1712.

PR #1703, opened 2024-03-09
MachRegister: Add interface for collecting and retrieving all registers

hainest:
This makes writing ABI rules much simpler. It could also be useful for some power users.

The registers returned are in the order they are declared in common/h/registers/<arch>_regs.h.

### DIFF
--- a/common/src/registers/MachRegister.C
+++ b/common/src/registers/MachRegister.C
@@ -8,19 +8,24 @@
 
 #include <cassert>
 #include <unordered_map>
+#include <map>
+#include <vector>
 
 namespace {
   std::unordered_map<signed int, std::string> names;
   const std::string invalid_reg_name{"<INVALID_REG>"};
+  std::map<Dyninst::Architecture, std::vector<Dyninst::MachRegister>> all_regs;
 }
 
 namespace Dyninst {
 
   MachRegister::MachRegister() : reg(0) {}
 
-  MachRegister::MachRegister(signed int r) : reg(r) {}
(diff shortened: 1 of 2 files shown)


common/src/registers/MachRegister.C
```diff
@@ -8,19 +8,24 @@
...
+  std::map<Dyninst::Architecture, std::vector<Dyninst::MachRegister>> all_regs;
 }
 
 namespace Dyninst {
 
   MachRegister::MachRegister() : reg(0) {}
 
-  MachRegister::MachRegister(signed int r) : reg(r) {}
+  MachRegister::MachRegister(signed int r) : reg(r) {
+    all_regs[getArchitecture()].push_back(*this);
```
kupsch, line 25: `MachRegister(Signed int r)` is used to construct a `MachRegister` from an id in more places than from a register definitions(`DEF_REGISTER`), for example `MachRegister::getBaseRegister`.  Moving the update to `all_regs` to `MachRegister(signed int r, std::string n)` would only add a new registers when a name is being added by `DEF_REGISTER`.

kupsch (changes requested):
This will likely have many duplicate registers in the `all_regs` vector. Updating in the specialized constructor should fix the problem.

common/src/registers/MachRegister.C
hainest: Bah. Yes, of course.

kupsch (approved):

PR #1707, opened 2024-03-11
Rewrite ABI class

hainest:
This makes it much easier to understand what is actually happening and corresponds to the concepts in the respective ABI standards.

### DIFF
--- a/dataflowAPI/h/ABI.h
+++ b/dataflowAPI/h/ABI.h
@@ -65,33 +65,6 @@ class ABI{
     DATAFLOW_EXPORT static ABI* getABI(int addr_width);
     DATAFLOW_EXPORT bitArray getBitArray();
  private:
-    static dyn_tls bitArray* callRead_;
-    static dyn_tls bitArray* callRead64_;
-
-    static dyn_tls bitArray* callWritten_;
-    static dyn_tls bitArray* callWritten64_;
-
-    static dyn_tls bitArray* returnRead_;
-    static dyn_tls bitArray* returnRead64_;
-
-    static dyn_tls bitArray* returnRegs_;
-    static dyn_tls bitArray* returnRegs64_;
-
-    static dyn_tls bitArray* callParam_;
-    static dyn_tls bitArray* callParam64_;
-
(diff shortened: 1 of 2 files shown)


hainest:
No test regressions.

kupsch (commented):
I would suggest changes to these variable names. They are names from the point of view of liveness analysis and would be easier to grasp if they were all from the point of view of the function.  Something like the following and a corresponding `syscall*` set.

Variables:

- `callParamRegs` - registers used to pass parameters to a function
- `callReturnRegs` - registers used to return parameters from a function
- `callPreservedRegs` - registers whose value is preserved on function return
- `callUnpreservedRegs` - registers that a function may overwrite
- `callGlobalRegs` - registers that act as global variable by the ABI/runtime system that are possibly read & written in the function and are valid after a function such as `%rsp`, `%rip` and `%fs`. This makes it so liveness and other analysis has correct values for these registers as these registers are always live/defined/...  

Invariants:

- `callPreservedRegs == ~callUnpreservedRegs` (one should be defined in terms of the other)
- `callReturnRegs & CallUnpreservedRegs == callReturnRegs` (return regs are all unpreserved)

Function return values:

- `getCallReadRegisters` returns `callParams | globalRegs` (registers that may contain meaningful values to the function)
- `getCallWrittenRegisters` returns `callUnpreservedRegs | globalRegs` (registers that a function may overwrite)
- `getReturnReadRegisters` returns `callReturnRegs | callPreservedRegs | globalRegs` (registers that may have a meaningful value after a function returns)
- `getCallParameterRegisters` return `callParamRegs`
- `getReturnRegisters` returns `callReturnRegs`

hainest:
>     * `callPreservedRegs` - registers whose value is preserved on function return

All of the ABI docs refer to these as callee-saved, so I think we should use that label for consistency.

>     * `callUnpreservedRegs` - registers that a function may overwrite

Since this is invariant to ~callUnpreservedRegs, I didn't list it explicitly.

hainest:
Subsumed by #1712.

PR #1710, opened 2024-03-11
Populate RegisterMaps using MachRegister::getAllRegistersForArch

hainest:
Can't be merged until after #1709


hainest:
RegisterMaps is removed by #1712.

PR #1718, opened 2024-03-22
Remove LivenessAnalyzer::query(ParseAPI::Location,Type,OutputIterator)

hainest:
It has never worked because it doesn't dereference the output iterator.


kupsch (approved):

hainest:
No test regressions.

PR #1719, opened 2024-03-24
Dead code cleanup

hainest:
A bit of weekend yak shaving. No urgency at all.


hainest:
No test regressions.

dyninstAPI/src/image.C
```diff
@@ -1049,11 +1037,7 @@ pdmodule *image::findModule(const string &name, bool wildcard)
       }
    }
 
-   //cerr << " (image::findModule) did not find module, returning NULL" << endl;
    if (found) {
-      // Just-in-time...
-      //if (parseState_ == symtab)
-      //analyzeImage();
       return found;
    }
```
kupsch, line 1043: replace this and the `return NULL;` with `return found;`

kupsch (approved):
one small comment to make code simpler/clearer.

dyninstAPI/src/image.C
hainest: Fixed. Thanks for the careful look!

PR #1721, opened 2024-04-08
improve cmake install dirs

### LINKED ISSUE #1720: release cmake configury: unable to install under /usr/lib64/...
I'm trying to package dyninst 13 for Fedora, wherein dyninst libraries should sit under /usr/lib64/dyninst (or even /usr/lib64).  However, the cmake files appear to override all my efforts to set DYNINST_INSTALL_LIBDIR or any other library-like path cmake variable, and end up with cmake --install plopping them under /usr/lib/.  This used to work back in the dyninst 12 days.

kupsch:
Use cmake's GNUInstallDirs to get the lib, bin and include directory names instead of always using 'lib', 'bin', and 'include'.  This results in two changes:

- lib directory becomes 'lib64' on platforms where this is the convention

- the lib, bin and include may be overwritten using the cmake variables CMAKE_INSTALL_BINDIR, CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_INCLUDEDIR respectively

Fixes #1720


kupsch:
@hainest I tested on rhel-based systems that use lib64 and it works as expected. Tested with packages that use cmake targets such as examples, unit-tests, testsuite and external-tests and they all work without changes as cmake searches both lib and lib64 for the `DyninstConfig.cmake`.  Also tested very non-standard names and those also worked with external packages if the unusual directory if you specified the exact directory instead of the prefix to find the `DyninstConfig.cmake`.

bigtrak:
> @hainest I tested on rhel-based systems that use lib64 and it
> works as expected. Tested with packages that use cmake targets such
> as examples, unit-tests, testsuite and external-tests and they all
> work without changes as cmake searches both lib and lib64 for the
> `DyninstConfig.cmake`.  Also tested very non-standard names and
> those also worked with external packages if the unusual directory if
> you specified the exact directory instead of the prefix to find the
> `DyninstConfig.cmake`.

I've run into this issue with cmake on some platforms, such as el8,
where it thinks lib64 and lib are different directories for some
software packages..   I've "worked around" the issue by just symlinking
the missing one, which solves the cmake configuration issues.

Jim's suggestion to use the cmake builtins to seperate the two
so it "just works" seems a good solution.  On other OSes
with combined dirs, it should work OK.

Jim -- did you happen to check on ubuntu, for example, to see if it
does an <architecture>/lib dir setup?  Such as ubuntu uses in /usr vs
the lib64 issue on redhat based systems?   I didn't look into that,
wondering if you saw something like than when testing the cmake stuff
you found.

Thanks
Bolo

PR #1723, opened 2024-04-09
Disable -Wdeprecated-declarations for Boost on clang > 17

hainest:
It's not clear why, but clang 18 and 19 do not correctly detect that /usr/include/boost is a system directory, and it emits warnings for use of deprecated std::unary_function.


cmake/tpls/DyninstBoost.cmake
```diff
@@ -63,6 +63,11 @@ if(NOT TARGET Dyninst::Boost)
                              INTERFACE ${Boost_INCLUDE_DIRS})
   target_compile_definitions(Dyninst::Boost_headers
                              INTERFACE BOOST_MULTI_INDEX_DISABLE_SERIALIZATION)
+  set(_id CMAKE_CXX_COMPILER_ID)
+  set(_ver CMAKE_CXX_COMPILER_VERSION)
+  if(${_id} STREQUAL Clang AND ${_ver} VERSION_GREATER 17)
+    target_compile_options(Dyninst::Boost_headers INTERFACE -Wno-deprecated-declarations)
```
kupsch, line 69: Since this is a warning suppression, this line should be

```
if(DYNINST_DISABLE_DIAGNOSTIC_SUPPRESSIONS)
  target_compile_options(Dyninst::Boost_headers INTERFACE -Wno-deprecated-declarations)
endif
```

That way the warnings will be displayed the DYNINST_DISABLE_DIAGNOSTIC_SUPPRESSIONS` option is set.

kupsch (approved):

cmake/tpls/DyninstBoost.cmake
hainest: My intent was to fix a compiler bug, but I agree it's probably good to put this under the control of DYNINST_DISABLE_DIAGNOSTIC_SUPPRESSIONS. I'll add the clang check to DyninstWarnings and make a separate variable for it.

kupsch (approved):

PR #1724, opened 2024-04-09
ignore .hmac files in system-libs github workflow

kupsch:
.hmac files are library files so do not try to parse

### DIFF
--- a/.github/workflows/system-libs.yaml
+++ b/.github/workflows/system-libs.yaml
@@ -46,8 +46,10 @@ jobs:
            done
            export LD_LIBRARY_PATH
            for dir in /usr/lib /usr/lib64; do                       \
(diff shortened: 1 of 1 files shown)


hainest:
Yeah, I ran into this with my updated CI stuff. The best part is that SymtabCodeSource raises no error, but prints to stderr. Naturally. It also seems to be impossible to parse archives using Symtab::openFile because the Symtab constructor only calls [Object::Object](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Symtab.C#L819) which ultimately only checks for complete ELF [Elf_X.C](https://github.com/dyninst/dyninst/blob/master/elf/src/Elf_X.C#L187). Double-fun, that check also uses `elf_kind` to set `isArchive`... which is always false.

hainest (approved):

PR #1727, opened 2024-04-11
CMake: update x86 platform detection

### LINKED ISSUE #1726: Upgrading SelectiveTaint: Replacing dyninst9.3 with dyninst13.0 in Ubuntu 14.04 Environment
I'm trying to replace dynisnt9.3 with dyninst13.0 in selectivetaint (running in ubuntu14.04 environment) and I'm getting the following error in the build from source short step, so that means that dynisnt9.3 onwards is not supported for installation in the ubuntu14.04 (because of its i686 architecture)?

`
CMake Error at cmake/DyninstPlatform.cmake:16 (message):
  Unsupported architecture: 'i686'
Call Stack (most recent call first):
  CMakeLists.txt:35 (include)
`

hainest:
Adds better detection of 32-bit x86 platforms and separates out the 64-bit ones.

Should fix #1726, but will need confirmation.

### DIFF
--- a/cmake/DyninstPlatform.cmake
+++ b/cmake/DyninstPlatform.cmake
@@ -11,17 +11,25 @@ endif()
 # Equivalent to CMAKE_HOST_SYSTEM_PROCESSOR and `uname -m` on Unixes
 cmake_host_system_information(RESULT _host_arch QUERY OS_PLATFORM)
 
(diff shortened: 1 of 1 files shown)


kupsch (approved):

hainest:
@kupsch It looks like this is working. Do you think we should merge it?

kupsch:
Yes. Merge it.

PR #1728, opened 2024-04-15
Fix AMDGPU register operand expansion

### LINKED ISSUE #1725: AMD GPU Instruction Decoder Failed to expand operand
**Describe the bug**

https://github.com/dyninst/dyninst/blob/e885d38c535976c92f108bce29803dcb6145476e/instructionAPI/src/AMDGPU/gfx908/appendOperands.C#L252-L263

The intention of this implementation is to expand a single operand into multiples, when the first one decode returns a registerAST expression.
The boost::dynamic_pointer_cast<RegisterAST::Ptr>(first) != NULL fails however, even if the variable first is indeed of type RegisterAST.
The reason why it wasn't discovered is because the Architecture Specific Formatter will expand on the first operand. 
Take S_LOAD_DWORDX2 S[0:1], S[4:5], 0x0 for example,
the output instruction will have 3 operands, s0, s4 and 0x0, but the formatter will correctly output s0 as s[0:1] and s4 as s[4:5], hiding the fact that we actually missed operands s1 and s5.

**Potential Solution**

The problem here is that we want a way to figure out the underlying type of a Expression::Ptr. 
One easy solution is to create a type field in Expression, that gets initialized with the proper type based on the constructor it uses,
but I'm not sure if this is a desirable solution.

bbiiggppiigg:
This commit fixes an issue with the use of dynmaic_pointer_cast in appendOperands.C, 
where the base type should be used in stead of the Ptr type.

Fixes #1725

### DIFF
--- a/instructionAPI/src/AMDGPU/gfx908/appendOperands.C
+++ b/instructionAPI/src/AMDGPU/gfx908/appendOperands.C
@@ -32,7 +32,7 @@ namespace InstructionAPI {
     {
         Expression::Ptr first = decodeOPR_ACCVGPR(input,vec_len);
         insn_in_progress->appendOperand(first,isRead,isWritten,isImplicit);
-        if (boost::dynamic_pointer_cast<RegisterAST::Ptr>(first) != NULL)
+        if (boost::dynamic_pointer_cast<RegisterAST>(first) != NULL)
         {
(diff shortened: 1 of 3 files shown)


bbiiggppiigg:
@rochauha This PR fixes the liveness problem you've seen in your branch.

kupsch (approved):

PR #1736, opened 2024-05-02
GithubCI: ignore libclang-cpp.so in system-libs test

hainest:
For some reason, the debuginfod server on Fedora 37 hangs when trying to fetch data for libclang-cpp.15.so. It's not critical for this test, so just don't parse it.

### DIFF
--- a/.github/workflows/system-libs.yaml
+++ b/.github/workflows/system-libs.yaml
@@ -48,7 +48,7 @@ jobs:
            for dir in /usr/lib /usr/lib64; do                       \
              for file in $(find $dir -type f -name "*.so.*"); do    \
                if ! expr "$file" : ".*\.hmac" >/dev/null 2>&1; then \
-                 if [[ $f =~ "libclang-cpp.so" ]]; then continue; fi\
+                 if [[ $f =~ "libclang-cpp.so" ]]; then continue; fi;\
                  echo $file;                                        \
                  ./simpleParser $file;                              \
                fi                                                   \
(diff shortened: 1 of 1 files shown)


.github/workflows/system-libs.yaml
```diff
@@ -48,6 +48,7 @@ jobs:
            for dir in /usr/lib /usr/lib64; do                       \
              for file in $(find $dir -type f -name "*.so.*"); do    \
                if ! expr "$file" : ".*\.hmac" >/dev/null 2>&1; then \
+                 if [[ $f =~ "libclang-cpp.so" ]]; then continue; fi\
```
kupsch: github actions use /bin/sh by default which disables bash constructs like `[[ ... ]]`.  You'll need to use expr like on line 50.

kupsch (changes requested):

kupsch:
Also on line 41, I noticed that `-n "LD_LIBRARY_PATH"` is just a string and should be `-n "$LD_LIBRARY_PATH"` to check the variable.

kupsch:
is there a reason to have the line continuations?  I think they can go away and then the `;` terminating statements at the end of lines can also go away as a new-line is also a statement terminator

hainest:
This is too much of a pain to fix here. Plus this whole thing is getting updated in https://github.com/hainest/dyninst/pull/7. I'll just delay this fix to there.

>github actions use /bin/sh by default

Fixed in PR7.

>Also on line 41, I noticed that -n "LD_LIBRARY_PATH" is just a string and should be -n "$LD_LIBRARY_PATH" to check the variable.

I'll add that to PR7.

>is there a reason to have the line continuations?

No. That's changed in PR7.

kupsch:
While you're make changes replacing the loops with a single` find` command would also work, be slightly more efficient and handle file names with spaces:

```
find /usr/lib /usr/lib64 -type f -name '*.so.*' \! -name '*.hmac' \! -name 'libclang-cpp.so*' -print -exec ./simpleParser {} \;
```

PR #1737, opened 2024-05-03
Replace utos with std::to_string

hainest:
This function doesn't need to exist since there is a standard way of doing it now.

### DIFF
--- a/dyninstAPI/src/BPatch_process.C
+++ b/dyninstAPI/src/BPatch_process.C
@@ -958,7 +958,7 @@ BPatch_object *BPatch_process::loadLibrary(const char *libname, bool)
          break;
       }
       if (bpfv.size() > 1) {
-         std::string msg = std::string("Found ") + utos(bpfv.size()) +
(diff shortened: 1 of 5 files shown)


kupsch (approved):
Looks good.  I think we should remove the the `utos` declaration and definition in util.{h,C} as it is unlikely that anyone is using this outside Dyninst.  If you think someone might be using it, then add a `DYNINST_DEPRECATED("replace with std::to_string")` annotation to the declaration and we can remove it in a future release.

kupsch:
Also do the same to `itos`, which is not used in Dyninst.

hainest:
> Looks good. I think we should remove the the `utos` declaration and definition in util.{h,C} as it is unlikely that anyone is using this outside Dyninst. If you think someone might be using it, then add a `DYNINST_DEPRECATED("replace with std::to_string")` annotation to the declaration and we can remove it in a future release.

It's undocumented, so I'm not worried. I'm gutting most of this file. It's causing build errors for in-tree systemtap builds because of a filename conflict. Ironically, 'util.h' has no utility to users- it's just more internals leaking out.

PR #1738, opened 2024-05-03
Consolidate visibility control macros

hainest:
There is no need to have one per toolkit.
    
This also removes the MSVC-specific check because we don't build there
and we don't have any of the visibility compiler switches set for it,
either.

### DIFF
--- a/common/h/dyninst_visibility.h
+++ b/common/h/dyninst_visibility.h
@@ -31,8 +31,6 @@
 #ifndef DYNINST_COMMON_DYNINST_VISIBILITY_H
 #define DYNINST_COMMON_DYNINST_VISIBILITY_H
 
-#ifndef DYNINST_EXPORT
-# define DYNINST_EXPORT __attribute__((visibility ("default")))
-#endif
+#define DYNINST_EXPORT __attribute__((visibility ("default")))
 
 #endif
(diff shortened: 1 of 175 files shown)


common/h/dyninst_visibility.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifndef DYNINST_COMMON_DYNINST_VISIBILITY_H
+#define DYNINST_COMMON_DYNINST_VISIBILITY_H
+
+#ifndef DYNINST_EXPORT
```
kupsch, line 34: why is this necessary?  Since there are include guards, doesn't this indicate a problem somewhere else?

kupsch (approved):
Looks good, but should `dyninstAPI/h/BPatch_dll.h` go away?  It appears useless, except to define `DYNINST_EXPORT` which is redundant with `dyninst_visibility.h` and is probably the reason for the define guard for `DYNINST_EXPORT`.  The other things like `SHM_SAMPLING`, windows and `extern "C" DYNINST EXPORT const char v_libdyninstAPI[];` aren't used.

common/h/dyninst_visibility.h
hainest: It shouldn't be. I just copy/pasted from the old definition. I'll remove it and see what breaks.

>should dyninstAPI/h/BPatch_dll.h go away?

Definitely. I'll do that in a separate PR.

PR #1741, opened 2024-05-09
Hide ompc_leftmost reducer if not using OpenMP

hainest:
This causes an unused function warning when USE_OpenMP=OFF.

We could also hide the `#prama omp declare` right after this and remove the CMake check (https://github.com/dyninst/dyninst/blob/master/cmake/DyninstWarnings.cmake#L93).

### DIFF
--- a/symtabAPI/src/dwarfWalker.C
+++ b/symtabAPI/src/dwarfWalker.C
@@ -108,9 +108,12 @@ DwarfWalker::DwarfWalker(Symtab *symtab, ::Dwarf *dbg, std::shared_ptr<ParsedFun
 DwarfWalker::~DwarfWalker() {
 }
 
+#ifdef _OPENMP
 static inline void ompc_leftmost(Module* &out, Module* &in) {
     out = out == NULL ? out : in;
 }
+#endif
+
 #pragma omp declare \
     reduction(leftmost : Module* : ompc_leftmost(omp_out, omp_in)) \
     initializer(omp_priv = NULL)


kupsch (approved):
Looks good.  I don't think the `-Wno-unknown-pragmas` can be removed as gcc warns about all instances of `#pragma omp ...` when compiling without OpenMP.

hainest:
> Looks good. I don't think the `-Wno-unknown-pragmas` can be removed as gcc warns about all instances of `#pragma omp ...` when compiling without OpenMP.

I was thinking of wrapping all of them in an _OPENMP check because I don't really like the asymmetry of guarding the reduction function and then ignoring pragma warnings. I guess it would be too noisy to add the check to all of the pragmas. Good enough here, then.

kupsch:
One thing that we could do is replace all the `#pragma omp ...` with `DYNINST_PRAGMA_OMP(...)` that conditionally expands to nothing or `_Pragma("omp ...")`.  Let me know if you think this would be worthwhile and I can make the change.

Seems like the compilers should not warn about this as an unknown pragma when omp is not enabled, it should know it and treat it as a NOP. Likely this is an implementation issue that the omp pragma is only in the list of known pragma's if omp is enabled.

hainest:
Of all the things we need to clean up, I think we can put this one fairly close to the bottom of the list.

>Seems like the compilers should not warn about this as an unknown pragma when omp is not enabled, it should know it and treat it as a NOP. 

I was thinking the same, but I guess they have an early bail-out if omp is disabled.

woodard:
I can ask jakub if we can get that changed in the compiler. I think it is arguable that it is "known" even if it is not enabled.

woodard:
OK Jakub is not buying the "it is known just currently disabled" argument. He says that  purpose of the warning is to warn about pragmas that don't do what the users might think they will do and that developers should be aware of that guard them with #ifdef _OPENMP if you don't want the warnings and it is intentional.

hainest:
@woodard Booo. Thanks for checking!

Issue #1285, opened 2022-07-11
Use thread-safe data structure for parsing Symtab line information


hainest:
See details in #1284. I don't see that TBB has a multi-index container. If we need a manual locking scheme against a boost::multi_index, we should make a proper data structure in Dyninst namespace.

@kupsch @bigtrak

jmellorcrummey:
@blue42u

hainest:
@blue42u

>I have serious doubts it will be feasible to find/implement a "proper" concurrent equivalent without sacrificing the performance.

Very much agree. When I said `we should make a proper data structure in Dyninst namespace`, I meant we should make a Dyninst data structure that is a wrapper around a `boost::mutli_index_container` rather than having a bunch of manual locking floating around.

>Back in https://github.com/dyninst/dyninst/pull/651 I originally tried wrapping the one used for indexing Symbols with a lock and the parallel SymtabAPI performance in hpcstruct got a <1x speedup.

What kind of locking scheme?

>In the end I sacrificed the general applicability and replaced it with a (fast and maintainable) series of concurrent hash maps

Given the complexity of [`RangeLookupType`](https://github.com/dyninst/dyninst/blob/35c289a2d38bd5b3c20c07cbe1c1f3c31ccc34a1/symtabAPI/h/RangeLookup.h#L99), I have a feeling it would be quite a bit of work to convert it to something else. I would be interested to know _why_ there is so much complexity needed, though.

blue42u:
@hainest 
> > I have serious doubts it will be feasible to find/implement a "proper" concurrent equivalent without sacrificing the performance.
> 
> Very much agree. When I said `we should make a proper data structure in Dyninst namespace`, I meant we should make a Dyninst data structure that is a wrapper around a `boost::mutli_index_container` rather than having a bunch of manual locking floating around.

That's how I interpreted your initial comment. I agree that adding a wrapper structure is definitely preferable to manual locking for code cleanliness, but just be aware that doing so will likely kill the parallel performance.

> What kind of locking scheme?

To be honest I don't remember (I wish I had written notes), but I think I tried both `std::mutex` and `std::shared_mutex`. I also vaguely recall having synchronization problems with `std::shared_mutex` (IIRC some `const` methods still altered the internal pointers for some undiscernible reason).

[...]

PR #1744, opened 2024-05-20
Fix Race Condition in Parsing Line Map for NVIDIA Binaries

hsmcintyre:
This patch addresses a race condition that occurs when parsing an NVIDIA binary where two modules point to the same object. The NVIDIA binary where the race condition was initially discovered is included in this pull request.

The issue arises when multiple threads attempt to open the object and parse the line map concurrently. This fix ensures proper handling for this specific use case, mitigating the lower-level race condition. However, it is important to note that a potential race condition exists at a higher level.

If two threads call parse on the same module, a race condition occurs as both threads attempt to set the line information for the module. Currently, there is no mechanism to ensure mutual exclusion in this scenario.

This patch addresses the lower-level race condition during object parsing. Specifically, while the object parse line info method is now thread-safe, the module parse line info method remains non-thread-safe. This fix resolves the issue for the specific use case in HPCToolkit, where concurrent calls to parse line info on the same module do not occur. The concurrent parsing on the object level is implicit, as it involves two modules pointing to the same object.

[recursion-sm_87-nvcc118-0.zip](https://github.com/dyninst/dyninst/files/15382254/recursion-sm_87-nvcc118-0.zip)

### DIFF
--- a/symtabAPI/src/Object-elf.h
+++ b/symtabAPI/src/Object-elf.h
@@ -148,7 +148,7 @@ class open_statement {
 };
 
 
-class Object : public AObject, private boost::basic_lockable_adapter<boost::mutex>
+class Object : public AObject, public boost::basic_lockable_adapter<boost::mutex>
 {
   friend class Module;
 
--- a/symtabAPI/src/Object-elf.h
+++ b/symtabAPI/src/Object-elf.h
@@ -148,7 +148,7 @@ class open_statement {
 };
 
 
-class Object : public AObject, public boost::basic_lockable_adapter<boost::mutex>
+class Object : public AObject, private boost::basic_lockable_adapter<boost::mutex>
 {
   friend class Module;
 
--- a/symtabAPI/src/Object-elf.h
+++ b/symtabAPI/src/Object-elf.h
@@ -148,7 +148,7 @@ class open_statement {
 };
 
 
-class Object : public AObject 
+class Object : public AObject, public boost::basic_lockable_adapter<boost::mutex>
 {
   friend class Module;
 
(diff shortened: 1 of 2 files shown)


symtabAPI/src/Object-elf.h
```diff
@@ -148,7 +148,7 @@ class open_statement {
 };
 
 
-class Object : public AObject 
+class Object : public AObject, public boost::basic_lockable_adapter<boost::mutex>
```
hainest, line 151: Could you change this to private inheritance? This represents a has-a rather than an is-a relationship. Generally, I'd prefer composition, but I know `basic_lockable_adapter` is generaly meant to be inherited from.

hainest (changes requested):
This is a consequence of unreliable thread safety in SymtabAPI (e.g., #1285 and #1446).

>two modules point to the same object

I thought CUDA binaries only had one module which is why [parseLineInfoForObject](https://github.com/dyninst/dyninst/blob/aa8eb5abcadf2f456bc4a8fecfdd7c897fca42cd/symtabAPI/src/Object-elf.C#L3425) slurps in the whole line table. Did I misremember this? The provided example binary only has one compilation unit (CU; i.e., module) with a PC range that covers all of the items in the line map, as well.

For the case of two modules, I would be very interested to see what the address ranges look like between them. It is possible for partial CUs to have overlapping ranges. However, Dyninst has never supported line maps for this case. I would like to fix that, but it's (unfortunately) a lower-priority issue. Definitely let us know if you encounter this.

P.S. We've never tested with the EDG compiler before, so this is interesting.

symtabAPI/src/Object-elf.h
hainest: Bah, sorry. I don't know what I was thinking. That adapter has to be publicly inherited from when used with `boost::lock_guard`. I'd like to move to using C++11 constructs for exclusivity, but that's for another day.

hsmcintyre: I've changed it back to public.

kupsch (approved):

PR #1745, opened 2024-05-23
Simplify RoseAST based on the associative property of addition

batuzovk:
Associative property of addition allows simplification of "const + something + const" RoseAST to "something + const" form. This can be particularly helpful in jump table analysis if PIC jump table uses some base to access both the table itself and targets.

A test example:

```
.text
.set	d, tbl - _start
.set	t1, .trg1 - _start
.set	t2, .trg2 - _start
_start:
    mov		$1, %eax
    lea		_start(%rip), %r10
    cmp 	$2, %rax
    ja		.nope
    mov		d(%r10, %rax, 4), %r9d
    add		%r10, %r9
    jmpq	*%r9
.trg1:
    mov		$0, %eax
    ret
.trg2:
    mov		$1, %eax
    ret	
.nope:
    ret

tbl: .long	t1
     .long	t2  
```

Without this change the jump table is not recognized:

```
[thread 0] In function _start, Address 101d, jump target format <add>(<deref>(<add>(<add>(<4096:64>,<uMult>(<V([x86_64::rax])>,<4:8>,),),<45:64>,),),<4096:64>,), index loc (@1012<[x86_64::r9]>[x86_64::rax]>[x86_64::r10]>H[]), index variable [x86_64::rax], memory read loc (@1012<[x86_64::r9]>[x86_64::rax]>[x86_64::r10]>H[]), isJumpTableFormat 0 
```

With this change it is recognized:

```
[thread 3] In function _start, Address 101d, jump target format <add>(<deref>(<add>(<uMult>(<V([x86_64::rax])>,<4:8>,),<4141:64>,),),<4096:64>,), index loc (@1012<[x86_64::r9]>[x86_64::rax]>[x86_64::r10]>H[]), index variable [x86_64::rax], memory read loc (@1012<[x86_64::r9]>[x86_64::rax]>[x86_64::r10]>H[]), isJumpTableFormat 1
```

A real life example:

```
   180155963:   4c 8d 15 96 a6 ea ff    lea    -0x15596a(%rip),%r10        # 0x180000000
   18015596a:   49 83 f8 0f             cmp    $0xf,%r8
   18015596e:   0f 87 0c 01 00 00       ja     0x180155a80
   180155974:   66 66 66 66 0f 1f 84    data16 data16 data16 nopw 0x0(%rax,%rax,1)
   18015597b:   00 00 00 00 00
   180155980:   47 8b 8c 82 00 c0 1d    mov    0x1dc000(%r10,%r8,4),%r9d
   180155987:   00
   180155988:   4d 03 ca                add    %r10,%r9
   18015598b:   41 ff e1                jmpq   *%r9
```

### DIFF
--- a/parseAPI/src/SymbolicExpression.C
+++ b/parseAPI/src/SymbolicExpression.C
@@ -174,7 +174,54 @@ AST::Ptr SymbolicExpression::SimplifyRoot(AST::Ptr ast, Address addr, bool keepM
                                                       }
                                                   }
                                               }
-                                          } 
+                                          }
+                                          // Case 4a: if (c1 + x) + c2 = x + (c1 + c2), where x is anything, c1 and c2 are constants
+                                          // Case 4b: if (x + c1) + c2 = x + (c1 + c2), where x is anything, c1 and c2 are constants
+                                          // Case 4c: if c1 + (c2 + x) = x + (c1 + c2), where x is anything, c1 and c2 are constants
+                                          // Case 4d: if c1 + (x + c2) = x + (c1 + c2), where x is anything, c1 and c2 are constants
+                                          if ((roseAST->child(0)->getID() == AST::V_RoseAST && roseAST->child(1)->getID() == AST::V_ConstantAST) ||
+                                              (roseAST->child(1)->getID() == AST::V_RoseAST && roseAST->child(0)->getID() == AST::V_ConstantAST)) {
+                                              RoseAST::Ptr rOp;
+                                              RoseAST::Ptr oldX = nullptr;
+                                              ConstantAST::Ptr oldC1 = nullptr, oldC2 = nullptr;
+                                              if (roseAST->child(0)->getID() == AST::V_RoseAST) {
+                                                  rOp = boost::static_pointer_cast<RoseAST>(roseAST->child(0));
+                                                  if (rOp->val().op == ROSEOperation::addOp) {
+                                                      if (rOp->child(0)->getID() == AST::V_ConstantAST) {
+                                                          // Case 4a
+                                                          oldX = boost::static_pointer_cast<RoseAST>(rOp->child(1));
+                                                          oldC1 = boost::static_pointer_cast<ConstantAST>(rOp->child(0));
+                                                          oldC2 = boost::static_pointer_cast<ConstantAST>(roseAST->child(1));
+                                                      } else if (rOp->child(1)->getID() == AST::V_ConstantAST) {
+                                                          // Case 4b
+                                                          oldX = boost::static_pointer_cast<RoseAST>(rOp->child(0));
+                                                          oldC1 = boost::static_pointer_cast<ConstantAST>(rOp->child(1));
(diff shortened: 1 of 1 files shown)


hainest:
This is definitely a necessary change, but I'm hesitant to add another set of special cases. I also don't want to explode this into a bunch of work to make a generic expression optimizer. I propose the following for reducing `ROSEOperation::addOp` expressions:

1. Regularize the form `c1 * x` to `x * c1`
  a. If `c1 == 0`, reduce to `0`.
2. Regularize the form `c1 + x` to `x + c1`
3. Regularize the form `c1 + x + c2` to `x + c1 + c2`
4. Carry out the reductions in Cases 4a-4d with the new expression format.

We can then subsume the existing `addOp` cases:

  Case 1: two constants
  Case 2: anything adding zero stays the same
  Case 3: if v + v * c = v * (c+1), where v is a variable and c is a constant

(1) should be added to the `ROSEOperation::{s,u}MultOp` cases and refactored to be used here, as well. As should the `keepMultiOne` behavior to make the distributive check in Case 2 consistent.

(1) combined with (2) handles Case 2 and makes it commutative.

(4) handles Case 1 when `x` is a constant. An additional check will be need to be added for this.

(4) handles Case 2 implicitly.

(1) combined with (3) simplifies the logic in (4).

Delaying the refactoring so that we can get the example case in the PR notes working for now is okay with me. We should at least make an issue for it, if we don't fix it here.

@kupsch Thoughts?

batuzovk:
If any refactoring of this function happens in the future, then there is a bunch of lines further down that need an attention in my opinion

https://github.com/dyninst/dyninst/blob/9cb8cba5e42c406d82f5e2d2a8c40785e9f9f675/parseAPI/src/SymbolicExpression.C#L328-L330

This code widens some 32-bit constants to 64 bits without any regard to the sizes of other operands. In 32-bit code it creates mismatched expressions where 64-bit immediate value is assigned to 32-bit register, for example.

I haven't done anything with this `if` because I can't figure out what it even tries to achieve. And I prefer not to touch things I don't understand when possible.

hainest:
> If any refactoring of this function happens in the future, then there is a bunch of lines further down that need an attention in my opinion
> 
> https://github.com/dyninst/dyninst/blob/9cb8cba5e42c406d82f5e2d2a8c40785e9f9f675/parseAPI/src/SymbolicExpression.C#L328-L330
> 
> This code widens some 32-bit constants to 64 bits without any regard to the sizes of other operands. In 32-bit code it creates mismatched expressions where 64-bit immediate value is assigned to 32-bit register, for example.

The treatment of 32/64 and signed/unsigned in Dyninst is not consistent, unfortunately. This is just another example. I'll add it to the other issue.

hainest (approved):

@bbiiggppiigg @kupsch Since this is needed for #2341, I'd say merge as-is. I can open an issue with the suggested extensions afterward.

bbiiggppiigg (approved):

PR #1749, opened 2024-05-24
x86: handle flag RF

batuzovk:
The only instruction supported by Dyninst that references flag RF is RSM. An attempt to getReadSet or getWriteSet for it fails an assertion.

Sample code:

```
_start:
    rsm
```

Dyninst without change:

```
ERROR: unhandled entry x86::rf
code_sample: /home/batuzovk/projects/dyninst-upstream/instructionAPI/src/Operation.C:541: Dyninst::InstructionAPI::Operation::SetUpNonOperandData()::<lambda()>: Assertion `0' failed.
Aborted
```


hainest:
No test regressions.

hainest (approved):

PR #1750, opened 2024-05-24
Handle ELF files that don't contain code

batuzovk:
When debug info is being split into a separate file, the resulting file has ELF format. Users may accidentally pass this file as an input to Dyninst. Usually these files' names end with .debug and are easy to distinguish from actual libraries, but lbunity, for example, names its debug file libunity.sym.so. Dyninst shouldn't crash on it.

Debug files have full set of sections you'd expect to see in a normal ELF file, but most of them are NOBITS. An attempt to read data from NOBITS section results in NULL pointer dereference. This happens in `read_except_table_gcc3` in particular.

This patch adds checks for NOBITS sections in two places: in `read_except_table_gcc3` where SIGSEGV happened, and when handling `.text` section to fail debug files early and avoid other potential segmentation faults that were not found yet.

To trigger mentioned SIGSEGV you need a separate debug file that has `.eh_frame` and `.gcc_except_table` sections. The debug file for libc.so.6 fits the requirements.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -674,6 +674,13 @@ bool Object::loaded_elf(Offset &txtaddr, Offset &dataddr,
         }
 
         if (strcmp(name, TEXT_NAME) == 0) {
+            // NOBITS .text section means that we are probably dealing with
+            // debug symbols being split into a separate ELF file.
+            // This file doesn't contain code for us to parse.
+            if (scn.sh_type() == SHT_NOBITS) {
+                log_elferror(err_func_, "section .text can't be NOBITS");
+                return false;
+            }
             text_addr_ = scn.sh_addr();
             text_size_ = scn.sh_size();
 
@@ -2679,6 +2686,12 @@ int read_except_table_gcc3(
     unsigned char lpstart_format, ttype_format, table_format;
     unsigned long value, table_end, region_start, region_size, landingpad_base;
     unsigned long catch_block, action, augmentor_len;
+
+    // An attempt to read data from NOBITS section will result in
+    // NULL pointer dereference
+    if (!eh_frame->isValid() || eh_frame->sh_type() == SHT_NOBITS) {
+        return false;
+    }
 
     Dwarf_Off offset = 0, next_offset, saved_cur_offset;
     int res = 0;


hainest:
> Usually these files' names end with .debug and are easy to distinguish from actual libraries, but lbunity, for example, names its debug file libunity.sym.so.

I'm curious what .dwo files look like. Fedora uses them, so it's something I should look at before merging.

> Debug files have full set of sections you'd expect to see in a normal ELF file, but most of them are NOBITS. An attempt to read data from NOBITS section results in NULL pointer dereference. This happens in `read_except_table_gcc3` in particular.

Which OS are you using? I'd like to take a closer look at these files.

> This patch adds checks for NOBITS sections in two places: in `read_except_table_gcc3`

Since we [check](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object-elf.C#L1479) for a valid gcc exception table section, I'd propose moving this check to avoid setting [`gcc_except`](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object-elf.C#L830) in `loaded_elf`.

> when handling `.text` section to fail debug files early and avoid other potential segmentation faults that were not found yet.

I think there is more to consider here. For example, we [skip](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object-elf.C#L605) debug sections that aren't executable. I'm not sure why such a check is necessary... We probably shouldn't [convert](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object-elf.C#L627) the section to a memory representation on ppc if .text isn't executable.

I'm curious if the `isFromDebugFile()` check on 619 couldn't be updated to handle this case on line 603 (the only place `setDebugFile` is ever called. I'll have a think. This code is so difficult to reason about...

> To trigger mentioned SIGSEGV you need a separate debug file that has `.eh_frame` and `.gcc_except_table` sections. The debug file for libc.so.6 fits the requirements.

Which version of libc? (`ldd --version`)

batuzovk:
I'm on Debian, bookworm more specifically. 

```
$ ldd --version
ldd (Debian GLIBC 2.36-9) 2.36
```

I've attached an archive with libc and its debug file: [glibc-with-dbg.tar.gz](https://github.com/dyninst/dyninst/files/15482537/glibc-with-dbg.tar.gz) An attempt to use disassemble from Dyninst examples on debug file leads to segfault.

> This code is so difficult to reason about...

It is indeed. This is why I ended up with this weird "fix where it breaks, then fix a bit more just in case" solution.

It is tempting to add `|| scn.sh_type() == SHT_NOBITS` to `!scn.isValid()` check [here](https://github.com/dyninst/dyninst/blob/07e41a479bce35c8b1c86b47feb869630918dd2d/symtabAPI/src/Object-elf.C#L575-L577), but I don't know if we need to process some of NOBITS sections.

Another option is to change `Elf_X_Shdr::isValid()` to return `false` for NOBITS sections. But this one raises even more concerns.

batuzovk:
Ping? I understand that this code is messy and it might require some refactoring in the future, but I think it is better to have an imperfect fix for a crash, than no fix at all.

hainest:
@kupsch I can't think of any other cases to handle here. Thoughts? I'm testing it now.

kupsch:
I think this patch will result not finding the debug data in separate debug files.  I have a fix for the core issue, but I need to test it and that will take a bit of time as I have other priorities that need to be completed first.

PR #1754, opened 2024-05-28
x86: Handle disambiguation between XOP prefix and POP instruction

batuzovk:
`0x8f` can be either the first byte of XOP prefix, or the opcode of POP instruction. The disambiguation is resolved by `map_select` field of XOP's second byte. If it is XOP prefix, than the field should be from [0x8, 0x1f] range. Otherwise it is ModRM byte, and `0x8f` is an opcode. The only instruction with this opcode is POP, and it requires `Reg` filed of ModRM to be 0.

`map_select` field is bits [4:0], `reg` field is [5:3]. Basically, if bits [4:3] are 0, then this is not a XOP prefix.

The group that POP belongs to is called "Group 1a" by AMD documentation, but this name is already used for "Group 1 opcode 80h", thus I called it "Group 1A".

Code sample to test:

```
_start:
    pop (%rax)
```

### DIFF
--- a/common/src/arch-x86.C
+++ b/common/src/arch-x86.C
@@ -163,7 +163,7 @@ unsigned int swapBytesIfNeeded(unsigned int i)
 
 // groups
 enum {
-  Grp1a=0, Grp1b, Grp1c, Grp1d, Grp2, Grp3a, Grp3b, Grp4, Grp5, Grp6, Grp7,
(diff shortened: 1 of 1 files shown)


hainest (approved):

hainest:
No test regressions.

PR #1757, opened 2024-05-30
[WIP] Implement parsing of PE files using pe-parse library

batuzovk:
This change adds support for parsing PE (Windows .EXE and .DLL) files. Compared to previous implementation it is not tied to any specific platform. With it you can parse PE files on Linux machine. It supports both 32- and 64-bit PE files. It detects file type based on "magic" bytes at the beginning, so you can feed .EXE/.DLL files to disassemble from Dyninst examples and it will work. The implementation is based on [pe-parse](https://github.com/trailofbits/pe-parse) library.

This change is marked as WIP because it requires tidying up before it can be merged. But before I do it I'd like to hear some feedback on few topics

* Windows builds. This change does not update anything related to Windows builds, and it is in obvious conflict with existing Object-nt.C. I tried to make Dyninst build on Windows machine, but the build was broken in more than one ways. Some of them require non-trivial effort and time to fix, but at the same time it doesn't seem to be used by anyone. For example, [P_cplus_demangle](https://github.com/dyninst/dyninst/blob/9daa8a09db09f8de153e351a6a18dd7ebcda1881/common/src/ntHeaders.h#L217-L234) in ntHeaders.h is missing semicolon in line 221 and closing brace in line 226. Those will obviously cause compilation error. They were introduced in 41ef642f3a, more than 3 years ago. Yet no one raised an issue about it. I have nearly zero motivation to spend efforts on something that is not used by anyone, but if you say that Windows build needs to work, then I'll try to fix it and incorporate this change into it.
* The change requires splitting into parts. It is too big for single review. There is an obvious split "refactoring or AObject" and "PE parsing", but this might not be enough. If you have any ideas how to further split the second part, I'll be glad to hear them.
* Possible change to AObject/Object interface. Currently I've dumped everything from ELF and PE together, but I'm sure there is a lot of trimming to do there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ include(DyninstLibIberty)
 include(DyninstThread_DB)
 include(DyninstValgrind)
 include(DyninstOpenMP)
+include(DyninstPeParse)
 
 add_subdirectory(common)
 add_subdirectory(elf)
--- a/cmake/DyninstConfig.cmake.in
+++ b/cmake/DyninstConfig.cmake.in
@@ -6,6 +6,7 @@ list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_LIST_DIR}/Modules"
 include(DyninstBoost)
 include(DyninstElfUtils)
 include(DyninstTBB)
+include(DyninstPeParse)
 
 set(CMAKE_MODULE_PATH ${_DYNINST_module_path_save})
 unset(_DYNINST_module_path_save)
--- a/symtabAPI/CMakeLists.txt
+++ b/symtabAPI/CMakeLists.txt
@@ -72,7 +72,8 @@ if(DYNINST_OS_UNIX)
     src/LinkMap.C
     src/emitElf.C
     src/emitElfStatic.C
-    src/dwarfWalker.C)
+    src/dwarfWalker.C
+    src/Object-pe.C)
 
   if(DYNINST_ARCH_x86_64 OR DYNINST_ARCH_i386)
     list(APPEND _sources src/emitElfStatic-x86.C src/relocationEntry-elf-x86.C)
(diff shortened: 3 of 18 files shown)


hainest:
This is a welcome change and moves us closer to our long-term goal of seamless cross-platform analysis. As you've noticed, a great deal of Dyninst code paths are tied to the machine on which it was compiled. Which is also why integrating this change will be more difficult than it seems. The code here works for parsing, but it will have a lot of impedance mismatch with the rest of Dyninst's analysis workflows. For example, liveness analysis won't work on PE binaries because the Windows ABI is only available when compiled on Windows. I don't think that should stop us from moving this work forward.

Dyninst never supported PE debug (PDB), so that's not a detriment per se. However, a large portion of Dyninst's analysis capabilities rely on debug information. I don't think we should put that as a priority, but I'm concerned that someone may try to push us toward that in the future. We are wholly unequipped to do that due to a lack of developers. Again, we shouldn't let that stop us from moving forward with this work. For future reference, the MS PDB "spec" is https://github.com/Microsoft/microsoft-pdb.

> I tried to make Dyninst build on Windows machine, but the build was broken in more than one ways.

Please don't spend your time trying to build the Windows code. It has been effectively abandoned. It hasn't worked correctly since the migration to CMake in 2013.

> The change requires splitting into parts. It is too big for single review.

Agreed. Thank you for doing that work. It makes our job so much easier.

> Possible change to AObject/Object interface. Currently I've dumped everything from ELF and PE together, but I'm sure there is a lot of trimming to do there.

Don't hesitate to reach out to us. It can be very easy to deep-dive into those old parts of Dyninst and get lost in the ancient, indecipherable code.

batuzovk:
I've converted this PR to Draft. I've split it into several changes and will make separate PRs for them (one by one since they depend on each other and will need to be applied in order anyway).

The first one is the refactoring #1758

batuzovk:
All changes have been merged.

The list of MR: #1758 , #1904 , #1896, #1908 , #1935, #1937

PR #1758, opened 2024-06-03
Refactroing of AObject class

batuzovk:
This change is made in preparation for implementation of PE files parsing support.

The base class AObject is renamed to Object and its interface is expanded.

Derived classes now have different names depending on the type of the object files they represent, e.g. ObjectELF

All non-specific parsing operations were changed to work with the basic class.

A new function that selects derived class based on magic bytes is implemented. It can only select ObjectELF for now.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -68,7 +68,7 @@ unsigned int elfHash(const char *name) {
 }
 
 template<class ElfTypes>
-emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, Object *obj_, void (*err_func)(const char *),
+emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, ObjectELF *obj_, void (*err_func)(const char *),
                                Symtab *st) :
         oldElfHandle(oldElfHandle_), newElf(NULL), oldElf(NULL),
         obj(st),
@@ -127,6 +127,8 @@ emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, Object *obj_,
         movePHdrsFirst = true;
         library_adjust = getpagesize();
     }
+
+    assert(obj && object && object == dynamic_cast<ObjectELF*>(obj->getObject()));
 }
 
 template<typename ElfTypes>
@@ -573,7 +575,7 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
             }
         }
 
-        vector<vector<unsigned long> > moveSecAddrRange = obj->getObject()->getMoveSecAddrRange();
+        vector<vector<unsigned long> > moveSecAddrRange = object->getMoveSecAddrRange();
 
         for (unsigned i = 0; i != moveSecAddrRange.size(); i++) {
             if ((moveSecAddrRange[i][0] == shdr->sh_addr) ||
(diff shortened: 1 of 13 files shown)


hainest:
I'm terribly sorry this has languished for so long. The conflict in Region.h is due to the addition of `operator<` and `struct less`. I'm curious why you're adding an ordering on a `Region`. Dyninst has a lot of these superfluous, and frankly meaningless, partial orderings defined on types that don't make sense just so they can be stored in ordered containers. I'd rather not perpetuate this.

I'm testing this PR now, so I'll update when I'm done.

batuzovk:
In PE symbols defined in Export Directory and relocations are described by their virtual address. In Dyninst they are tied to regions. To find enclosing region faster I sort vector of `Region *` by the Regions' memory offset.

I don't want to update the branch while you are testing it. Once some (intermediate) conclusion is reached, I'll update it and resolve the conflict.

hainest:
> In PE symbols defined in Export Directory and relocations are described by their virtual address. In Dyninst they are tied to regions. To find enclosing region faster I sort vector of `Region *` by the Regions' memory offset.

That's a property of a Region that is specific to the analysis of PE symbols, so let's move it out of the class.

> I don't want to update the branch while you are testing it. Once some (intermediate) conclusion is reached, I'll update it and resolve the conflict.

I'm seeing a lot of test regressions with this branch. I'm working on debugging them.

batuzovk:
Sorry it took so long to reply, Dyninst's testsuite isn't easy to figure out.

I think I've found the cause. `FunctionBase::expandLocation` tries to access dwarf, which is only usable in ObjectELF. But it uses `auto obj` which resolves to `Object *`. I'll rerun the testsuite and push an updated branch if everything is fine.

hainest:
> Sorry it took so long to reply, Dyninst's testsuite isn't easy to figure out.

It's not really usable for anyone outside of the core Dyninst devs. Even then, we still struggle to get it to work correctly. I had made a reproducer, but got sidetracked on another project. I should have posted it here, sorry about that.

> I think I've found the cause. `FunctionBase::expandLocation` tries to access dwarf, which is only usable in ObjectELF. But it uses `auto obj` which resolves to `Object *`. I'll rerun the testsuite and push an updated branch if everything is fine.

That makes sense. I was seeing bad `this` access. This suggests there is an architectural issue around discriminating between ObjectELF and anything else. We should reconsider some aspects of the `Object` interface- particularly anything around the old `a.out` format. We don't need that here. Getting PE parsing working for your work is more important right now.

batuzovk:
I've updated the branch. The issue in Function.C is resolved. Region comparison is removed (will be added back later in a different file). Branch is rebased on top of current master.

Some of the tests still don't work, but they don't work for master as well. I guess I wasn't able to set up testsuite correctly in the end.

> This suggests there is an architectural issue around discriminating between ObjectELF and anything else. 

This particular case was my mistake. While deciding where to put dwraf handle I ended up with it in both `Object` and `ObjectELF`, and since class fields are not virtual, the wrong one was used. I've removed it from the base class where it doesn't make sense.

symtabAPI/src/emitElf.C
```diff
@@ -573,7 +573,8 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
             }
         }
 
-        vector<vector<unsigned long> > moveSecAddrRange = obj->getObject()->getMoveSecAddrRange();
+        ObjectELF *elf_object = dynamic_cast<ObjectELF*>(obj->getObject());
```
hainest, line 576: All of these casts for `obj->getObject()` can be replaced with `this->object` since `object` being an instance of `ObjectELF` is a class invariant. I have no idea why the original authors didn't do this.

hainest (changes requested):
We require clang's `-Winconsistent-missing-override`. Here's a patch for that.
[override.patch.tar.gz](https://github.com/user-attachments/files/18913201/override.patch.tar.gz)

symtabAPI/src/emitElf.C
batuzovk: Technically, emitElf constructor allows situation where `obj->getObject()` is different from `object`. I don't think it is an intended use case though, so I've added an assert to the constructor.

batuzovk:
I've updated the branch with all the suggestions.

symtabAPI/src/emitElf.C
batuzovk: I've changed the constructor parameter to be `ObjectELF *`. I kept the assert because it checks a stronger constraint that `ObjectELF` and `Symtab` are describing the same entity.

hainest:
Minor item, but this is should probably be removed: https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Symbol.h#L78

hainest:
No test regressions on 17 machines.

batuzovk:
> Minor item, but this is should probably be removed: https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Symbol.h#L78

Fixed.

hainest:
> > Minor item, but this is should probably be removed: https://github.com/dyninst/dyninst/blob/master/symtabAPI/h/Symbol.h#L78
> 
> Fixed.

Thanks so much for your work on this.

hainest (approved):

PR #1767, opened 2024-07-15
Revert "Add MultiRegisterAST to instructionAPI for AMDGPU"

bbiiggppiigg:
Reverts dyninst/dyninst#1733


bbiiggppiigg:
There are missing return statements after assert in Result.h that causes compile errors.
Should I revert the commit and fix, or add the fix and update?

PR #1769, opened 2024-07-16
Fix vector segmentation fault introduced in #1733

bbiiggppiigg:
The use of reserve instead of resize for the operand vector somehow cause Dyninst to segmentation fault on test cases with GETPC, SETPC and SWAPPC.

### DIFF
--- a/dataflowAPI/src/RoseInsnFactory.C
+++ b/dataflowAPI/src/RoseInsnFactory.C
@@ -443,8 +443,8 @@ void RoseInsnAMDGPUFactory::massageOperands(const Instruction &insn,
     case amdgpu_gfx940_op_S_SWAPPC_B64: {
         assert(operands.size() == 4);
         operands.reserve(6);
-        operands[4] = operands[2];
-        operands[5] = operands[3];
+        operands.push_back(operands[2]);
+        operands.push_back(operands[3]);
         MultiRegisterAST::Ptr src_regs  = boost::dynamic_pointer_cast<MultiRegisterAST>(operands[1].getValue());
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
[test.tar.gz](https://github.com/user-attachments/files/16245463/test.tar.gz)

dataflowAPI/src/RoseInsnFactory.C
```diff
@@ -442,7 +442,7 @@ void RoseInsnAMDGPUFactory::massageOperands(const Instruction &insn,
     case amdgpu_gfx90a_op_S_SWAPPC_B64: 
     case amdgpu_gfx940_op_S_SWAPPC_B64: {
         assert(operands.size() == 4);
-        operands.reserve(6);
+        operands.resize(6);
         operands[4] = operands[2];
         operands[5] = operands[3];
```
kupsch, line 447: A better solution would be to use `reserve` to make sure the memory is allocated at most once, and then use `push_back` to expand the vector without constructing an operand that is then overwritten.

```cpp
        operands.reserve(6);
        operands.push_back(operands[2]);
        operands.push_back(operands[3]);
```

kupsch (commented):
I think you should use `reserve` and then `push_back` to add elements to the vector.

dataflowAPI/src/RoseInsnFactory.C
bbiiggppiigg: @kupsch I've implemented the suggested fix, and tested that it works.

kupsch (approved):
looks good. commit after CI passes.

PR #1772, opened 2024-07-25
Add liveness support for AMDGPU GFX908/90A/940

bbiiggppiigg:
Add initial suport for the AMDGPU GFX908, GFX90A and GFX940 architectures.

Not every architecturally available register is listed in the registerMap.

Adds new interfaces to livenessAnalyzer/getABI with architecture as an argument.


dataflowAPI/h/ABI.h
```diff
@@ -61,8 +61,10 @@ class ABI{
 
     DYNINST_EXPORT static void initialize32();
     DYNINST_EXPORT static void initialize64();
+    DYNINST_EXPORT static void initialize64(Architecture arch);
 
     DYNINST_EXPORT static ABI* getABI(int addr_width);
+    DYNINST_EXPORT static ABI* getABI(Architecture arch);
```
kupsch, line 67:
```suggestion
    // WARNING: the architecture is fixed on first use and cannot be reset
    //
    // if not set, set the architecture to the host. return addr_width ABI.
    DYNINST_EXPORT static ABI* getABI(int addr_width);
    // if not set, set the architecture to arch. return the 64-bit ABI.
    DYNINST_EXPORT static ABI* getABI(Architecture arch);
```

kupsch (approved):
This change is dangerous in that the ABI is fixed on the first call to getABI, and returns the set value on subsequent calls and will give the wrong ABI if anyone is using multiple ABI architectures. For now, add the comments to ABI.h.  This will need to get fixed in the work to support cross/multiple architectures.

PR #1776, opened 2024-09-19
Add support for 96 bit MultiRegister

bbiiggppiigg:


kupsch (changes requested):
You can't move the two static `Result` methods that were defined in `Result.h` (I don't know why someone chose to define all of the methods in the header file). These are part of the public interface and could be used by user code.  So the methods need remain and be updated in the header file since there isn't a `Result.C` to put them in.

bbiiggppiigg:
I added them in the commit that add support for MultiRegisterAST on July-15, and they are not intended to be used as public interface.

https://github.com/dyninst/dyninst/pull/1733/files#diff-42cd4da7ab534524d1eb297a79bdd930a3a79177764e032fb6140766ae698cbb

kupsch (approved):

PR #1782, opened 2024-09-25
warning fix: ambiguous-reversed-operator

kupsch:
In C++20 and later the signature of operator== should be T::operator==(const T&) const otherwise it is ambiguous as C++ 20 and later also looks for matching function with the arguments reversed and if only one is const then reversing results in two distinct function signatures that match equally well.  Some of these were reported as compiler warnings, others from manual inspection.


bbiiggppiigg (approved):

hainest:
I didn't do a careful diff, but how is this different from #1777?

kupsch:
It's the same.  I renamed the branch and github closed the old one.

PR #1783, opened 2024-09-26
update compiler-multibuild with new OS & compilers

kupsch:
- replace ubuntu 23.10 with ubuntu 24.04

- add new compilers:  gcc 14 and clang 18

- test C++20 and C++23 with clang as it now builds without warnings

- use gcc-14 and clang-18 to test C++ 11, 14, 17, 20 and 23


hainest (approved):

hainest:
Since this PR originates from your repo, I can't manually run the compiler multibuild job against it. Could you do that after you merge here?

kupsch:
I manually started the workflow.  I on dyninst/dyninst, and it ran successfully on my fork.

Thanks

Issue #1763, opened 2024-06-14
Missing entries in DWARF line information lookup


mplegendre:
When looking up line information entries by address, SymtabAPI is failing to return some line entries that should exist.

Here's output from the attached test case:
 
```
Source to address lookup for outline.c:9 to outline.c:12:
        outline.c:9 ranges are 10ad-10c0, 10c0-10c0, 10c0-10c3,
        outline.c:10 ranges are 10c0-10c0,
        outline.c:11 ranges are 10c0-10c0,
        outline.c:12 ranges are 10c0-10c0,
Address to source line lookups:
        10ad: /home/legendre/outline.c:9 is 10ad-10c0;
        10ae: /home/legendre/outline.c:9 is 10ad-10c0;
        10af: /home/legendre/outline.c:9 is 10ad-10c0;
        10b0: /home/legendre/outline.c:9 is 10ad-10c0;
        10b1: /home/legendre/outline.c:9 is 10ad-10c0;
        10b2: /home/legendre/outline.c:9 is 10ad-10c0;
        10b3: /home/legendre/outline.c:9 is 10ad-10c0;
        10b4: /home/legendre/outline.c:9 is 10ad-10c0;
        10b5: /home/legendre/outline.c:9 is 10ad-10c0;
        10b6: /home/legendre/outline.c:9 is 10ad-10c0;
        10b7: /home/legendre/outline.c:9 is 10ad-10c0;
        10b8: /home/legendre/outline.c:9 is 10ad-10c0;
        10b9: /home/legendre/outline.c:9 is 10ad-10c0;
        10ba: /home/legendre/outline.c:9 is 10ad-10c0;
        10bb: /home/legendre/outline.c:9 is 10ad-10c0;
        10bc: /home/legendre/outline.c:9 is 10ad-10c0;
        10bd: /home/legendre/outline.c:9 is 10ad-10c0;
        10be: /home/legendre/outline.c:9 is 10ad-10c0;
        10bf: /home/legendre/outline.c:9 is 10ad-10c0;
        10c0: /home/legendre/outline.c:9 is 10c0-10c3;
        10c1: /home/legendre/outline.c:9 is 10c0-10c3;
        10c2: /home/legendre/outline.c:9 is 10c0-10c3;
 ```
 
[...]

PR #1785, opened 2024-09-30
support single address line map entries

kupsch:
Single address line map entries can be passed to this function as a range with the low and high address being equal.  If so, adjust the range to the correct closed, open range [low, low + 1)

Partial fix for #1763.


bbiiggppiigg:
I don't know enough about dwarf but the change make sense to me.

bbiiggppiigg (approved):

PR #1786, opened 2024-10-02
Remove elfutils from public dependencies

hainest:
It's only used internally, so it shouldn't be exported. This will also make supporting other files formats (e.g., Windows PE) easier since we can properly hide those details.

This will cause link failures for consumers that explicitly pass `-ldyn{Dwarf,Elf}`. I'll run the consumers workflow to check for this, but I'm fairly certain I'll need to send patches to systemtap and TAU.

This will also cause a couple of failures in dyninst/external-tests because it tries to compile the dyn{Elf,Dwarf} headers. I'll post a PR to remove those tests.


kupsch (approved):

hainest:
I've sent a patch to the TAU folks. I'll merge here once I've heard back from them.

PR #1789, opened 2024-10-07
remove singleton_object_pool

kupsch:
- All uses of singleton_object_pool use the standard allocator, so this serves no purpose

- replace uses of the function ::make_shared with boost::make_shared as ::make_shared can leak if exceptions are thrown, and it requires two memory allocation for the shared_ptr's metadata and the object itself where boost::make_shared combines them into one

- remove pool_allocators.h as it is unused


hainest (approved):
LGTM. I take it you are working your way through the AST construction for x86?

PR #1794, opened 2024-10-10
Fix duplicate representations for AMDGPU registers

hainest:
Is hw_reg_ib_sts intended to be aliased to hw_reg_status for 9{0a,40}? What about for 908? If so, we can close this.


bbiiggppiigg:
@hainest 
Yeah I think it is a bug.
It will fixed as part of  #1791, but that PR requires a bit more work to be done.
If you think it is urgent I should approve this PR otherwise maybe we can wait a little bit ?

hainest:
> If you think it is urgent I should approve this PR otherwise maybe we can wait a little bit ?

Yup, that's fine. I just noticed it, so I made a fix. I'll close this.

Issue #1115, opened 2021-10-21
Remove support for PPC32


hainest:
This architecture is no longer in use. We must be careful to check that the PPC64 code does not implicitly rely upon any of the explicit PPC32 code.

hainest:
`InstructionDecoder` constructor does not set is64 for PPC.

hainest:
Notes on ppc32 support:

[fedora](https://developer.fedoraproject.org/deployment/secondary_architectures/powerpc.html): not supported after Fedora 22 (EOL since 2016)

[debian](https://wouter.pages.debian.net/webwml/ports/powerpc/index): not supported since Debian 8 ([EOL](https://www.debian.org/releases/) since 2025).

[gentoo](https://wiki.gentoo.org/wiki/Project:PowerPC): still supported, but most mentions are for PowerMacs. I'm not sure if they support Power8 or newer (that's the minimum Dyninst supports).

[openSUSE](https://en.opensuse.org/Portal:PowerPC): dropped after Evergreen 11.4. The last Evergreen release was [13.1](https://en.opensuse.org/openSUSE:Evergreen) which had its support dropped in 2016.

[RHEL](https://access.redhat.com/solutions/509373): dropped in RHEL 7 (last [updated](https://access.redhat.com/articles/red-hat-enterprise-linux-release-dates#RHEL7) in 2020).

https://powerpc.dev/linux/distros.html has a few others, but no citations for their sources.

PR #1803, opened 2024-10-25
Import and update ppc64le opcodes from Capstone

hainest:
Capstone is missing several opcodes, most of which are from the Decimal Floating Point extension. I'm going to work with them  to update their importer to include them.

ppc32 is explicitly not included because it's no longer supported (#1115, #1145).

Zero hurry on this.


hainest:
No test regressions on Power{9,10}. I don't have access to a Power8 machine.

kupsch (approved):

PR #1804, opened 2024-10-27
Fix Boost not found error on RHEL 9

blue42u:
Currently Dyninst can only find Boost if it has a `BoostConfig.cmake` (implied by the `find_package(... HINTS ...)` keyword). Some notable distros, specifically RHEL 9.x, have a new enough Boost to build Dyninst but do not package `BoostConfig.cmake`, which breaks the build.

This PR removes the `HINTS` syntax and allows using the `FindBoost.cmake` module shipped with CMake, which supports common variables like `BOOST_ROOT`, works without `BoostConfig.cmake`, and supports older versions of Boost.

### DIFF
--- a/cmake/tpls/DyninstBoost.cmake
+++ b/cmake/tpls/DyninstBoost.cmake
@@ -17,8 +17,12 @@
 
 include_guard(GLOBAL)
 
-# Need at least 1.71 for a usable BoostConfig.cmake
 set(_min_version 1.71.0)
+
+# Silence a warning about CMake 3.30 removing the FindBoost module
+if(POLICY CMP0167)
+  cmake_policy(SET CMP0167 NEW)
+endif()
 
 # Use multithreaded libraries
 set(Boost_USE_MULTITHREADED ON)
@@ -26,6 +30,7 @@ set(Boost_USE_MULTITHREADED ON)
 # Don't use libraries linked statically to the C++ runtime
 set(Boost_USE_STATIC_RUNTIME OFF)
 
+# Follow convention with custom Find modules
 if(Boost_ROOT_DIR)
   set(Boost_NO_SYSTEM_PATHS ON)
   set(Boost_ROOT ${Boost_ROOT_DIR})
@@ -36,15 +41,7 @@ set(Boost_NO_WARN_NEW_VERSIONS ON)
 
 # Library components that need to be linked against
 set(_boost_components atomic chrono date_time filesystem thread timer)
-find_package(
-  Boost
-  ${_min_version}
-  QUIET
-  REQUIRED
-  HINTS
-  ${PATH_BOOST}
-  ${BOOST_ROOT}
-  COMPONENTS ${_boost_components})
+find_package(Boost ${_min_version} QUIET REQUIRED COMPONENTS ${_boost_components})
 
 # Don't let Boost variables seep through
 mark_as_advanced(Boost_DIR)


cmake/tpls/DyninstBoost.cmake
```diff
@@ -17,33 +17,27 @@
 
 include_guard(GLOBAL)
 
-# Need at least 1.71 for a usable BoostConfig.cmake
-set(_min_version 1.71.0)
```
hainest, line 21: We use this to synchronize the version requirements with the containers (see [docker/Dockerfile.ubuntu](https://github.com/dyninst/dyninst/blob/master/docker/Dockerfile.ubuntu).

cmake/tpls/DyninstBoost.cmake
```diff
@@ -17,33 +17,27 @@
...
+  cmake_policy(SET CMP0167 NEW)
+endif()
 
 # Use multithreaded libraries
 set(Boost_USE_MULTITHREADED ON)
 
 # Don't use libraries linked statically to the C++ runtime
 set(Boost_USE_STATIC_RUNTIME OFF)
 
-if(Boost_ROOT_DIR)
```
hainest, line 29: Because CMake's [search rules](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) are complex, all of our custom finders use the convention of `Package_ROOT_DIR` to force CMake to use the given location. Disabling searching system paths is necessary (but not necessarily sufficient) to make that happen. Translating from Boost_ROOT_DIR to Boost_ROOT also alleviates needing an extra HINT.

hainest (changes requested):
> Currently Dyninst can only find Boost if it has a `BoostConfig.cmake` (implied by the `find_package(... HINTS ...)` keyword).

Boost added a broken one in v1.70.0 which is why the minimum supported version is 1.71.0 where they fixed it.

>Some notable distros, specifically RHEL 9.x, have a new enough Boost to build Dyninst but do not package `BoostConfig.cmake`, which breaks the build.

Oh, RedHat, you do everything in your power to make things worse...

We test with Fedora 41 using their provided packages for our dependencies. How far is that from RHEL9? Also, why would RHEL9 not ship the same Boost package as Fedora? That seems janky.
 
> This PR removes the `HINTS` syntax

`BOOST_ROOT` shouldn't have been in there. That was my error. However, `PATH_BOOST` was part of the old build system. We can *probably* throw it away without disrupting anyone (https://xkcd.com/1172/) since hints are searched after the Boost_ROOT-style variables.

>and allows using the `FindBoost.cmake` module shipped with CMake, which supports common variables like `BOOST_ROOT`, works without `BoostConfig.cmake`, and supports older versions of Boost.

But enabling CMP0167 makes FindBoost.cmake no longer available. *If* I've RTFM correctly,

> CMake 3.30 and above prefer to not provide the FindBoost module so that find_package(Boost) calls, without the CONFIG or NO_MODULE options, find the upstream BoostConfig.cmake directly.

would mean that enabling CMP0167 wouldn't fix the build on RHEL9 because it doesn't provide a BoostConfig.cmake.

cmake/tpls/DyninstBoost.cmake
blue42u: Whoops, fixed.

blue42u:
I've pushed a new (smaller) version of the patch that should address the mentioned compatibility issues.

> We test with Fedora 41 using their provided packages for our dependencies. How far is that from RHEL9? Also, why would RHEL9 not ship the same Boost package as Fedora? That seems janky.

Pretty far AFAICT, RHEL 9 is 1.75.x while Fedora is at 1.83.x ([ref](https://pkgs.org/download/boost-devel)). The rules for RHEL are a bit incomprehensible, some packages (apps? e.g. `podman`) get aggressive updates while others (libraries, e.g. `boost-devel`) don't.

> But enabling CMP0167 makes FindBoost.cmake no longer available. *If* I've RTFM correctly, [...] would mean that enabling CMP0167 wouldn't fix the build on RHEL9 because it doesn't provide a BoostConfig.cmake.

For compatibility with CMake >3.30 the policy has to be set to `NEW`, because now that the policy is in place CMake can drop the FindBoost module at any time and generate a fatal error if the policy is set to `OLD` (or left unset). In other words, the Dyninst CMake code is fully compatible with both the module (`OLD`) and the config search (`NEW`), and setting the policy to `NEW` expresses this to CMake.

The expectation here is that users of CMake >=3.30 will use a Boost that has a suitable `BoostConfig.cmake` (i.e. >=1.71.0 and not RHEL 9.x). RHEL 9 (and even today's Fedora) don't ship a new enough CMake for this to matter, but this requirement is something 3rd-party package managers (esp. Spack) will need to be made aware of if they aren't already.

cmake/tpls/DyninstBoost.cmake
blue42u: (IMHO using such a convention is evil and confusing. I understand the desire to assert e.g. "Boost comes from this directory" but muddying CMake's already heady conventions with your own is an abominable approach. But not my project nor a feature I require, so I'll be quiet now.)

The latest version of the patch has the intended behavior, either `-DBoost_ROOT_DIR` or `-DBoost_ROOT` will find an externally-installed Boost and take precedence over a system-installed Boost.

woodard:
> > Currently Dyninst can only find Boost if it has a `BoostConfig.cmake` (implied by the `find_package(... HINTS ...)` keyword).
> 
> Boost added a broken one in v1.70.0 which is why the minimum supported version is 1.71.0 where they fixed it.
> 
> > Some notable distros, specifically RHEL 9.x, have a new enough Boost to build Dyninst but do not package `BoostConfig.cmake`, which breaks the build.
> 
> Oh, RedHat, you do everything in your power to make things worse...

We can ask for that to be fixed.  That is the first time I've heard of this problem.
I think that I may need a bit more detail that what is in this case to request a fix though. Can you send me an email?

> 
> We test with Fedora 41 using their provided packages for our dependencies. How far is that from RHEL9? Also, why would RHEL9 not ship the same Boost package as Fedora? That seems janky.

In particular, this suggests to me that there are some additional details that need further explanation in a bug report.

woodard:
> I've pushed a new (smaller) version of the patch that should address the mentioned compatibility issues.
> 
> > We test with Fedora 41 using their provided packages for our dependencies. How far is that from RHEL9? Also, why would RHEL9 not ship the same Boost package as Fedora? That seems janky.
> 
> Pretty far AFAICT, RHEL 9 is 1.75.x while Fedora is at 1.83.x ([ref](https://pkgs.org/download/boost-devel)). The rules for RHEL are a bit incomprehensible, some packages (apps? e.g. `podman`) get aggressive updates while others (libraries, e.g. `boost-devel`) don't.
> 

General rule of thumb:
podman is an application and doesn't export libraries used outside of the application and so what stays stable is the UI and any scriptable interfaces. Podman also underlies many of our newest technology and customers really are pushing it finding bugs and requesting features and so it is under heavy development right now.

boost-devel and other things that customers may rely on to build in-house software stay the same through a whole major release. We don't want to break someone's internal build environment. In extreme cases of conflicting priorities, we will ship two versions of something. e.g. gcc the system compiler version and gcc-toolset-<some higher number>

There are some exceptions to these broad general rules. It is all really heavily documented. If you have questions, I can send you a ton of material explaining how and why certain things change and why other things don't. It is fairly well thought out and tries to balance stability and supportability with new features and capability.

cmake/tpls/DyninstBoost.cmake
hainest: >(IMHO using such a convention is evil and confusing. I understand the desire to assert e.g. "Boost comes from this directory" but muddying CMake's already heady conventions with your own is an abominable approach. But not my project nor a feature I require, so I'll be quiet now.)

Worry not, I share the sentiment. But this addresses two of my biggest gripes with CMake: a lack of configurability and a lack of "standard" conventions. I spent an unreasonable amount of time researching this and took up yet another unreasonable amount of Jim's time deciding on what we really wanted to do. But "use only this directory" was most important to us. Given what I've learned from CMake talks in the interim, I would have just added a dyninst-specific flag to constrain the search process instead of using different _ROOT specifiers.

>The latest version of the patch has the intended behavior: either -DBoost_ROOT_DIR or -DBoost_ROOT will find an externally-installed Boost, the former will fail if not found and the latter can fall back to a system-installed Boost.

Well, I *was* going to say that's always been the case by design, but I just noticed that `Boost_ROOT_DIR` didn't properly fail when pointed to a bad Boost install without this PR. I'm fairly certain I had tested that while I was working on the CMake updates. Ugh. I'll add more tests.

```
$ cmake .. -DBoost_ROOT=/home/tim/Downloads/boost_1_86_0/install
                                                  ^^^^^
...
Found Boost 1.86.0
Boost include directories: /home/tim/Downloads/boost_1_86_0/install/include

$ cmake .. -DBoost_ROOT=/home/tim/Downloads/boost_1_66_0/install
                                                  ^^^^^
...
Found Boost 1.74.0
-- Boost include directories: /usr/include
```

hainest (changes requested):

hainest (approved):

blue42u:
Amended to fix `cmake-format` CI error.

PR #1806, opened 2024-10-29
Add Option to Override the detection of host archiecture

bbiiggppiigg:
This option allows us to override the default target detection with user input 
architecture.

It must be one of the known archtecture specified in known_arches specified in
DyninstPlatform.cmake


cmake/DyninstPlatform.cmake
```diff
@@ -47,16 +49,39 @@ if(DYNINST_OS_Linux OR DYNINST_OS_FreeBSD)
...
 
-if(${_host_arch} IN_LIST _64bit_x86_arches)
-  set(DYNINST_ARCH_x86_64 TRUE)
-elseif(${_host_arch} STREQUAL "aarch64")
-  set(DYNINST_ARCH_aarch64 TRUE)
-elseif(${_host_arch} STREQUAL "ppc64le")
-  set(DYNINST_ARCH_ppc64le TRUE)
-elseif(${_host_arch} IN_LIST _32bit_x86_arches)
-  set(DYNINST_ARCH_i386 TRUE)
+if(DYNINST_CODEGEN_ARCH)
```
hainest, line 52: To prevent code duplication, let's try

```suggestion
if(DYNINST_CODEGEN_ARCH)
  set(${_host_arch} ${DYNINST_CODEGEN_ARCH})
endif()
```

and then do the checks using `${_host_arch}`.

hainest (changes requested):
This doesn't build because macros like [RTlinux.c:381](https://github.com/dyninst/dyninst/blob/master/dyninstAPI_RT/src/RTlinux.c#L326) don't check for unknown architectures.

bbiiggppiigg:
@hainest 

The problem seems to be that Dyninst is using the ucontext.h of the host, which has different ways of accessing the PC in the ucontext, when cap_mutatee_trap is enabled, which is when DYNINST_OS_Linux is set.

I'm not sure what is the right solution for this problem.
It seems like the architecture for execution needs to be separated from the architecture for code-generation.

@kupsch Any suggestions?

hainest:
> It seems like the architecture for execution needs to be separated from the architecture for code-generation.

I would suggest that CMake only be used to determine the environment that will be used for execution vis a vis proccontrol and dyninstRT. Those are the only places where we actually have to communicate with the OS and hardware.

For rewriting/codegen, we should determine the platform for the input file (e.g., ELF on Aarch64) at runtime. If a user wants to rewrite a ppc64le binary on an x86_64 host, that should just work. If we force a codegen type at build time and then don't check the input platform, we could get into a situation of a user trying to rewrite an Aarch64 binary on an x86_64 host using ppc64le codegen which will not and should not work. Of course, this is a monumental change. But that flag has already been raised.

PR #1809, opened 2024-11-08
Use correct semantics when constructing AST for return instructions

hainest:
Instruction::makeReturnInstruction always assumes a 32-bit unsigned value is read from the top of the stack. This is not true for Aarch64 and ppc64le (and probably not for RISCV). In fact, ppc and x86 (kind of) already create AST for return instructions, but it gets ignored by Instruction::getControlFlowTarget.


kupsch (approved):

hainest:
This approach won't work because `Instruction::getCategory()` calls `InstructionDecoder_<ARCH>::decodeOperands`.

PR #1813, opened 2024-11-13
Add category helpers to Instruction class

hainest:
getCategory() assumes there is only one category per instruction, but that restriction will be removed when using the Capstone decoder and the function deprecated. These helpers will make it easier for users to check behavior without having to know how it's being tracked in Instruction.

This makes IA_IAPI::hasCFT() more expensive, but it will be much cheaper as we move more architectures to Capstone.

### DIFF
--- a/dataflowAPI/src/liveness.C
+++ b/dataflowAPI/src/liveness.C
@@ -526,10 +526,7 @@ ReadWriteInfo LivenessAnalyzer::calcRWSets(Instruction curInsn, Block *blk, Addr
       }
     }
   }
-  InsnCategory category = curInsn.getCategory();
-  switch(category)
-  {
-  case c_CallInsn:
+  if(curInsn.isCall()) {
       // Call instructions not at the end of a block are thunks, which are not ABI-compliant.
       // So make conservative assumptions about what they may read (ABI) but don't assume they write anything.
       ret.read |= (abi->getCallReadRegisters());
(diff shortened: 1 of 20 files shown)


hainest:
Compiles on all three architectures. No test suite regressions.

dataflowAPI/src/liveness.C
```diff
@@ -526,43 +526,34 @@ ReadWriteInfo LivenessAnalyzer::calcRWSets(Instruction curInsn, Block *blk, Addr
...
-      const bool isSyscall = Dyninst::InstructionAPI::isSystemCall(curInsn);
-
-      if (isInterrupt || isSyscall) {
-	ret.read |= (abi->getSyscallReadRegisters());
-	ret.written |= (abi->getSyscallWrittenRegisters());
-      }
+  }
+  else {
+    auto is_interrupt = Dyninst::InstructionAPI::isSoftwareInterrupt(curInsn);
+    if (is_interrupt || curInsn.isSyscall()) {
```
kupsch, line 552: For uniformity, any reason not to add a `isSoftwareInterrup()` method to the instruction?

```suggestion
    if (curInsn. isSoftwareInterrupt() || curInsn.isSyscall()) {
```

kupsch (approved):
One small thing to consider, otherwise looks good.

dataflowAPI/src/liveness.C
hainest: There isn't a category for interrupts. I plan to update that next.

kupsch: Sure, the `isSoftwareInterrupt` method can be added when you add it as a category instead of now.

parseAPI/src/IA_power.C
```diff
@@ -312,7 +312,7 @@ bool IA_power::isReturnAddrSave(Address& retAddr) const
 bool IA_power::isReturn(Dyninst::ParseAPI::Function * context, Dyninst::ParseAPI::Block* currBlk) const
 {
   /* Check for leaf node or lw - mflr - blr pattern */
-  if (curInsn().getCategory() != c_ReturnInsn) {
+  if (curInsn().isReturn()) {
```
bbiiggppiigg, line 315: This is missing a not.

parseAPI/src/IA_x86.C
```diff
@@ -814,8 +811,7 @@ bool IA_x86::isIATcall(std::string &calleeName) const
 
 bool IA_x86::isNopJump() const
 {
-    InsnCategory cat = curInsn().getCategory();
-    if (c_BranchInsn != cat) {
+    if (curInsn().isBranch()) {
```
bbiiggppiigg, line 814: This is also missing a not.

parseAPI/src/IA_IAPI.C
```diff
@@ -461,27 +459,27 @@ bool IA_IAPI::isAbsoluteCall() const
...
 {
-    Instruction ci = curInsn();
-    return (ci.getOperation().getID() == e_sysenter);
+    return curInsn().isSysEnter();
 }
 
 bool IA_IAPI::isIndirectJump() const {
     Instruction ci = curInsn();
-    if(ci.getCategory() != c_BranchInsn) return false;
+    if(ci.isBranch()) return false;
```
bbiiggppiigg, line 482: Missing not here.

parseAPI/src/IA_power.C
bbiiggppiigg: @hainest
It seems like this check is removed in #1929.
But I'm not sure if you moved this logic somewhere else in that PR.
If you did, please let me know so I can include the fix in #2074

hainest: That was checking for idioms in ancient PowerPC and isn't needed anymore.

PR #1816, opened 2024-11-15
Add hasControlFlowTarget to Instruction class

hainest:
Checks if the instruction has a control flow target. For an Instruction object insn, these are equivalent:

  if(ins.hasControlFlowTarget())
  if(ins.getControlFlowTarget())

but the first does not do any expression evaluation and can be substantially cheaper.

### DIFF
--- a/dataflowAPI/src/stackanalysis.C
+++ b/dataflowAPI/src/stackanalysis.C
@@ -2459,7 +2459,7 @@ bool StackAnalysis::handleNormalCall(Instruction insn, Block *block,
       }
    }
 
-   if (!insn.getControlFlowTarget()) return false;
+   if (!insn.hasControlFlowTarget()) return false;
 
    // Must be a thunk based on parsing.
    if (off != block->lastInsnAddr()) return false;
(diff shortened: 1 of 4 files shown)


kupsch (commented):
I don't think this is equivalent and making it equivalentn eliminates the efficiency of this new method. `getControlFlowTarget` executes the code in the macro `DECODE_OPERANDS` which is what causes `m_Successors` to be populated.  `hasControlFlowTarget` does not execute the macro, so if prior to `hasControlFlowTarget` being called there is no other function called that includes `DECODE_OPERANDS`, then the result will always be false as `m_Successors` will always be empty.

hainest:
> I don't think this is equivalent and making it equivalentn eliminates the efficiency of this new method. `getControlFlowTarget` executes the code in the macro `DECODE_OPERANDS` which is what causes `m_Successors` to be populated. `hasControlFlowTarget` does not execute the macro, so if prior to `hasControlFlowTarget` being called there is no other function called that includes `DECODE_OPERANDS`, then the result will always be false as `m_Successors` will always be empty.

You are quite right. I'm not sure how I missed that.

PR #1818, opened 2024-11-16
Remove InstructionDecoder_x86::decode

hainest:
This doesn't need to be implemented since it just calls the one in the base.


hainest:
No test regressions.

bbiiggppiigg (commented):
I'm not entirely sure, but shouldn't it be the otherway around?
Move the x86 implementation into InstructionDecoder-x86.C and make it virtual on the base class?

hainest:
> Move the x86 implementation into InstructionDecoder-x86.C and make it virtual on the base class?

It is virtual in the base class (https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/InstructionDecoderImpl.h#L54).

> I'm not entirely sure, but shouldn't it be the otherway around?

The confusion is completely understandable. The `InstructionDecoderImpl` class is just weird.

`InstructionDecoderImpl::decode` calls the protected pure virtual `decodeOpcode` in the derived class.

For aarch64, `InstructionDecoder_aarch64::decode` does real work and *doesn't* call 
`InstructionDecoderImpl::decode`. ppc does the same.

For x86, we have `InstructionDecoder_x86::decode` -> `InstructionDecoderImpl::decode` -> `InstructionDecoder_x86::decodeOpcode`. This is equivalent to x86 *not* overriding `decode` and just use the default behavior of `InstructionDecoderImpl::decode`. That's what this PR does. I have no idea why it's like this.

bbiiggppiigg:
> For aarch64, `InstructionDecoder_aarch64::decode` does real work and _doesn't_ call `InstructionDecoderImpl::decode`. ppc does the same.
> 
> For x86, we have `InstructionDecoder_x86::decode` -> `InstructionDecoderImpl::decode` -> `InstructionDecoder_x86::decodeOpcode`. This is equivalent to x86 _not_ overriding `decode` and just use the default behavior of `InstructionDecoderImpl::decode`. That's what this PR does. I have no idea why it's like this.

AMDGPU decoders also implements their own decode and does not call InstructionDecoderImpl::decode.
So I was wondering if it is better that we just move the implementation from
https://github.com/dyninst/dyninst/blob/de98a78e00642c240d19f38933a4eff0f1ddba7b/instructionAPI/src/InstructionDecoderImpl.C#L59-L64
to InstructionDecoder_x86::decode, and remove InstructionDecoderImpl::decode completely.
This way the decoding logic for each architecture are all implemented in the architecture specific decoder implementation.

bbiiggppiigg (approved):
LGTM

PR #1819, opened 2024-11-20
Separate operand interface in Instruction class

hainest:
There are many places in Dyninst where only one type of operand is needed. This removes the need to do the filtering at those code points and allows for possibly expanding what an implicit register is (e.g., vector mask registers in some cases).

Deprecates Instruction::getOperands(std::vector<Operand>).

### DIFF
--- a/instructionAPI/src/syscalls.C
+++ b/instructionAPI/src/syscalls.C
@@ -130,8 +130,7 @@ namespace x86 {
      *  so we'll consider it as our special case here.
      */
     if(id == e_call) {
-      std::vector<di::Operand> operands;
-      ins.getOperands(operands);
+      auto operands = ins.getExplicitOperands();
 
       if(operands.size() != 1) {
         di::decode_printf("[%s:%d] Incorrect number of arguments to 'call'. Found %lu, expected 1\n",
--- a/dataflowAPI/src/AbslocInterface.C
+++ b/dataflowAPI/src/AbslocInterface.C
@@ -520,8 +520,7 @@ void AssignmentConverter::convert(const Instruction &I,
         // SGPR_PAIR[0] = PC & 0xffffffff
         // SGPR_PARI[1] = PC >> 32
         //
-        std::vector<Operand> operands;
-        I.getOperands(operands);
+        auto operands = I.getAllOperands();
         assert(operands.size() == 2);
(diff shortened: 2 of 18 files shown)


hainest:
The deprecation does not generate any warnings in the known consumers (https://github.com/dyninst/dyninst/actions/runs/11957537634).

instructionAPI/src/syscalls.C
```diff
@@ -130,8 +130,7 @@ namespace x86 {
      *  so we'll consider it as our special case here.
      */
     if(id == e_call) {
-      std::vector<di::Operand> operands;
-      ins.getOperands(operands);
+      auto operands = ins.getExplicitOperands();
```
bbiiggppiigg, line 133: Should this bet getAllOperands?

hainest: This code is checking for the special `call gs:[0x10]` for 32-bit x86. Technically, that is the same as `call [gs * 16 + 0x10]` where the `16` is an implicit operand, but I haven't decided if I want to implement it in AST that way. I just guarded against it here.

bbiiggppiigg: The next line explicitly checks  if(operands.size() != 1)  return false.
So I was just wandering if changing from the default behavior of checking against all operands to explicit operands would make a difference.
If there is only one operand for this instruction, then maybe it doesnt matter.

hainest: Oh! Now I remember. The old decoder didn't include `gs` in the operands even though it is *definitey* one. In fact, it's mentioned in the comment on line 128. It's a good thing I put that there, so I could not read it in the future. At least old me was paying attention. /shrug

The updated AST using Capstone has both `gs` and `0x10`, but `gs` is taken to be implicit. I actually don't think that's correct, either. I'll double-check and update that and this test for the Capstone branch.

kupsch (approved):
In the documentation for `getOperand` it would be good to explicitly state that this is all the operands by changing "_...  returns the operand at position \code{index}, ..._" to "_...  returns the operand at position \code{index} **of the vector returned by getAllOperands()**, ..._".

hainest:
> In the documentation for `getOperand` it would be good to explicitly state that this is all the operands by changing "_... returns the operand at position \code{index}, ..._" to "_... returns the operand at position \code{index} **of the vector returned by getAllOperands()**, ..._".

Updated.

PR #1820, opened 2024-11-21
Make InstructionDecoder::doDelayedDecode private

hainest:
This allows for safer assumptions in the upcoming Capstone-base decoders. It also makes no sense for users to call this, and was never documented.


bbiiggppiigg (commented):
Should we try to fix the indentation of this file?
Or is it fine the way it is?

hainest:
> Should we try to fix the indentation of this file? Or is it fine the way it is?

I'm actually working on a PR for that. :)

bbiiggppiigg (approved):

hainest:
Consumers are building fine (https://github.com/dyninst/dyninst/actions/runs/11958633928).

PR #1821, opened 2024-11-22
Clean up InstructionAPI decoder and AST classes

hainest:
These classes will need non-trivial modifications for the move to Capstone. This makes them easier to work with.

### DIFF
--- a/instructionAPI/h/Dereference.h
+++ b/instructionAPI/h/Dereference.h
@@ -81,95 +81,12 @@ namespace Dyninst
       virtual std::string format(formatStyle) const
       {
     std::string retVal;
-#if defined(DEBUG_MEMORY_ACCESS_WIDTH)
-        switch(Expression::userSetValue.type)
-        {
-            case u8:
-                retVal += "u8 @ ";
-                break;
-            case s8:
-                retVal += "s8 @ ";
-                break;
-            case u16:
-                retVal += "u16 @ ";
-                break;
-            case s16:
-                retVal += "s16 @ ";
-                break;
-            case u32:
-                retVal += "u32 @ ";
-                break;
-            case s32:
-                retVal += "s32 @ ";
-                break;
-            case u64:
-                retVal += "u64 @ ";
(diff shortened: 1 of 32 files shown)


instructionAPI/h/Dereference.h
```diff
@@ -31,211 +31,83 @@
...
-#endif
-    retVal += "[" + addressToDereference->format() + "]";
-        return retVal;
-      }
+namespace Dyninst { namespace InstructionAPI {
+  using std::vector;
 
-        virtual std::string format(Architecture arch, formatStyle) const
-        {
-#if defined(DEBUG_MEMORY_ACCESS_WIDTH)
```
bbiiggppiigg, line 160: Are you sure these debug will never be used?

hainest: These formatters are going away with Capstone. There are so many of these little debug macros everywhere that have clearly not been used since the original implementation. For example, this one doesn't cover all of the possible sizes *and* `Result` has its own formatter method.

bbiiggppiigg (approved):

instructionAPI/src/BinaryFunction.C
```diff
@@ -32,425 +32,375 @@
...
+                               arg2.convert<Result_type2type<u32>::type>());
+      case s48:
+      case u48:
+        return Result(u48, arg1.convert<Result_type2type<u48>::type>() |
+                               arg2.convert<Result_type2type<u48>::type>());
+      case s64:
+      case u64:
+        return Result(u64, arg1.convert<Result_type2type<u64>::type>() |
+                               arg2.convert<Result_type2type<u64>::type>());
+      /*case sp_float:
```
bbiiggppiigg, line 263: This comment is kind of hard to read.

hainest: That looks like dead code. I'll remove it and the ones above.

bbiiggppiigg (approved):

PR #1822, opened 2024-11-22
Fix null pointer access in Operand::get{Read,Write}Set

hainest:


bbiiggppiigg (approved):
Looks like a bug I introduced.

hainest:
> Looks like a bug I introduced.

I have introduced more than one...

PR #1828, opened 2024-12-05
Fix crash in DwarfFrameParser::create

hainest:
When running with ~4 or more threads, there is an intermittent crash during the destruction of frameParsers.


hainest:
Running it 100 times using 1, 2, or 4 threads, I only saw crashes with 4 threads. After this fix, I couldn't get a crash even with 8 threads after 100 runs.

kupsch (approved):

PR #1831, opened 2024-12-06
Add RISC-V instruction mnemonics and registers

wxrdnx:
Added RISC-V register and instruction mnemonic definitions.
Added RISC-V support to the Symtab API so that it can parse and understand RISC-V ELF binaries.


kupsch:
@wxrdnx, a more descriptive title for the PR and the commit would be `add RISC-V instruction mnemonics and registers`.  You can change the PR title with the `edit` button, and you change the commit using `commit --amend` or interactive rebase and a forced push.

wxrdnx:
> @wxrdnx, a more descriptive title for the PR and the commit would be `add RISC-V instruction mnemonics and registers`. You can change the PR title with the `edit` button, and you change the commit using `commit --amend` or interactive rebase and a forced push.

All done. The wrong comment in `riscv64_regs.h` is also fixed.

hainest:
Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.

wxrdnx:
> Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.

Got it.
Also I have noticed that `DwarfEncToReg` is added to `MachRegister.C`, so I need to add the translation code for Dwarf register.
I'll probably do them tomorrow.

hainest:
> > Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.
> 
> Got it. Also I have noticed that `DwarfEncToReg` is added to `MachRegister.C`, so I need to add the translation code for RISC-V Dwarf registers. I'll probably do them tomorrow and also rebase my branch to master.

I had thought about that after I commented yesterday, but didn't have a chance to leave a comment. Much like the ROSE translations, I plan to move the DWARF translations out of MachRegister. I'll make a draft PR to do that and wait to merge it until you have time to add the translation tables.

wxrdnx:
> > > Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.
> > 
> > 
> > Got it. Also I have noticed that `DwarfEncToReg` is added to `MachRegister.C`, so I need to add the translation code for RISC-V Dwarf registers. I'll probably do them tomorrow and also rebase my branch to master.
> 
> I had thought about that after I commented yesterday, but didn't have a chance to leave a comment. Much like the ROSE translations, I plan to move the DWARF translations out of MachRegister. I'll make a draft PR to do that and wait to merge it until you have time to add the translation tables.

I've just added the translation tables, so I think #1845 becomes mergeable.

hainest:
> > > > Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.
> > > 
> > > 
> > > Got it. Also I have noticed that `DwarfEncToReg` is added to `MachRegister.C`, so I need to add the translation code for RISC-V Dwarf registers. I'll probably do them tomorrow and also rebase my branch to master.
> > 
> > 
> > I had thought about that after I commented yesterday, but didn't have a chance to leave a comment. Much like the ROSE translations, I plan to move the DWARF translations out of MachRegister. I'll make a draft PR to do that and wait to merge it until you have time to add the translation tables.
> 
> I've just added the translation tables, so I think #1845 becomes mergeable.

If I merge that, then you'll need to rebase this PR and move the decode tables out of MachRegister.C and into dataflowAPI/rose/registers. If we merge this PR first, then I can take care of that code motion. Either way works for me. I just don't want to add work for you unnecessarily.

wxrdnx:
> > > > > Ah, sorry. I didn't look at this before merging #1842. You'll need to move the ROSE register stuff into dataflowAPI. Let me know if you have any questions.
> > > > 
> > > > 
> > > > Got it. Also I have noticed that `DwarfEncToReg` is added to `MachRegister.C`, so I need to add the translation code for RISC-V Dwarf registers. I'll probably do them tomorrow and also rebase my branch to master.
> > > 
> > > 
> > > I had thought about that after I commented yesterday, but didn't have a chance to leave a comment. Much like the ROSE translations, I plan to move the DWARF translations out of MachRegister. I'll make a draft PR to do that and wait to merge it until you have time to add the translation tables.
> > 
> > 
> > I've just added the translation tables, so I think #1845 becomes mergeable.
> 
> If I merge that, then you'll need to rebase this PR and move the decode tables out of MachRegister.C and into dataflowAPI/rose/registers. If we merge this PR first, then I can take care of that code motion. Either way works for me. I just don't want to add work for you unnecessarily.

Okay. I don't think moving the decoder table is a difficult task, so maybe #1845 can be merged first and I'll handle the decoder table on my own.

wxrdnx:
@kupsch I've fixed the issue you've mentioned in today's meeting.

common/src/registers/MachRegister.C
```diff
@@ -769,6 +790,9 @@ namespace Dyninst {
           return true;
       }
     }
+    if(arch == Arch_riscv64) {
```
hainest, line 793: We don't need this check since `getZeroFlag` will return `InvalidReg`. This puts the documentation for riscv in one spot (i.e., in `getZeroFlag`).

common/src/registers/MachRegister.C
```diff
@@ -864,6 +891,12 @@ namespace Dyninst {
         return is_vec || is_fpr || is_ctl || is_sts;
       }
 
+      case Arch_riscv64: {
+        auto const is_vec = isVector();
+        auto const is_fpr = (category == riscv64::FPR);
+        return is_vec || is_fpr;
```
hainest, line 897: `isFloatingPoint` includes all control/status registers modified or used by FP instructions.

```suggestion
        auto const is_fcsr = (*this == riscv64::fcsr);
        return is_vec || is_fpr || is_fcsr;
```

common/src/registers/MachRegister.C
```diff
@@ -947,6 +980,10 @@ namespace Dyninst {
         return false;
       }
 
+      case Arch_riscv64: {
```
hainest, line 983: If Capstone supports them, we should probably include the **un**privileged registers at some point. We don't have to do that now.

common/src/registers/MachRegister.C
```diff
@@ -1058,6 +1095,9 @@ namespace Dyninst {
                category == aarch64::SVE2 ||
                category == aarch64::SME;
 
+      case Arch_riscv64:
+        return false; // vector currently not supported
```
hainest, line 1099: Are they in Capstone?

instructionAPI/src/Instruction.C
```diff
@@ -92,6 +92,8 @@ namespace Dyninst { namespace InstructionAPI {
           return id != amdgpu_gfx90a_op_INVALID;
         case Arch_amdgpu_gfx940:
           return id != amdgpu_gfx940_op_INVALID;
+        case Arch_riscv64:
```
hainest, line 95: Could you add a blank line above this to visually separate it from the amdgpu ones?

hainest (changes requested):
Please add the unit tests from https://github.com/dyninst/unit-tests/pull/14.

For the DWARF registers, we should probably think about aliased registers at some point. I should do the same for arm. This is pretty much never used, so it's not a priority.

common/src/registers/MachRegister.C
```diff
@@ -165,6 +165,12 @@ namespace Dyninst {
 
         return *this;
       }
+      case Arch_riscv64:
+         switch(category) {
+             case riscv64::GPR: return riscv64::x0;
```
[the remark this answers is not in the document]
wxrdnx: I thought there were only 32 general-purpose registers (x0~x31) in RISC-V and all their basic registers are mapped to x0?

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,255 @@
...
+   *
+   *  These values/layout are not guaranteed to remain the same as part of the
+   *  public interface, and may change.
+   **/
+
+  // 0xff000000  0x00ff0000      0x0000ff00      0x000000ff
+  // arch        reg cat:GPR     alias&subrange  reg ID
+  const int32_t GPR = 0x00010000; // general purpose registers
+  const int32_t FPR = 0x00020000; // floating point registers
+  //const int32_t VTR = 0x00040000; // vector registers
```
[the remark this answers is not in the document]
wxrdnx: Capstone currently does not support vector registers

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
hainest: x0 is hardwired to contain only zeros, so I don't think it could serve as even a logical base register for x1-31. There is a lot of chatter in the [manual](https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-25d7f16-2025-02-11/riscv-unprivileged.pdf) in `2.1. Programmers' Model for Base Integer ISA`, but my primary influence was

```text
For these reasons, we chose a conventional size of 32 integer registers for RV32I.
Dynamic register usage tends to be dominated by a few frequently accessed
registers, and regfile implementations can be optimized to reduce access energy for
the frequently accessed registers (Tseng & Asanović, 2000).
```

Even if the hardware uses register renaming on some smaller register file, the integer registers (GPRs) should not alias each other. Thoughts?

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: From my reading, it looks like the Q extension needs D and Zhf needs only F, but neither needs the vector extension. Are the Capstone folks treating Q and Zhf as vector extensions?

common/src/registers/MachRegister.C
```diff
@@ -165,6 +165,12 @@ namespace Dyninst {
...
+             case riscv64::FPR: return riscv64::f0;
```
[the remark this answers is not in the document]
wxrdnx: Capstone distinguishes between `f<N>_32` and `f<N>_64`. Dwarf, on the other hand, does not distinguish between the two. In fact, in riscv64, `f<N>_32` is Nan-boxed to `f<N>_64`.

> When multiple floating-point precisions are supported, then valid values of narrower n-bit types,
n<FLEN, are represented in the lower n bits of an FLEN-bit NaN value, in a process termed NaN-
boxing. The upper bits of a valid NaN-boxed value must be all 1s. Valid NaN-boxed n-bit values
therefore appear as negative quiet NaNs (qNaNs) when viewed as any wider m-bit value, n < m ≤ FLEN.
Any operation that writes a narrower result to an 'f' register must write all 1s to the uppermost FLEN-n
bits to yield a legal NaN-boxedvalue

So when accessing f0_32, you are actually accessing the entire register, not only just half of it. Therefore, I introduced `f<N>` since there is no need to distinguish between `f<N>_32` and `f<N>_64`.

f<N>_32` and `f<N>_64` exist only because Capstone differentiate the two. Personally, I prefer using `f<N>`  because it feels more succinct compared to `f<N>_64`. What do you think?

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
wxrdnx: It is true that Q depends on D and Zhf depends on F but not on D.

Capstone folks only supports I (Integer), M (Multiplication), A (Atomic), F (Floating Points), D (Double Floating Points), and C (Compressed) extensions. Q extension, Zhf extension, and vector extensions are all not supported.

common/src/registers/MachRegister.C
wxrdnx: No

common/src/registers/MachRegister.C
wxrdnx: Unfortunately Capstone does not support them

instructionAPI/src/Instruction.C
wxrdnx: Fixed

common/src/registers/MachRegister.C
wxrdnx: Fixed

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,255 @@
...
+  DEF_REGISTER(        sp,   2 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        gp,   3 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        tp,   4 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t0,   5 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t1,   6 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t2,   7 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        fp,   8 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        s0,   8 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        s1,   9 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        a0,  10 | GPR | Arch_riscv64, "riscv64");
```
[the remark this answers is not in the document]
wxrdnx: Ok. Should I add aliases for `f<N>` registers only or all `f<N>`, `f<N>_32`, and `f<N>_64`? (See [#discussion_r1955600709](discussion_r1955600709))

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
wxrdnx: Interesting. My view point is that `baseRegister(x<N>) = x<0>` because in terms of register opcode encoding, x<N> is indeed encoded to N, so naturally the base register of x<N> can be thought of as x0, kind of like isomorphism. Personally I think Register File is a low-level, hardware implementation stuff. It has little to do with this matter.

common/src/registers/MachRegister.C
wxrdnx: Fixed

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: Thanks for the details.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
hainest: Oooh, sorry. This has been a complete miscommunication on my part. A base register is a register that contains some other named register. The main use for this is to know if two instructions write to the same location.

On x86,
```asm
mov  al, 0x1
mov rax, 0x2
```

Here the two registers have different names, but the value is written to the same location such that reading `al` after the second instruction would return `0x2` because `ah` refers to bits 0 to 7 of `rax`. In this case, we need to make sure `ah.getBaseRegister() == rax.getBaseRegister()`.

Contrastingly on RISC-V,
```asm
addi x1, x0, 0x1
addi x2, x0, 0x2
```

A read of `x1` after the second instruction would still return 0x1, so we need to make sure `x1.getBaseRegister() != x2.getBaseRegister()`.

Does that help clarify? :smiley: 

---

This is an aside, but something that caught my attention. I just noticed that the 32-bit RV32I registers seem to alias the 64-bit registers, even though they have identical names. For example,

```text
LUI (load upper immediate) uses the same opcode as RV32I. LUI
places the 32-bit U-immediate into register rd, filling in the lowest
12 bits with zeros. The 32-bit result is sign-extended to 64 bits.
```

This seems to suggest that `x1.getSize() == 4` for RVI32, but is 8 for RVI64 even though they are the same register name. I'm assuming this is part of the portability goal to have one binary run on both RVI32 and RVI64. I don't think this would affect our calculation of the base register, but might cause Dyninst to run into issues if it checks the size of the register instead of checking the base register. This is akin to ppc32 vs ppc64, but we do a terrible job of separating those two architectures. Maybe we don't need to worry about this right now.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
hainest: > Capstone distinguishes between `f<N>_32` and `f<N>_64`. Dwarf, on the other hand, does not distinguish between the two. In fact, in riscv64, `f<N>_32` is Nan-boxed to `f<N>_64`.
> 
> > When multiple floating-point precisions are supported, then valid values of narrower n-bit types,
> > n<FLEN, are represented in the lower n bits of an FLEN-bit NaN value, in a process termed NaN-
> > boxing. The upper bits of a valid NaN-boxed value must be all 1s. Valid NaN-boxed n-bit values
> > therefore appear as negative quiet NaNs (qNaNs) when viewed as any wider m-bit value, n < m ≤ FLEN.
> > Any operation that writes a narrower result to an 'f' register must write all 1s to the uppermost FLEN-n
> > bits to yield a legal NaN-boxedvalue

I have no idea how I missed that. Though I have to say they really twisted themselves into knots to minify FP support here. I've never seen anything like this before.

> So when accessing f0_32, you are actually accessing the entire register, not only just half of it. Therefore, I introduced `f<N>` since there is no need to distinguish between `f<N>_32` and `f<N>_64`.

Agreed. We should also explicitly make `f<N>_32` and `f<N>_64` alias `f<N>` in `riscv_regs.h`. Then we can just check `f<N>` here.

> `f<N>_32` and `f<N>_64` exist only because Capstone differentiate the two. Personally, I prefer using `f<N>` because it feels more succinct compared to `f<N>_64`. What do you think?

I think `f<N>` is a good approach. I would add some comments as to why we're doing it that way.

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: If we alias `f<N>_32` and `f<N>_64` to `f<N>`, then we can just have the ABI registers alias `f<N>`.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
wxrdnx: Oh sorry I misunderstood the meaning of base registers. If that's the case, I'll just simply map all GPRs and FPRs to itself

```c
case Arch_riscv64:
   return *this;
```

is that ok?
---
Since I'm focusing on RI64 right now, I don't worry too much about stuff related to RI32.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
wxrdnx: The comments in `riscv64_regs.h`, which I wrote in the past before I knew about the Nan-boxed thing, were indeed confusing. I updated the comments in that file.

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,255 @@
...
+  DEF_REGISTER(       f26,  26 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       f27,  27 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       f28,  28 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       f29,  29 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       f30,  30 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       f31,  31 | FPR | Arch_riscv64, "riscv64");
+
+  // Aliases (ABI Name)
+  DEF_REGISTER(      zero,   0 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        ra,   1 | GPR | Arch_riscv64, "riscv64");
```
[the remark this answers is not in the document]
wxrdnx: AFAIK the "alternate link register" is used for implementing millicode routines, particularly prologue/epilogue routine mentioned in [Waterman's thesis](https://people.eecs.berkeley.edu/~krste/papers/EECS-2016-1.pdf) on page 66 ~ 67.
It seems that the intention of function prologue/epilogue routines mentioned in that paper is to reduce the overall code size by implementing the function prologues/epilogues as a function subroutine itself. However, since we are dealing with `x1`(`ra`) in the function prologue/epilogue, we use `x5`(`t0`) instead of `x1`(`ra`) to call the function prologue/epilogue function routine in order to avoid register conflicts.
This is very interesting, and if I ever find myself seeing it in the wild, I'll definitely take notes of it. Regardless, based on the examples he gave in his paper, I don't think we should make `ra` an alias for `x5`.

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
wxrdnx: Updated

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
hainest: This should be fine once we have the aliases in place.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
hainest: After a bit more consideration, I think we can just eliminate the 64/32 suffixed versions and just directly map both Capstone representations to the `f<N>` inside the instruction decoder. That would save us some cognitive load without losing any information. Thoughts?

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: Let me read over this some more. The more I get to know RISCV, the less I understand why people like it so much. I've never seen anything add so much complexity to save a few transistors.

common/src/registers/MachRegister.C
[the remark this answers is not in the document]
wxrdnx: Fixed

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: At the time of Waterman's thesis, there didn't seem to be a convention for calling millicode routines. I think x5 has become that convention, but I agree we should probably leave that out for now. We should include `ra` as an alias for `x1` and `sp` as an alias for `x2`.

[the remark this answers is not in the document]
wxrdnx: Okay.
But my question is: If I change `riscv64::ra` aliasa to `riscv64::x5`, wouldn't that make `riscv64::ra` ambiguous?

tests/MachRegister/base_registers/riscv64.cpp
```diff
@@ -0,0 +1,75 @@
...
+  BASEREG_CHECK(riscv64::f23, riscv64::f23);
+  BASEREG_CHECK(riscv64::f24, riscv64::f24);
+  BASEREG_CHECK(riscv64::f25, riscv64::f25);
+  BASEREG_CHECK(riscv64::f26, riscv64::f26);
+  BASEREG_CHECK(riscv64::f27, riscv64::f27);
+  BASEREG_CHECK(riscv64::f28, riscv64::f28);
+  BASEREG_CHECK(riscv64::f29, riscv64::f29);
+  BASEREG_CHECK(riscv64::f30, riscv64::f30);
+  BASEREG_CHECK(riscv64::f31, riscv64::f31);
+
```
wxrdnx: @hainest 
Is the unit test okay?
Do I need to add/delete anything?

wxrdnx:
Hi @kupsch and @hainest:

I think this PR is ready to review.
The only testers that does not pass are the Libabigail ones. This is because

1. New RISC-V mnemonics are added
3. Code that parses RISC-V specific sections are also added to the objectELF class.

wxrdnx:
Hi @kupsch and @hainest:

I've removed the RISC-V unit test in this PR, and I'm going to move it to another separate PR.

hainest:
> Hi @kupsch and @hainest:
> 
> I've removed the RISC-V unit test in this PR, and I'm going to move it to another separate PR.

I'd rather have the unit tests here so we can run them during the CI for this PR.

To keep this PR more constrained, I would prefer the ELF parsing be moved to a separate PR. Naturally, it will be dependent on this one.

wxrdnx:
@hainest 

I've moved the ELF parsing to a separate PR #1966
I've also added unit tests in this PR

hainest:
> @hainest
> 
> I've moved the ELF parsing to a separate PR #1966 I've also added unit tests in this PR

Thanks for making the suggested updates. 

> The only testers that does not pass are the Libabigail ones. This is because

If you add the ABI-BREAKER label, the libabigail tests won't run.

---

On first glance, this seems to be good. I'll take a deeper look in the next few days. If Jim approves them first, let's merge. I don't want to delay your progress any further.

common/h/registers/riscv64_regs.h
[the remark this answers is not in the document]
hainest: Aliases need to be declared with [`DEF_REGISTER_ALIAS`](https://github.com/dyninst/dyninst/blob/master/common/h/registers/reg_def.h#L59) so that `riscv64::ra == riscv64::x5` is always true.

common/src/registers/MachRegister.C
```diff
@@ -579,6 +588,7 @@ namespace Dyninst {
       case Arch_ppc64: return ppc64::r0;
       case Arch_aarch64: return aarch64::x8;
       case Arch_aarch32:
+      case Arch_riscv64: return riscv64::a7;
```
hainest, line 597: I didn't see this in The ISA manual (20240411). Did it come from the linux kernel source? If so, could we add a link for reference?

instructionAPI/src/syscalls.C
```diff
@@ -179,6 +179,15 @@ namespace aarch64 {
   }
 }
 
+namespace riscv64 {
+  bool isSystemCall(di::Instruction const& ins) {
+    auto const id = ins.getOperation().getID();
+
+    return id == riscv64_op_ecall;
```
hainest, line 185: What about ebreak?

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,264 @@
...
+  DEF_REGISTER(        s9,  25 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       s10,  26 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(       s11,  27 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t3,  28 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t4,  29 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t5,  30 | GPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(        t6,  31 | GPR | Arch_riscv64, "riscv64");
+
+  // 32 bit FPRs f<N>_32 (Mainly for Capstone compatibility)
+  DEF_REGISTER(     f0_32,   0 | FPR | Arch_riscv64, "riscv64");
```
hainest: `f<N>_32` and `f<N>_64` should be declared with DEF_REGISTER_ALIAS to make them explicit aliases.

tests/unit/MachRegister/base_registers/riscv64.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+int main() {
+
+  // GPR 64 -> 64
+  BASEREG_CHECK(riscv64::x0, riscv64::x0);
+  BASEREG_CHECK(riscv64::x8, riscv64::x8);
+  BASEREG_CHECK(riscv64::x31, riscv64::x31);
+
+  // FPR 64 Boxed -> 64 Boxed
+  BASEREG_CHECK(riscv64::f0, riscv64::f0);
```
hainest, line 12: These are actually alias checks (see my comments for riscv64_regs.h).

wxrdnx:
> Have you encountered any serious parsing issues due to the absence of ROSE data? It looks like even the most recent version of ROSE still doesn't have any riscv64 stuff.

ROSE does not support RISC-V unfortunately, so the RISC-V ROSE files were written by me, similar to the AMDGPU ROSE files. Since RISC-V is not supported by ROSE, we're using [SAIL](https://github.com/riscv/sail-riscv) as our instruction semantics. However, SAIL is written in Ocaml, meaning that it is extremely difficult for Dyninst to use the information directly. Thus, I've written scripts that convert SAIL to ROSE C++ class, and those scripts will come later.

> I'm a bit confused. This adds ROSE registers in external/rose/riscv64InstructionEnum.h, but the ExpressionConversionVisitor doesn't implement any expansion for them. Will that come later?

Yes, they will come later. Do you think they should also go in this PR?

> Will instruction [categories](https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/Instruction.C#L428) be added in the instruction decoding PR?

Yes, they will in the future

> It looks like the ELF stuff was extracted into [1966](https://github.com/dyninst/dyninst/pull/1966). Should this be removed here?

Sorry, I'm not sure what "this" means here. Do you want me to move all files in `symtabAPI` to #1966 ?

common/src/registers/MachRegister.C
wxrdnx: Yes, from the [linux kernel source](https://elixir.bootlin.com/linux/v6.15.7/source/arch/riscv/include/asm/syscall.h#L27-L31).

instructionAPI/src/syscalls.C
wxrdnx: `ebreak` is the equivalent of `int3` in x64. It is not used to perform syscall.

common/h/registers/riscv64_regs.h
wxrdnx: Got it, now they're declared using `DEF_REGISTER_ALIAS`.

tests/unit/MachRegister/base_registers/riscv64.cpp
wxrdnx: Got it, Now they're changed to alias checks.

hainest:
> > Have you encountered any serious parsing issues due to the absence of ROSE data? It looks like even the most recent version of ROSE still doesn't have any riscv64 stuff.
> 
> ROSE does not support RISC-V unfortunately, so the RISC-V ROSE files were written by me, similar to the AMDGPU ROSE files. Since RISC-V is not supported by ROSE, we're using [SAIL](https://github.com/riscv/sail-riscv) as our instruction semantics. However, SAIL is written in Ocaml, meaning that it is extremely difficult for Dyninst to use the information directly. Thus, I've written scripts that convert SAIL to ROSE C++ class, and those scripts will come later.

What a pain. It's good you were able to get it working. 

> > I'm a bit confused. This adds ROSE registers in external/rose/riscv64InstructionEnum.h, but the ExpressionConversionVisitor doesn't implement any expansion for them. Will that come later?
> 
> Yes, they will come later. Do you think they should also go in this PR?

No, they should go in a separate PR.
 
> > Will instruction [categories](https://github.com/dyninst/dyninst/blob/master/instructionAPI/src/Instruction.C#L428) be added in the instruction decoding PR?
> 
> Yes, they will in the future

:+1: 

> > It looks like the ELF stuff was extracted into [1966](https://github.com/dyninst/dyninst/pull/1966). Should this be removed here?
> 
> Sorry, I'm not sure what "this" means here. Do you want me to move all files in `symtabAPI` to #1966 ?

Yeah, let's move the symtab stuff into 1966. That way it's all in one place to make it easier to review.

wxrdnx:
> Yeah, let's move the symtab stuff into 1966. That way it's all in one place to make it easier to review.

@hainest Done!

common/src/registers/MachRegister.C
hainest: Did you have a chance to check on this one?

instructionAPI/src/syscalls.C
hainest: Did you have a chance to check on this one?

The RISC-V Instruction Set Manual Volume I
Version 20240411
2.8. Environment Call and Breakpoints

wxrdnx: Checked

> EBREAK was primarily designed to be used by a debugger to cause execution to stop and
fall back into the debugger. EBREAK is also used by the standard gcc compiler to mark
code paths that should not be executed.

So `ebreak` isn't used for syscalls. It is used to fallback into the debugger.

common/src/registers/MachRegister.C
wxrdnx: @hainest  Unfortunately I just figured out that the syscall calling convention is not defined in the doc. (See [this discussion](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/13))

TLDR: the syscall nr is OS-specific, and Linux chooses `a7` ~while FreeBSD chooses `t0`~.

FreeBSD also use `a7`. See [the FreeBSD Manual](https://man.freebsd.org/cgi/man.cgi?apropos=0&manpath=Debian+13.0.0&query=syscall&sektion=2), scroll down to the `Architecture calling conventions` section

hainest:
@wxrdnx Have you had a chance to take a look at the two items I mentioned? [1](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219583399), [2](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219660929). No rush. With just two small items to go, we could get this merged and let you move on to more important things.

wxrdnx:
> @wxrdnx Have you had a chance to take a look at the two items I mentioned? [1](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219583399), [2](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219660929). No rush. With just two small items to go, we could get this merged and let you move on to more important things.

Hi @hainest :

For [1](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219583399), the syscall calling convention is nowhere defined in the ISA manual. [This discussion](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/13) says that the syscall nr is OS-specific. For Linux, the syscall nr is defined in the [source code](https://elixir.bootlin.com/linux/v6.15.7/source/arch/riscv/include/asm/syscall.h#L27-L31), which is `a7`. I'm not sure if Dyninst runs on FreeBSD, but for FreeBSD, the syscall nr can be seen from [the FreeBSD Manual](https://man.freebsd.org/cgi/man.cgi?apropos=0&manpath=Debian+13.0.0&query=syscall&sektion=2) (scroll down to the `Architecture calling conventions` section). It is also `a7`.

For [2](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219660929), `ecall` is the RISC-V equivalent of `int 0x80` in x86 or `syscall` in x86_64, while `ebreak` is the RISC-V equivalent of `int3` in x86. `ebreak` is used to generate a breakpoint exception, not to perform syscall. Like `int3`, `ebreak` triggers a user-mode trap, transferring the control to a trap handler, so `ebreak` does not go through the kernel in the same way `ecall` does. Thus, I don't think `ebreak` qualifies as a syscall. 

Source: section 2.8 of the [RISC-V Instruction Set Manual Volume I: Unprivileged ISA](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view):

> EBREAK was primarily designed to be used by a debugger to cause execution to stop and fall back into the debugger. EBREAK is also used by the standard gcc compiler to mark code paths that should not be executed.

hainest:
> > @wxrdnx Have you had a chance to take a look at the two items I mentioned? [1](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219583399), [2](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219660929). No rush. With just two small items to go, we could get this merged and let you move on to more important things.
> 
> Hi @hainest :
> 
> For [1](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219583399), the syscall calling convention is nowhere defined in the ISA manual. [This discussion](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/13) says that the syscall nr is OS-specific. For Linux, the syscall nr is defined in the [source code](https://elixir.bootlin.com/linux/v6.15.7/source/arch/riscv/include/asm/syscall.h#L27-L31), which is `a7`.

Could you add a link to the linux kernel source with a comment that a7 isn't uniform across all OSes? That sounds like a real pain that I hope we never have to deal with.

> I'm not sure if Dyninst runs on FreeBSD, but for FreeBSD, the syscall nr can be seen from [the FreeBSD Manual](https://man.freebsd.org/cgi/man.cgi?apropos=0&manpath=Debian+13.0.0&query=syscall&sektion=2) (scroll down to the `Architecture calling conventions` section). It is also `a7`.

I would be surprised if the FreeBSD-specific parts of Dyninst have been compiled in the last decade, so I'd say 'no'.

> 
> For [2](https://github.com/dyninst/dyninst/pull/1831#discussion_r2219660929), `ecall` is the RISC-V equivalent of `int 0x80` in x86 or `syscall` in x86_64, while `ebreak` is the RISC-V equivalent of `int3` in x86. `ebreak` is used to generate a breakpoint exception, not to perform syscall. Like `int3`, `ebreak` triggers a user-mode trap, transferring the control to a trap handler, so `ebreak` does not go through the kernel in the same way `ecall` does. Thus, I don't think `ebreak` qualifies as a syscall.
> 
> Source: section 2.8 of the [RISC-V Instruction Set Manual Volume I: Unprivileged ISA](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view):
> 
> > EBREAK was primarily designed to be used by a debugger to cause execution to stop and fall back into the debugger. EBREAK is also used by the standard gcc compiler to mark code paths that should not be executed.

Derp. I have no idea how I missed that in the docs. Since it is technically a software interrupt, let's add it to [isSoftwareInterrupt](https://github.com/dyninst/dyninst/blob/v13.0.0/instructionAPI/src/interrupts.C).

wxrdnx:
Hi @hainest :

> Could you add a link to the linux kernel source with a comment that a7 isn't uniform across all OSes? That sounds like a real pain that I hope we never have to deal with.

Done.

> Derp. I have no idea how I missed that in the docs. Since it is technically a software interrupt, let's add it to [isSoftwareInterrupt](https://github.com/dyninst/dyninst/blob/v13.0.0/instructionAPI/src/interrupts.C).

Done.

hainest (approved):
LGTM. Thanks so much for putting up with all of the questions and edits.

@kupsch Any thoughts?

wxrdnx:
@hainest @kupsch 

I did not know that if I comment directly instead of quote reply on "Files Changed", the comments are in pending state.
I need to click "Review Changes" -> "Submit Review" to submit the review.
No wonder you can't see my comments months ago.
Guess I've learned something every day.

tests/MachRegister/base_registers/riscv64.cpp
hainest: These look fine. I don't think it's necessary to test every register.

hainest:
> @hainest @kupsch
> 
> I did not know that if I comment directly instead of quote reply on "Files Changed", the comments are in pending state. I need to click "Review Changes" -> "Submit Review" to submit the review. No wonder you can't see my comments months ago. Guess I've learned something every day.

Ah, I was curious about that. You have to resubmit every time. It's the price that is paid so you don't have to do the entire review all at once.

common/src/registers/MachRegister.C
```diff
@@ -578,6 +588,11 @@ namespace Dyninst {
       case Arch_ppc32: return ppc32::r0;
       case Arch_ppc64: return ppc64::r0;
       case Arch_aarch64: return aarch64::x8;
+      // Note that a7 isn't uniform across all OSes
+      // The syscall calling convention for RISC-V is not defined in the ISA manual
+      // For Linux, the syscall NR is defined in the source code, which is a7:
+      // https://elixir.bootlin.com/linux/v6.15.7/source/arch/riscv/include/asm/syscall.h#L27-L31
```
kupsch: You can remove this comment and add the following to the top of the function

```
// The syscall register for all platforms can be found in the 
// linux syscall(2) man page.
```

external/rose/riscv64InstructionEnum.h
```diff
@@ -0,0 +1,367 @@
+//
+// Created by angus on 07/25/24.
+//
+
```
kupsch: Delete this as git records who added the code and when.

```suggestion
```

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,264 @@
...
+  DEF_REGISTER(       frm,   2 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(      fcsr,   3 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(    vstart,   4 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(     vxsat,   5 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(      vxrm,   6 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(      vcsr,   7 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(      seed,   8 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(     cycle,   9 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(      time,  10 | SPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(   instret,  11 | SPR | Arch_riscv64, "riscv64");
```
kupsch: You should move these 3 after `fcsr` as these are all `CSR addresses`. Perhaps these 3 and `fflags`, `frm`, `fcsr` should be there own `CSR` class?  @hainest what do you think?

common/src/registers/MachRegister.C
```diff
@@ -165,6 +165,8 @@ namespace Dyninst {
 
         return *this;
       }
+      case Arch_riscv64:
+        return *this;
```
kupsch, line 172: The base of `fflags` and `frm` should be `fcsr`, as `fcsr` contains these other two registers.

common/src/registers/MachRegister.C
```diff
@@ -634,6 +652,7 @@ namespace Dyninst {
       case Arch_ppc64: return ppc64::cr0e;
       case Arch_aarch64: return aarch64::z;
       case Arch_aarch32:
+      case Arch_riscv64: // RISC-V does not have flag registers
```
kupsch: for consistency remove the comment or add a blank line before `Arch_aarch32` and the comment `// These Architecture do not have a zero flag:`

common/src/registers/MachRegister.C
```diff
@@ -701,6 +720,7 @@ namespace Dyninst {
         return regC == aarch64::FLAG ||
                regC == aarch64::PSTATE;
       }
+      case Arch_riscv64: return false; // RISC-V does not have flag registers
```
kupsch: `fcsr` and `fflags` are flags.

common/src/registers/MachRegister.C
```diff
@@ -849,6 +872,13 @@ namespace Dyninst {
         return is_vec || is_fpr || is_ctl || is_sts;
       }
 
+      case Arch_riscv64: {
+        auto const is_vec = isVector();
+        auto const is_fpr = (category == riscv64::FPR);
+        auto const is_fcsr = (*this == riscv64::fcsr);
```
kupsch: Also `fflags` and `frm`?

common/src/registers/MachRegister.C
```diff
@@ -932,6 +962,10 @@ namespace Dyninst {
         return false;
       }
 
+      case Arch_riscv64: {
+        return false; // csrs currently not supported
```
kupsch: Since you define the register you should add them here.  All the `SPR`s except `PC`?

kupsch (commented):
Several comments that should be adjusted and then changes around SPRs.

common/h/registers/riscv64_regs.h
hainest: I don't have any strong opinions. If adding a `CSR` class would simplify or enrich the `MachRegister` properties, then it would be a good idea.

common/src/registers/MachRegister.C
wxrdnx: Done

common/src/registers/MachRegister.C
wxrdnx: Done

common/h/registers/riscv64_regs.h
wxrdnx: I agree they should be their own class.
I think I should also add the performance-monitoring registers (hpmcounterNN), as they are also unprivileged CSRs.

wxrdnx: Update: I added all unprivileged CSRs mentioned in The RISC-V Instruction Set Manual: Volume II Privileged Architecture, Section 2.2, regardless of whether Capstone supports it or not. (except for those end in "h", which are only used in RV32).

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,264 @@
...
+   *  public interface, and may change.
+   **/
+
+  // 0xff000000  0x00ff0000      0x0000ff00      0x000000ff
+  // arch        reg cat:GPR     alias&subrange  reg ID
+  const int32_t GPR = 0x00010000; // general purpose registers
+  const int32_t FPR = 0x00020000; // floating point registers
+  //const int32_t VTR = 0x00040000; // vector registers
+  const int32_t SPR = 0x00080000; // special purpose registers
+
```
wxrdnx, line 60: Add CSR register group
```suggestion
const int32_t CSR = 0x00100000; // control and status registers
```

common/h/registers/riscv64_regs.h
wxrdnx: Added all unprivileged CSRs mentioned in The RISC-V Instruction Set Manual: Volume II Privileged Architecture, Section 2.2:
```suggestion
  DEF_REGISTER(           pc,   0 | SPR | Arch_riscv64, "riscv64");

  // From The RISC-V Instruction Set Manual: Volume II Privileged Architecture, Section 2.2
  DEF_REGISTER(       fflags,   0 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(          frm,   1 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(         fcsr,   2 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(       vstart,   3 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(        vxsat,   4 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(         vxrm,   5 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(         vcsr,   6 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(           vl,   7 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(        vtype,   8 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(        vlenb,   9 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(          ssp,  10 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(         seed,  11 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(          jvt,  12 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(        cycle,  13 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(         time,  14 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(      instret,  15 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter3,  16 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter4,  17 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter5,  18 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter6,  19 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter7,  20 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter8,  21 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(  hpmcounter9,  22 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter10,  23 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter11,  24 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter12,  25 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter13,  26 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter14,  27 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter15,  28 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter16,  29 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter17,  30 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter18,  31 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter19,  32 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter20,  33 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter21,  34 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter22,  35 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter23,  36 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter24,  37 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter25,  38 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter26,  39 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter27,  40 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter28,  41 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter29,  42 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter30,  43 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter31,  44 | CSR | Arch_riscv64, "riscv64");
  // RV32 only
  DEF_REGISTER(       cycleh,  45 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(        timeh,  46 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(     instreth,  47 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter3h,  48 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter4h,  49 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter5h,  50 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter6h,  51 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter7h,  52 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter8h,  53 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER( hpmcounter9h,  54 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter10h,  55 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter11h,  56 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter12h,  57 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter13h,  58 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter14h,  59 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter15h,  60 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter16h,  61 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter17h,  62 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter18h,  63 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter19h,  64 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter20h,  65 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter21h,  66 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter22h,  67 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter23h,  68 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter24h,  69 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter25h,  70 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter26h,  71 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter27h,  72 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter28h,  73 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter29h,  74 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter30h,  75 | CSR | Arch_riscv64, "riscv64");
  DEF_REGISTER(hpmcounter31h,  76 | CSR | Arch_riscv64, "riscv64");
```

common/src/registers/MachRegister.C
wxrdnx: Done

wxrdnx:
```suggestion
      case Arch_riscv64: {
        if(*this == riscv64::fflags || *this == riscv64::frm) {
          return riscv64::fcsr;
        }
        return *this;
      }
```

common/src/registers/MachRegister.C
wxrdnx: Since `fcsr` includes `fflags`, should I also treat `fcsr` as a flag register?
```suggestion
      case Arch_riscv64: return regC == riscv64::fflags || regC == riscv64::fcsr;
```

common/src/registers/MachRegister.C
wxrdnx: Makes sense.
```suggestion
        auto const is_fcsr = (*this == riscv64::fflags) || (*this == riscv64::frm) || (*this == riscv64::fcsr);
```

common/src/registers/MachRegister.C
kupsch: sure

common/src/registers/MachRegister.C
wxrdnx: I've moved all CSRs to its own category, so comparing the category should suffice.
```suggestion
        return category == riscv64::CSR;
```

wxrdnx (commented):
I've moved the CSRs to its own category.
Just one question: Some CSR registers that end in "h" (e.g. `cycleh`, `timeh`) are for RV32 only. Should I also include them?

kupsch:
I don't think that it would hurt to add them.  You should add a comment that they are RV32 only.

wxrdnx:
> I don't think that it would hurt to add them. You should add a comment that they are RV32 only.

Done. Can you review the suggestions?
Does the code look good to you?

wxrdnx:
@kupsch 
I have sorted CSR registers according to their value instead of their appearance in which they appear in the ISA manual.
I have also added CSR register values as comments.

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,339 @@
...
+  DEF_REGISTER_ALIAS(    s7, x23, "riscv64");
+  DEF_REGISTER_ALIAS(    s8, x24, "riscv64");
+  DEF_REGISTER_ALIAS(    s9, x25, "riscv64");
+  DEF_REGISTER_ALIAS(   s10, x26, "riscv64");
+  DEF_REGISTER_ALIAS(   s11, x27, "riscv64");
+  DEF_REGISTER_ALIAS(    t3, x28, "riscv64");
+  DEF_REGISTER_ALIAS(    t4, x29, "riscv64");
+  DEF_REGISTER_ALIAS(    t5, x30, "riscv64");
+  DEF_REGISTER_ALIAS(    t6, x31, "riscv64");
+
```
kupsch, line 278: Add the floating point register aliases also:

```suggestion

  // Floating point register aliases
  DEF_REGISTER_ALIAS(   ft0,  f0, "riscv64");
  DEF_REGISTER_ALIAS(   ft1,  f1, "riscv64");
  DEF_REGISTER_ALIAS(   ft2,  f2, "riscv64");
  DEF_REGISTER_ALIAS(   ft3,  f3, "riscv64");
  DEF_REGISTER_ALIAS(   ft4,  f4, "riscv64");
  DEF_REGISTER_ALIAS(   ft5,  f5, "riscv64");
  DEF_REGISTER_ALIAS(   ft6,  f6, "riscv64");
  DEF_REGISTER_ALIAS(   ft7,  f7, "riscv64");
  DEF_REGISTER_ALIAS(   fs0,  f8, "riscv64");
  DEF_REGISTER_ALIAS(   fs1,  f9, "riscv64");
  DEF_REGISTER_ALIAS(   fa0, f10, "riscv64");
  DEF_REGISTER_ALIAS(   fa1, f11, "riscv64");
  DEF_REGISTER_ALIAS(   fa2, f12, "riscv64");
  DEF_REGISTER_ALIAS(   fa3, f13, "riscv64");
  DEF_REGISTER_ALIAS(   fa4, f14, "riscv64");
  DEF_REGISTER_ALIAS(   fa5, f15, "riscv64");
  DEF_REGISTER_ALIAS(   fa6, f16, "riscv64");
  DEF_REGISTER_ALIAS(   fa7, f17, "riscv64");
  DEF_REGISTER_ALIAS(   fs2, f18, "riscv64");
  DEF_REGISTER_ALIAS(   fs3, f19, "riscv64");
  DEF_REGISTER_ALIAS(   fs4, f20, "riscv64");
  DEF_REGISTER_ALIAS(   fs5, f21, "riscv64");
  DEF_REGISTER_ALIAS(   fs6, f22, "riscv64");
  DEF_REGISTER_ALIAS(   fs7, f23, "riscv64");
  DEF_REGISTER_ALIAS(   fs8, f24, "riscv64");
  DEF_REGISTER_ALIAS(   fs9, f25, "riscv64");
  DEF_REGISTER_ALIAS(  fs10, f26, "riscv64");
  DEF_REGISTER_ALIAS(  fs11, f27, "riscv64");
  DEF_REGISTER_ALIAS(   ft8, f28, "riscv64");
  DEF_REGISTER_ALIAS(   ft9, f29, "riscv64");
  DEF_REGISTER_ALIAS(  ft10, f30, "riscv64");
  DEF_REGISTER_ALIAS(  ft11, f31, "riscv64");
```

instructionAPI/capstone/riscv64/mnemonics.py
kupsch, line 1: Add the dyninst copyright header

instructionAPI/capstone/riscv64/registers.py
kupsch, line 1: Add the dyninst copyright header

common/src/registers/MachRegister.C
```diff
@@ -578,6 +592,9 @@ namespace Dyninst {
       case Arch_ppc32: return ppc32::r0;
       case Arch_ppc64: return ppc64::r0;
       case Arch_aarch64: return aarch64::x8;
+      // The syscall register for all platforms can be found in the 
+      // linux syscall(2) man page.
```
kupsch: move this comment to the function definition or at the beginning of the body (before the `switch`) as it applies to all architecture.

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,372 @@
...
+  DEF_REGISTER(           pc,   0 | SPR | Arch_riscv64, "riscv64");
+
+  // control status register
+  // From The RISC-V Instruction Set Manual: Volume II Privileged Architecture, Section 2.2
+
+  // Unprivileged Floating-Point CSRs
+  DEF_REGISTER(       fflags,   0 | CSR | Arch_riscv64, "riscv64"); // 0x001
+  DEF_REGISTER(          frm,   1 | CSR | Arch_riscv64, "riscv64"); // 0x002
+  DEF_REGISTER(         fcsr,   2 | CSR | Arch_riscv64, "riscv64"); // 0x003
+  // Unprivileged Vector CSRs, read/write
```
kupsch, line 156: Adding a blank line before comments that start a new section would make this more readable. Same thing applies to many other places in this file.

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,372 @@
...
+  DEF_REGISTER(          f28,  28 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(          f29,  29 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(          f30,  30 | FPR | Arch_riscv64, "riscv64");
+  DEF_REGISTER(          f31,  31 | FPR | Arch_riscv64, "riscv64");
+
+  // special purpose register
+  DEF_REGISTER(           pc,   0 | SPR | Arch_riscv64, "riscv64");
+
+  // control status register
+  // From The RISC-V Instruction Set Manual: Volume II Privileged Architecture, Section 2.2
```
kupsch: Add the manual version number as there are and will be many versions of this document, also a comment describing the comment
```suggestion
  // Control Status Registers (CSRs) - the comment after each is its CSR Id
  // From The RISC-V Instruction Set Manual: Volume II Privileged Architecture,
  // Version 20250805, Section 2.2
```

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,372 @@
...
+  // D extension widens f0 ~ f31, to 64 bits
+
+  // In riscv64, 32-bit FPRs are NaN-boxed to 64-bit FPRs.
+  // In other words, in RI64, when accessing 32-bit FPRs, you are actually accessing the entire register.
+
+  // Capstone distinguishes between 32-bit FPRs and 64-bit FPRs.
+  // So far, other libraries such as Dwarf does not distinguish between the two.
+  // So here f<N>_32 and f<N>_64 are used to handle Capstone registers
+  // On the other hand, f<N> are the "actual" FPRs that include both the NaN-boxed f<N>_32 and f<N>_64.
+  // f<N>_32 and f<N>_64 are both aliases of f<N>
```
kupsch: Try to keep lines that are just comments to 80 columns, and no need to have each sentence be on its own line here they are sentences in a paragraph, so format it as one.
```suggestion
  // In riscv64, 32-bit FPRs are NaN-boxed to 64-bit FPRs.  In other words, in
  // RI64, when accessing 32-bit FPRs, you are actually accessing the entire
  // register.

  // Capstone distinguishes between 32-bit FPRs and 64-bit FPRs.  So far, other
  // libraries such as Dwarf does not distinguish between the two.  So here
  // f<N>_32 and f<N>_64 are used to handle Capstone registers On the other
  // hand, f<N> are the "actual" FPRs that include both the NaN-boxed f<N>_32
  // and f<N>_64.  f<N>_32 and f<N>_64 are both aliases of f<N>
```

common/h/registers/riscv64_regs.h
```diff
@@ -0,0 +1,372 @@
...
+#include "Architecture.h"
+#include "registers/reg_def.h"
+
+namespace Dyninst { namespace riscv64 {
+
+  /**
+   * For interpreting constants:
+   *  Lowest 16 bits (0x000000ff) is base register ID
+   *  Next 16 bits (0x0000ff00) are used to distinguish between 32/64 bit floating point registers
+   *  Next 16 bits (0x00ff0000) are the register category, GPR/FPR/MMX/...
```
kupsch: Wrap the long line, and change x86 categories to to RISC-V:
```suggestion
   *  Next 16 bits (0x0000ff00) are used to distinguish between 32/64 bit
   *                            floating point registers
   *  Next 16 bits (0x00ff0000) are the register category, GPR/FPR/SPR/CSR
```

common/h/registers/riscv64_regs.h
wxrdnx: Done

common/src/registers/MachRegister.C
wxrdnx: Done

instructionAPI/capstone/riscv64/registers.py
wxrdnx: Done

instructionAPI/capstone/riscv64/mnemonics.py
wxrdnx: Done

kupsch (approved):

hainest (approved):

Thanks so much for all the effort on handling the many change requests!

PR #1832, opened 2024-12-15
Remove unused ROSE files from dataflowAPI

hainest:
Most of these were added by 558227822 in 2016, but are no longer are used or even compile.


hainest:
Confirmed builds correctly on all platforms, unit and external tests pass, and consumer projects build successfully.

kupsch (approved):

PR #1834, opened 2024-12-15
Separate dataflowAPI files from parseAPI CMakeLists.txt

hainest:
This should have been done as part of 739ad5788 in 2023.

- [x] Clean up regex in DyninstLibrary.cmake
- [x] Requires https://github.com/dyninst/external-tests/pull/32


hainest:
Of course there are stupid anomalies I didn't think to check for...

> This change compiles 1 additional file:
> 
> ```
> dataflowAPI/rose/semantics/SMTSolver.C
> ```
Fun fact: The only function ever called from this class is `virtual Satisfiable satisfiable(const SymbolicExpr::Ptr&)`. I can delete SMTSolver.C, and everything still compiles and links. SMTSolver is never used as a base class anywhere. The blood runs from my eyes. I'll see if I can figure out what's going on.
 
> and does not compile two files that were compiled:
> 
> ```
> dataflow/rose/ExtentMap.C
The only member functions called from this class were non-overridden ones from the base class. It can just get replaced by `using ExtentMap = RangeMap<Extent>`.

> dataflow/rose/rangemap.C
This just contained a bunch of explicit instantiations of `Range` that were never used. I deleted this and the declarations.

> This is why the `tmpnam` warning arose. Can all three of these be removed?

I had assumed something jostled the disabled warnings, but didn't feel like chasing it down rather than just fixing it. At least I can delete that PR.

hainest:
@kupsch 

I found the problem. You are allowed to declare and use a virtual function without a definition. I had no idea this was legal. I guess it could be understandable, but I would have thought a warning could be issued.

```c++
struct Foo {
    virtual ~Foo() {}
    virtual int foo();
};

int bar(Foo *f) {
  return f->foo();
}

int main() {
    Foo f;
    return bar(&f);
}
```

SMTSolver was only used in `BinarySymbolicExpr` which, itself, is never used. It can all get deleted. I'll make a separate PR for that.

kupsch (approved):

hainest:
@kupsch What are your thoughts on the regex changes in DyninstLibrary? I think they'll be ok for now. Could you also take a look at https://github.com/dyninst/external-tests/pull/32 when you have a moment? Thanks so much!

kupsch:
The regex changes look OK.

hainest:
Confirmed builds on aarch65 and ppc.

PR #1838, opened 2024-12-16
Treat linker warnings as errors when DYNINST_WARNINGS_AS_ERRORS is on

hainest:
I've confirmed this works with gcc-14 and clang-15. We should run this against the compiler multibuild workflow before merging.


kupsch (approved):

hainest:
Confirmed multibuild workflow passed.

PR #1840, opened 2024-12-17
GitHubCI: re-use source checkout in build action

hainest:
This will allow the build action to be used in other repos.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -29,13 +29,9 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v4
-      with:
-        path: "build-checkout"
-
     - name: Build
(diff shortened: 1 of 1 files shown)


.github/actions/build/action.yaml
```diff
@@ -48,7 +44,7 @@ runs:
 
     - name: Build
       if: ${{ startsWith(inputs.os, 'ubuntu') }}
-      uses: ./build-checkout/.github/actions/build-ubuntu
+      uses: ${{ inputs.src-dir }}/.github/actions/build-fedora
```
kupsch, line 47: is this a copy/paste error in changing ubuntu to fedora?

hainest: Derp. Good catch. Fixed.

kupsch (approved):

PR #1842, opened 2024-12-18
Move MachRegister -> ROSE register conversion into dataflowAPI

hainest:
The conversion is only ever used in dataflowAPI, so that's where it belongs. It also simplifies the conversion by directly creating a `RegisterDescriptor` which is the register vocabulary for interacting with ROSE.

There are several bugs that I will submit fixes for in separate PRs, notably in `SymEvalSemantics::RegisterStateAST::convert`.

@bbiiggppiigg Could you double-check the GPU changes?

### DIFF
--- a/common/h/registers/MachRegister.h
+++ b/common/h/registers/MachRegister.h
@@ -81,8 +81,6 @@ namespace Dyninst {
     bool isFlag() const;
     bool isZeroFlag() const;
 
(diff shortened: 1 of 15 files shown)


bbiiggppiigg (commented):
All the AMDGPU changes looks good to me.

kupsch (approved):

PR #1843, opened 2024-12-18
Fix aarch64 MachRegister -> ROSE register conversion

hainest:
Requires #1842


hainest:
This fixes several test failures on zeroah.

kupsch (approved):

PR #1850, opened 2024-12-26
Unify valid instruction detection

hainest:
These only detect actual decode failures on x86.

'isValid' was added by 410f34166 in 2008 and 'isLegalInsn' by b490edd40 in 2009. It's unclear why the latter was needed.

@bbiiggppiigg Is there a reason the AMD GPU opcodes don't have an invalid entryID?


kupsch (approved):

hainest:
This kicked up a lot of instruction decoder bugs that will need to be fixed. https://github.com/dyninst/unit-tests/pull/8

PR #1851, opened 2024-12-27
Fix MachRegister::getBaseRegister for x86 vector types and IP/PC

hainest:


common/src/registers/MachRegister.C
```diff
@@ -46,33 +46,53 @@ namespace Dyninst {
...
-            // Keep the register number, but change the category to ZMM (e.g., YMM1 -> ZMM1).
-            return MachRegister((reg & ~0x00ff0000) | x86::ZMM);
+          // XMM<N> -> ZMM<N>
+          auto const id = val() & 0xff;
+          auto const arch = getArchitecture();
+          return MachRegister(id | x86::ZMMS | x86::ZMM | arch);
+        }
+        else if(category == x86::YMM) {
+          // assume CPU is new enough that it always has AVX-512 registers
+          // YMM<N> -> ZMM<N>
```
kupsch, line 64: To eliminate code duplication, delete previous `else if(category == x86::XMM)` and change to:

```suggestion
        else if(category == x86::XMM || category = x86::YMM) {
          // assume CPU is new enough that it always has AVX-512 registers
          // XMM<N> and YMM<N> -> ZMM<N>
```

common/src/registers/MachRegister.C
```diff
@@ -46,33 +46,53 @@ namespace Dyninst {
...
+        else if(category == x86_64::XMM) {
+          // assume CPU is new enough that it always has AVX registers
+          // XMM<N> -> ZMM<N>
+          auto const id = val() & 0xff;
+          auto const arch = getArchitecture();
+          return MachRegister(id | x86_64::ZMMS | x86_64::ZMM | arch);
+        }
+        else if(category == x86_64::YMM) {
+          // assume CPU is new enough that it always has AVX-512 registers
+          // YMM<N> -> ZMM<N>
```
kupsch, line 91: To eliminate code duplication, delete previous `else if(category == x86::XMM)` and change to:

```suggestion
        else if(category == x86::XMM || category = x86::YMM) {
          // assume CPU is new enough that it always has AVX-512 registers
          // XMM<N> and YMM<N> -> ZMM<N>
```

kupsch (approved):
You could combine XMM and YMM conversions, otherwise looks good

common/src/registers/MachRegister.C
hainest: Yeah, I noticed that, but thought it might be clearer to leave them as separate branches. It's probably not, though. Fixed.

hainest:
No test regressions on x86.

PR #1854, opened 2024-12-30
Don't use base register in isZeroFlag

hainest:
The individual flag registers (e.g., x86::zf) get mapped to the whole register (e.g., x86::flags).

### DIFF
--- a/common/src/registers/MachRegister.C
+++ b/common/src/registers/MachRegister.C
@@ -596,19 +596,18 @@ namespace Dyninst {
 
   bool MachRegister::isZeroFlag() const {
     if(*this == InvalidReg) return false;
-    auto const base = getBaseRegister();
     switch(getArchitecture()) {
       case Arch_ppc32:
       case Arch_ppc64: {
         // For power, we have a different register representation.
         // We do not use the subrange field for MachReigsters
         // and all lower 32 bits are base ID
-        int baseID = base.val() & 0x0000FFFF;
+        int baseID = val() & 0x0000FFFF;
         return (baseID <= 731 && baseID >= 700 && baseID % 4 == 2) ||
                (baseID <= 628 && baseID >= 621);
       }
       default:
-        return base == getZeroFlag(getArchitecture());
+        return *this == getZeroFlag(getArchitecture());
     }
     return false;
   }


kupsch (approved):
The x86 change looks correct. Should the ppc say that CR<n> (621-628) be a zero flag as these select all 4 flags of the 8 groups?  This would be the equivalent of x86_64 return true for for rflags (if there was only one group).

Of course the code that uses isZeroFlag (if it gets called on ppc) doesn't understand that there are 8 different zero flags and they are not equivalent.

hainest:
> The x86 change looks correct. Should the ppc say that CR (621-628) be a zero flag as these select all 4 flags of the 8 groups? This would be the equivalent of x86_64 return true for for rflags (if there was only one group).

The docs say that "A specified CR field can be set as the result of a Compare instruction.", so any of the CR groups can be used as a "zero" register. That said, cr1 is the implicit one for fixed-point instructions, cr1 for FP ones, and cr6 for vector ones. Maybe a solution is to update #1855 to ensure that something like `reg.isZeroFlag() && reg.isFloatingPoint() && !reg.isVector() == ppc{32,64}::cr1`?

> Of course the code that uses isZeroFlag (if it gets called on ppc) doesn't understand that there are 8 different zero flags and they are not equivalent.

Naturally.

kupsch:
I just meant that `cr3` represent the set of 4 registers `cr3l`, `cr3g`, `cr3e`, and `cr3s`, so `cr3` isn't a zero flag, but it does contain one. `cr` contains `cr3` along with the other 7 sets of condition codes.  I think only `crNe` should be a zero flag while `crN` and `cr` should not be.

Fixing the code that uses `isZeroFlag()` for ppc would require checking that the group used to set the code matched the group used in the test (the code is looking for a specific sequence of instruction properties, so maybe they don't occur on ppc, or when they do the groups are the same).

hainest:
> I just meant that `cr3` represent the set of 4 registers `cr3l`, `cr3g`, `cr3e`, and `cr3s`, so `cr3` isn't a zero flag, but it does contain one. `cr` contains `cr3` along with the other 7 sets of condition codes. I think only `crNe` should be a zero flag while `crN` and `cr` should not be.

Oh, sorry. Yes, I agree. Since `getZeroFlag` returns `cr0e`, then we should just check the single bit for each group. Of course, this wouldn't be an invertable mapping, but that just seems to be the nature of the CR.

> Fixing the code that uses `isZeroFlag()` for ppc would require checking that the group used to set the code matched the group used in the test (the code is looking for a specific sequence of instruction properties, so maybe they don't occur on ppc, or when they do the groups are the same).

I'm almost certain that's not the only place where subregisters would need to be properly handled. At least using `cr<N>e` will make things like [this](https://github.com/dyninst/dyninst/blob/master/parseAPI/src/JumpTableIndexPred.C#L183) work correctly (assuming this is ever used on ppc).

PR #1855, opened 2024-12-30
Add type queries to MachRegister

hainest:
This replaces explicit category checks in user code.

unit tests: https://github.com/dyninst/unit-tests/pull/9

Requires #1867 #1866 #1864


kupsch:
Are there definitions for types.  I'm wondering if you had the intentions that floating point and vector are those types of registers **and related registers**. For instance KMSK isn't a vector register and FPCTL types are not a floating point register? This implies that places that check for `isFloatingPoint()` now including floating control registers (maybe not right) and vector registers (maybe OK); and the places that now chek `isVectorRegister` now include KMSK register types on x86_64 (maybe not right) and more vector types (likely an improvement).

@bbiiggppiigg: should VGPR and ACC_VGPR types be a GPR (seems more like a vector register), and there are masks for AMDGPU vectors in VCC (which serve a function similar to KMSK, and there is a also the EXEC register which functions like a mask.

hainest:
> Are there definitions for types. I'm wondering if you had the intentions that floating point and vector are those types of registers **and related registers**.

I made a note in the comments for `isFloatingPoint`, but just noticed it's not there for `isVector` (I'll update that).

>For instance KMSK isn't a vector register and FPCTL types are not a floating point register? This implies that places that check for `isFloatingPoint()` now including floating control registers (maybe not right) and vector registers (maybe OK); and the places that now chek `isVectorRegister` now include KMSK register types on x86_64 (maybe not right) and more vector types (likely an improvement).

The idea is that `isVector` returns true for any register that could be used directly or indirectly by a vector instruction. This includes both the arithmetic and control/status registers. `isControlStatus` can then be used to filter out which of these are actually wanted. For example, `reg.isVector() && !reg.isControlStatus()` would produce the mm and x/y/z/mm registers on x86[1]. Similarly, `reg.isFloatingPoint() && !reg.isControlStatus() && !reg.isVector()` would give the st* ones. One place this falls down is for predicate registers on aarch64 since the only way to find them is with `reg.isVector() && reg.isControlStatus() && reg.getLengthID() == aarch64::PREDS` which is gross and not at all how I intended `getLengthID` to be used. I had considered an `isVectorPredicate` that would parallel the `trueP`/`falseP` in `Instruction::appendOperand`, but wasn't sure it would be useful. Thoughts?

> @bbiiggppiigg: should VGPR and ACC_VGPR types be a GPR (seems more like a vector register)

The GPU architectures do diverge a bit from the CPU ones as most GPU registers are vectors, too. However, I could see this distinction being useful during rewriting to store temporaries in the correct type of GPR. It would require extra care for the register allocator to know that GPU registers can be used for both general-purpose and floating-point operations. In fact, this might be something to add for the CPU architectures. Although we are a long way from being able to rewrite vector operations.

>and there are masks for AMDGPU vectors in VCC (which serve a function similar to KMSK, and there is a also the EXEC register which functions like a mask.

I noticed they aren't [flags](https://github.com/dyninst/unit-tests/pull/9#issuecomment-2565732992), but I *should* have considered if they might be control registers. Good catch.

---

[1] I just noticed that `mxcsr` isn't being considered as a vector register. That's inconsistent with including the predicate registers. I'll add it.

kupsch (approved):

hainest:
@bbiiggppiigg Can the [TTMP](https://github.com/dyninst/dyninst/blob/master/common/h/registers/AMDGPU/amdgpu_gfx908_regs.h#L130) registers be used for floating-point calculations? If so, I should add them to `isFloatingPoint`.

bbiiggppiigg:
> @bbiiggppiigg Can the [TTMP](https://github.com/dyninst/dyninst/blob/master/common/h/registers/AMDGPU/amdgpu_gfx908_regs.h#L130) registers be used for floating-point calculations? If so, I should add them to `isFloatingPoint`.

I assume they can be used for floating-point calculations. But they are also only writable in trap handler.

PR #1857, opened 2024-12-31
Fix MachRegister::isZeroFlag for ppc

hainest:
Currently, this treats every field in each group as being a zero flag. Since we have representations for each bit in the field, then only the EQ flag should be checked here.


hainest:
I just noticed that I already had tests for the in https://github.com/dyninst/unit-tests/pull/9.

kupsch (approved):

PR #1858, opened 2025-01-05
Use MachRegister::isFlag in RegisterAST::isFlag

hainest:
This isn't currently used inside of Dyninst, but this class is intended to be overridden by users creating their own AST. In the base case, it should do the right thing here instead of only working on x86.


kupsch (approved):

hainest:
No regressions on x86.

PR #1864, opened 2025-01-12
Fix base register calculation for ppc

hainest:
OpenPOWER Foundation
Power Instruction Set Architecture
Version 3.1C
May 26, 2024
4.2 Floating-Point Facility Registers
3.2 Fixed-Point Facility Registers
6.3 Vector Facility Registers
7.2 VSX Registers

Unit tests are passing: https://github.com/dyninst/unit-tests/pull/12

### DIFF
--- a/common/src/registers/MachRegister.C
+++ b/common/src/registers/MachRegister.C
@@ -166,8 +166,34 @@ namespace Dyninst {
           default: return *this;
         }
 
-      case Arch_ppc32:
-      case Arch_ppc64:
+      case Arch_ppc32: {
+        auto ppc_id = [](MachRegister r) {
+          return r.val() & 0x0000FFFF;
+        };
+
+        auto const id = ppc_id(*this);
+
+        if(category == ppc32::FSR) {
+          // fsr<N> -> fpr<N>[0:31]
+          auto const base_id = ppc_id(ppc32::fsr0);
+          auto const offset = id - base_id;
+          auto const new_id = ppc_id(ppc32::fpr0) + offset;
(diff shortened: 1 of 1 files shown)


hainest:
The unit tests won't pass because there's a circular dependency with the unit test PR. I should figure out how to put the unit tests in the Dyninst source tree instead of a separate repo.

kupsch:
Why do r<N>, fpr<N>, and vsr<N> map to their respective 0 register? Shouldn't they map to themselves?

What are fsr<N>? I'm having problems find any information about them.

kupsch:
Since the fsr registers are 32-bit float registers and they share bits with the fpr registers, their base should be the corresponding fpr register.

hainest:
> Since the fsr registers are 32-bit float registers and they share bits with the fpr registers, their base should be the corresponding fpr register.

They'll get removed when migrating to Capstone, but I'll do this.

kupsch (approved):
Looks good.  It looks like all the test failures are due to unit-tests needed to be updated to match the base register changes.

kupsch:
one quick change is that `vr<N>` seems to map to `vr<N+32>`, so the base reqister for vr-register should be a vsr-register (see chapter 6.3 "Vector Registers" of PowerISA_public.v3.1).

kupsch:
Never mind the last comment, dyninst doesn't have `vr` registers, just `vsr`.

hainest:
CI tests are passing after merging required unit tests. https://github.com/dyninst/dyninst/actions/runs/12791997056

PR #1866, opened 2025-01-12
Update aarch64 flag and pstate registers

hainest:
This gives the same semantics as for flag registers on the other platforms.

It doesn't look like Capstone has a representation for 'pstate'. I'll update the importer once the aarch64 decoder is migrated.


kupsch (approved):

hainest:
@kupsch Sorry, I forgot to make the size of the DAIF flags one bit.

kupsch (approved):

PR #1868, opened 2025-01-12
Update base registers for AMDGPU

hainest:
TGT registers aren't included because the comments say their usage isn't yet known.

unit tests: https://github.com/dyninst/unit-tests/pull/11


bbiiggppiigg (approved):

common/src/registers/MachRegister.C
```diff
@@ -166,32 +166,56 @@ namespace Dyninst {
...
-        switch(category) {
-          case amdgpu_gfx908::SGPR: return MachRegister((reg & 0x000000ff) | amdgpu_gfx908::s0);
-          case amdgpu_gfx908::VGPR: return MachRegister((reg & 0x000000ff) | amdgpu_gfx908::v0);
-          case amdgpu_gfx908::HWR: return MachRegister(reg);
-
-          default: return *this;
+      case Arch_amdgpu_gfx908: {
+        if(category == amdgpu_gfx908::MISC) {
+          switch(val()) {
+            case amdgpu_gfx908::ivcc_lo:
```
bbiiggppiigg, line 172: I'm not sure why there is an i infront of these amdgpu registers. Same comment applies to all other registers and cases for 90a and 940

bbiiggppiigg: These names seem to be introduced in #1855, not sure if is intentional

hainest: The `i` prefix is the variable that is [intended](https://github.com/dyninst/dyninst/blob/v13.0.0/common/h/registers/reg_def.h#L37) to be used in switch cases. For some reason, no one really uses it- instead, relying on the terribly implicit conversion [operator](https://github.com/dyninst/dyninst/blob/v13.0.0/common/h/registers/MachRegister.h#L59).

bbiiggppiigg (approved):
LGTM

PR #1870, opened 2025-01-16
Add missing invalid mnemonics

### LINKED ISSUE #1869: broken CFG support for AMD GPUs
I think a recent change to `Instruction.C` broke the CFG support for AMD GPUs.
```
commit b8199b220f8bd3394c7b734380ccfa0e9e91c23c
Author: Tim Haines <thaines.astro@gmail.com>
Date:   Sat Dec 28 16:16:31 2024 -0600

    Unify valid instruction detection (#1850)

M       instructionAPI/src/Instruction.C
```
The cases for the AMD GPU types all return `false` for `is_valid_mnemonic()`.
This results in every GPU function having only one machine instruction
with an out edge to the sink.  So, `parse()` basically misses the entire
function.

For example, my debug output for hpcstruct run on quicksilver reports:
```
func:  0x14c74  (1/1)  link='_Z17CycleTrackingGutsP10MonteCarloiP13ParticleVaultS2_'
parse:  'CycleTrackingGuts'
file:   '/scratch/krentel/hpctoolkit-tutorial-examples/gpu/amd/quicksilver.rocm/quicksilver.rocm/quicksilver/src/CycleTracking.cc'

block:
stmt:  0x14c74 (4)  l=16  f='/scratch/krentel/hpctoolkit-tutorial-examples/gpu/amd/quicksilver.rocm/quicksilver.rocm/quicksilver/src/CycleTracking.cc'
out edges:  0xffffffffffffffff (direct)
```

That's the entire function, one 4-byte instruction.

I should check, AFAIK dyninst supports CFGs for AMD GPUs, is that right?
With the patch below, I get a full CFG with loops and calls.

Anyway, I suggest the following patch, but if you want something else,
feel free.  Also feel free to ask me to write the MR if you're happy
with my suggestion.

```
diff --git a/instructionAPI/src/Instruction.C b/instructionAPI/src/Instruction.C
index 03d4fabf3..2ec081dfd 100644
--- a/instructionAPI/src/Instruction.C
+++ b/instructionAPI/src/Instruction.C
@@ -83,10 +83,12 @@ namespace Dyninst { namespace InstructionAPI {
         case Arch_aarch32:
         case Arch_cuda:
         case Arch_intelGen9:
+          return false;
+
         case Arch_amdgpu_gfx908:
         case Arch_amdgpu_gfx90a:
         case Arch_amdgpu_gfx940:
-          return false;
+          return true;
       }
       return false;
     }
```

An example program is quicksilver compiled on ufront.  This is a fat
binary and contains two gpubins.  I'll attach the gpubin I was using
for testing.  If someone needs updated access to ufront, let me know.

[6167c8.gpubin.gz](https://github.com/user-attachments/files/18443503/6167c8.gpubin.gz)

@hainest  @bbiiggppiigg

hainest:
Fixes #1869

@bbiiggppiigg Does the work with your importer?

@mwkrentel Could you try this with HPCToolkit?

### DIFF
--- a/common/h/mnemonics/AMDGPU/gfx908_entryIDs.h
+++ b/common/h/mnemonics/AMDGPU/gfx908_entryIDs.h
@@ -1,3 +1,4 @@
+amdgpu_gfx908_op_INVALID,
 amdgpu_gfx908_op_BUFFER_ATOMIC_ADD,
 amdgpu_gfx908_op_BUFFER_ATOMIC_ADD_F32,
 amdgpu_gfx908_op_BUFFER_ATOMIC_ADD_X2,
--- a/common/h/mnemonics/AMDGPU/gfx90a_entryIDs.h
+++ b/common/h/mnemonics/AMDGPU/gfx90a_entryIDs.h
@@ -1,3 +1,4 @@
+amdgpu_gfx90a_op_INVALID,
 amdgpu_gfx90a_op_BUFFER_ATOMIC_ADD,
 amdgpu_gfx90a_op_BUFFER_ATOMIC_ADD_F32,
 amdgpu_gfx90a_op_BUFFER_ATOMIC_ADD_F64,
--- a/common/h/mnemonics/AMDGPU/gfx940_entryIDs.h
+++ b/common/h/mnemonics/AMDGPU/gfx940_entryIDs.h
@@ -1,3 +1,4 @@
+amdgpu_gfx940_op_INVALID,
 amdgpu_gfx940_op_BUFFER_ATOMIC_ADD,
 amdgpu_gfx940_op_BUFFER_ATOMIC_ADD_F32,
 amdgpu_gfx940_op_BUFFER_ATOMIC_ADD_F64,
--- a/common/h/mnemonics/IntelGPU/generic_entryIDs.h
+++ b/common/h/mnemonics/IntelGPU/generic_entryIDs.h
@@ -1,1 +1,2 @@
+intel_gpu_op_INVALID,
 intel_gpu_op_general,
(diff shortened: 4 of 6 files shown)


bbiiggppiigg (approved):

bbiiggppiigg:
> Fixes #1869
> 
> @bbiiggppiigg Does the work with your importer?

I've added it to the importer so it should be reflected when the implementation based on the next drop is done.
 
> @mwkrentel Could you try this with HPCToolkit?

This along won't work until we have #1871

kupsch (approved):

mwkrentel:
Yes, AFAICT this fixes all the known problems with AMD GPUs.
I get real CFGs, including loops and calls, not just the single
machine instruction.

I was about to report another problem where this would throw an
assert in `RoseInsnFactory.C`, but between the two patches,
this also fixes that.

Thanks !!

hainest:
> Yes, AFAICT this fixes all the known problems with AMD GPUs. I get real CFGs, including loops and calls, not just the single machine instruction.
> 
> I was about to report another problem where this would throw an assert in `RoseInsnFactory.C`, but between the two patches, this also fixes that.

Very glad that fixed it quickly. Could you post the program you were using to grab the CFG? I'd really like to add even a tiny test for AMD GPU stuff.

mwkrentel:
> Could you post the program you were using to grab the CFG?
> I'd really like to add even a tiny test for AMD GPU stuff.

Maybe.  I was using debug output embedded in the hpcstruct program and
you probably don't want to use that.

I have some 8-year old unit tests, probably one would be useful for
you.  Probably, there's one that simply dumps the CFG + inline info
into text.  Either that or something close to it.
But it'll be before spack and I'll have to see if it needs some tweaks
to run with current dyninst.

Right now, we're all focused on getting rocm ready for a LLNL hack-a-thon.
I'll get to it at some point, but if you want to look now, try:

https://github.com/HPCToolkit/hpctoolkit-tests/tree/master/dyninst

One of the tests is probably close to what you want.

PR #1871, opened 2025-01-16
Fix AMDGPU rose to use explicit operands

bbiiggppiigg:
Previous implementation assumed the existence of allOperands instead of just explicitOperands.
This change is required to reflect the changes made in #1859.


kupsch (commented):
Looks OK.  The comment on line 436 should get updated to reflect 2 operands.

kupsch (approved):

PR #1886, opened 2025-01-30
Disable simplifyStack for AMDGPU

bbiiggppiigg:
Disable simplifyStack for AMDGPU as it is causing segmentation fault when paring certain AMDGPU binaries, 
when OMP_NUM_THREADS!=1. T
The problem seems to be related the maps in Annotatable.h needs to be thread safe.


hainest (approved):
No detected regressions.

PR #1891, opened 2025-02-06
Clean up dyninstAPI codegen/instrumentation files

hainest:
This makes it much easier to identify common interface types and functions.


hainest:
This is building/testing fine on all 19 machines.

PR #1895, opened 2025-02-20
Fix ROSE expansion of aarch64 ldp, ldurs instructions

bbiiggppiigg:
Fix the handling for ldp_gen_execute
(address should be args[2], immediate args[3])

https://developer.arm.com/documentation/ddi0602/2024-12/Base-Instructions/LDP--Load-pair-of-registers-

Fix the handling for ldursw
(immediate args[2] is optional)
https://developer.arm.com/documentation/ddi0602/2024-12/Base-Instructions/LDURSW--Load-register-signed-word--unscaled--?lang=en#simm

### DIFF
--- a/dataflowAPI/rose/semantics/DispatcherARM64.C
+++ b/dataflowAPI/rose/semantics/DispatcherARM64.C
@@ -2921,7 +2921,7 @@ namespace rose {
 
                 struct IP_ldp_gen_execute : P {
                     void p(D d, Ops ops, I insn, A args, B raw) {
-                        BaseSemantics::SValuePtr address = d->effectiveAddress(args[1]);
+                        BaseSemantics::SValuePtr address = d->effectiveAddress(args[2]);
                         BaseSemantics::SValuePtr data1;
                         BaseSemantics::SValuePtr data2;
                         int dbytes = d->getDatasize(raw) / 8;
@@ -2975,7 +2975,7 @@ namespace rose {
                             if (wb_unknown) {
                                 address = ops->unspecified_(1);
                             } else if ((EXTR(11, 11) == 0 && EXTR(24, 24) == 0)) {
-                                address = ops->add(address, d->read(args[2]));
+                                address = ops->add(address, d->read(args[3]));
                             }
 
                             if (EXTR(5, 9) == 31) {
@@ -3892,7 +3892,7 @@ namespace rose {
                             if (wb_unknown) {
                                 address = ops->unspecified_(1);
                             } else if ((EXTR(11, 11) == 0 && EXTR(24, 24) == 0)) {
-                                address = ops->add(address, d->read(args[2]));
+                                address = ops->add(address, ops->number_(EXTR(12,20),9));
                             }
 
                             if (EXTR(5, 9) == 31) {


bbiiggppiigg:
It seems like the CI is down because of some payment issue.

hainest:
I'd say Github CI is broken. As an open-source project, we don't have any CI limits. The usage statements confirm (https://github.com/organizations/dyninst/settings/billing/summary). Let's keep trying to manually run them until it works.

hainest:
I was recently chasing this down, but found myself beating my head against the wall. I found the ROSE semantics were working, but never stopped to check to see if they were actually correct. We *REALLY* need to work on using ROSE as an external dependency so we can get updates.

Which machine did you find this bug on?

bbiiggppiigg:
> Which machine did you find this bug on?

John reported to us a binary for ARM binary for NVIDIA Grace CPU that he fails to parse.

I don't know whether it is becuase the rose execution of these instructions were never tested, or something changed recently with how these instructions are dealt with.

One extra thing note mentioning:
ldp_gen_execute fails consistently, but ldursw does not always fail (especially when OMP_NUM_THREADS set to 1).
That being said, after fixing the ldursw I didn't see observe any failures across multiple runs with parallel parsing (or maybe I just didn't run it enough times lul).

> I was recently chasing this down, but found myself beating my head against the wall. I found the ROSE semantics were 
> working, but never stopped to check to see if they were actually correct. We REALLY need to work on using ROSE as an
> external dependency so we can get updates.

Does ROSE (the maintained version) actually support these instructions ?

hainest:
> > Which machine did you find this bug on?
> 
> John reported to us a binary for ARM binary for NVIDIA Grace CPU that he fails to parse.

That's where I was seeing it, as well.

> I don't know whether it is becuase the rose execution of these instructions were never tested, or something changed recently with how these instructions are dealt with.

I can't be certain, but I believe gcc is only emitting them for grace/hopper CPUs. I've not seen them anywhere else.

> One extra thing note mentioning: ldp_gen_execute fails consistently, but ldursw does not always fail (especially when OMP_NUM_THREADS set to 1). That being said, after fixing the ldursw I didn't see observe any failures across multiple runs with parallel parsing (or maybe I just didn't run it enough times lul).

That's interesting. Not surprising, but interesting.

> > I was recently chasing this down, but found myself beating my head against the wall. I found the ROSE semantics were
> > working, but never stopped to check to see if they were actually correct. We REALLY need to work on using ROSE as an
> > external dependency so we can get updates.
> 
> Does ROSE (the maintained version) actually support these instructions ?

[Yes](https://github.com/rose-compiler/rose/blob/weekly/src/Rose/BinaryAnalysis/InstructionSemantics/DispatcherAarch64.C#L777), but they use a completely different representation now. The one we're using is 12+ years old.

bbiiggppiigg:
> > > Which machine did you find this bug on?
> > 
> > 
> > John reported to us a binary for ARM binary for NVIDIA Grace CPU that he fails to parse.
> 
> That's where I was seeing it, as well.
> 
> > I don't know whether it is becuase the rose execution of these instructions were never tested, or something changed recently with how these instructions are dealt with.
> 
> I can't be certain, but I believe gcc is only emitting them for grace/hopper CPUs. I've not seen them anywhere else.
> 
> > One extra thing note mentioning: ldp_gen_execute fails consistently, but ldursw does not always fail (especially when OMP_NUM_THREADS set to 1). That being said, after fixing the ldursw I didn't see observe any failures across multiple runs with parallel parsing (or maybe I just didn't run it enough times lul).
> 
> That's interesting. Not surprising, but interesting.
> 
> > > I was recently chasing this down, but found myself beating my head against the wall. I found the ROSE semantics were
> > > working, but never stopped to check to see if they were actually correct. We REALLY need to work on using ROSE as an
> > > external dependency so we can get updates.
> > 
> > 
> > Does ROSE (the maintained version) actually support these instructions ?
> 
> [Yes](https://github.com/rose-compiler/rose/blob/weekly/src/Rose/BinaryAnalysis/InstructionSemantics/DispatcherAarch64.C#L777), but they use a completely different representation now. The one we're using is 12+ years old.

Our Ldp has 4 arguments and will fail with assertions in the new implementation. The decoder probably needs to be updated accordingly.

hainest:
This didn't fix the failure I'm seeing. If it fixes John's issue, then we should merge it.

hainest:
> > > > Which machine did you find this bug on?
> > > 
> > > 
> > > John reported to us a binary for ARM binary for NVIDIA Grace CPU that he fails to parse.
> > 
> > 
> > That's where I was seeing it, as well.
> > > I don't know whether it is becuase the rose execution of these instructions were never tested, or something changed recently with how these instructions are dealt with.
> > 
> > 
> > I can't be certain, but I believe gcc is only emitting them for grace/hopper CPUs. I've not seen them anywhere else.
> > > One extra thing note mentioning: ldp_gen_execute fails consistently, but ldursw does not always fail (especially when OMP_NUM_THREADS set to 1). That being said, after fixing the ldursw I didn't see observe any failures across multiple runs with parallel parsing (or maybe I just didn't run it enough times lul).
> > 
> > 
> > That's interesting. Not surprising, but interesting.
> > > > I was recently chasing this down, but found myself beating my head against the wall. I found the ROSE semantics were
> > > > working, but never stopped to check to see if they were actually correct. We REALLY need to work on using ROSE as an
> > > > external dependency so we can get updates.
> > > 
> > > 
> > > Does ROSE (the maintained version) actually support these instructions ?
> > 
> > 
> > [Yes](https://github.com/rose-compiler/rose/blob/weekly/src/Rose/BinaryAnalysis/InstructionSemantics/DispatcherAarch64.C#L777), but they use a completely different representation now. The one we're using is 12+ years old.
> 
> Our Ldp has 4 arguments and will fail with assertions in the new implementation. The decoder probably needs to be updated accordingly.

We'll move to Capstone for aarch64 long before we get around to updating ROSE.

bbiiggppiigg:
> This didn't fix the failure I'm seeing. If it fixes John's issue, then we should merge it.

This fixed the issue parsiing John's binary.
If you have other binaries that are failing because of ROSE, let me know.

hainest (approved):

Issue #1892, opened 2025-02-06
Discussion on cross-platform code generation


hainest:
@kupsch @bbiiggppiigg @wxrdnx @rochauha

Everything below pertains only to code generation and parsing in dyninstAPI. I'm ignoring proccontrol, stackwalker, and dyninstAPI_RT for now. The analysis components in ParseAPI, InstructionAPI, and SymtabAPI seem to be working in a cross-platform sense already- although in a very reduced capacity on anything other than x86 (that's a conversation for another day).

Note: I use "conditional compilation" here to mean selectively adding source/header files in CMake rather than the act of checking the corresponding macros in the sources (e.g., `#ifdef DYNINST_HOST_ARCH_I386`). The latter is a related but more difficult problem (see #583).

I very much understand the approach taken in #1802, #1806, #1830, and #1831. They follow the existing convention of using conditional compilation to achieve some degree of polymorphic functionality. This has worked for a long time because no one was asking to do cross-platform codegen (e.g., rewriting an x86 binary on an aarch64 host). Mapping the codegen to the host made sense (although the design choice to achieve is arguably suboptimal). With GPUs, this is not the case. AMD gpus alone can be used on all three platforms we support (@wxrdnx do you know what general GPU support is like on RISC-V?). This requires a different approach.

As I understand it, there are two sources of AMD GPU object files:

1. Feed GPU-specific source code into the ROCM compiler to create a standalone standard ELF DSO that can be loaded onto the device
2. Use device offloading (e.g., OpeMP) to create a [bundled code object](https://clang.llvm.org/docs/ClangOffloadBundler.html) that mixes host and device code via separate .o files linked into a single ELF.

Case (1) would mostly fit into the current convention of having the "host" architecture be the same as the codegen one. However, rewriting a binary that links against such a DSO wouldn't be able to rewrite it's dependencies because HOST_ARCH != CODEGEN_ARCH. This would require two builds of Dyninst- one for the host binaries and one for the GPU binaries.

Case (2) is a degenerate case of (1) because embeddeding host and device code means no build of Dyninst could rewrite the entire binary.

In both cases, any tool that uses Dyninst would have to have two different link dependencies when doing rewriting since the symbol names for any code created via conditional compilation will collide (e.g., insnCodeGen::generateTrap). That means the tool author needs to either have two builds of their tool with the different Dyninst builds RPATHed accordingly or use LD_PRELOAD to fix our design. I wouldn't use a library that forced either of these on me.

[...]

PR #1896, opened 2025-02-27
Define ObjectPE

batuzovk:
This is a next step towards support of PE files parsing.

A dummy ObjectPE class is defined. Actual parsing is not implemented yet.

### DIFF
--- a/symtabAPI/CMakeLists.txt
+++ b/symtabAPI/CMakeLists.txt
@@ -32,6 +32,7 @@ set(_private_headers
     src/emitWin.h
     src/LinkMap.h
     src/Object-elf.h
+    src/Object-pe.h
     src/Object.h
     src/Object-nt.h
     src/Type-mem.h
@@ -72,7 +73,8 @@ if(DYNINST_OS_UNIX)
     src/LinkMap.C
     src/emitElf.C
     src/emitElfStatic.C
-    src/dwarfWalker.C)
+    src/dwarfWalker.C
+    src/Object-pe.C)
 
   if(DYNINST_HOST_ARCH_X86_64 OR DYNINST_HOST_ARCH_I386)
     list(APPEND _sources src/emitElfStatic-x86.C src/relocationEntry-elf-x86.C)
--- a/symtabAPI/h/Function.h
+++ b/symtabAPI/h/Function.h
@@ -193,6 +193,7 @@ class DYNINST_EXPORT InlinedFunction : public FunctionBase
    friend class DwarfWalker;
    friend class Object;
    friend class ObjectELF;
+   friend class ObjectPE;
   public:
    InlinedFunction(FunctionBase *parent);
    virtual ~InlinedFunction();
--- a/symtabAPI/h/Region.h
+++ b/symtabAPI/h/Region.h
@@ -47,6 +47,7 @@ class Symtab;
 class DYNINST_EXPORT Region : public AnnotatableSparse {
    friend class Object;
    friend class ObjectELF;
+   friend class ObjectPE;
    friend class Symtab;
    friend class SymtabTranslatorBase;
    friend class SymtabTranslatorBin;
--- a/symtabAPI/h/Symtab.h
+++ b/symtabAPI/h/Symtab.h
@@ -91,6 +91,7 @@ class DYNINST_EXPORT Symtab : public LookupInterface,
    friend class relocationEntry;
    friend class Object;
    friend class ObjectELF;
+   friend class ObjectPE;
 
    // Hide implementation details that are complex or add large dependencies
    const std::unique_ptr<symtab_impl> impl;
@@ -236,6 +237,7 @@ class DYNINST_EXPORT Symtab : public LookupInterface,
    bool isCode(const Offset where) const;
    bool isData(const Offset where) const;
    bool isValidOffset(const Offset where) const;
+   FileFormat getFileFormat() const;
 
    bool getMappedRegions(std::vector<Region *> &mappedRegs) const;
 
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -2298,7 +2298,7 @@ ObjectELF::ObjectELF(MappedFile *mf_, bool, void (*err_func)(const char *),
         containingFunc(nullptr)
 {
     li_for_object = NULL; 
-
+    file_format_ = FileFormat::ELF;
 #if defined(TIMED_PARSE)
     struct timeval starttime;
   gettimeofday(&starttime, NULL);
--- a/symtabAPI/src/Object-pe.C
+++ b/symtabAPI/src/Object-pe.C
@@ -1,0 +1,113 @@
+/*
+ * See the dyninst/COPYRIGHT file for copyright information.
+ * 
+ * We provide the Paradyn Tools (below described as "Paradyn")
+ * on an AS IS basis, and do not warrant its validity or performance.
+ * We reserve the right to update, modify, or discontinue this
+ * software at any time.  We shall have no obligation to supply such
+ * updates or modifications or any other form of support to you.
+ * 
+ * By your use of Paradyn, you understand and agree that we (or any
+ * other person or entity with proprietary rights in Paradyn) are
+ * under no obligation to provide either maintenance services,
+ * update services, notices of latent defects, or correction of
+ * defects for Paradyn.
+ * 
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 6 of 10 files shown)


batuzovk:
This is a follow-up to https://github.com/dyninst/dyninst/pull/1758

Copyright notice at the start of the new files was copied from Object-elf, thus it can be outdated.

symtabAPI/h/symutil.h
```diff
@@ -70,7 +70,9 @@ typedef enum {
    obj_Unknown,
    obj_SharedLib,
    obj_Executable,
-   obj_RelocatableFile
+   obj_RelocatableFile,
+   obj_PEExecutable,
+   obj_PESharedLib
```
hainest, line 75: It's not clear why we need special types for PE files here. The [old](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object-nt.C#L1081) PE implementation used the existing types.

hainest (changes requested):
Ultimately, we don't want users to be forced to have `peparse` installed when they aren't analyzing PE binaires. Naturally, this will need some kind of `-DDYNINST_ENABLE_PE` which will lead to needing to hide the PE implementation. However, we do not want to have a macro guard around local code like Object.C:313. 

Without that kind of code firewall, we're kind of only left with needing a conditionally-compiled stub implementation for PE files. That's not great, but I'm very much open to other solutions.

symtabAPI/h/symutil.h
batuzovk: For the same purpose this enum exists - to be able to tell which type of binary file we are working on. I started by reusing the approach from the old PE implementation, but the moment we needed on the user side (i.e. in the application using Dyninst as a library) to tell if we are working with ELF or PE it felt inadequate. I think having different types of object file here is better than trying to present both .dll and .so as SharedLibrary and having a separate variable for PE-ELF distinction. After all, a relocatable executable and a shared library in ELF format have more in common than a .dll and an .so.

As for pe-parse library, I was planning to raise this question myself in the next patch in the series. It is a very small library, but it is uncommon, and analyzing PE files probably aren't going to be the most common use case. The decent options I see is either have a stub (copy of the code from this MR will work), or bring pe-parse in as a submodule.

hainest: > It was added for the same purpose this enum exists - to be able to tell which type of binary file we are working on.

The public interface for this is defined by `Symbtab::is{Exec,Executable,SharedLibrary}`. `Symtab::getObjectType` really should have been deprecated in #1042. It is a clear instance of incidental design and leaking internal information.

>I started by reusing the approach from the old PE implementation, but the moment we needed on the user side (i.e. in the application using Dyninst as a library) to tell if we are working with ELF or PE it felt inadequate.

This is a point we didn't address in #1042 because the Windows code is effectively non-functioning. I agree there could be utility in allowing a user to know the type of `Object` being parsed (i.e., ELF vs PE). However, that information should be orthogonal to `Symbtab::is{Exec,Executable,SharedLibrary}`. Yet, I am hesitant to make that available because it detracts from having an abstract `Object` interface.

>After all, a relocatable executable and a shared library in ELF format have more in common than a .dll and an .so.

I don't know much about the Windows internals. Could you elaborate on why a DLL couldn't be agnostically treated the same way as shared libraries are using the current `Object` interface (e.g., having PIC, dynamic linkage, etc.)?

>As for pe-parse library, I was planning to raise this question myself in the next patch in the series. It is a very small library, but it is uncommon, and analyzing PE files probably aren't going to be the most common use case. The decent options I see is either have a stub (copy of the code from this MR will work), or bring pe-parse in as a submodule.

We've discussed submodules in the past. The tradeoffs have typically not been favorable to us. However, this might be an exception. In the future, we might also need to include ROSE as a submodule, so now could be a good time to revisit the conversation. @kupsch @bbiiggppiigg @wxrdnx, thoughts?

batuzovk: > Could you elaborate on why a DLL couldn't be agnostically treated the same way as shared libraries are using the current Object interface (e.g., having PIC, dynamic linkage, etc.)?

Some differences I can immediately think of:

* Different relocation types. We keep the "type" from the original, but the same value means different things in PE and ELF. For example, the value `4` means `R_386_PLT32` for 32-bit ELF, but `IMAGE_REL_BASED_HIGHADJ` for 32-bit PE.
* Different implementation of PIC for 32-bit binaries. ELF uses thunks, PE has relocation in the code that are adjusted during loading process. In practice this can have a significant effect on the user. Basically, with PE you need additional checks to make sure the code you are seeing is the same code that will be executed. It creates a lot of additional problems if any code rewriting is considered.
* Different loading process (e.g. existence of preferred base in PE, file alignment that doesn't allow direct mapping from file to memory for PE, etc.). The situations where it would matter for the user are a bit exotic, but not impossible.

Basically, most operations don't depend on whether we are working with ELF or PE (e.g. disassemble from examples repository will work with PE without any modifications), but some times there are use cases where the users need to know which one they are dealing with at the moment.

I agree that some kind of interface like `isPE` or `isELF` is needed. The question is how to better store this information internally, i.e. to expand the existing enum, or to add another one, like `format_PE`, `format_ELF`, etc. Or we can just dynamically cast `this`, but it sounds a bit ugly to me.

hainest: Thanks for taking the time to write this up. It was very helpful. I left a few comments.


> > Could you elaborate on why a DLL couldn't be agnostically treated the same way as shared libraries are using the current Object interface (e.g., having PIC, dynamic linkage, etc.)?
> 
> Some differences I can immediately think of:
> 
>Different relocation types. We keep the "type" from the original, but the same value means different things in PE and ELF. For example, the value `4` means `R_386_PLT32` for 32-bit ELF, but `IMAGE_REL_BASED_HIGHADJ` for 32-bit PE.

That makes sense since those values effectively can't be globally unique. We have format-specific code for relocation (e.g., symtabAPI/src/relocationEntry-elf-x86.C) with a common interface (symtabAPI/h/relocationEntry.h), so we can handle PE in the same way. Although the `relocationEntry` class likely needs to be redesigned since it talks about PLT stuff.


>Different implementation of PIC for 32-bit binaries. ELF uses thunks, PE has relocation in the code that are adjusted during loading process. In practice this can have a significant effect on the user. Basically, with PE you need additional checks to make sure the code you are seeing is the same code that will be executed. It creates a lot of additional problems if any code rewriting is considered.

I'll add this to the notes on cross-platform code generation in #1892.


>Different loading process (e.g. existence of preferred base in PE

Are the [`Object::get*Address()`](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object.h#L152) members sufficient for this?

> file alignment that doesn't allow direct mapping from file to memory for PE, etc.).

Will this be an issue when reading a PE file into memory on Linux using mmap [`MappedFile::map_file`](https://github.com/dyninst/dyninst/blob/master/common/src/MappedFile.C#L306)?


> Basically, most operations don't depend on whether we are working with ELF or PE (e.g. disassemble from examples repository will work with PE without any modifications), but some times there are use cases where the users need to know which one they are dealing with at the moment.
> I agree that some kind of interface like `isPE` or `isELF` is needed. The question is how to better store this information internally, i.e. to expand the existing enum, or to add another one, like `format_PE`, `format_ELF`, etc.

There are several places in Dyninst where we are making compile-time decisions based on the current host platform (e.g., the macros `os_windows` and `i386_unknown_nt4_0`) that will need to be turned into runtime checks to properly handle PE files. I'll make a list of those in a separate comment so as not to completely clog this one. The question becomes where file format information should be kept. I _think_ the `Object` class should work, but there are some places in `image` and `MappedFile` that might need that information before an `Object` is created.

I would suggest we move `ObjectType` into ObjectElf and then have `ObjectPE` define its own. Those values can then be used to compute `isOnlyExecutable()`, `isExecutable()`, `isSharedLibrary()`, `isOnlySharedLibrary()`, `isDebugOnly()`.

Then add 

```c++
enum class file_format {
  ELF,
  PE
};
```

to `Object`.


> Or we can just dynamically cast `this`, but it sounds a bit ugly to me.

Definitely agree that's not a good solution.

batuzovk: > Are the [Object::get*Address()](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/Object.h#L152) members sufficient for this?

`getPrefferedBase()` was added for this. Both `LoadAddress` and `BaseAddress` have their own separate meaning in this context.

> Will this be an issue when reading a PE file into memory on Linux using mmap [MappedFile::map_file](https://github.com/dyninst/dyninst/blob/master/common/src/MappedFile.C#L306)?

Not at all. We only need to read data from a file, not load it in the way it can be executed.

I've updated the `ObjectType`/`FileType` part.

hainest (approved):

Issue #1071, opened 2021-07-01
heuristics fail to find main in stripped ppcle executable


stanfordcox:
The heuristics for the dyninst ppcle target fail in findMain to find main inside striped binaries.  This is the case for most any standard executable.

PR #1898, opened 2025-03-05
Add regression tests for image::findMain

hainest:
This is needed for upcoming changes to this function. In particular, there is no implementation for aarch64.

Requires #1897


kupsch (approved):

hainest:
`findMain` doesn't work at all on ppc. It seems this was a known issue (#1071) for stripped binaries. I'll close this and bundle it with the required changes to fix that.

PR #1904, opened 2025-03-06
Deprecate Object::getObjectType()

batuzovk:
ObjectType is internal to Object class. It can be different for different object file formats. Other code should be using accessors isExecutable(), isSharedLibrary(), etc.

### DIFF
--- a/symtabAPI/src/Object-elf.h
+++ b/symtabAPI/src/Object-elf.h
@@ -196,7 +196,6 @@ public:
      return (char *)mf->base_addr();
   }
 
-  DYNINST_EXPORT ObjectType objType() const override;
   const char *interpreter_name() const override;
 
 
@@ -286,6 +285,7 @@ public:
     DYNINST_EXPORT bool isOnlySharedLibrary() const override;
     DYNINST_EXPORT bool isDebugOnly() const override;
     DYNINST_EXPORT bool isLinuxKernelModule() const override;
+    DYNINST_EXPORT bool isRelocatableFile() const override;
 
     bool getSegments(std::vector<Segment> &segs) const override;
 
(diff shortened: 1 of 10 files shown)


batuzovk:
The semantic has changed slightly in a couple of places because `getObjectType() == obj_SharedLib` is not exactly the same as `isSharedLibrary()`, but it should be close enough.

symtabAPI/src/Object.h
```diff
@@ -171,6 +170,9 @@ class Object : public boost::basic_lockable_adapter<boost::mutex>
     DYNINST_EXPORT virtual bool isOnlySharedLibrary() const { return false; }
     DYNINST_EXPORT virtual bool isDebugOnly() const { return false; }
     DYNINST_EXPORT virtual bool isLinuxKernelModule() const { return false; }
+    DYNINST_EXPORT virtual bool isRelocatableFile() const { return false; }
+    DYNINST_EXPORT virtual bool isELF() const { return false; }
+    DYNINST_EXPORT virtual bool isPE() const { return false; }
```
hainest, line 175: I know it's a prevalent idiom in Dyninst, but base classes should never know about there possible derived classes. Let's add these to the interface:

```c++
enum class file_format {
  ELF,
  PE
};

DYNINST_EXPORT file_format getFileFormat() const { ... };
```

Right now, this is equivalent to what you have written. However, the file format doesn't have to match the derived object type. For example, we might want to have `file_format::CUDA_FATBIN` that is actually an ELF file.

symtabAPI/src/Object-elf.h
```diff
@@ -373,10 +374,19 @@ class ObjectELF final : public Object
   Dyninst::DwarfDyninst::DwarfHandle::ptr dwarf;
   private:
 
+  enum ObjectType {
+   obj_Unknown,
+   obj_SharedLib,
+   obj_Executable,
+   obj_RelocatableFile,
+  };
+
```
hainest, line 383: A minor item that we definitely don't have to fix now, but C-style enums with prefix notation are better expressed using `enum class`.

hainest (changes requested):

symtabAPI/src/Object.h
batuzovk: Oops, those two methods shouldn't be in this MR. Sorry.

I'll keep that in mind when updating https://github.com/dyninst/dyninst/pull/1896

symtabAPI/src/Object-elf.h
batuzovk: This isn't hard to do. We can as well change it now. I've updated the PR.

hainest:
Those `static_ctor_dtor` tests are sometimes spurious failures. I'm re-running that [job](https://github.com/dyninst/dyninst/actions/runs/13701456010) against HEAD. I'll post here with the results.

hainest:
Testing looks good on all machines. Thanks!

hainest (approved):

PR #1905, opened 2025-03-07
Clean up common/pathName

hainest:


common/src/pathName.C
```diff
@@ -1,326 +1,107 @@
...
-	#define S_ISDIR(x) ((x) & _S_IFDIR)
-#else
-	#include <sys/types.h>
-	#include <sys/stat.h>
-	#include <unistd.h>
-	#include <pwd.h>
-#endif
-
-std::string expand_tilde_pathname(const std::string &dir) {
-#ifdef os_windows
```
bbiiggppiigg, line 51: Shouldn't we keep it a no-op for windows?

hainest: I made a note about that (https://github.com/dyninst/dyninst/pull/1905/commits/141591b6f9199e53655c0ea3caa1674b830e3e5f). These functions are only ever used in linux/freebsd files right now. My goal is to move away from doing OS-specific code by guarding blocks with macros.

bbiiggppiigg (approved):

kupsch:
I think not using the password file is incorrect. Let's discuss this afternoon.

hainest:
> I think not using the password file is incorrect. Let's discuss this afternoon.

I'm free if/when you have time.

common/src/pathName.C
hainest: Now that I've added back the lookup in passwd, I've also needed to add back the windows check. It's at least at the file level instead of inside the functions.

kupsch (approved):
Looks good except an unknown user should return the tilde expression unchanged.

common/src/pathName.C
```diff
@@ -1,326 +1,169 @@
...
+    // If no name is given, use current effective user
+    if(username.empty()) {
+      getpwuid_r(geteuid(), &pwd, buf.data(), size, &result);
+    } else {
+      getpwnam_r(username.c_str(), &pwd, buf.data(), size, &result);
+    }
+
+    if(!result) {
+      // failed to find an entry
+      return {};
```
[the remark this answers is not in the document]
hainest: Good catch. Lots of corner cases here. I added the check in `expand_tilde` and had it return the complete path so that it matches the logic of the `~/x` case.

PR #1906, opened 2025-03-07
Rename Symtab::isRelocatableFile -> isIntermediateObject

hainest:
The idea of a relocatable file as only an intermediate, unlinked file
is a concept from ELF/Unix. For me, a relocatable file is any binary
that has relocations. This includes all formats Dyninst currently
supports except non-PIEs. It also appears that Mach-O files have
an intermediate form (see [`MH_OBJECT`](https://github.com/aidansteele/osx-abi-macho-file-format-reference)).

@batuzovk It seems like this is an acceptable indicator for PE .obj files. Thoughts?


bbiiggppiigg:
I don't know much about this topic, but I think maybe renaming isRelocatableObj to isRelocatableFile for the image Class would be more natural, as at the end of the day it is checking whether the ObjectType is RelocatableFile

rochauha:
I think `isRelocatableFile` is self explanatory.`isIntermediateObject` requires explanation/documentation.

Also, linked executables may also retain relocation records by setting appropriate flags (`--emit-relocs`) when building a project/tool. But these files are not intermediate files (they also aren't relocatable objects in the sense that static relocations have been resolved, but those relocation records are still in the binary)

hainest:
> I think `isRelocatableFile` is self explanatory.`isIntermediateObject` requires explanation/documentation.

For someone familiar with ELF, that makes sense. But what about PE .obj or Mach-O '.o'? Neither of them uses the 'relocatable file' phrasing. For Mach-O, they are called intermediate files. The research I did for PE just calls them objects- which is why I asked Kirill for feedback.

What bothers me is that `isSharedLibrary() == isRelocatableFile()` is *never* true. To which I ask, why isn't a shared library relocatable?! The answer, of course, is that the ELF nomenclature has leaked into symtab- which should be agnostic of the file type (cue discussion about the pointy corners of the Symtab interface).

I'm ok with spending a little bit of time bikeshedding on the name, if it helps to make things clearer.

> Also, linked executables may also retain relocation records by setting appropriate flags (`--emit-relocs`) when building a project/tool. But these files are not intermediate files (they also aren't relocatable objects in the sense that static relocations have been resolved, but those relocation records are still in the binary)

This is consistent with the current meaning. If an ELF file has `ET_EXEC`, then Dyninst treats it like an executable (i.e., it has type `ObjectType::Executable`, `Symtab::isExecutable() == true`, etc.). The only time we treat is as relocatable is when it's a PIE. I have a different fix coming for that.

batuzovk:
> It seems like this is an acceptable indicator for PE .obj files. Thoughts?

I like the new naming a lot better. It makes a lot more sense for people not familiar with ELF-specific terminology and communicates the intent a lot better.

It should work for PE `.obj` files as well, though I've never tried parsing those.

batuzovk:
> Also, linked executables may also retain relocation records by setting appropriate flags (--emit-relocs) when building a project/tool. But these files are not intermediate files (they also aren't relocatable objects in the sense that static relocations have been resolved, but those relocation records are still in the binary)

Those files won't count as `isIntemediateObject()` or `isRelocatableFile()`. Only those files with `e_type == ET_REL` count as "relocatable file" (which used to make sense when ELF was introduced, but not any more).

This is the reason I think this change is good. The intuitive meaning of the old name was different from the actual one.

PR #1909, opened 2025-03-11
Reduce usage of 'util.h' in headers

hainest:
This reduces transitive includes, making it easier to add tests.


bbiiggppiigg (approved):
I assume as long as the CI compiles then it should be good.

hainest:
> I assume as long as the CI compiles then it should be good.

There are a few architecture-specific files that I'm testing on their respective platforms. Thus far, I've not seen anything break.

kupsch (approved):

PR #1912, opened 2025-03-11
Move all stats declarations into 'stats.h'

hainest:
More cleanup found when adding unit tests. This can be merged once the CI passes.

### DIFF
--- a/dyninstAPI/src/debug.h
+++ b/dyninstAPI/src/debug.h
@@ -72,22 +72,12 @@ extern int dyn_debug_regalloc;
 extern int dyn_debug_ast;
 extern int dyn_debug_write;
 extern int dyn_debug_infmalloc;
-extern int dyn_stats_instru;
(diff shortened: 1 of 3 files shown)


dyninstAPI/src/debug.h
```diff
@@ -178,7 +168,4 @@ DECLARE_PRINTF_FUNC(stackmods_printf_int);
 // And initialization
 extern bool init_debug();
 
-extern bool init_stats();
-extern bool print_stats();
```
bbiiggppiigg, line 182: If you remove these declarations because they are unused, should the definitions in debug.C be removed as well?

hainest: Oops. That's a rebase error. I'll add them back. Although they aren't actually used, but that's for another PR.

bbiiggppiigg (approved):

PR #1915, opened 2025-03-12
GithubCI: enable unit, integration, and regression tests in PR workflow

hainest:


docker/build.sh
```diff
@@ -38,7 +38,7 @@ build_dir=$(mktemp -d "/tmp/XXXXXX")
 mkdir -p ${dest_dir}
 
 if test "${run_tests}" = "Y"; then
-  cmake_args="-DDYNINST_ENABLE_TESTS=ON ${cake_args}"
+  cmake_args="-DDYNINST_ENABLE_TESTS=ALL ${cake_args}"
```
kupsch, line 41: `cake_args` -> `cmake_args`

kupsch (approved):
looks good, but there was an error/typo in the existing code that you should fix.

docker/build.sh
hainest: Good catch. I guess I was thinking about cake...

hainest:
Working now https://github.com/dyninst/dyninst/actions/runs/13820798109.

PR #1918, opened 2025-03-12
Correctly test PIE binaries in regression tests

hainest:
This only happened to work on ubuntu because the gcc there accepts `-f[no-]PIE` as an equivalent to `-[no-]pie`. Fedora's gcc does not use that.

### DIFF
--- a/tests/regression/symtabAPI/Symtab/file_properties/CMakeLists.txt
+++ b/tests/regression/symtabAPI/Symtab/file_properties/CMakeLists.txt
@@ -7,6 +7,9 @@ if(POLICY CMP0083)
   cmake_policy(SET CMP0083 NEW)
 endif()
 
+include(CheckPIESupported)
+check_pie_supported()
+
 set(_all_tests "")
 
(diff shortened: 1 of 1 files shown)


tests/regression/symtabAPI/Symtab/file_properties/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ if(POLICY CMP0083)
   cmake_policy(SET CMP0083 NEW)
 endif()
 
+include(CheckPIESupported)
+check_pie_supported()
+
```
bbiiggppiigg, line 12: Are these function / checks builtin to cmake??
I can't find where there are defined in the dyninst repo.

hainest: It's provided by CMake. https://cmake.org/cmake/help/latest/module/CheckPIESupported.html

bbiiggppiigg (approved):

kupsch (approved):

PR #1921, opened 2025-03-13
Remove elfutils dependency from dyninstAPI

hainest:
It's not actually needed. Binary rewriting uses dynElf and dynDwarf, though.

### DIFF
--- a/dyninstAPI/src/image.C
+++ b/dyninstAPI/src/image.C
@@ -61,11 +61,6 @@
 #include <sys/time.h>
 #endif
 
-#if defined( cap_dwarf )
-#include "dwarf.h"
-#include "elfutils/libdw.h"
-#endif
-
 #if defined(_MSC_VER)
 #include <dbghelp.h>
 #endif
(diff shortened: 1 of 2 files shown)


dyninstAPI/src/image.C
```diff
@@ -61,11 +61,6 @@
 #include <sys/time.h>
 #endif
 
-#if defined( cap_dwarf )
```
bbiiggppiigg, line 64: In ./dyninstAPI/src/BPatch_collections.C there is also a #if defined(cap_dwarf)
I'm not sure if it is related, but it seems like the guard there can be removed as well.

bbiiggppiigg: based on my understanding of the logic of that piece of  code.

hainest: That check can probably be removed, but it's beyond the scope of what I want to change here.

bbiiggppiigg (approved):

PR #1923, opened 2025-03-13
Add 'std::string const& getName() const' to BPatch_module

hainest:
The old name interfaces are clumsy to use.

### DIFF
--- a/dyninstAPI/src/BPatch_module.C
+++ b/dyninstAPI/src/BPatch_module.C
@@ -96,6 +96,14 @@ BPatch_object *BPatch_module::getObject() {
    return img->findOrCreateObject(mod->obj());
 }
 
+std::string const& BPatch_module::getName() const {
+  static std::string const empty_string{};
+  if(!mod) {
+    return empty_string;
+  }
+  return mod->fileName();
+}
+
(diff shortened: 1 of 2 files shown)


dyninstAPI/src/BPatch_module.C
```diff
@@ -96,6 +96,14 @@ BPatch_object *BPatch_module::getObject() {
    return img->findOrCreateObject(mod->obj());
 }
 
+std::string const& BPatch_module::getName() const {
+  static std::string const empty_string{};
```
bbiiggppiigg, line 100: Would it be better to check for 
if (mod) return mod->fileName(); first?
It is probably more likely then the case of empty_string?

hainest: Guessing what the compiler will do here is a good way to make yourself go insane. The function-local static is required to be thread-safe, so the mutex check is going to dominate. On second thought, I'm just going to have it return by value. I deliberated that, but decided against it since we aren't using C++17 that guarantees copy elision here. For c++11, it will be a copy, but at least we'll get a free optimization once we move to c++17. I'll update it.

kupsch (approved):

PR #1926, opened 2025-03-14
Don't include platform-specific Object headers in Object.h

hainest:
This was causing unnecessary build failures.

### DIFF
--- a/symtabAPI/src/Archive-elf.C
+++ b/symtabAPI/src/Archive-elf.C
@@ -32,7 +32,8 @@
 
 #include "symtabAPI/h/Symtab.h"
 #include "symtabAPI/h/Archive.h"
-#include "symtabAPI/src/Object-elf.h"
+#include "Elf_X.h"
+#include "MappedFile.h"
 
 using namespace std;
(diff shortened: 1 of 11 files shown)


bbiiggppiigg:
> This was causing unnecessary build failures.

I don't understand. I suppose the current master should have no build failures in order to pass CI?

symtabAPI/src/Archive-elf.C
```diff
@@ -32,7 +32,7 @@
 
 #include "symtabAPI/h/Symtab.h"
 #include "symtabAPI/h/Archive.h"
-#include "symtabAPI/src/Object.h"
+#include "symtabAPI/src/Object-elf.h"
```
bbiiggppiigg, line 35: I don't see any use of the class Object-elf in this class.
Should it be Elf_X.h or maybe other relevant header like elf.h instead?

hainest: Agreed. I updated it to "Elf_X.h" and fixed another missing transitive include.

kupsch (approved):

PR #1928, opened 2025-03-17
Fix ROSE register size for ppc64 CR registers

hainest:
Dyninst treats them as the whole register, but ROSE splits them up.

The values come from `RegisterDictionary::dictionary_powerpc`.


bbiiggppiigg (approved):

bbiiggppiigg:
The reason why we need to specify the size here is because the register class does not encode the size information accurately I assume?

hainest:
> The reason why we need to specify the size here is because the register class does not encode the size information accurately I assume?

MachRegister reports sizes in bytes, but the cr<N> registers are only 4 bits so Dyninst just assumes that cr<N> aliases the whole cr.

PR #1929, opened 2025-03-17
Fix PowerPC branch instruction semantics

### LINKED ISSUE #1072: jump table analysis fails checking bclr opcode for ppcle so
For the dyninst ppcle target, jump table analysis during code parsing fails with:
Dyninst::DataflowAPI::RoseInsnPPCFactory::handleSpecialCases(entryID, SgAsmInstruction*, SgAsmOperandList*): Assertion `power_op_bclr == iapi_opcode' failed.

This patch bypasses the problem:
--- dyninst-11.0.0/dataflowAPI/src/RoseInsnFactory.C.orig	2021-06-09 15:54:21.753883619 -0400
+++ dyninst-11.0.0/dataflowAPI/src/RoseInsnFactory.C	2021-06-23 14:17:37.854933719 -0400
@@ -317,5 +317,5 @@
     // It looks like the ROSE semantics code will infer the target from 
     // the bo field. So, what is passed in as the third operands does not matter
-    if(branch_target) {
+    if(branch_target || iapi_opcode == power_op_bc) {
       rose_operands->append_operand(new SgAsmDoubleWordValueExpression(branch_target));
     } else if(power_op_bcctr == iapi_opcode) {

The problem is dependent on how libdw is built
 https://scox.fedorapeople.org/dyninst/libdw-0.185.ppc64le.so

hainest:
This should have been part of 64499aa304.

Fixes #1072

### DIFF
--- a/instructionAPI/src/InstructionDecoder-power.C
+++ b/instructionAPI/src/InstructionDecoder-power.C
@@ -370,12 +370,82 @@ which are both 0).
     }
 
     if(current->op == power_op_bcctr) {
-      insn_in_progress->addSuccessor(makeRegisterExpression(ppc32::ctr), field<31, 31>(insn) == 1,
-                                     true, bcIsConditional, false);
-      if(bcIsConditional) {
-        insn_in_progress->addSuccessor(makeFallThroughExpr(), false, false, false, true);
+      /*********************************************************************
+       *  Power ISA Version 3.1C, Book I. May 2024
+       *  2.4 Branch Instructions
+       *
+       *    bcctr   BO,BI,BH (LK=0)
+       *    bcctrl  BO,BI,BH (LK=1)
+       *
+       *    Extended form examples:
+       *      blctr
+       *      bnectr cr0
+       *
+       *  Programming Note (pg 38)
+       *  ------------------------
+       *
+       *  1. Computed goto’s, case statements, etc.
+       *
+       *     Put the branch address in CTR and set LK=0 (BH=0b11, if appropriate)
+       *
+       *  2. Indirect subroutine linkage
+       *
+       *     Place the final target address in CTR and use bcctr (LK=0).
+       *
+       *     NOTE: This form is not meant to fall through (e.g., it's used in PLT
+       *     thunks), so usually also has BH=0b1z1zz as per Figure 2.5 to indicate
+       *     "branch always".
+       *
+       *  3. Function call
+       *
+       *     For direct calls, put the target address in CTR and use bcctrl (LK=1).
+       ****************************************************************************/
+      bool const LK = field<31, 31>(insn);
+
+      auto add_lr_operand = [this, LK](bool const is_read) {
+        auto link_register = makeRegisterExpression(ppc32::lr);
+        constexpr bool is_implicit = true;
+        bool const is_written = (LK == 1);
+        insn_in_progress->appendOperand(link_register, is_read, is_written, is_implicit);
+      };
+
+      if(LK == 1) {
+        // Case 3 - Function call
+        auto ctr = makeRegisterExpression(ppc32::ctr);
+        constexpr bool is_indirect = true;
+        constexpr bool is_call = true;
+        bool const is_fallthrough = false;
+        constexpr bool is_conditional = false;
+        insn_in_progress->addSuccessor(std::move(ctr), is_call, is_indirect, is_conditional, is_fallthrough);
+        add_lr_operand(true);
+      } else {
+        bool const always_branch = [this]() {
+          auto const bo_field = field<6,10>(insn);
+          auto const mask = 0x14;  // bits 0,1,3 are ignored (0b10100)
+          return (bo_field & mask) == mask;
+        }();
+
+        if(always_branch) {
+          // Case 2 - Indirect subroutine linkage
+          auto ctr = makeRegisterExpression(ppc32::ctr);
+          constexpr bool is_indirect = true;
+          constexpr bool is_call = false;
+          bool const is_fallthrough = false;
+          constexpr bool is_conditional = false;
+          insn_in_progress->addSuccessor(std::move(ctr), is_call, is_indirect, is_conditional, is_fallthrough);
+          add_lr_operand(false);
+        } else {
+          // Case 1 - Computed goto’s, case statements, etc.
+          auto ctr = makeRegisterExpression(ppc32::ctr);
+          constexpr bool is_indirect = true;
+          constexpr bool is_call = false;
+          bool const is_fallthrough = true;
+          constexpr bool is_conditional = true;
+          insn_in_progress->addSuccessor(std::move(ctr), is_call, is_indirect, is_conditional, is_fallthrough);
+        }
       }
     }
+
     if(current->op == power_op_addic || current->op == power_op_andi_rc ||
        current->op == power_op_andis_rc || current->op == power_op_stwcx_rc ||
        current->op == power_op_stdcx_rc) {
--- a/instructionAPI/src/InstructionDecoder-power.C
+++ b/instructionAPI/src/InstructionDecoder-power.C
@@ -1336,14 +1336,6 @@ which are both 0).
       return false;
     }();
 
-    if(writes_ctr) {
-      if(field<9, 9>(insn)) {
-        insn_in_progress->getOperation().mnemonic = "bdz";
-      } else {
-        insn_in_progress->getOperation().mnemonic = "bdn";
-      }
-    }
-
     // The count register (ctr) is conditionally read and written
     {
       auto ctr = makeRegisterExpression(makePowerRegID(ppc32::ctr, 0));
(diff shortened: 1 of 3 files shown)


bbiiggppiigg (approved):

instructionAPI/src/InstructionDecoder-power.C
```diff
@@ -305,7 +305,13 @@ which are both 0).
       auto target = makeRegisterExpression(ppc32::lr);
       insn_in_progress->addSuccessor(std::move(target), false, true, bcIsConditional, false);
 
-      if(bcIsConditional) {
+      bool const is_subroutine_return = [this]() {
```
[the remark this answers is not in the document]
hainest: I prefer self-documenting code when I can. Also, I made a couple of mistakes when I did the short version, so it was better for me, as well.

instructionAPI/src/InstructionDecoder-power.C
```diff
@@ -302,11 +302,29 @@ which are both 0).
...
+      }();
+
+      bool const is_conditional = !is_subroutine_return;
+
+      {
+        auto target = makeRegisterExpression(ppc32::lr);
+        constexpr bool is_call = false;
+        constexpr bool is_indirect = true;
+        constexpr bool is_fallthrough = false;
+        insn_in_progress->addSuccessor(std::move(target), is_call, is_indirect, is_conditional,
```
kupsch, line 317: Is `std::move` useful in this case?

kupsch (approved):

instructionAPI/src/InstructionDecoder-power.C
hainest: Strictly, it's a micro-optimization, but it saves a single ref update in the shared_ptr.

instructionAPI/src/InstructionDecoder-power.C
```diff
@@ -294,28 +294,159 @@ which are both 0).
...
+          constexpr bool is_fallthrough = false;
+          insn_in_progress->addSuccessor(lr, is_call, is_indirect, is_conditional, is_fallthrough);
+        }
+      }
 
       if(bcIsConditional) {
-        insn_in_progress->addSuccessor(makeFallThroughExpr(), false, false, false, true);
+        auto target = makeFallThroughExpr();
+        constexpr bool is_call = false;
+        constexpr bool is_indirect = true;
```
bbiiggppiigg, line 365: I assume fallthrough should be indirect as in the original code?

parseAPI/src/IA_power.C
```diff
@@ -316,12 +316,18 @@ bool IA_power::isReturn(Dyninst::ParseAPI::Function * context, Dyninst::ParseAPI
...
-  parsing_printf
-    ("isblrReturn at 0x%lx Addr 0x%lx 0x%lx Function addr 0x%lx leaf %d \n",
-     current, getAddr (), currBlk->start (), context->addr (),
-     func->_is_leaf_function);
-  if (!func->_is_leaf_function)
+
+  parsing_printf("Checking return idioms for %s@0x%lx in block=0x%lx, context=0x%lx\n",
+                 context->name().c_str(), getAddr(), currBlk->start(), context->addr());
+
+  if(func->_is_leaf_function) {
```
bbiiggppiigg, line 325: You added a check for the case of leaf function and return early, but it seems like you didnt remove the original code in the else clause of if(_is_leaf_function) that does the same thing?

And maybe we should stick to just one of func or context for consistency (even though they seem to have the same value as func is assigned in line 319).

instructionAPI/src/InstructionDecoder-power.C
```diff
@@ -1165,23 +1296,31 @@ which are both 0).
...
+
     if(!(field<6, 6>(insn))) {
       invertBranchCondition = !field<7, 7>(insn);
       if(insn_in_progress->getOperation().mnemonic == "bc") {
         insn_in_progress->getOperation().mnemonic = "b";
       }
       insn_in_progress->appendOperand(makeBIExpr(), true, false);
     }
-    if(field<8, 8>(insn) && field<6, 6>(insn)) {
+    if(writes_ctr && field<6, 6>(insn)) {
```
bbiiggppiigg, line 1323: writes_ctr checks for field<8,8>(insn) ==0, so here it should be (!writes_ctr) I assume?

instructionAPI/src/InstructionDecoder-power.C
hainest: It's unclear what "indirect" means here. Except for printing, Dyninst only uses that information in only one place: https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/Relocation/Widgets/CFWidget.C#L150. 

The comment there is

```
   // Indirect branch:
   //   1) Just go for it... we have no control, really
```

Super helpful. Given this, I interpret an "indirect" control flow target to be contained in some location that must be dereferenced before being used (e.g., branching through a register) as opposed to being a literal offset or absolute target. Hence, fallthrough targets are always indirect since they go through the PC (or the LR for ppc).

parseAPI/src/IA_power.C
hainest: This is still a draft, so that code is actually going away. I wish GitHub would let you remove a review request, but it won't. Sorry.

instructionAPI/src/InstructionDecoder-power.C
bbiiggppiigg: I thiink fallthrough targets are always direct snice they are literally the next instruction and does not need to go through the process of resolving indirect jump target.

hainest: My biggest problem is that `InstructionDecoder_power::makeFallThroughExpr` assumes that the fallthrough target can be addressed via the PC, but Power doesn't have one[1]. Since it's not direct, then I considered it indirect. But a re-read of the decode algorithms for conditional branch instructions doesn't actually indicate that. I'll change this back to being direct to make it consistent with the other architectures. As an aside, I plan to change the  CFT interface when updating to Capstone to make all of this more explicit.

[1] From the docs:
```
Obtaining the address of the next instruction:
Use the following form of Branch and Link.

    bcl 20,31,$+4
```
where `$` is an assembler directive.

hainest:
With https://github.com/dyninst/testsuite/pull/256, this reduces the failures from ~99% to <1%.

parseAPI/src/IA_power.C
```diff
@@ -309,91 +309,9 @@ bool IA_power::isReturnAddrSave(Address& retAddr) const
   return ret;
 }
 
-bool IA_power::isReturn(Dyninst::ParseAPI::Function * context, Dyninst::ParseAPI::Block* currBlk) const
+bool IA_power::isReturn(Dyninst::ParseAPI::Function *, Dyninst::ParseAPI::Block*) const
 {
-  /* Check for leaf node or lw - mflr - blr pattern */
```
bbiiggppiigg, line 314: Is this pattern no longer being used in powerpc so we don't need to check it anymore?

hainest: Correct. This is from ancient POWER and hasn't been used since v2.0 (~Power7) which is the oldest we support.

instructionAPI/src/InstructionDecoder-power.C
```diff
@@ -294,28 +294,158 @@ which are both 0).
...
+        bool const is_written = (LK == 1);
+        insn_in_progress->appendOperand(link_register, is_read, is_written, is_implicit);
+      };
+
+      if(LK == 1) {
+        // Case 3 - Function call
+        auto ctr = makeRegisterExpression(ppc32::ctr);
+        constexpr bool is_indirect = true;
+        constexpr bool is_call = true;
+        bool const is_fallthrough = false;
```
bbiiggppiigg, line 417: Any reason why this is not constexpr like the others?
And just a thought, would it be more clear if we define is_indirect/is_call/is_fallthrough/is_conditoinal all as true the start of the function, and then do something like appendOperand(..., is_call, is_indirect, !is_conditiona, !is_fallthrough);

hainest: > Any reason why this is not constexpr like the others?

Nope, that was just a mistake.

> And just a thought, would it be more clear if we define is_indirect/is_call/is_fallthrough/is_conditoinal all as true the start of the function, and then do something like appendOperand(..., is_call, is_indirect, !is_conditiona, !is_fallthrough);

Good idea! I should do that since I did that in the tests.

hainest: Updated.

kupsch:
@hainest I see the same test suite problem of it not running most of the tests.  18 tests run and PASS and then it stops.  It looks like ptrace is being restricted due to an update or a CSL configuration change. I found this in `stderr.log`:

```
Warning: your Linux system provides limited ptrace functionality as a security
measure. This measure prevents ProcControl and Dyninst from attaching to binaries.
To temporarily disable this measure (until a reboot), execute the following command:
        echo 0 > /proc/sys/kernel/yama/ptrace_scope
To permanently disable this measure, edit the file "/etc/sysctl.d/10-ptrace.conf"
and follow the directions in that file.
For more information, see https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening
```

I'll open an issue with the CSL.

hainest:
Thanks for the feedback, all.

@kupsch I just tried to run the test suite on follis, but it still failed early. I'm going to go ahead and merge this since I've not seen any regressions on the other x86 machines.

kupsch:
@hainest, I too only see 18 tests run and PASSed. The ptrace setting was made, and I can attach to another process with gdb `gdb -p ...` so the setting seems to be active.  The ptrace failure message is no longer present in the testsuite logs, so something else must be going wrong.

hainest:
@kupsch Have you tried running without this commit?

kupsch:
WIth the previous commit (fb151...), 18 run with 2 failures (power_cft and power_decode).  I'll try going back further.

kupsch:
@hainest I tried instances of dyninst and the testsuite from many months ago, where there were 1552 PASSED tests and 0 FAILED on x86, and I still see 18 PASSED and 0 FAILED. I'll investigate further next week.

PR #1931, opened 2025-03-19
Clean up and refactor string handling in parts of SymtabAPI and DyninstAPI

hainest:
I was chasing down an unrelated bug and kept running into corrupted `std::string`s- probably from a life-extended temporary getting modified. This fixes that issue.

### DIFF
--- a/dyninstAPI/src/BPatch_binaryEdit.C
+++ b/dyninstAPI/src/BPatch_binaryEdit.C
@@ -87,7 +87,7 @@ BPatch_binaryEdit::BPatch_binaryEdit(const char *path, bool openDependencies) :
 
   startup_printf("[%s:%d] - Opening original file %s\n",
                  FILE__, __LINE__, path);
-  origBinEdit = BinaryEdit::openFile(std::string(path));
+  origBinEdit = BinaryEdit::openFile(path ? path : "");
 
   if (!origBinEdit){
      startup_printf("[%s:%d] - Creation error opening %s\n",
--- a/dyninstAPI/src/image.C
+++ b/dyninstAPI/src/image.C
@@ -1005,13 +1005,12 @@ pdmodule *image::findModule(const string &name, bool wildcard)
       //
       dyn_hash_map <string, pdmodule *>::iterator mi;
       string str; pdmodule *mod;
-      std::string pds = name.c_str();
 
       for(mi = modsByFileName.begin(); mi != modsByFileName.end() ; mi++)
       {
          str = mi->first;
          mod = mi->second;
-         if (wildcardEquiv(pds, mod->fileName())) {
+         if (wildcardEquiv(name, mod->fileName())) {
             found = mod; 
             break;
          }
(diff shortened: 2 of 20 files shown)


hainest:

dyninstAPI/src/image.C
```diff
@@ -1613,7 +1612,7 @@ void pdmodule::dumpMangled(std::string &prefix) const
       parse_func * pdf = (parse_func*)*fit;
       if (pdf->pdmod() != this) continue;
 
-      if( ! strncmp( pdf->symTabName().c_str(), prefix.c_str(), strlen( prefix.c_str() ) ) ) {
+      if(pdf->symTabName() != prefix) {
```
bbiiggppiigg, line 1615: It seems like this line is comparing to the prefix length instead of full comparison.
Would this be an issue?

dyninstAPI/src/BPatch_binaryEdit.C
```diff
@@ -87,7 +87,7 @@ BPatch_binaryEdit::BPatch_binaryEdit(const char *path, bool openDependencies) :
 
   startup_printf("[%s:%d] - Opening original file %s\n",
                  FILE__, __LINE__, path);
-  origBinEdit = BinaryEdit::openFile(std::string(path));
+  origBinEdit = BinaryEdit::openFile(path ? path : "");
```
bbiiggppiigg, line 90: Should we do the 
if(path == nullptr) return 
or something like the next commit?

dyninstAPI/src/image.C
hainest: Derp. You are quite right. I don't know how I didn't see that. The correct substitution would be `if(pdf->symTabName().find(prefix) != 0UL)` which is effective [`std::string::starts_with`](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) from C++20.

dyninstAPI/src/BPatch_binaryEdit.C
hainest: >Should we do the
>if(path == nullptr) return

Since the next like checks the success of `openFile` _and_ `openFile` will fail for an invalid path, then this change at least avoids throwing an exception when `path` is null.

>or something like the next commit?

The early return from `BPatch::openBinary` is fine because it's a (sort-of) factory function, so errors can be expected. A constructor should never leave an object in a partially-constructed state. If I had the time, I'd rip out all of that dangerous nonsense in these BPatch classes.

bbiiggppiigg (approved):
LGTM

hainest:
> LGTM

Btw, I hope you don't mind me asking you to do reviews. I greatly appreciate your careful analysis, but don't feel like you must do them. I don't want to detract from your work too much.

bbiiggppiigg:
> > LGTM
> 
> Btw, I hope you don't mind me asking you to do reviews. I greatly appreciate your careful analysis, but don't feel like you must do them. I don't want to detract from your work too much.

Its fine. I'm learning C++ in this process as well.

PR #1935, opened 2025-03-20
Parse symbols in PE files

batuzovk:


hainest (approved):
If this is working for you, then we can merge.

batuzovk:
Yes, this change works for me. The easiest way to test this is to parse some .dll 
(e.g. [libmpv-2.dll.gz](https://github.com/user-attachments/files/19387648/libmpv-2.dll.gz) from mpv media player) and see that two things happened.

* The number of discovered functions increased substantially compared to the revision without this patch.
* There are named functions exactly where symbols from export directory are. To independently read export directory `readpe` tool can be used. The final address can be calculated from its output as `ImageBase + SymbolAddress`.

PS: there is a recurring pattern in `.dll` files where a function jumps to a trampoline at the end of the `.text` section. This kind of breaks current implementation of `disassemble` from the examples repository. I should probably fix that to make testing of PE patches easier.

hainest (approved):

hainest:
> Yes, this change works for me. The easiest way to test this is to parse some .dll (e.g. [libmpv-2.dll.gz](https://github.com/user-attachments/files/19387648/libmpv-2.dll.gz) from mpv media player) and see that two things happened.

I need to update the pr testing workflow to including building with PE support. I'll add this as a test case.

PR #1937, opened 2025-03-21
Handle relocations in PE files

batuzovk:


batuzovk:
I decided to put those two patches in one review because the second one helps to test the first one.

Ways to test things really work. For input we need 32-bit PE file (those have a lot of relocations, e.g. 
[libass-9.dll.gz](https://github.com/user-attachments/files/19390351/libass-9.dll.gz))

* Disassemble to the text file.
* Rebase and disassemble again. Addresses should shift, immediate values in the instructions should change. Everything else should be unchanged.
* Rebase back to `preferedBase_` and disassemble again. The text file should be identical to the first one.

symtabAPI/src/Object-pe.C
```diff
@@ -366,3 +382,82 @@ int ObjectPE::getArchWidth() const
...
+            writeRelocTarget(ptr, peparse::RELOC_HIGH, val);
+            writeRelocTarget((uint8_t *)ptr + 2, peparse::RELOC_LOW, val);
+            break;
+        case peparse::RELOC_DIR64:
+            *(uint64_t *)ptr = (uint64_t)val;
+            break;
+        case peparse::RELOC_MIPS_JMPADDR:
+        case peparse::RELOC_MIPS_JMPADDR16:
+        default:
+            std::cout << "Unsupported relocation in PE file\n";
```
hainest, line 433: same as above.

symtabAPI/src/Object-pe.C
```diff
@@ -366,3 +382,82 @@ int ObjectPE::getArchWidth() const
...
+    for (Region *r : regions_) {
+        for (relocationEntry &rel : r->getRelocations()) {
+            if ((peparse::reloc_type)rel.getRelType() == peparse::RELOC_ABSOLUTE) {
+                continue;
+            }
+            uint8_t *ptr = (uint8_t *)r->getPtrToRawData() + ((Offset)rel.rel_addr() - (Offset)r->getMemOffset());
+            Offset val = readRelocTarget(ptr, (peparse::reloc_type)rel.getRelType());
+            assert(rel.target_addr() == val);
+            writeRelocTarget(ptr, (peparse::reloc_type)rel.getRelType(), val + delta);
+            assert(val + delta == readRelocTarget(ptr, (peparse::reloc_type)rel.getRelType()));
```
hainest, line 450: Does `readRelocTarget` have side effects? If so, it should be called outside of the assert.

hainest (changes requested):

symtabAPI/src/Object-pe.C
batuzovk: No, no side effects.

symtabAPI/src/Object-pe.C
```diff
@@ -366,3 +382,82 @@ int ObjectPE::getArchWidth() const
...
+        case peparse::RELOC_HIGHLOW:
+            return *(const uint32_t *)ptr;
+        case peparse::RELOC_HIGHADJ:
+            return readRelocTarget(ptr, peparse::RELOC_HIGH) | readRelocTarget((const uint8_t *)ptr + 2, peparse::RELOC_LOW);
+        case peparse::RELOC_DIR64:
+            return *(const uint64_t *)ptr;
+        case peparse::RELOC_MIPS_JMPADDR:
+        case peparse::RELOC_MIPS_JMPADDR16:
+        default:
+            std::cerr << "Unsupported relocation in PE file";
```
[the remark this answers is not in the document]
hainest: That's probably the better one to use.

hainest (approved):

PR #1940, opened 2025-03-24
Remove unnecessary copy of 'allowedEdges' in parseAPI::BuildEdgesAux

batuzovk:


batuzovk:
Usually it doesn't matter because slices are short and sets are small. But occasionally it can spin out of control resulting in hundreds percent increase in parse time.

We've observed it on `libLLVM-14.so.1`, but in conjunction with other changes. Basic `parse()` doesn't seem to be affected.

hainest (approved):

PR #1941, opened 2025-03-26
Track unreliable jump table bounds

cheremnov:
When analyzing a jump table, the parser determines the bounds optimistically. When it can't determine bounds, it arbitrary declares there are 256 entires; the parser limits bounds when it found the bad entry. This means we can end up with jump tables which have potentially incorrect parameters.

Having any, even optimistically determined, boundaries is beneficial for analysis. But, when it comes to binary rewriting, having potentially incorrect parameters given to the user in the jump table info, may lead to hidden pitfalls.

Say, the user is transforming binary code via DynInst. They receive the info: there are only 256 entries. They transform those 256 entries but all entries beyond this remain unprocessed.
If a jump leads to one of these discarded entries, the target will be determined incorrectly and the binary won't work.

Let's extend the API for jump tables and track the cases when the bounds are unreliable. We introduce an additional flag for jump tables which tells the user whether they can rely upon the boundaries.

### DIFF
--- a/parseAPI/h/CFG.h
+++ b/parseAPI/h/CFG.h
@@ -402,6 +402,8 @@ class DYNINST_EXPORT Function : public AnnotatableSparse, public boost::lockable
         int indexStride;
         int memoryReadSize;
         bool isZeroExtend;
+        // If there's something wrong with targets
+        bool areBoundsUnreliable;
         std::map<Address, Address> tableEntryMap;
         Block* block;
     };
--- a/parseAPI/src/IndirectAnalyzer.C
+++ b/parseAPI/src/IndirectAnalyzer.C
@@ -154,6 +154,7 @@ bool IndirectControlFlowAnalyzer::NewJumpTableAnalysis(std::vector<std::pair< Ad
         }
     }
 
+    bool is_unbounded = false;
     StridedInterval b;
     if (!variableArguFormat) {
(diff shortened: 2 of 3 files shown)


hainest:
Thanks for the PR. It's not immediately clear to me how this would be useful to *users*. If the table is unparseable, then there's nothing they can do to change how it gets rewritten. It would be very helpful if you could provide an example case where these changes enabled you to do better analysis. Attaching a binary demonstrating the issue would be even better.

I'll have a deeper think about this issue, and get back with you.

batuzovk:
Presence of incompletely parsed jump tables is an indicator by itself. It means that the function is better not to be changed. It is similar to how LLVM Bolt handles things. They have `isSimple()` method for functions. Anything with partially unresolved CFG is marked as not simple and is ignored during optimization process.

Another example would be static binary translation. Completely parsed jump table with all their targets known can be translated to host jump tables. For all other indirect edges dynamic dispatch has to be performed, and landing pads have to be prepared (because you can't enter binary translated code at a random unprepared spot).

All of the above can be achieved by examining indirect edges themselves. So, another possible solution would be to drop unreliable jump tables after parsing is complete and possibly add an indirect edge to sink node. But this would require larger code changes and will introduce a change in what Dyninst considers a jump table (from "something with the known format" to "something with the known format and bounds").

Internally partially parsed jump tables are useful to Dyninst because they help to discover code.

PR #1942, opened 2025-03-26
Fix copying missing constant instruction literal

bbiiggppiigg:
AMDGPU instructions could use a 32 bit instruction literall following the current instruction.
However, whether a instruction uses a literal can only be determined after parsing the operands.
Previous approach crafts a instruction first and then update the bytes if a literal is beng used, but the 
copy of raw bytes is copied only during the construction of the instruction.

This PR updated the implementation of updateSize to copy the literal bytes.

### DIFF
--- a/instructionAPI/h/Instruction.h
+++ b/instructionAPI/h/Instruction.h
@@ -168,7 +168,7 @@ namespace Dyninst { namespace InstructionAPI {
     typedef boost::shared_ptr<Instruction> Ptr;
 
   private:
-    void updateSize(const unsigned int new_size) { m_size = new_size; }
+    void updateSize(const unsigned int new_size, const unsigned char * raw);
 
     void decodeOperands() const;
     void addSuccessor(Expression::Ptr e, bool isCall, bool isIndirect, bool isConditional, bool isFallthrough,
--- a/instructionAPI/h/Instruction.h
+++ b/instructionAPI/h/Instruction.h
@@ -181,7 +181,7 @@ namespace Dyninst { namespace InstructionAPI {
     mutable Operation m_InsnOp;
     bool m_Valid;
     raw_insn_T m_RawInsn;
-    unsigned int m_size;
+    unsigned int m_size{};
     Architecture arch_decoded_from;
     mutable std::list<CFT> m_Successors;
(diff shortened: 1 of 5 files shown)


kupsch (approved):
Looks OK.  You should move the buffer management from `updateSize` to `copyRaw` as in the comment.

hainest:
I really don't like this business of fiddling with the guts of constructed objects. If these changes help you get your work done for now, that's fine. Just make sure we leave these changes undocumented so we can update them with impunity later. As an aside, it seems like the GPU decoders are really using the `Instruction` to just hold some state as it does its work. It might be better to split those into member variables and then produce a single instruction at the end. That's a change for later.

bbiiggppiigg:
> I really don't like this business of fiddling with the guts of constructed objects. If these changes help you get your work done for now, that's fine. Just make sure we leave these changes undocumented so we can update them with impunity later. As an aside, it seems like the GPU decoders are really using the `Instruction` to just hold some state as it does its work. It might be better to split those into member variables and then produce a single instruction at the end. That's a change for later.

The whole decoding process (for aarch64 and power as well) is filling up this insn_in_progress object:
Adding operands, adding successors (and other control flow information), updating mnemonic.
So I don't see why updating rawbytes associated with the object is any different.

hainest:
> > I really don't like this business of fiddling with the guts of constructed objects. If these changes help you get your work done for now, that's fine. Just make sure we leave these changes undocumented so we can update them with impunity later. As an aside, it seems like the GPU decoders are really using the `Instruction` to just hold some state as it does its work. It might be better to split those into member variables and then produce a single instruction at the end. That's a change for later.
> 
> The whole decoding process (for aarch64 and power as well) is filling up this insn_in_progress object: Adding operands, adding successors (and other control flow information), updating mnemonic. So I don't see why updating rawbytes associated with the object is any different.

It's not any different. It's just as wrong as all the other classes in Dyninst with member functions poking at the guts of objects. It violates class invariants and makes code difficult to reason about. This is particularly true over in the DWARF parser where this stuff is done concurrently. Uf dah. Your extension here is fine since it helps you get your work done. It would be good to eventually move away from these idioms, but I doubt that will happen until there is more than one non-grad-student working on Dyninst.

PR #1943, opened 2025-03-27
Fix possible use of uninitialized value in Result::operator<

ilil01:
When Result object is not defined, its `type` field may be uninitialized or contain a meaningless value. Thus check the `defined` field first.
Also the case `type > o.type` was missing, which may lead to an incorrect comparison.

The issue was highlighted by Valgrind.


hainest:

hainest (approved):

PR #1944, opened 2025-03-27
add lock_guard on Block accesses in slicing

ilil01:
Slicing and parsing phases of analysis may overlap. Since they share structures like Block objects, and parsing have write accesses to these structures, all slicing accesses to the same structures should be preceded with a `lock_guard` to ensure no iterator gets invalidated while being used.

This issue caused a rare spurious segmentation fault during analysis of some binaries.

### DIFF
--- a/dataflowAPI/src/slicing.C
+++ b/dataflowAPI/src/slicing.C
@@ -677,6 +677,7 @@ Slicer::getSuccessors(
         // will not transform any of them (besides changing
         // the location)
 
+        boost::lock_guard<Block> lg(*cand.loc.block);
         const Block::edgelist & targets = cand.loc.block->targets();
         Block::edgelist::const_iterator eit = targets.begin();
         for( ; eit != targets.end(); ++eit) {
@@ -846,6 +847,8 @@ Slicer::handleCall(
     ParseAPI::Edge * funlink = NULL;
     bool widen = false;
 
+    {
+        boost::lock_guard<Block> lg(*cur.loc.block);
     const Block::edgelist & targets = cur.loc.block->targets();
     Block::edgelist::const_iterator eit = targets.begin();
     for( ; eit != targets.end(); ++eit) {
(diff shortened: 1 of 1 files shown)


hainest:
Thanks for looking into this. Unfortunately, there's no way to guarantee this solution won't deadlock because it's possible to slice into the same block in the same thread (shared blocks are fragile in the parser right now). That aside, it is a deeper problem that slicing is being requested before parsing is complete because it's not possible to do accurate slicing until after parsing has finished.

Do you have a reproducer we could examine?

ilil01:
> Do you have a reproducer we could examine?

I'm afraid I don't have a reproducer that can be examined. The issue initially was discovered while testing a static analyzer on 32-bit glibc. The analyzer segfaulted 2-3 times per 200 runs, but only on one of the PCs used for testing. Still I'm sure the problem lies there since I was able to catch the segfault in GDB and it pointed to the iterator use in `EndsWithConditionalJump`.

> there's no way to guarantee this solution won't deadlock because it's possible to slice into the same block in the same thread

The `Block` class represents a recursive mutex due to inheriting from `boost::lockable_adapter<boost::recursive_mutex>`, so a thread cannot deadlock with itself over a block. As for deadlocks between different slicing threads, it's impossible since a thread can hold no more than one lock at the same time.

> it is a deeper problem that slicing is being requested before parsing is complete because it's not possible to do accurate slicing until after parsing has finished

Slicing is used for jump table analysis, which is required in order to finish the parsing process. The parsing process cannot be finished without slicing, and the slicing process cannot be done correctly before the parsing is finished. This "the chicken and the egg" problem doesn't have an easy solution.

ilil01:
By the way, `getSuccessors` and `getPredecessors` use different locking schemes. I don't know which one is better.

hainest:
> If there are still any concerns with the approach I will be glad to answer the questions and update the PR. If there are none, may I ask for a review?

The only actual Dyninst developer is on medical leave right now. I'd like to confer with him before making any parser changes, particularly without a detailed root cause analysis.

PR #1946, opened 2025-04-02
GitHubCI: fix libabigail workflow

hainest:


hainest:
This is working with the latest sources. For some reason, building with clang is not ABI-compatible with Dyninst built with gcc. I don't think that's a compatibility we should strive for, unless someone disagrees.

kupsch (approved):

PR #1947, opened 2025-04-02
GitHubCI: add pe-parse

hainest:


hainest:
This won't build on fedora because the build-fedora action needs the changes here. I've tested it locally using Fedora 41, and it's working fine.

kupsch (approved):

PR #1951, opened 2025-04-07
GitHubCI: add /usr/lib to LD_LIBRARY_PATH

hainest:
This is needed on Fedora where, for a yet-unknown reason, the RUNPATH does not include /usr/lib even though

1) DYNINST_FORCE_RUNPATH=ON

2) `ldd libsymtabAPI.so` clearly shows that pe-parse.so resolves to /usr/lib/pe-parse.so.

Even more vexxing is that it works correctly when the idential container and commands are run locally (i.e., not in GitHub).

### DIFF
--- a/.github/workflows/pr-tests.yaml
+++ b/.github/workflows/pr-tests.yaml
@@ -135,6 +135,7 @@ jobs:
         shell: bash
         run: |
           set -ex
+          export LD_LIBRARY_PATH=/usr/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


hainest:
@bbiiggppiigg @kupsch This is finally working under CI. It's infuriating that it works fine when running the containers locally using the identical steps. This is needed before I can merge any other PRs. If either of you has a moment, it would be really appreciated!

bbiiggppiigg (approved):

kupsch (approved):

PR #1952, opened 2025-04-07
GitHubCI: don't build extrae on ubuntu-24.10

hainest:
There's a bug in their config I haven't been able to fix, so just exclude it for now.


hainest:
This is working (https://github.com/dyninst/dyninst/actions/runs/14317796787). There is an unrelated failure with systemtap- it seems the sourceforge git server is down at the moment.

kupsch (approved):

PR #1957, opened 2025-04-11
Refactor InstructionAST

hainest:
As part of the move to Capstone, I want to replace `Instruction::get{Read,Write}Set(std::set<RegisterAST::Ptr>&)` with something like `std::vector<RegisterAST::Ptr> get{Read,Written}Registers()`. This is because 1) it makes no sense to have an ordering on `RegisterAST`s and 2) in-out parameters are terrible for usability at callsites. I'd like to remove them and the ordering on the ASTs, but it's all in the documentation and would likely break somebody's world. I'll deprecate them for now.

To accomplish this, I need to update the interfaces throughout the AST hierarchy. While I'm at it, I'm also doing some simple refactoring to modernize the code as I go. This is the first step (of many) to do that. I'll make a separate PR for deprecating the `get{Uses,Children}` interfaces after this is merged.

Requires #1956


bbiiggppiigg (approved):

kupsch (approved):

hainest:
The libabigail checks are failing due to additional inlining of member functions. I hadn't considered that. I've added the appropriate label to the PR.

PR #1959, opened 2025-04-15
Make Expression the base of the InstructionAPI AST hierarchy

hainest:
The InstructionAST abstraction doesn't provide any functionality beyond
its immediate child, Expression. This is particularly evident by

1) all uses of InstructionAST in Dyninst are immediately converted to
   an Expression

2) the lackluster description provided in the docs:

  The InstructionAST class is the base class for all nodes in the ASTs
  used by the Operand class. It defines the necessary interfaces for
  traversing and searching an abstract syntax tree representing an
  operand.

But the real problem is with 'getChildren'. Because both

InstructionAST::getChildren(std::vector<InstructionAST::Ptr> const&)=0

and

Expression::getChildren(std::vector<Expression::Ptr> const&)=0

exist, then every AST type has to implement both which is redundant and
provides no logical separation between an InstructionAST and an
Expression. In C++, these are referred to as overloaded virtual
functions and are generally a sign of an error since the one from the
base class will be hidden (see -Woverloaded-virtual). The author
"overcame" this by making them pure virtual. That doesn't remove the
logical incorrectness, though.

InstructionAST should have implemented 'getChildren' (or even better,
'getChildNodes') and Expression should have implemented
'getSubexpressions'. Then all of this would have been separated in a
useful way.

This merger is an ABI break only. No user code changes will be needed.

---

This PR _does_ incidentally add the requirements to implement

```c++
void getUses(std::set<Expression::Ptr> &uses)
bool isUsed(Expression::Ptr findMe)
std::string format(Dyninst::Architecture, formatStyle = defaultStyle)
std::string format(formatStyle = defaultStyle)
void getChildren(std::vector<Expression::Ptr>& children)
bool isStrictEqual(const Expression &rhs)
```

These requirements will be removed in a sequence of upcoming PRs.


hainest:

kupsch (approved):

PR #1963, opened 2025-06-22
adjust handling for indirect jump analysis for AMD GPUs

jmellorcrummey:
Don't assert or SEGV during indirect jump analysis if analysis can't proceed because an AST node is NULL.

If analysis yields a NULL AST node, analysis is simply over.


jmellorcrummey:
@bbiiggppiigg These were the changes needed to analyze a LLNL AMD GPU binary without dying.

bbiiggppiigg (approved):
LGTM.

PR #1965, opened 2025-06-26
CMake: update detection of static libraries

hainest:
Since we require CMake >= v3.14.0, then we can use CMAKE_REQUIRED_LINK_OPTIONS. Update names to follow standard conventions.

### DIFF
--- a/cmake/Modules/DyninstLinkProperties.cmake
+++ b/cmake/Modules/DyninstLinkProperties.cmake
@@ -1,15 +1,29 @@
 include_guard(GLOBAL)
 
-include(CheckCSourceCompiles)
-include(CheckCXXSourceCompiles)
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
 
 # Check if the CMAKE_<LANG>_COMPILER supports static linking
-set(_CMAKE_REQUIRED_LIBRARIES_BAK ${CMAKE_REQUIRED_LIBRARIES})
-set(CMAKE_REQUIRED_LIBRARIES "-static")
 
-set(_src_code "int main(){ return 0; }")
+# gcc and clang both use '-static', but allow for other compilers
+set(_possible_static_flags "-static")
 
-check_c_source_compiles("${_src_code}" DYNINST_C_STATIC_LINK)
-check_cxx_source_compiles("${_src_code}" DYNINST_CXX_STATIC_LINK)
+foreach(_flag ${_possible_static_flags})
+  set(_saved_link_opts ${CMAKE_REQUIRED_LINK_OPTIONS})
 
-set(CMAKE_REQUIRED_LIBRARIES ${_CMAKE_REQUIRED_LIBRARIES_BAK})
+  set(CMAKE_REQUIRED_LINK_OPTIONS "${_flag}")
+
+  check_c_compiler_flag("${_flag}" DYNINST_C_HAVE_STATIC_LINK_FLAG)
+
+  if(DYNINST_C_HAVE_STATIC_LINK_FLAG AND NOT DYNINST_C_STATIC_LINK_FLAG)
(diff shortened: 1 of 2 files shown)


cmake/Modules/DyninstCheckStaticExe.cmake
```diff
@@ -0,0 +1,37 @@
...
+foreach(_flag ${_possible_static_flags})
+  set(CMAKE_REQUIRED_LINK_OPTIONS "${_flag}")
+
+  foreach(_lang C CXX)
+    set(_have_flag_name DYNINST_${_lang}_HAVE_STATIC_EXE_LINK_FLAG)
+    set(_flag_name DYNINST_${_lang}_STATIC_EXE_LINK_FLAG)
+
+    dyninst_check_compiler_flag(${_lang} ${_flag} ${_have_flag_name})
+
+    if(${_have_flag_name} AND NOT ${_flag_name})
```
bbiiggppiigg: I really have a hard time understanding the logic here.
If dyninst_check_compiler_flag passes for a language, ${_have_flag_name} will be set to true.
Why is ${_flag_name} relevant here?

bbiiggppiigg: _flag_name was set to DYNINST_${_lang}_STATIC_EXE_LINK_FLAG. But when the _flag_name is false, it is set to ${_flag} which is one of the items of the ${_possible_static_flags}?

hainest: >I really have a hard time understanding the logic here.
>If dyninst_check_compiler_flag passes for a language, ${_have_flag_name} will be set to true.
>Why is ${_flag_name} relevant here?

Because `add_executable(foo STATIC foo.c)` isn't legal CMake, we have to track the flags needed to statically link executables and then manually add them to a target using `target_compiler_options`. By convention, CMake's `check_LANG_compiler_flag` accepts a `HAVE_FLAG_FOO` as an output argument (of course, it can really be anything). I kept that convention here while also keeping track of the actual value of the flag. For gcc/clang, we'd have `DYNINST_CXX_HAVE_STATIC_EXE_LINK_FLAG=TRUE` and `DYNINST_CXX_STATIC_EXE_LINK_FLAG=-static`.

>flag_name was set to DYNINST${_lang}_STATIC_EXE_LINK_FLAG.
>But when the _flag_name is false, it is set to ${_flag} which is one of the items of the ${_possible_static_flags}?

Ah, yes, sorry. I meant to add some comments on this. I wanted to save only the first flag that is usable. In the current case, this simply isn't used because we only check for `-static`.

---

I agree this is a bit convoluted. Let me refactor it. What do you think of the following user experience? It's comparable to how [`check_pie_supported`](https://cmake.org/cmake/help/latest/module/CheckPIESupported.html) works with the added need to manually track the flag.

```CMake
include(DyninstCheckStaticExe)
dyninst_check_static_exe_supported(LANGUAGES C)

add_executable(foo foo.c)
if(DYNINST_C_STATIC_EXE_LINK_SUPPORTED)
  target_compile_options(foo PRIVATE ${DYNINST_C_STATIC_EXE_LINK_FLAG})
  target_link_libraries(foo PRIVATE ${DYNINST_C_STATIC_EXE_LINK_FLAG})
endif()
```

bbiiggppiigg: The user experience looks good to me.
The main issue i had i guess is understanding the cmake logic that was written.

hainest: Updated. Please re-check.

> The user experience looks good to me. The main issue i had i guess is understanding the cmake logic that was written.

The check is now `if(${_flag_name})` which is effectively `if(DYNINST_${_lang}_STATIC_EXE_LINK_FLAG NOT STREQUAL "")`. Does that clarify?

bbiiggppiigg (approved):
LGTM.

hainest:

PR #1966, opened 2025-07-12
Add RISC-V Specific Elf Section Parsing

wxrdnx:
I've added code that parses the `.riscv_attribute` section.
This section contains RISC-V sepcific meta information like which extensions the binary support, stack alignment size, and whether it allows unaligned memory access.

### DIFF
--- a/symtabAPI/src/Object-elf.C
+++ b/symtabAPI/src/Object-elf.C
@@ -4327,7 +4327,7 @@ bool ObjectELF::parse_attrs(const char *attr_string,
                 return false;
             }
             int attr_end = curr + attr_size - 1;
-            curr += sizeof(int);
+            curr += sizeof(attr_size);
 
             // Parse all attribute data in the attribute
             while (curr < attr_end) {


wxrdnx:
I added the `ABI-BREAKER` tag to prevent the libabigail test case from running

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4203,167 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+            uint64_t tag = *p++;
+            uint32_t size;
+            memcpy(&size, p, 4);
+
+            if (size > attr_len) {
+                create_printf("%s[%d]:  Bad subsection length (%u > %u)\n", FILE__, __LINE__, size, attr_len);
+                return false;
+            }
+            if (size < 6) {
+                create_printf("%s[%d]:  Bad subsection length (%u < 6)\n", FILE__, __LINE__, size);
```
kupsch: missing `return false;`?

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4203,167 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+    }
+    return true;
+}
+
+void ObjectELF::get_riscv_extensions() {
+    // Obtain information from .riscv.attributes
+    if (!riscv_attrs.empty() && riscv_attrs.find(Tag_RISCV_arch) != riscv_attrs.end()) {
+        std::string arch_string = riscv_attrs[Tag_RISCV_arch].sval;
+
+        std::regex ext_regex(R"(([a-z0-9]+)(\d+)p(\d+))");
```
wxrdnx: @hainest 
Do you think using regex to parse the RISC-V extension string is a good idea?
Or would a manual parsing approach be better?

wxrdnx:
@hainest @kupsch 
I have removed the regex and created a struct for RISC-V attributes
I think this PR can be reviewed. Thanks.

symtabAPI/src/Object-elf.C
```diff
@@ -52,6 +52,8 @@
 
 #include "Object-elf.h"
 
+#include <regex>
```
bbiiggppiigg: If regex is no longer used, consider remove this?

symtabAPI/src/Object.C
```diff
@@ -297,16 +297,16 @@ DYNINST_EXPORT Object::~Object()
...
-   data_ptr_(0), data_off_(0), data_len_(0),
-   code_vldS_(0), code_vldE_(0),
-   data_vldS_(0), data_vldE_(0),
-   loader_off_(0), loader_len_(0),
-   is_aout_(false), is_dynamic_(false),
-   has_error(false), is_static_binary_(false),
-   no_of_sections_(0), no_of_symbols_(0),
-  deferredParse(false), parsedAllLineInfo(false), err_func_(err_func), addressWidth_nbytes(4),
-  associated_symtab(st), file_format_(FileFormat::Unknown)
+    code_ptr_(0), code_off_(0), code_len_(0),
```
bbiiggppiigg: Revert white space only change

symtabAPI/src/Object.h
```diff
@@ -41,11 +41,13 @@
  * header files.
 ************************************************************************/
 
+
 // trace data streams
 #include <iosfwd>
 #include <utility>
 #include <string>
 #include <vector>
+#include <unordered_set>
```
bbiiggppiigg: why is this included?

symtabAPI/src/Object-elf.C
wxrdnx: Ok, removed

symtabAPI/src/Object.C
wxrdnx: Ok, reverted

symtabAPI/src/Object.h
wxrdnx: Ok, removed

hainest:
Has this been rebased to include #2007? If so, would it make sense to add it to the list of codegen architectures (https://github.com/dyninst/dyninst/blob/master/.github/workflows/codegen-cross-build.yaml#L65)?

wxrdnx:
> Has this been rebased to include #2007? If so, would it make sense to add it to the list of codegen architectures (https://github.com/dyninst/dyninst/blob/master/.github/workflows/codegen-cross-build.yaml#L65)?

@hainest 
It makes sense to add RISC-V to the list of codegen architectures.
However, Dyninst currently does not compile for RISC-V codegen because some code (mainly `#ifdef`s) has not yet been merged into the master branch.
I can try creating another PR to make the compilation work

hainest:
> > Has this been rebased to include #2007? If so, would it make sense to add it to the list of codegen architectures (https://github.com/dyninst/dyninst/blob/master/.github/workflows/codegen-cross-build.yaml#L65)?
> 
> @hainest It makes sense to add RISC-V to the list of codegen architectures. However, Dyninst currently does not compile for RISC-V codegen because some code (mainly `#ifdef`s) has not yet been merged into the master branch. I can try creating another PR to make the compilation work

I think taking an incremental approach like what is happening with the amdgpu work is best. Do you have a branch ready for doing code generation?

wxrdnx:
> > > Has this been rebased to include #2007? If so, would it make sense to add it to the list of codegen architectures (https://github.com/dyninst/dyninst/blob/master/.github/workflows/codegen-cross-build.yaml#L65)?
> > 
> > 
> > @hainest It makes sense to add RISC-V to the list of codegen architectures. However, Dyninst currently does not compile for RISC-V codegen because some code (mainly `#ifdef`s) has not yet been merged into the master branch. I can try creating another PR to make the compilation work
> 
> I think taking an incremental approach like what is happening with the amdgpu work is best. Do you have a branch ready for doing code generation?

@hainest 
I don't have a separate branch for codegen, but it is in the `angushe/riscv` branch (the main RISC-V branch).
My current thought is that I can create another PR containing the backbone necessary for compilation. Does that work for you?

symtabAPI/src/Object-elf.C
```diff
@@ -835,6 +837,8 @@ bool ObjectELF::loaded_elf(Offset &txtaddr, Offset &dataddr,
             hasModinfo_ = true;
         } else if (strcmp(name, GNU_LINKONCE_THIS_MODULE_NAME) == 0) {
             hasGnuLinkonceThisModule_ = true;
+        } else if (strcmp(name, RISCV_ATTRIBUTES) == 0) {
```
kupsch: Instead of the section name, you should prefer to use the program header type `PT_RISCV_ATTRIBUTES` (0x70000003) as this should always be present, while section names can be stripped. You could keep the section as a backup, if the program header isn't available. If so you should compare the section's headers type to `SHT_RISCV_ATTRIBUTES` (0x70000003), instead of the name.

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +91,28 @@ class Object;
...
+        Tag_RISCV_arch = 5,
+        Tag_RISCV_unaligned_access = 6,
+        Tag_RISCV_priv_spec = 8,
+        Tag_RISCV_spec_minor = 10,
+        Tag_RISCV_priv_spec_revision = 12,
+        Tag_RISCV_atomic_abi = 14,
+        Tag_RISCV_x3_reg_usage = 16
+    };
+    // RISC-V Attributes
+    std::map<int, riscv_attr_t> raw_attrs;
```
kupsch: Add member for each tag with a type `std::optional<std::string>` or `std::optional<uint32_t>`. Get rid of raw_attrs

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +91,28 @@ class Object;
...
+    // Supported RISC-V extensions
+    std::unordered_set<std::string> extensions;
```
kupsch: make this a map from extension name to a `std::pair<uint32_t>` containing the major and minor versions.

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +91,28 @@ class Object;
 class ObjectELF;
 class InlinedFunction;
 
+// RISC-V attributes
+struct RiscvAttributes {
```
kupsch, line 124: add a member for each of the riscv attributes in the e_flags, most will booleans, except the floatAbi will need to be an `enum class RiscvFloatAbiEnum { soft, single, double, quad}`.

symtabAPI/src/Object-elf.h
```diff
@@ -483,6 +506,12 @@ class ObjectELF final : public Object
   void parse_dynamicSymbols( Elf_X_Shdr *& dyn_scnp, Elf_X_Data &symdata,
                              Elf_X_Data &strdata, bool shared_library);
 
+  bool parse_riscv_attributes(Elf_X_Shdr *);
+
+  void get_riscv_extensions();
+
+  bool getUseRVC() override { return riscv_attrs.extensions.find("c") != riscv_attrs.extensions.end(); }
```
kupsch: A better name would be `riscvUsesRVC` instead of `getUseRVC`.  The implementation should `return riscv_attrs.rvc;`.

symtabAPI/src/Object-elf.h
```diff
@@ -517,7 +546,10 @@ class ObjectELF final : public Object
   Function* containingFunc;
   std::unordered_map<void*, std::vector<open_statement> > contextMap;
 
-        };
+  // RISC-V Attributes
```
kupsch: uesless comment

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4201,167 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+                }
+            }
+        }
+    }
+
+    // Obtain information from e_flags
+    unsigned long e_flags = elfHdr->e_flags();
+    if (e_flags & EF_RISCV_RVC) {
+        riscv_attrs.extensions.insert("c");
+    }
```
kupsch, line 4465: I don't think that there is any need for this and it can be deleted.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4201,167 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+                    end++;
+                    std::string ext = arch_string.substr(start, end - start);
+                    riscv_attrs.extensions.insert(ext);
+                }
+            }
+        }
+    }
+
+    // Obtain information from e_flags
+    unsigned long e_flags = elfHdr->e_flags();
```
kupsch, line 4439: assign all the RISC-V fields in flags to the riscv_attrs.

symtabAPI/src/Object-elf.C
```diff
@@ -2296,7 +2310,8 @@ ObjectELF::ObjectELF(MappedFile *mf_, bool, void (*err_func)(const char *),
         obj_type_(ObjectType::Unknown),
         DbgSectionMapSorted(false),
         soname_(NULL),
-        containingFunc(nullptr)
+        containingFunc(nullptr),
+        riscv_attrs()
```
kupsch: No need to default initialize a class/struct as the language does this if not present, the only time you need to explicitly call a default constructor/initializer is for scalar types. If the member is being initialized to a constant it is better to use default member initialization.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4201,167 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
    return dwarf->frameParser()->getRegValueAtFrame(pc, reg, reg_result, reader, frame_error);
 
 }
+
+bool ObjectELF::parse_riscv_attributes(Elf_X_Shdr *riscv_attr_scnp) {
```
kupsch: We should talk about this function.

symtabAPI/src/relocationEntry-elf-riscv64.C
```diff
@@ -0,0 +1,142 @@
...
+	// Some Platforms does not have R_RISCV_PLT32, R_RISCV_SET_ULEB128, and/or R_RISCV_SUB_ULEB128
+#ifdef R_RISCV_PLT32
+        CASE_RETURN_STR(R_RISCV_PLT32);
+#endif
+#ifdef R_RISCV_SET_ULEB128
+        CASE_RETURN_STR(R_RISCV_SET_ULEB128);
+#endif
+#ifdef R_RISCV_SUB_ULEB128
+        CASE_RETURN_STR(R_RISCV_SUB_ULEB128);
+#endif
```
kupsch: It would be better to handle this adding the following to an appropriate .C or .h file:

```
// define RISCV macros if missing in <elf.h> 
#ifndef R_RISCV_PLT32
    #define R_RISCV_PLT32 the_value_from_the_abi
#endif
...
```
Then the code will handle these relocation types everywhere.

symtabAPI/src/Object-elf.C
wxrdnx: Yes, a `return false;` is missing. Added.

symtabAPI/src/Object-elf.C
wxrdnx: Got it. Deleted

symtabAPI/src/Object-elf.C
wxrdnx: Got it.
Removed the default initialization of `riscv_attrs`.

symtabAPI/src/Object-elf.h
wxrdnx: Removed.

symtabAPI/src/Object-elf.h
wxrdnx: Okay. Changed the function name and implementation.

symtabAPI/src/Object-elf.C
wxrdnx: Done

symtabAPI/src/Object-elf.h
wxrdnx: I cannot create an enum entry called `double` because `double` is a reserved word.
So I use all caps: `enum class RiscvFloatAbiEnum { soft, single, double, quad}`

symtabAPI/src/relocationEntry-elf-riscv64.C
wxrdnx: They are now moved to `relocationEntry.h`

symtabAPI/src/Object-elf.h
wxrdnx: Done

symtabAPI/src/Object-elf.h
wxrdnx: Done.
The new structure:
```c
struct RiscvAttributes {                                        
    bool rvc;                                                   
    bool rve;                                                   
    bool tso;                                                   
    RiscvFloatAbiEnum floatABI;                                 
    std::map<std::string, std::pair<int, int>> riscv_extensions;
    std::string riscv_attr_string;                              
    std::optional<int64_t> stack_align;                         
    std::optional<bool> unaligned_access;                       
    std::optional<int64_t> priv_spec;          // Deprecated    
    std::optional<int64_t> priv_spec_minor;    // Deprecated    
    std::optional<int64_t> priv_spec_revision; // Deprecated    
    std::optional<int64_t> atomic_abi;                          
    std::optional<int64_t> x3_reg_usage;                        
};                                                              

```

symtabAPI/src/Object-elf.C
wxrdnx: Done

wxrdnx:
@kupsch @hainest 
I rewrote the parsing of RISC-V `.riscv.attribute` and the extension string (in function `parse_riscv_attrs` and `get_riscv_extensions` respectively) using C++ strings.
Could you please review the code and share your feedback?
Thanks

wxrdnx:
@kupsch 
`std::optional` is available in c++17 but not available in c++11.
Should I implement `struct RiscvAttribute` without optional, or make Dyninst support c++17?

kupsch:
use boost::optional for now.

symtabAPI/src/Object-elf.h
kupsch: ok

wxrdnx:
> use boost::optional for now.

Okay. Changed to`boost::optional`.

symtabAPI/h/relocationEntry.h
```diff
@@ -39,6 +39,20 @@
 #include "Annotatable.h"
 #include "Region.h"
 
+// The minimum version of libelf Dyninst supports (0.173) does not define:
```
hainest, line 42: Dyninst [requires](https://github.com/dyninst/dyninst/blob/v13.0.0/cmake/tpls/DyninstElfUtils.cmake#L29) 0.186 or newer. We could increase that (0.194 is coming out next month), as long as it wouldn't need to be a really recent version. Do you happen to know when those constants were added?

symtabAPI/src/emitElfStatic-riscv64.C
```diff
@@ -0,0 +1,213 @@
...
+    return false;
+}
+
+inline
+static bool adjustValInRegion(Region *reg, Offset offInReg, Offset addressWidth, int adjust) {
+    Offset newValue;
+    unsigned char *oldValues;
+
+    oldValues = reinterpret_cast<unsigned char *>(reg->getPtrToRawData());
+    memcpy(&newValue, &oldValues[offInReg], addressWidth);
```
hainest, line 184: This assumes an `Offset` can hold `addressWidth` bytes. That seems like a weird constraint, but is probably one of those implicit, unsubstantiated assumptions Dyninst uses all the time.

symtabAPI/src/emitElfStatic-riscv64.C
```diff
@@ -0,0 +1,213 @@
...
+    Region *targetRegion = obj->findEnclosingRegion(rel.rel_addr());
+    if (targetRegion == NULL) {
+        rewrite_printf("Failed to find enclosing Region for relocation");
+        return false;
+    }
+    unsigned addressWidth = obj->getAddressWidth();
+    switch (rel.getRelType()) {
+        case R_RISCV_RELATIVE:
+        case R_RISCV_IRELATIVE:
+            rel.setAddend(rel.addend() + library_adjust);
```
hainest, line 199: If `R_RISCV_RELATIVE` corresponds to a `DT_REL` entry, then this will create an addend (with value `library_adjust`) for an entry that logically doesn't have one. Does riscv use `DT_RELA`?

symtabAPI/src/Object-elf.C
```diff
@@ -417,6 +418,11 @@ bool ObjectELF::loaded_elf(Offset &txtaddr, Offset &dataddr,
             foundInterp = true;
         } else if (elfPhdr.p_type() == PT_LOAD) {
             hasProgramLoad_ = true;
+        } else if (elfPhdr.p_type() == PT_RISCV_ATTRIBUTES) {
+            size_t len;
+            riscv_attrs.riscv_attr_string = std::string(
+                    (const char *)elfHdr->e_rawfile(len) + elfPhdr.p_offset(),
```
hainest: Is the attribute string guaranteed to exist? I ask because constructing a `std::string` from a null pointer will throw an exception.

symtabAPI/src/Object-elf.C
```diff
@@ -835,6 +841,10 @@ bool ObjectELF::loaded_elf(Offset &txtaddr, Offset &dataddr,
             hasModinfo_ = true;
         } else if (strcmp(name, GNU_LINKONCE_THIS_MODULE_NAME) == 0) {
             hasGnuLinkonceThisModule_ = true;
+        } else if (scn.sh_type() == SHT_RISCV_ATTRIBUTES) {
+            riscv_attrs.riscv_attr_string = std::string(
+                    (const char *)scnp->get_data().d_buf(),
+                    scnp->get_data().d_size());
```
hainest: Same as above.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1576,94 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+
+            // The attribute data contains a series of <tag, value> pair.
+            // The tag is a uleb128 integer indication the attribute tag. The
+            // value is the attribute value corresponding to the tag. The RISC-V
+            // ABI defines that if the tag is odd, the attribute value will be
+            // an integer. Otherwise, it will be a string. 
+
+            // See https://github.com/bminor/binutils-gdb/blob/master/binutils/readelf.c
+            std::string attr_section_name = "riscv";
+            auto handle_riscv_attr = [this](std::string &attr, int curr) -> int {
```
hainest: A lambda this big should probably be refactored to a free function.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1576,94 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+                unsigned long long attr_tag = 0;
+                unsigned int attr_shift = 0;
+                while (1) {
```
hainest: An infinite loop w/o side effects is undefined behavior (UB) in c++.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1576,94 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+                        break;
+                    attr_shift += 7;
+                }
+
+                // RISC-V attributes have a string value if the tag number is odd
+                // and an integer value if the tag number is even
+                if (attr_tag % 2 != 0) {
+                    // a string value
+                    unsigned slen = attr.find('\0', curr) - curr;
+                    std::string sval = attr.substr(curr, slen);
```
hainest: These two lines make me uneasy. Why are we searching for a null terminator in a `std::string`? Using embedded nulls is tolerable for C-style strings, but not `std::string`. If there are embedded nulls, then they should be parsed as a collection of strings before being passed to this lambda.

Also, `find` returns a `std::string::size_type` which is usually a `size_t`. Since `curr` is signed, you get mixed-sign arithmetic which is a great way to go to bad places. I just had a horror show of that in the arm staticElf stuff last week. Dyninst is _terrible_ about keeping signed and unsigned types separate.

I would suggest trying to see if something in [Boost.Algorithm](https://www.boost.org/doc/libs/1_64_0/doc/html/string_algo/reference.html#header.boost.algorithm.string.find_hpp) would make this safer/easier.

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +93,50 @@ class Object;
...
+#ifndef EF_RISCV_RVE
+#define EF_RISCV_RVE 0x0008
+#endif
+
+#ifndef EF_RISCV_TSO
+#define EF_RISCV_TSO 0x0010
+#endif
+
+// RISC-V attributes
+enum RiscvAttrTags {
```
hainest: Just a couple style notes:

1. Make it a scoped enum
2. Rename `RiscvAttrTags` -> `RiscvAttrTag`
3. `RiscvAttrTags::Tag_RISCV_stack_align` is better written `RiscvAttrTag::stack_align`. No need to repeat the type in the enumerator's name.

Personally, I would put this in a namespace to have something like `riscv::AttributeTag::stack_align`, but I know the other architectures aren't set up that way. Feel free to leave these as they are. These are just my thoughts on coding style.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1576,94 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+                            break;
+                    }
+                }
+                else {
+                    // an integer value
+
+                    // The integer values are supposed to be ULEB128 encoded
+                    unsigned long long ival = 0;
+                    unsigned int ival_shift = 0;
+                    while (1) {
```
hainest: Same UB as above.

symtabAPI/src/Object-elf.h
```diff
@@ -483,6 +529,12 @@ class ObjectELF final : public Object
   void parse_dynamicSymbols( Elf_X_Shdr *& dyn_scnp, Elf_X_Data &symdata,
                              Elf_X_Data &strdata, bool shared_library);
 
+  bool parse_riscv_attrs(std::string &, std::string &, std::function<int(std::string &, int)>);
```
hainest: In/out parameters (especially stringy data) hurt my soul. Could this not just return the parsed attribute?

symtabAPI/src/Object.h
```diff
@@ -178,6 +178,8 @@ class Object : public boost::basic_lockable_adapter<boost::mutex>
     virtual Offset getTOCoffset(Offset) const { return 0; }
     virtual void setTOCoffset(Offset) { }
 
+    virtual bool riscvUsesRVC() { return false; }
```
hainest: Having architecture-specific interfaces at the `ObjectElf` level I can mostly understand, but I'm curious why this is at the `Object` level? This is supposed to be a generic interface into a file blob. Admittedly, though, it is very centered on elf/dwarf.

symtabAPI/src/relocationEntry-elf-riscv64.C
```diff
@@ -0,0 +1,135 @@
...
+    }
+    if (sym->getType() == Symbol::ST_FUNCTION) {
+        return R_RISCV_JUMP_SLOT;
+    }
+    else {
+        return R_RISCV_64;
+    }
+
+	//should never return this.
+    return relocationEntry::dynrel;
```
hainest: This will run afoul of clang's [warning](https://clang.llvm.org/docs/DiagnosticsReference.html#wunreachable-code) for unreachable code. gcc has one, as well.

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +93,50 @@ class Object;
...
+    Tag_RISCV_priv_spec_minor = 10,
+    Tag_RISCV_priv_spec_revision = 12,
+    Tag_RISCV_atomic_abi = 14,
+    Tag_RISCV_x3_reg_usage = 16
+};
+enum class RiscvFloatAbiEnum { SOFT, SINGLE, DOUBLE, QUAD };
+struct RiscvAttributes {
+    bool rvc;
+    bool rve;
+    bool tso;
```
hainest: Truly, the hardest part of all of programming is naming things. Could these get more descriptive names? If they are canonical names in the docs could you add a quick comment describing each one?

symtabAPI/src/Object-elf.C
wxrdnx: No, the attribute does not always exists.
But I'm not sure what this has to do with constructing a `std::string` from a null pointer. If `PT_RISCV_ATTRIBUTES` does not exists, it won't even enter the `else if ` statement. Also, `elfHdr->e_rawfile(len)` is the raw pointer of the ELF image file, handled by libelf, which is not null.

hainest: > No, the attribute does not always exists.

I know. That's why I asked if the _string_ always exists. In other words, could there be the case that `elfPhdr.p_type() == PT_RISCV_ATTRIBUTES` is true, but the string isn't actually there?

symtabAPI/src/Object-elf.h
wxrdnx: The parsed attribute is handled in the lambda function (third argument). I discussed with Jim on Tuesday, and he suggested this function to be architecture agnostic. So I think we need at least three parameters for this function: the attribute string itself, the name of the attribute (which corresponds to the attribute we want to parse), and a function that parses the attribute data,  which is architecture dependent, and assign the parsed to the corresponding structure in ObjectELF). Since the parsed data is architecture dependent, I don't think returning parsed attribute is a good idea.

(By the way, since this function is architecture agnostic, I forgot to rename it to `parse_riscv_attrs` instead of `parse_attrs`).

symtabAPI/src/Object.h
wxrdnx: Other RISC-V object formats besides ELF (e.g., RISC-V PE) might also need to know whether compressed instructions are supported. Although Dyninst does not currently support PE, it still makes sense conceptually for this information to be in Object.

symtabAPI/src/relocationEntry-elf-riscv64.C
wxrdnx: Removed.

symtabAPI/src/Object-elf.h
wxrdnx: How about
```
struct RiscvAttributes {
    bool compressed_extension;
    bool embedded_abi;
    bool total_store_ordering;
```

symtabAPI/src/Object-elf.C
wxrdnx: Oh, I got it.
I think there is a possibility that `elfPhdr.p_type() == PT_RISCV_ATTRIBUTES` is true, but `elfPhdr.p_offset()` is junk, which makes `elfHdr->e_rawfile(len) + elfPhdr.p_offset()` point to junk. `elfHdr->e_rawfile(len)` I think will never be null because it is the pointer to the raw ELF image.

symtabAPI/src/emitElfStatic-riscv64.C
wxrdnx: `R_RISCV_RELATIVE` corresponds to `RELA` relocations, not `REL`.

symtabAPI/src/Object-elf.h
hainest: Much better. Thanks! :+1:

symtabAPI/src/Object.h
hainest: Ah, ok. Let's rename it to something like `usesCompressedInstructionFormat`?

>  RISC-V PE
> Although Dyninst does not currently support PE

Let's not think about that...

symtabAPI/src/emitElfStatic-riscv64.C
wxrdnx: I believe so

symtabAPI/src/emitElfStatic-riscv64.C
hainest: That seems to be more the default these days, but I wanted to double-check.

symtabAPI/src/Object-elf.C
hainest: Yikes. There's not much we can do about that.

symtabAPI/h/relocationEntry.h
wxrdnx: Just checked, 0.186 also does not have these three macros defined. It appears starting from 0.190

hainest: Hmm. 0.190 was released in 2023. That might be "too new" for some folks. @kupsch What do you think about bumping the min version?

symtabAPI/src/Object-elf.h
wxrdnx: Done

symtabAPI/h/relocationEntry.h
kupsch: Probably too new.  I think defining the couple of macros if not defined is OK until we do update. It probably wouldn't hurt to put a comment in `tpls/DyninstElfUtils.cmake` (above the minimum version): `delete the RISCV elf macros if version upgraded >= 0.190`

symtabAPI/src/Object-elf.C
wxrdnx: There is a free function `read_uleb128` in `Object-elf.C`.
I think I can call that function directly

symtabAPI/src/Object-elf.C
wxrdnx: Ok, I'll change back to C-style string

symtabAPI/h/relocationEntry.h
hainest: >Probably too new. I think defining the couple of macros if not defined is OK until we do update.

That works for me.

> It probably wouldn't hurt to put a comment in tpls/DyninstElfUtils.cmake (above the minimum version): delete the RISCV elf macros if version upgraded >= 0.190

@wxrdnx Could you add this?

symtabAPI/src/Object-elf.C
hainest: Well, I wouldn't necessarily go directly to that. Does the structure of the attribute string actually use or allow for embedded nulls? For example, if it doesn't then you can construct the string directly without providing a size.

```c++
std::string("abc\0def\0") == "abc";  // true
std::string("abc\0def\0", 8) == "abc";  // false
std::string("abc\0def\0", 8).size() == 8UL; // true
```

Does that make sense?

symtabAPI/src/Object-elf.C
wxrdnx: I think it's better to make it a member function because the function will assign attributes to an `ObjectELF` instance.
Perhaps something like
```
std::function<int(const char *)> handle_riscv_attr_f =
    std::bind(&ObjectELF::handle_riscv_attr, this,
            std::placeholders::_1);
```
Where `handle_riscv_attr` is a function member of ObjectELF
```
int ObjectELF::handle_riscv_attr(const char *attr_str) {
...
}
```

symtabAPI/h/relocationEntry.h
wxrdnx: Done

symtabAPI/src/Object-elf.C
wxrdnx: `std::string` does allow for embedded nulls, so that's why I construct the string by providing the size.

I do agree that if we are dealing with embedded nulls, it's better to use C-style strings, so for parsing, I changed the implementation of attribute parsing back to using C-style strings. However, ~for the `riscv_attrs` structure, I think storing the attribute string as `std::string` is better.~

symtabAPI/src/Object.h
wxrdnx: OK. Renamed to `usesCompressedInstructionFormat`

symtabAPI/src/Object-elf.h
wxrdnx: Makes sense.
Renamed `RiscvAttrTags` to `RiscvAttrTag`, as well as changing `Tag_RISCV_stack_align` to `stack_align`.

symtabAPI/src/Object-elf.C
wxrdnx: Same, used the free function `read_uleb128` in Object-elf.C

symtabAPI/src/Object-elf.C
```diff
@@ -417,6 +418,11 @@ bool ObjectELF::loaded_elf(Offset &txtaddr, Offset &dataddr,
             foundInterp = true;
         } else if (elfPhdr.p_type() == PT_LOAD) {
             hasProgramLoad_ = true;
+        } else if (elfPhdr.p_type() == PT_RISCV_ATTRIBUTES) {
+            size_t len;
+            riscv_attrs.riscv_attr_string = std::string(
```
kupsch: This is not good. You're assigning `riscv_attrs.riscv_attr_string` to the entire ELF attribute data structure here. Then later you you replace this value with the contents of the `RiscvAttrTag::arch` tag's value (this is what the value should be).  Make two new member that are the pointer to the beginning of the data and it length (just like the other `*_addr_`, and `*_size_` members) instead of making a string and overloading this member.

wxrdnx: @kupsch
Done. Added `riscv_attrs.riscv_attr_addr`, which stores the pointer to the beginning data of RISC-V attribute, and `riscv_attrs.riscv_attr_size`, which stores the length of RISC-V attribute,
I also changed `riscv_attrs.riscv_attr_string` to `riscv_extension_string`.

symtabAPI/src/Object-elf.C
wxrdnx: @hainest 
Update: As Jim mentioned below, I replaced the original `riscv_attrs.riscv_attr_string` with:
- `riscv_attrs.riscv_attr_addr`, which stores the pointer to the beginning data of RISC-V attribute
- `riscv_attrs.riscv_attr_size`, which stores the length of RISC-V attribute.

symtabAPI/src/Object-elf.C
kupsch: I meant to add them to the ObjectElf like the other member of ObjectElf that contain the start and length of section/segments (Object-elf.h line 382-409). Then you can add a methods to get these values like line 319-320

wxrdnx: @kupsch 
Done.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4223,240 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+                reinterpret_cast<const void*>(&attr_str[curr])), &attr_bytes_read);
+    curr += attr_bytes_read;
+
+    // RISC-V attributes have a string value if the tag number is odd
+    // and an integer value if the tag number is even
+    if (attr_tag % 2 != 0) {
+        // a string value
+        const char *cstr = &attr_str[curr];
+        std::string sval = std::string(cstr);
+        curr += strlen(cstr) + 1;
```
kupsch: This code and the code from the `else` to `switch` should be `parse_attrs` function as it is independent of the riscv attributes (all ELF attributes work like this). `parse_attrs` should take *2* lambdas as parameters

1. bool(int tag, int value)
2. bool(int tag, std::string value)

Each function returns `true` on unknown tag value and `false` otherwise. Then each of your switch statements becomes its own function that takes a ref to the riscv_attrs in addition to the parameters above, returning the unknown tag value. `parse_attrs` can also take an optional parameter turn unknown tags into an error or ignore unknown tags (the default?).

This way the lambda are independent of the ELF attributes format and calls to ELF that would have to be repeated if we want to parse other ELF attribute sections in the future.

If you have questions we can meet.

symtabAPI/src/Object-elf.C
wxrdnx: @hainest 
I have made the changes suggested by Jim. See [here](#pullrequestreview-3362686772)

symtabAPI/src/Object-elf.h
wxrdnx: @hainest 
I have made the changes suggested by Jim. See [here](#pullrequestreview-3362686772)

symtabAPI/src/Object-elf.C
wxrdnx: @kupsch 
Done

wxrdnx:
@hainest 
May I ask do you have any further thoughts or feedback about this PR?

symtabAPI/src/Object-elf.C
```diff
@@ -835,6 +839,11 @@ bool ObjectELF::loaded_elf(Offset &txtaddr, Offset &dataddr,
             hasModinfo_ = true;
         } else if (strcmp(name, GNU_LINKONCE_THIS_MODULE_NAME) == 0) {
             hasGnuLinkonceThisModule_ = true;
+        } else if (scn.sh_type() == SHT_RISCV_ATTRIBUTES) {
+            riscv_attr_size_ = scnp->get_data().d_size();
+            riscv_attr_addr_ = scnp->sh_offset();
+            std::cout << std::hex << scnp->sh_offset() << std::endl;
```
hainest: If this debugging is useful, it should use `create_printf`. Otherwise, let's remove it.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1575,68 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+                        return true;
+                }
+            };
+            bool result = parse_attrs(riscv_attr_addr,
+                    riscv_attr_size,
+                    "riscv",
+                    riscv_parse_attr_int,
+                    riscv_parse_attr_string);
+            if (!result) {
+                create_printf("%s[%d]: riscv attributes missing or corrupted\n", FILE__, __LINE__);
```
hainest, line 1633: Will `get_riscv_extensions` work if `parse_attrs` fails? If not, then we should probably return early here.

symtabAPI/src/Object-elf.C
```diff
@@ -1566,6 +1575,68 @@ void ObjectELF::load_object(bool alloc_syms) {
...
+                        riscv_attrs.stack_align = boost::make_optional(value);
+                        return false;
+                    default:
+                        return true;
+                }
+            };
+            auto riscv_parse_attr_string = [this](int tag, std::string value) -> bool {
+                switch (tag) {
+                    case RiscvAttrTag::arch:
+                        riscv_attrs.riscv_extension_string = value;
```
hainest:
```suggestion
                        riscv_attrs.riscv_extension_string = std::move(value);
```

Might as well avoid a copy here.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4257,210 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
    return dwarf->frameParser()->getRegValueAtFrame(pc, reg, reg_result, reader, frame_error);
 
 }
+
+bool ObjectELF::parse_attrs(const char *attr_string,
```
hainest, line 4260: Oh, but for want of a [`string_view`](https://www.boost.org/doc/libs/latest/libs/utility/doc/html/utility/utilities/string_view.html)...

That would allow using iterators instead of keeping an array index around. You don't need to update this now, but it's definitely something to consider for the future.

In general, this function still makes me uneasy. I don't like the nested loops, global variables, myriad "size" variables, and the general lack of structure. It's very difficult to follow which would make it almost impossible to debug. Again, I won't stop this PR for it, but I would strongly encourage considering a refactoring in the future. I'd recommend at least a [Compose Method](https://scrutinizer-ci.com/docs/refactorings/compose-method). Do you have any tests that you used to verify its correctness? If so, we should try to adapt them into a unit test.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4257,210 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+    //         2) A 32-bit integer indicating the size of the attribute
+    //         3) A series of "attribute data". The attribute data is platform dependent
+
+    int section_end = attr_string_size;
+
+    // Parse all attribute sections
+    while (curr < section_end) {
+
+        // Get the attribute section size
+        int attr_section_size = read_memory_as<int>(&attr_string[curr]);
```
hainest: Since the standard says `32-bit integer`, let's use a fixed-size type just to make sure there aren't any portability accidents in the future as we move toward decoupling host and codegen architectures.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4257,210 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+        if (attr_section_size > section_end) {
+            create_printf("%s[%d]:  Bad attribute section length (%d > %d)\n",
+                    FILE__, __LINE__, attr_section_size, section_end);
+            return false;
+        }
+        int attr_section_end = curr + attr_section_size;
+        curr += sizeof(int);
```
hainest:
```suggestion
        curr += sizeof(attr_section_size);
```

et al.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4257,210 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+        // The next null-terminated string indicates the name of the current attribute
+        const char *curr_attr_section_name = &attr_string[curr];
+        if (strcmp(curr_attr_section_name, attr_section_name)) {
```
hainest:
```suggestion
        if (strcmp(curr_attr_section_name, attr_section_name) != 0) {
```

Just for clarity because I had to stop to remember that `strcmp` returns 0 on "equal". *string_view*...

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4257,210 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+            riscv_attrs.floatABI = RiscvFloatAbiEnum::SINGLE;
+            break;
+        case EF_RISCV_FLOAT_ABI_DOUBLE:
+            riscv_attrs.floatABI = RiscvFloatAbiEnum::DOUBLE;
+            break;
+        case EF_RISCV_FLOAT_ABI_QUAD:
+            riscv_attrs.floatABI = RiscvFloatAbiEnum::QUAD;
+            break;
+        default:
+            break;
```
hainest, line 4458: Logging a warning message here would probably be a good idea since `-Wswitch` can't save us from a missing any future `EF_RISCV_FLOAT_*` values added to ELF because they just use `#define`s.

symtabAPI/src/Object-elf.h
```diff
@@ -91,6 +93,50 @@ class Object;
 class ObjectELF;
 class InlinedFunction;
 
+#ifndef PT_RISCV_ATTRIBUTES
```
hainest, line 96: Include these in the note in DyninstElfUtils.cmake what needs to be modified when upgrading to a newer version of elfutils.

symtabAPI/src/Object.h
```diff
@@ -178,6 +178,8 @@ class Object : public boost::basic_lockable_adapter<boost::mutex>
     virtual Offset getTOCoffset(Offset) const { return 0; }
     virtual void setTOCoffset(Offset) { }
 
+    virtual bool usesCompressedInstructionFormat() { return false; }
```
hainest:
```suggestion
    virtual bool usesCompressedInstructionFormat() const { return false; }
```

symtabAPI/src/Object-elf.C
wxrdnx: Done

symtabAPI/src/Object-elf.C
wxrdnx: Agree, let's return early.

symtabAPI/src/Object-elf.C
wxrdnx: Done

symtabAPI/src/Object-elf.C
wxrdnx: Done.
Change`int` to `int32_t`

symtabAPI/src/Object-elf.C
wxrdnx: Done

symtabAPI/src/Object-elf.h
wxrdnx: Done

symtabAPI/src/Object-elf.C
wxrdnx: Yes, I do have some test attributes to verify the correctness.
@hainest Where should I add the unit test?

hainest: I would recommend 'tests/unit/symtabAPI/Object/riscv_attributes.cpp'.

symtabAPI/src/Object-elf.C
```diff
@@ -4184,3 +4256,211 @@ bool ObjectELF::getRegValueAtFrame(Dyninst::Address pc, Dyninst::MachRegister re
...
+            }
+
+            int32_t attr_size = read_memory_as<int32_t>(&attr_string[curr]);
+            if (attr_size > attr_section_end) {
+                create_printf("%s[%d]:  Bad attribute length (%d > %d)\n",
+                        FILE__, __LINE__, attr_size, attr_section_end);
+                return false;
+            }
+            int attr_end = curr + attr_size - 1;
+            curr += sizeof(int);
```
hainest: Sorry, I missed this one in my other review. Let's replace it, as well.

```suggestion
            curr += sizeof(attr_size);
```

hainest (approved):
There's one last `sizeof` to update, but everything looks good, otherwise.

hainest:
The `parse_attr_string` lambda is running afoul gcc's [`-Wnoexcept`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61372) (see https://github.com/dyninst/dyninst/actions/runs/19023368220). I _think_ the correct solution is to follow the compiler's suggestion of just making it noexcept.

```c++
auto riscv_parse_attr_string = [this](int tag, std::string value) noexcept -> bool
```

That might run into `-Wnoexcept-type` since we're not using C++17. :\

---

```
In file included from /usr/include/c++/14/bits/stl_pair.h:60,
                 from /usr/include/c++/14/tuple:38,
                 from /usr/include/c++/14/mutex:40,
                 from /__w/dyninst/dyninst/src/common/src/vgannotations.h:60,
                 from /__w/dyninst/dyninst/src/symtabAPI/src/Object-elf.C:36:
/usr/include/c++/14/type_traits: In instantiation of 'constexpr bool std::__call_is_nt(__invoke_other) [with _Fn = Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, string)>&; _Args = {int, __cxx11::basic_string<char, char_traits<char>, allocator<char> >}]':
/usr/include/c++/14/type_traits:3129:34:   required from 'struct std::__call_is_nothrow<std::__invoke_result<Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >'
 3129 |         std::__call_is_nt<_Fn, _Args...>(typename _Result::__invoke_type{})
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/type_traits:183:35:   required by substitution of 'template<class ... _Bn> std::__detail::__first_t<std::integral_constant<bool, true>, typename std::enable_if<(bool)(_Bn::value), void>::type ...> std::__detail::__and_fn(int) [with _Bn = {std::integral_constant<bool, true>, std::__call_is_nothrow<std::__invoke_result<Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >}]'
  183 |                                       __enable_if_t<bool(_Bn::value)>...>;
      |                                                               ^~~~~
/usr/include/c++/14/type_traits:199:42:   required from 'struct std::__and_<std::integral_constant<bool, true>, std::__call_is_nothrow<std::__invoke_result<Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >'
  199 |     : decltype(__detail::__and_fn<_Bn...>(0))
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
/usr/include/c++/14/type_traits:3222:12:   required from 'struct std::is_nothrow_invocable_r<bool, Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >'
 3222 |     struct is_nothrow_invocable_r
      |            ^~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/type_traits:3547:52:   required from 'constexpr const bool std::is_nothrow_invocable_r_v<bool, Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >'
 3547 |     = is_nothrow_invocable_r<_Ret, _Fn, _Args...>::value;
      |                                                    ^~~~~
/usr/include/c++/14/bits/invoke.h:105:14:   required from 'constexpr std::enable_if_t<((bool)is_invocable_r_v<_Res, _Callable, _Args ...>), _Res> std::__invoke_r(_Callable&&, _Args&& ...) [with _Res = bool; _Callable = Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, string)>&; _Args = {int, __cxx11::basic_string<char, char_traits<char>, allocator<char> >}; enable_if_t<((bool)is_invocable_r_v<_Res, _Callable, _Args ...>), _Res> = bool]'
  105 |     noexcept(is_nothrow_invocable_r_v<_Res, _Callable, _Args...>)
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/std_function.h:290:30:   required from 'static _Res std::_Function_handler<_Res(_ArgTypes ...), _Functor>::_M_invoke(const std::_Any_data&, _ArgTypes&& ...) [with _Res = bool; _Functor = Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>; _ArgTypes = {int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >}]'
  290 |         return std::__invoke_r<_Res>(*_Base::_M_get_pointer(__functor),
      |                ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  291 |                                      std::forward<_ArgTypes>(__args)...);
      |                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/std_function.h:451:21:   required from 'std::function<_Res(_ArgTypes ...)>::function(_Functor&&) [with _Functor = Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>&; _Constraints = void; _Res = bool; _ArgTypes = {int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >}]'
  451 |               _M_invoker = &_My_handler::_M_invoke;
      |                            ^~~~~~~~~~~~~~~~~~~~~~~
/__w/dyninst/dyninst/src/symtabAPI/src/Object-elf.C:1632:38:   required from here
 1632 |             bool result = parse_attrs(riscv_attr_addr,
      |                           ~~~~~~~~~~~^~~~~~~~~~~~~~~~~
 1633 |                     riscv_attr_size,
      |                     ~~~~~~~~~~~~~~~~  
 1634 |                     "riscv",
      |                     ~~~~~~~~          
 1635 |                     riscv_parse_attr_int,
      |                     ~~~~~~~~~~~~~~~~~~~~~
 1636 |                     riscv_parse_attr_string);
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/type_traits:3123:14: error: noexcept-expression evaluates to 'false' because of a call to 'Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>' [-Werror=noexcept]
 3123 |       return noexcept(std::declval<_Fn>()(std::declval<_Args>()...));
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/__w/dyninst/dyninst/src/symtabAPI/src/Object-elf.C:1623:44: note: but 'Dyninst::SymtabAPI::ObjectELF::load_object(bool)::<lambda(int, std::string)>' does not throw; perhaps it should be declared 'noexcept'
 1623 |             auto riscv_parse_attr_string = [this](int tag, std::string value) -> bool {
      |    
```

PR #1968, opened 2025-07-14
CMake: update detection of 32-bit runtime

hainest:
Requires #1965 

This also fixes a bug in the filename generated for the 32-bit static dyninstAPI_RT.


bbiiggppiigg (approved):

hainest:
@bbiiggppiigg I changed this to not use DyninstCheckStaticExeSupported since that's a separate concept from a static 32-bit runtime.

hainest:
@bbiiggppiigg Ok. *Now* I'm done updating this...

bbiiggppiigg (approved):

hainest:
No regressions on 19 machines- though there are only a couple with 32-bit libraries.

PR #1969, opened 2025-07-14
CMake: add detection of PIE support

hainest:
Requires #1965


tests/regression/symtabAPI/Symtab/file_properties/CMakeLists.txt
```diff
@@ -64,31 +60,29 @@ foreach(strip_t "stripped" "none")
...
+        endif()
       else()
-        # some compilers use PIE by default
-        target_compile_options(${_targ} PRIVATE "-no-pie")
-        target_link_options(${_targ} PRIVATE "-no-pie")
+        set(_use_pic FALSE)
+        if(pic_t STREQUAL "pic")
+          set(_use_pic TRUE)
+        endif()
+        set_property(TARGET ${_targ} PROPERTY POSITION_INDEPENDENT_CODE ${_use_pic})
```
bbiiggppiigg, line 78: Is there a reason not to just
if  (pic_t STREQUAL "pic")
set_property(...  POSITION_INDEPENDENT_CODE TRUE).
endif()))
and does POSITION_INDEPENDENT_CODE needs to set for static-pie?

hainest: There are tests for non-pic binaries.

```CMake
foreach(pic_t "pic" "none")
```

In this case, it's necessary to explicitly do `set_property(... POSITION_INDEPENDENT_CODE FALSE)` so that CMake will use `-no-pie`. Many compilers now default to using PIE, so this circumvents that.

>and does POSITION_INDEPENDENT_CODE needs to set for static-pie?

I've not seen that it's necessary. I _did_ find that CMake will always append `-fPIE` to the link line even if this is called before `target_link_options(... ${DYNINST_CXX_STATIC_PIE_FLAG})` which breaks gcc since `-pie` overrides `-static-pie` which is annoying.

tests/cmake/DyninstCheckPIESupported.cmake
```diff
@@ -0,0 +1,87 @@
...
+    # If a flag for this language has been found, don't re-check.
+    if(${_spf})
+      continue()
+    endif()
+
+    set(_sps DYNINST_${_lang}_STATIC_PIE_SUPPORTED)
+    foreach(_flag ${_possible_staticpie_flags})
+      dyninst_check_compiler_flag(${_lang} "${_flag}" "${_sps}")
+
+      set(${_spf}
```
bbiiggppiigg, line 89: It seems like you decide to not check the ${_sps} before setting $_spf?

hainest: Woops. I mean to add a check to set it to an empty string.

hainest:
No test regressions on 19 machines. Is also working on machines that have `-static` but not `-static-pie` (e.g., ppc64le).

PR #1970, opened 2025-07-17
Refactor Codegen as a first step for removing ifdefs

bbiiggppiigg:
- [ ] Make NS_instruction virtual
- [ ] Resolve conflicts betweeen NS_* name spaces

- [ ] Make insnCodeGen virtual
- [ ] Make all per architecture insnCodeGen classes a subclass of insnCodeGen.
- [ ] Change how code gen's are created

Cross Compilation Testng
Mahir
- [x] x86_64
- [x] x86
- [x] aarch64
- [x] ppc64le

Zeroah
- [ ] x86_64
- [ ] x86
- [ ] aarch64
- [ ] ppc64le

LLNL
- [ ] x86_64
- [ ] x86
- [ ] aarch64
- [ ] ppc64le

### DIFF
--- a/common/src/arch-aarch64.h
+++ b/common/src/arch-aarch64.h
@@ -56,7 +56,7 @@ namespace NS_aarch64 {
 //#define UNCOND_BR_REG_MASK  (0xfe000000)
 //#define UNCOND_BR_REG       (0xd6000000)
 
-#define BREAK_POINT_INSN 0xd4200000
+const uint64_t BREAK_POINT_INSN=0xd4200000;
 
 #define BOp             0x05
 #define BCondOp         0x2A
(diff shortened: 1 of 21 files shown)


bbiiggppiigg:
The eventual goal is to allow us to instantiate a target specific subclass of codegen (like what we do for instruction-decoder). 
As a first step, we need to allow the codegen (insnCodeGen) for different architectures to coexist.

@hainest 
While there are some more items need to be worked out, can you skim through the PR and see if the way I'm doing it make sense ?

hainest:
@bbiiggppiigg Do you have some time to chat about this? I have several questions that are probably more easily answered that way.

common/src/arch-aarch64.h
```diff
@@ -95,17 +95,17 @@ namespace NS_aarch64 {
...
 #define MAX_IMM8    (127)
-#define MIN_IMM16   (-32768)
-#define MAX_IMM16   (32767)
-#define MIN_IMM32   (-2147483647 - 1)
-#define MAX_IMM32   (2147483647)
+const uint64_t MIN_IMM16   =(-32768);
+const uint64_t MAX_IMM16   = (32767);
+const uint64_t MIN_IMM32   = (-2147483647 - 1);
+const uint64_t MAX_IMM32   (2147483647);
 #define MAX_IMM48   ((long)(-1 >> 17))
```
bbiiggppiigg, line 102: Wrong constant implementations

PR #1972, opened 2025-07-17
Add support for RISC-V CODEGEN_ARCH

wxrdnx:
@bbiiggppiigg

I added support for RISC-V CODGEN_ARCH (https://github.com/dyninst/dyninst/pull/1830). Now we can perform cross platform code generation on RISC-V.

I've tested cross compiling x64 codegen on RISC-V host and cross compiling RISC-V codegen on x64 host. Both worked fine.


cmake/DyninstCapArchDef.cmake
```diff
@@ -53,6 +53,9 @@ elseif(DYNINST_CODEGEN_ARCH_AMDGPU_GFX90A)
   set(ARCH_DEFINES_CODEGEN -Darch_amdgpu_gfx90a -Darch_64bit)
 elseif(DYNINST_CODEGEN_ARCH_AMDGPU_GFX940)
   set(ARCH_DEFINES_CODEGEN -Darch_amdgpu_gfx940 -Darch_64bit)
+elseif(DYNINST_CODEGEN_ARCH_RISCV64)
+  set(ARCH_DEFINES_CODEGEN -Darch_riscv64 -Darch_64bit)
+  set(CAP_DEFINES ${CAP_DEFINES} -Dcap_registers)
```
bbiiggppiigg: Consider move the case to RISCV to be right after the other cpu architectures (between arch and amdgpu).

cmake/DyninstCapArchDef.cmake
```diff
@@ -70,6 +73,8 @@ if(DYNINST_OS_Linux)
     set(BUG_DEFINES ${BUG_DEFINES} -Dbug_registers_after_exit)
   elseif(DYNINST_HOST_ARCH_AARCH64)
     set(OLD_DEFINES -Daarch64_unknown_linux)
+  elseif(DYNINST_HOST_ARCH_RISCV64)
+    set(OLD_DEFINES -Driscv64_unknown_linux)
```
hainest: Is this define used in Dyninst? The `OLD_DEFINES` are really just there to support the existing weird conditional compilation in the test suite. The testsuite doesn't have anything for riscv specifically, so this wouldn't be needed in that case.

cmake/DyninstPlatform.cmake
```diff
@@ -99,6 +103,8 @@ if(${_host_os} STREQUAL "Linux")
       set(DYNINST_PLATFORM x86_64-unknown-linux2.4)
     elseif(${_host_arch} STREQUAL "aarch64")
       set(DYNINST_PLATFORM aarch64-unknown-linux)
+    elseif(${_host_arch} STREQUAL "riscv64")
+      set(DYNINST_PLATFORM riscv64-unknown-linux)
```
hainest: Same comment as above. This isn't needed since the test suite doesn't have a config for riscv.

cmake/DyninstCapArchDef.cmake
wxrdnx: Okay! Fixed

cmake/DyninstCapArchDef.cmake
wxrdnx: The define is not used in Dyninst. Removed!

cmake/DyninstPlatform.cmake
wxrdnx: Got it. Removed!

wxrdnx:
@hainest 
Alright, I just removed all RISC-V old defines.

hainest (approved):
Before you merge, could you re-run your tests just to triple check those changes didn't break anything?

wxrdnx:
> Before you merge, could you re-run your tests just to triple check those changes didn't break anything?

No problem

PR #1974, opened 2025-07-18
CMake: move static exe detection into tests

hainest:
This module is only used for tests, so put it there.


hainest:

PR #1802, opened 2024-10-25
Codegen for AMDGPU

rochauha:
This branch contains codegen work for AMDGPU.

To build for AMDGPU, pass an additional option `-DDYNINST_ARCH_amdgpu=ON` to CMake.

PR #1979, opened 2025-08-11
[BPatch] Add API function std::string BPatch_image::getProgramFileName()

rochauha:
This method enables getting the filename without knowing the length.

It is used in #1802 for spitting out a file with instrumented kernel names from the GPU code object.

### DIFF
--- a/dyninstAPI/src/BPatch_image.C
+++ b/dyninstAPI/src/BPatch_image.C
@@ -919,20 +919,33 @@ char *BPatch_image::getProgramName(char *name, unsigned int len)
 
 char *BPatch_image::getProgramFileName(char *name, unsigned int len)
 {
-   std::vector<AddressSpace*> as;
-   addSpace->getAS(as);
-   AddressSpace *aout = as[0];
+  std::string fileName = this->getProgramFileName();
 
-   if (!aout->mappedObjects().size()) {
+  if (len > fileName.size())  {
+    std::strcpy(name, fileName.c_str());
+  } else if (len > 0u) {
+    name[0] = '\0';
+  }
+
+  return name;
+}
+
+std::string BPatch_image::getProgramFileName() const {
+   std::vector<AddressSpace*> addrSpaces;
+   addSpace->getAS(addrSpaces);
+
+   assert(!addrSpaces.empty());
+   AddressSpace *addrSpace = addrSpaces[0];
+
+   assert(addrSpace);
+   if (addrSpace->mappedObjects().empty()) {
       // No program defined yet
-      strncpy(name, "<no program defined>", len);
+      return "<no program defined>";
    }
 
-   string imname =  aout->getAOut()->fileName();
-   if (imname.empty()) imname = "<unnamed image file>";
-
-   strncpy(name, imname.c_str(), len);
-   return name;
+   string fileName =  addrSpace->getAOut()->fileName();
+   if (fileName.empty()) fileName = "<unnamed file>";
+   return fileName;
 }
 
 BPatch_module *BPatch_image::findModule(mapped_module *base) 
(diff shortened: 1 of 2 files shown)


dyninstAPI/src/BPatch_image.C
```diff
@@ -935,6 +935,21 @@ char *BPatch_image::getProgramFileName(char *name, unsigned int len)
    return name;
 }
 
+std::string BPatch_image::getProgramFileName() const {
+   std::vector<AddressSpace*> addrSpaces;
+   addSpace->getAS(addrSpaces);
+   AddressSpace *addrSpace = addrSpaces[0];
```
hainest, line 938: I know the other implementation doesn't check, but are we guaranteed to have at least one address space here?

dyninstAPI/src/BPatch_image.C
```diff
@@ -935,6 +935,21 @@ char *BPatch_image::getProgramFileName(char *name, unsigned int len)
...
+
+   if (!addrSpace->mappedObjects().size()) {
```
hainest:
```suggestion
   if (addrSpace->mappedObjects().empty()) {
```

hainest (changes requested):
We should implement one of these two versions in terms of the other. I would recommend keeping this one and changing the old one to

```C++
char *BPatch_image::getProgramFileName(char *name, unsigned int len)
{
   std::string fileName = getProgramFileName();
   strncpy(name, fileName.c_str(), len);
   return name;
}
```

kupsch:
I agree with @hainest suggestion to implement one of the functions in terms of the others, but the proposed and the current version using `strncpy` is broken. `strncpy` does not guarantee that the string is null-terminated (it is also inefficient as it zero fills the buffer that is unwritten).  Here is an improved version

```C
char *BPatch_image::getProgramFileName(char *name, unsigned int len)
{
   std::string fileName = getProgramFileName();
   if (filename.size() < len)  {
      strcpy(name, filelanen.c_str());
   }  else if (len > 0) {
      name[0] = '\0';  // return empty string if len is too small
   }
   return name;
}
```

rochauha:
> I agree with @hainest suggestion to implement one of the functions in terms of the others, but the proposed and the current version using `strncpy` is broken. `strncpy` does not guarantee that the string is null-terminated (it is also inefficient as it zero fills the buffer that is unwritten). Here is an improved version
> 
> ```c
> char *BPatch_image::getProgramFileName(char *name, unsigned int len)
> {
>    std::string fileName = getProgramFileName();
>    if (filename.size() < len)  {
>       strcpy(name, filelanen.c_str());
>    }  else if (len > 0) {
>       name[0] = '\0';  // return empty string if len is too small
>    }
>    return name;
> }
> ```

Should this be a separate PR or do you think it should be in this PR?

dyninstAPI/src/BPatch_image.C
rochauha: Now we are.

hainest:
> > I agree with @hainest suggestion to implement one of the functions in terms of the others, but the proposed and the current version using `strncpy` is broken. `strncpy` does not guarantee that the string is null-terminated (it is also inefficient as it zero fills the buffer that is unwritten). Here is an improved version
> > ```c
> > char *BPatch_image::getProgramFileName(char *name, unsigned int len)
> > {
> >    std::string fileName = getProgramFileName();
> >    if (filename.size() < len)  {
> >       strcpy(name, filelanen.c_str());
> >    }  else if (len > 0) {
> >       name[0] = '\0';  // return empty string if len is too small
> >    }
> >    return name;
> > }
> > ```
> 
> Should this be a separate PR or do you think it should be in this PR?

It should go here. A tiny change to ward off a possible future `-Wsign-compare` warning (if ever we turn that on), would be `else if (len > 0)` -> `else if (len > 0u)`.

rochauha:
I've made the suggested changes.

kupsch (approved):

PR #1981, opened 2025-08-11
Add new interfaces to the emitter

rochauha:
Most of these are used in #1802.

Also, the functionality of `emitMovePCtoReg` is used by other architectures as well, but it is implemented in various places in the codebase. It would be good to have this behavior implemented in respective emitter implementations. AMDGPU doesn't use it yet, but will switch to using it soon.

### DIFF
--- a/dyninstAPI/src/emitter.h
+++ b/dyninstAPI/src/emitter.h
@@ -122,13 +122,16 @@ class Emitter {
     virtual bool emitTOCJump(block_instance *, codeGen &) { assert(0); return false; }
     virtual bool emitTOCCall(block_instance *, codeGen &) { assert(0); return false; }
 
-    virtual void emitNops(unsigned numNops, codeGen &gen) {}
-    virtual void emitEndProgram(codeGen &gen) {}
-    virtual void emitMovLiteral(Register reg, uint32_t value, codeGen &gen) {}
-    virtual void emitConditionalBranch(bool onConditionTrue, int16_t wordOffset,
-                                     codeGen &gen) {}
-    virtual void emitShortJump(int16_t wordOffset, codeGen &gen) {}
-    virtual void emitLongJump(Register reg, uint64_t toAddress, codeGen &gen) {}
+    virtual void emitNops(unsigned /* numNops */, codeGen & /* gen */) {}
+    virtual void emitEndProgram(codeGen & /* gen */) {}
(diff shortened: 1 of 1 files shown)


dyninstAPI/h/BPatch_image.h
```diff
@@ -258,6 +258,9 @@ class DYNINST_EXPORT BPatch_image: public BPatch_sourceObj {
 
   char * getProgramFileName(char *name, unsigned int len);
 
+  // Same as above, but doesn't require knowing the length.
+  std::string getProgramFileName() const;
+
```
kupsch: Why is this here if it is in #1979?  Don't include stuff from another PR.

dyninstAPI/src/BPatch_image.C
```diff
@@ -935,6 +935,21 @@ char *BPatch_image::getProgramFileName(char *name, unsigned int len)
...
+   if (!addrSpace->mappedObjects().size()) {
+      // No program defined yet
+      return "<no program defined>";
+   }
+
+   string fileName =  addrSpace->getAOut()->fileName();
+   if (fileName.empty()) fileName = "<unnamed file>";
+   return fileName;
+}
+
```
kupsch: Why is this here if it is in #1979?  Don't include stuff from another PR.

kupsch (commented):
Remove changes that are in another PR.

hainest:
A couple of items:

1. These need to be carefully documented.
2. Aside from `emitMovePCtoReg`, are these features useful to architectures outside of amdgpu? I'm thinking about how much work we're adding to #1970.

rochauha:
> A couple of items:
> 
> 1. These need to be carefully documented.
> 2. Aside from `emitMovePCtoReg`, are these features useful to architectures outside of amdgpu? I'm thinking about how much work we're adding to [Refactor Codegen as a first step for removing ifdefs #1970](https://github.com/dyninst/dyninst/pull/1970).

The method `emitMovePCtoReg` is used by Power and AArch64. `dyninstAPI/src/emit-power.h` and `dyninstAPI/src/emit-aarch64.h` have this for their respective emitters and also implement it. But the method is specific to those emitters only. This PR also brings it as an option to all emitters.

X86 uses `emitMovPCRMToReg` and `emitMovPCRMToReg64` functions for PC-relative computations.

PR #1982, opened 2025-08-20
Handle s_endpgm as an aborting instruction

bbiiggppiigg:
This is the correct fix for the problem of duplicated s_endpgm, which happens if a basic block starts with s_endpgm.

### DIFF
--- a/parseAPI/src/IA_IAPI.C
+++ b/parseAPI/src/IA_IAPI.C
@@ -321,7 +321,7 @@ bool IA_IAPI::isAbort() const
     entryID e = curInsn().getOperation().getID();
     return e == e_int3 ||
         e == e_hlt ||
-        e == e_ud2 || curInsn().isGPUKernelExit();
+        e == e_ud2;
 }
 
 bool IA_IAPI::isInvalidInsn() const
--- a/parseAPI/src/IA_IAPI.C
+++ b/parseAPI/src/IA_IAPI.C
@@ -321,7 +321,7 @@ bool IA_IAPI::isAbort() const
     entryID e = curInsn().getOperation().getID();
     return e == e_int3 ||
         e == e_hlt ||
-        e == e_ud2;
+        e == e_ud2 || curInsn().isGPUKernelExit();
 }
 
 bool IA_IAPI::isInvalidInsn() const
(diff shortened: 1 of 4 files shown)


parseAPI/src/IA_IAPI.C
```diff
@@ -321,7 +321,7 @@ bool IA_IAPI::isAbort() const
     entryID e = curInsn().getOperation().getID();
     return e == e_int3 ||
         e == e_hlt ||
-        e == e_ud2;
+        e == e_ud2 || curInsn().isGPUKernelExit();
```
hainest: I feel like a better solution would be to override `isAbort` in `IA_amdgpu`.

bbiiggppiigg: I totally forgot about IA_amdgpu. let me check if that works.

bbiiggppiigg:
@hainest I don't understand why I suddenly start triggering -Winconsistent-missing-override in the libabigail tests. Should I add override to every single function that was overriden in IA_amdgpu?

parseAPI/src/IA_amdgpu.h
```diff
@@ -49,22 +49,23 @@ class IA_amdgpu : public IA_IAPI {
...
-	virtual bool isStackFramePreamble() const;
-	virtual bool cleansStack() const;
-	virtual bool sliceReturn(ParseAPI::Block* bit, Address ret_addr, ParseAPI::Function * func) const;
-	virtual bool isReturnAddrSave(Address& retAddr) const;
-	virtual bool isReturn(Dyninst::ParseAPI::Function * context, Dyninst::ParseAPI::Block* currBlk) const;
-	virtual bool isFakeCall() const;
-	virtual bool isIATcall(std::string &) const;
-	virtual bool isLinkerStub() const;
-	virtual bool isNopJump() const;
+	IA_amdgpu* clone() const override;
```
hainest: Let's not fix the `-Winconsistent-override` warnings in this PR. Instead, just don't explicitly mark `isAbort` as `override`. All of the `InstructionAdaptor` classes need to be cleaned up like this, though.

hainest (changes requested):

bbiiggppiigg:
@hainest @kupsch this PR should be ready

kupsch (approved):

hainest (approved):

PR #1983, opened 2025-08-20
Fix errors in separation of host and codegen arch

bbiiggppiigg:


hainest:
This PR is all over the place. What is the main goal here?

bbiiggppiigg:
> This PR is all over the place. What is the main goal here?

So we are trying to merge ronak's pull request #1802 
During that process, we discovered several places where the conditional compile should be based on the HOST_ARCH instead of CODEGEN_ARCH.
Mainly, anything that is proccontrol / stackwalker related should be HOST_ARCH dependent.
So these changes are required for the AMDGPU PR to build.
So far this is tested on mahir (x86_64) and llnl(power), zeroah is down for the moment.
I'll request review after I cross compiled on all three of them.

bbiiggppiigg:
@hainest @kupsch 

Tested cross compile on llnl mahir and zeroah

PR #1987, opened 2025-08-26
Preliminary work for AMDGPU codegen

rochauha:
This branch currently targets AMDGPU GFX908 only. Use -DDYNINST_CODEGEN_ARCH=amdgpu_gfx908 when building with CMake.

This PR is a continuation of my older PR - #1802 (Initial work for supporting AMDGPU codegen); with a few fixes from @bbiiggppiigg for host-codgen separation.


rochauha:
@hainest @kupsch, this is ready for review.

hainest:
There's a lot here, so I'd like to go through it in stages. My first concern is putting architecture-specific code in public headers. That's a habit that can quickly spiral out of control as more architectures are added. If we can avoid that here, it would be best.

`BPatch_addressSpace`
1. Why does an `addressSpace` for a GPU binary need to have a cache of the instrumented functions when the other architectures don't (cf. `instrumentedFunctions`)?
2. `insertPrologue` is `insertSnippet:866` with `when`=`BPatch_callBefore` and `order`=`BPatch_firstSnippet`, but forces `ipOrder` returned by `BPatchToInternalArgs` to be `orderFirstAtPoint`. Yet, that's what `BPatchToInternalArgs` returns when `when`=`BPatch_firstSnippet`. `insertPrologue` also doesn't insert the point into the cache mentioned in (1). Is that intentional?

`BPatch_snippet`
1. I am in support of adding an overload for the constructor that takes a name, but it should take the `std::string` by value and move it in. The ctor that takes a C string should then delegate to the new one. If we want to take an ABI break, I'd just replace the existing C string parameter with `std::string`. But that should be in a separate PR.
2. The added constructor doesn't need to be here. The same would be better achieved by creating a free function in the AMDGPU instrumentation code that takes the smaller parameter set, does the type size check, constructs the `AstNodePtr`, and then calls the existing `BPatch_variablExpr` constructor.

rochauha:
> There's a lot here, so I'd like to go through it in stages. My first concern is putting architecture-specific code in public headers. That's a habit that can quickly spiral out of control as more architectures are added. If we can avoid that here, it would be best.

I think this makes sense. `insertPrologue` and `insertEpilogue` are required internally only. The user doesn't need them. Do you have any suggestions for refactoring these changes into something self-contained?

> 
> `BPatch_addressSpace`
> 
> 1. Why does an `addressSpace` for a GPU binary need to have a cache of the instrumented functions when the other architectures don't (cf. `instrumentedFunctions`)?
>
We need to insert a prologue for instrumented kernel functions (i.e GPU functions that can be called from the CPU). For this, we need to track the instrumented functions.

We also spit out a file that contains names of instrumented kernels. This file is used by other tools to further rewrite metadata in the GPU binary after instrumentation.

> 2. `insertPrologue` is `insertSnippet:866` with `when`=`BPatch_callBefore` and `order`=`BPatch_firstSnippet`, but forces `ipOrder` returned by `BPatchToInternalArgs` to be `orderFirstAtPoint`. Yet, that's what `BPatchToInternalArgs` returns when `when`=`BPatch_firstSnippet`. 

`insertPrologue` inserts the prologue into each instrumented kernel. The prologue must be before any other instrumentation that was inserted at the entrypoint of a function.

> `insertPrologue` also doesn't insert the point into the cache mentioned in (1). Is that intentional?

We track the instrumented functions and insert the prologue at entry points of those functions in `insertPrologueInInstrumentedFunctions`.

> 
> `BPatch_snippet`
> 
> 1. I am in support of adding an overload for the constructor that takes a name, but it should take the `std::string` by value and move it in. The ctor that takes a C string should then delegate to the new one. If we want to take an ABI break, I'd just replace the existing C string parameter with `std::string`. But that should be in a separate PR.

Got it.

> 2. The added constructor doesn't need to be here. The same would be better achieved by creating a free function in the AMDGPU instrumentation code that takes the smaller parameter set, does the type size check, constructs the `AstNodePtr`, and then calls the existing `BPatch_variablExpr` constructor.

Got it. I will make the change.

dyninstAPI/src/registerSpace.C
```diff
@@ -796,6 +821,8 @@ bool registerSpace::writeProgramRegister(codeGen &gen,
             gen.codeEmitter()->emitMoveRegToReg(source, destination, gen);
         return true;
         break;
+
+#if defined(arch_x86) || defined(arch_x86_64) // framePointer only defined for x86 and x86_64
```
bbiiggppiigg: If these are only defined for x86 and x86_64, would this case even hit for other architectures?

dyninstAPI/src/registerSpace.C
```diff
@@ -748,6 +770,8 @@ bool registerSpace::readProgramRegister(codeGen &gen,
         gen.codeEmitter()->emitMoveRegToReg(src, dest, gen);
         return true;
         break;
+
+#if defined(DYNINST_CODEGEN_ARCH_X86) || defined(DYNINST_CODEGEN_ARCH_X86_64) // framePointer only defined for x86 and x86_64
```
bbiiggppiigg: Will this case even hit for other architectures?

dyninstAPI/src/parse-cfg.h
```diff
@@ -323,7 +323,7 @@ class parse_func : public ParseAPI::Function
         }
     };
 
-#if defined(DYNINST_HOST_ARCH_POWER) || defined(DYNINST_HOST_ARCH_AARCH64)
+#if defined(DYNINST_HOST_ARCH_POWER) || defined(DYNINST_HOST_ARCH_AARCH64) || defined(DYNINST_CODEGEN_ARCH_AMDGPU_GFX908)
    bool savesReturnAddr() const { return saves_return_addr_; }
```
bbiiggppiigg, line 327: I believe this function only used in stackwalker, since we removed every stack walker stuff for AMDGPU, we shouldn't need this for AMDGPU.

dyninstAPI/src/inst-amdgpu.h
```diff
@@ -0,0 +1,57 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include <cassert>
+#include <stdint.h>
+
+#ifndef INST_AMDGPU_H
+#define INST_AMDGPU_H
+
+inline int TRAMP_GPR_OFFSET(int /* x */) {
```
bbiiggppiigg: I believe none of these macros are used, please remvoe if they are indeed unused.

dyninstAPI/src/inst-amdgpu.h
```diff
@@ -0,0 +1,57 @@
...
+}
+
+inline int TRAMP_SPR_OFFSET(int /* x */) {
+  assert(false && "Not implemented for AMDGPU");
+  return 0;
+}
+
+class codeGen;
+
+void pushStack(codeGen &gen);
```
bbiiggppiigg: Same, these seems to be unused.

dyninstAPI/src/function.h
```diff
@@ -261,7 +261,8 @@ class func_instance : public patchTarget, public Dyninst::PatchAPI::PatchFunctio
 
 
 
-#if defined(DYNINST_HOST_ARCH_POWER) || defined(DYNINST_HOST_ARCH_AARCH64)
+#if defined(DYNINST_HOST_ARCH_POWER) || defined(DYNINST_HOST_ARCH_AARCH64) \
```
bbiiggppiigg: Probably unused, remove if unuse.

bbiiggppiigg (changes requested):
I've left some comment for the files that are easier to review.

bbiiggppiigg:
@rochauha 

Right now, this PR would not build for AMDGPU on ARM / Power.

The problem is that linux-<arch>.C uses getMaxBranch, which is conditionally included from inst-<arch>.h
But, getMaxBranch depends on MAX_BRANCH/MAX_BRANCH_OFFSET, which are defined in common/src/arch-<arch>.h, and MAX_BRANCH and MAX_BRANCH_OFFSET are used in both codegen and linux (so both target arch and host arch)

So one solution I came up with is to move getMaxBranch to the linux-<arch>.C but also make a copy of the definitions (constant + comments) of MAX_BRANCH/MAX_BRANCH_OFFSET there.
I've tested this on zeroah (arm), and I suppose this would work for llnl (power) as well.

And also put a comment to say this is a temporary hack, if you decide to do it this way.

dyninstAPI/src/registerSpace.C
rochauha: This is something that probably got added sometime and got missed during rebase. I'm not sure whether other architectures would hit this case.

dyninstAPI/src/registerSpace.C
rochauha: This is something that probably got added sometime and got missed during rebase. I'm not sure whether other architectures would hit this case.

dyninstAPI/src/inst-amdgpu.h
rochauha: The entire file inst-amdgpu.h contains functions that are not used. For now, I'm cleaning up the file and pushing the changes.

I think we shouldn't have that file at all for the AMDGPU port. Thoughts?

dyninstAPI/src/registerSpace.C
rochauha: For now, I'm correcting the check to use codegen-arch.

dyninstAPI/src/linux-aarch64.C
```diff
@@ -45,6 +45,19 @@
...
+// This file is compiled when host architecture is AArch64. If codegen isn't for AArch64, we define the following:
+#if !defined(DYNINST_CODEGEN_ARCH_AARCH64)
+
+#define MAX_BRANCH_OFFSET      0x07ffffff  // 128MB  Used for B
+
+Address getMaxBranch() {
+    return MAX_BRANCH_OFFSET;
+}
+
+
```
bbiiggppiigg, line 62: Where is ur endif ?
I asked Jim and he says u do have account on zeroah, so you should at least test to make sure u can build for AMDGPU on  zeroah

bbiiggppiigg (changes requested):

dyninstAPI/src/inst-amdgpu.h
bbiiggppiigg: > The entire file inst-amdgpu.h contains functions that are not used. For now, I'm cleaning up the file and pushing the changes.
> 
> I think we shouldn't have that file at all for the AMDGPU port. Thoughts?

Yeah, please go ahead and remove it if it is unneeded.

hainest:
@rochauha @kupsch @bbiiggppiigg I don't mean to be a pain, but I still have unanswered questions about the changes to BPatch. I feel like a chat would be more quickly productive. I can meet today or anytime next week.

bbiiggppiigg:
> @rochauha @kupsch @bbiiggppiigg I don't mean to be a pain, but I still have unanswered questions about the changes to BPatch. I feel like a chat would be more quickly productive. I can meet today or anytime next week.

Yeah I've sent an email to schedule a meeting, lets see if people are available.

dyninstAPI/src/registerSpace.C
rochauha: Given that current master is good, this conditional should be removed?

dyninstAPI/src/registerSpace.C
rochauha: Given that current master is good, this conditional should be removed?

bbiiggppiigg:
@rochauha 

Please rebase your branch and let us know when this PR is ready for review again.

rochauha:
The prologue and epilogue variants using `BPatch_snippet` are not required, but make the additional functions for inserting prologue/epilogue at points self explanatory and typed such that they can only be used for AMDGPU prologue/epilogue.

Let me know your thoughts on the current implementation for inserting prologue/epilogue. It can further be simplified to use the AST nodes directly.

bbiiggppiigg:
> The prologue and epilogue variants using `BPatch_snippet` are not required, but make the additional functions for inserting prologue/epilogue at points self explanatory and typed such that they can only be used for AMDGPU prologue/epilogue.
> 
> Let me know your thoughts on the current implementation for inserting prologue/epilogue. It can further be simplified to use the AST nodes directly.

Is there a reason why you cannot do FuncEntry FuncExit instrumentation, but need to use points?

rochauha:
> > The prologue and epilogue variants using `BPatch_snippet` are not required, but make the additional functions for inserting prologue/epilogue at points self explanatory and typed such that they can only be used for AMDGPU prologue/epilogue.
> > Let me know your thoughts on the current implementation for inserting prologue/epilogue. It can further be simplified to use the AST nodes directly.
> 
> Is there a reason why you cannot do FuncEntry FuncExit instrumentation, but need to use points?

`insertSnippet` for `BPatch_snippets` just attaches / queues ASTs to points. My approach specially handles this for prologue and epilogue . The PatchAPI code emission/patching is happens towards the end along with user snippets.

I also tested with a mutator that inserts at end of function on AMDGPU. The epilogue appears after the user snippet as expected.

bbiiggppiigg:
> > > The prologue and epilogue variants using `BPatch_snippet` are not required, but make the additional functions for inserting prologue/epilogue at points self explanatory and typed such that they can only be used for AMDGPU prologue/epilogue.
> > > Let me know your thoughts on the current implementation for inserting prologue/epilogue. It can further be simplified to use the AST nodes directly.
> > 
> > 
> > Is there a reason why you cannot do FuncEntry FuncExit instrumentation, but need to use points?
> 
> `insertSnippet` for `BPatch_snippets` just attaches / queues ASTs to points. My approach specially handles this for prologue and epilogue .
> 
> I also tested with a mutator that inserts at end of function on AMDGPU. The epilogue appears after the user snippet as expected.

I asume you are saying that your code works, but I'm asking why it needs to be done this way.
Anyway you should try to schedule a meeting with Jim, Tim  and I.

dyninstAPI/h/BPatch_snippet.h
```diff
@@ -386,8 +386,7 @@ class DYNINST_EXPORT BPatch_variableExpr : public BPatch_snippet
 						 std::string name,
 						 void* offset,
 						 BPatch_type* type);
-    
-    
+
```
bbiiggppiigg, line 390: Remove redundant space.

rochauha:
Resolved @bbiiggppiigg's comment. Rebased on master.

rochauha:
Dropped 29b3e9f and rebased on master.

rochauha:
Rebased on master and removed redundant files.

rochauha:
This PR is ready for review.

rochauha:
@hainest @kupsch @bbiiggppiigg, this PR is ready for review.

hainest:
@rochauha Please merge in master to get the latest CI fixes.

rochauha:
Merged in master. This branch is now ready for review.

dyninstAPI/src/opcode.h
```diff
@@ -36,6 +36,7 @@
 typedef enum { 
    invalidOp,
    plusOp,
+   atomicAddOp,
```
bbiiggppiigg: Where is this thing used?

bbiiggppiigg (changes requested):
I don't have much comments other than that it seems like you decide to reuse existing opcode and encapsulate in BPatch_atomcOperationStmt.

I'm not entirely sure whether it would be better to have use a dedicated atomic versions of the opcodes (atomicAddOp vs plusOp).

@hainest @kupsch  What are your thoughts on this.

And in the case that we decide to reuse existing opcode, please remove the unused atomicAddOp.

dyninstAPI/h/BPatch_addressSpace.h
```diff
@@ -408,7 +411,9 @@ class DYNINST_EXPORT BPatch_addressSpace {
   // Returns true if the underlying image represents a 
   // statically-linked executable, false otherwise
   bool  isStaticExecutable();
-};
 
+  // THIS IS SPECIFIC TO AMDGPU
+  static std::set<BPatch_function *> instrumentedFunctions;
```
hainest: This can now be a non-static member variable in `AmdgpuGfx908PointHandler`.

rochauha: Done. Turns out I missed removing this. We already use the non-static member in `AmdgpuGfx908PointHandler`.

dyninstAPI/src/opcode.h
rochauha: I missed removing this! Done.

rochauha:
Addressed comments from @bbiiggppiigg and @hainest. Merged current master into this branch.

Please let me know if you have further comments.

rochauha:
@hainest, @kupsch, @bbiiggppiigg, please let me know if there are any changes you'd suggest. If not, then it would really help me if this PR gets merged sooner rather than later.

hainest (approved):
All tests are passing. LGTM. Jim is out until Monday, but we should wait for him to review.

dyninstAPI/src/RegisterConversion-amdgpu.C
```diff
@@ -0,0 +1,1247 @@
...
+
+MachRegister convertRegID(Register r, Dyninst::Architecture arch) {
+  // clang-format off
+  if(arch == Arch_amdgpu_gfx908) {
+    switch(r) {
+      case NS_amdgpu::sgpr0: return amdgpu_gfx908::s0;
+      case NS_amdgpu::sgpr1: return amdgpu_gfx908::s1;
+      case NS_amdgpu::sgpr2: return amdgpu_gfx908::s2;
+      case NS_amdgpu::sgpr3: return amdgpu_gfx908::s3;
+      case NS_amdgpu::sgpr4: return amdgpu_gfx908::s4;
```
kupsch: is there a reason that this switch statement is not replaced with a use of `regToMachReg32`? Seems very duplicative.

common/src/arch-amdgpu.C
```diff
@@ -0,0 +1,158 @@
...
+
+instruction *instruction::copy() const {
+  assert(!"Not implemented for AMDGPU");
+  return nullptr;
+}
+
+unsigned instruction::getTargetReg() const {
+  assert(!"Not implemented for AMDGPU");
+  return 0;
+}
```
kupsch: Comment out or delete all of the functions in this file that are unimplemented (the `assert` calling functions).  Right now they are run-time time bombs if they are used. Not defining them turns them into compile time errors if they are used and not implemented.

`isUncondBranch` is the only function used. It should be implemented or the function that uses it removed or somthing.

dyninstAPI/src/AmdgpuKernelDescriptor.C
```diff
@@ -0,0 +1,1011 @@
...
+  return kdRepr.kernel_code_entry_byte_offset;
+}
+
+void AmdgpuKernelDescriptor::setKernelCodeEntryByteOffset(int64_t value) {
+  kdRepr.kernel_code_entry_byte_offset = value;
+}
+
+#define GET_VALUE(MASK) ((fourByteBuffer & MASK) >> (MASK##_SHIFT))
+#define SET_VALUE(MASK) (fourByteBuffer | ((value) << (MASK##_SHIFT)))
+#define CLEAR_BITS(MASK) (fourByteBuffer & (~(MASK)))
```
kupsch: `CLEAR_BITS` should always be called before `SET_VALUE` (and is), so `CLEAR_BITS` should be removed and be incorporated into `SET_VALUE`

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -0,0 +1,231 @@
...
+
+  AstNodePtr prologueNodePtr =
+        boost::make_shared<AmdgpuPrologueNode>(prologuePtr);
+
+  AmdgpuPrologueSnippet prologueSnippet(prologueNodePtr);
+  insertPrologueAtPoints(prologueSnippet, entryPoints);
+}
+
+void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function) {
+  // If this function is a kernel, insert a s_dcache_wb instruction at its exit points.
```
kupsch, line 156: Eventually you'll need to insert an epilogue at all points before all instructions that cause the kernel to exit (`s_endpgm`, others?), not just the kernel function itself to handle the case of `s_endpgm` in other functions. Should this be fixed now, or add comments and open an issue?

dyninstAPI/src/ast-amdgpu.C
```diff
@@ -0,0 +1,3202 @@
...
+  // Now we have s[88:89] with address of the variable. Emit appropriate atomic instruction.
+  switch (opcode) {
+  case plusOp:
+    emitter->emitAtomicAdd(88, src0, gen);
+    break;
+  case minusOp:
+    emitter->emitAtomicSub(88, src0, gen);
+    break;
+  default:
+    assert(!"atomic operation for this opcode is not implemented");
```
kupsch: why are `divideOp` and `timesOp` implemented in the `BPatch_atomicOperationStmt` constructor, but not here?

dyninstAPI/src/ast-amdgpu.C
kupsch, line 1: Is this file largely a copy of ast.C? It seems like it is, but it was reformatted and lines were slightly rearranged. Why was there a need to make a copy, and how isn't this going to be a maintenance problem with two files and bigger with the reformatting.

dyninstAPI/src/RegisterConversion-amdgpu.C
rochauha: Got it. Used `regToMachReg32` in next patch.

common/src/arch-amdgpu.C
rochauha: Done.

dyninstAPI/src/ast-amdgpu.C
rochauha: `ast.C` implements lowering for the internal representation of ASTs.

As discussed earlier, other architectures redundantly implement code generation in different places (`emitA`, `emitR`, `emitT`, etc). `insnCodegen` contains some functionality. `Emitter` instances contain some functionality. Some of these are overlapping, others aren't.

Early on we decided that for AMDGPU, all codegen functionality should implemented in one place -- the AMDGPU emitter, with the `gfx-908-details` files doing all bit-twiddling. But existing AST-lowering code uses the other functions along with additional checks (see `emitA`, `emitR`, etc). Before host-codegen separation, this caused a lot of conflict with other architectures, and required a lot of conditional compilations. Hence the need for `ast-amdgpu.C`.

I think eventually the approach to code generation should be entirely architecture-specific i.e each architecture has its own pass over the internal ASTs, and generate code. These internal ASTs will be specific to architectures (like the placeholder-reg + offset AST only makes sense for architectures that have prologues, and hence doesn't show up elsewhere). In that case, each architecture should have its own `ast-<arch>.C` anyways. Many variants of *internal ASTs* won't be created at all for some architectures.

dyninstAPI/src/ast-amdgpu.C
rochauha: There is no atomic multiply instruction on gfx908. Atomic multiply can be simulated using critical sections.

There is no scalar division on gfx908. We need to use critical section and vector registers for this.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: I'll add a comment for now.

Only for `s_endpgm` because it causes the control to return to host. I'm not sure which other instructions would require the epilogue.

dyninstAPI/src/AmdgpuKernelDescriptor.C
rochauha:
```
+#define SET_VALUE(MASK) ((fourByteBuffer = fourByteBuffer & (~(MASK))), \
                                                   (fourByteBuffer | ((value) << (MASK##_SHIFT))))
````

Is the above ok? It introduces a side effect inside a macro.

File: dyninstAPI/src/ast-amdgpu.C
Comment (rochauha) [reply dyninstAPI/src/ast-amdgpu.C:1]: A lot of the contents of `ast-amdgpu.C` will be deleted once I have some benchmarks running. This is because many of the internal AST representations aren't relevant to AMDGPU, and will never show up in any of the snippets.

So for example, `RegOffset`,  `VariableAddr`, `VariableValue`, etc will never be created. So those parts in `ast-amdgpu.C` will be deleted. We instead use `AddressAsPlaceholderRegAndOffset` throughout for AMDGPU because we compute the base address in the prologue and store it in a register pair, which we then want to assign to the `PlaceHolderReg` subtree of the AST.

Other architectures don't ever need this for variables, so they won't implement it anyways.

**Comment by rochauha** [comment]
@kupsch, I have addressed all your comments (through responses, and commits). Please let me know if you have any more.

**Comment by rochauha** [comment]
@kupsch, I have updated the ast-amdgpu.C to be the unformatted version. I have only kept the changes that are needed for atomic operation statements.

File: dyninstAPI/src/AmdgpuKernelDescriptor.C [hunk 16]
```diff
@@ -0,0 +1,1011 @@
...
+  assert(isGfx90aOr942());
+  const uint32_t fourByteBuffer = kdRepr.compute_pgm_rsrc3;
+  return GET_VALUE(COMPUTE_PGM_RSRC3_GFX90A_TG_SPLIT);
+}
+
+void AmdgpuKernelDescriptor::setCOMPUTE_PGM_RSRC3_TgSplit(bool value) {
+  assert(isGfx90aOr942());
+  uint32_t fourByteBuffer = kdRepr.compute_pgm_rsrc3;
+  fourByteBuffer = CLEAR_BITS(COMPUTE_PGM_RSRC3_GFX90A_TG_SPLIT);
+  kdRepr.compute_pgm_rsrc3 = SET_VALUE(COMPUTE_PGM_RSRC3_GFX90A_TG_SPLIT);
```
kupsch: You can remove of the macro functions and all uses of `CLEAR_BITS`, `SET_VALUE` and `CHECK_WIDTH`. Then you can replace them with an inlined function that has no side effects and a macro function to get the mask shift value:

```cpp
static inline uint32_t insert_field(uint32_t op, uint32_t value, uint32_t mask, uint32_t shift)
{
    assert(value >> shift && "value contains more bits than specified");
    op &= ~mask;
    op |= value << shift;
    return op;
}
#define INSERT_FIELD(op,value,mask) insert_field(op, value, mask, mask##_SHIFT)
```
Then you use the following idiom where the old macros were used:
```suggestion
  assert(isGfx90aOr942());
  kdRepr.compute_pgm_rsrc3 = INSERT_FIELD(kdRepr.compute_pgm_rsrc3, value, COMPUTE_PGM_RSRC3_GFX90A_ACCUM_OFFSET);
```
This turns 3 calls into 1 and get rid of macros with side effects.

dyninstAPI/src/ast-amdgpu.C
kupsch: I'm still not sure this file shouldn't be merged/unified with ast.C. Conceptually I don't see why AMD GPUs are that different that they require a whole new file. This will also complicate having all CODEGEN_ARCHs compiled so that the build time option can go away. I'd be OK with this for now, but eventually I think it is going to need to change.

@hainest, @bbiiggppiigg any thoughts?

rochauha:
@kupsch, I have simplified bit manipulation based on your suggestion.

dyninstAPI/src/ast-amdgpu.C
rochauha: I'd still vote for having a separate ast-amdgpu.C for the near future. When we introduce per wave / lane instrumentation, we will need different register allocation / tracking mechanisms throughout, that other architectures simply don't have or need right now. Conditional compilation in selected places will mess up already messy ast-lowering code.

Once we have a clearer picture, we can decide how a unified code base could look like.

hainest: I had mostly skipped over this file when I did my review because I assumed it was purely gpu stuff that I didn't have enough knowledge about to effectively comment on. I see now that was not the case. The complete duplication of the `AstNode*` members is a massive maintenance hazard and a hard no-go. Mea culpa on that one.

It looks like the most substantive change is the addition of a real definition for `AstAtomicOperationStmtNode::generateCode_phase2`. I see no particular reason that couldn't be  moved into `ast.C` at this point.

---

>Once we have a clearer picture, we can decide how a unified code base could look like.

At the time we are deciding on expanding the ast stuff, we'll do it for all architectures (otherwise we're making technical debt for future devs which will likely still the the same folks here). In my opinion, *that* would be the easier code-change point to extract out the changes here. Plus, if those other ast types never come to pass, the changes here will make life unnecessarily harder down the road.

rochauha: @hainest, @kupsch, I can move the current code in `ast.C` with some conditional compilation if you want.

But I believe that things become messy when we introduce a type system and AMDGPU-specific code generation in `generateCode_phase2` of relevant ASTs. The messy parts aren't seen here because atomic operation is a single self-contained AST node with no sub-operations.

Many of these internal ASTs don't make sense on AMDGPUs and will not show up at all. So we'd want to have an error message for them or an assertion, and ensure that the parts of `BPatch` not meant for GPUs aren't used on GPUs. But that eventually introduces `#ifdef`s in different places.

Consider `AstOperatorNode` for example. `AstOperatorNode::generateCode_phase2` is already over 500 loc - [1266-1791](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/ast.C#L1266).

When extended with types, and concepts like vector + scalar = vector, the implementation will be purely emitter-based code generation, which only works for AMDGPU, because other architectures implement code generation in different places. Further, since we require register pairs, or single registers depending on our operands, we also end up allocating, freeing, and passing around registers differently. All this will introduce heavy conditional compilation and makes the code hard to follow.

Another case - relational operations have implicit destination registers on AMDGPU, and they differ based on whether it is a vector or scalar relational operation. So this will again introduce some bespoke logic that goes under an #ifdef, because other architectures deal with relational operations just like arithmetic operations.

Right now we have gfx908 only. If other AMDGPU architectures are added, with with slight quirks of their own, it would be cleaner to make those changes in `ast-amdgpu.C` instead of further cluttering `ast.C`.

A separate `ast-amdgpu.C` allows to cleaning up all the irrelevant clutter, and have codegen for AMDGPU only. A similar approach could be taken for other architectures as well, if necessary.

hainest: > I can move the current code in ast.C with some conditional compilation if you want.

Please do. Even if it turns out to be a pain later, it will at least save everyone from the maintenance nightmare that is duplicating the other ~3k loc in the interim.

>Many of these internal ASTs don't make sense on AMDGPUs and will not show up at all. ... But that eventually introduces #ifdefs in different places.

It requires an actual structure to the codegen/emitter classes instead of the tangled web of conditional compilation that is currently in place. I'm happy to start on some designs for that once we merge the riscv codegen.

> When extended with types, and concepts like vector + scalar = vector, ... All this will introduce heavy conditional compilation and makes the code hard to follow.

There are several design patterns that may be useful here: [Template Method](https://refactoring.guru/design-patterns/template-method), [Composite](https://refactoring.guru/design-patterns/composite), and [Decorator](https://refactoring.guru/design-patterns/decorator). They are essentially more powerful forms of the shim stuff we've been doing.

As you work more on expanding the AST types, you'll want to keep everyone updated on your design ideas. You may come up with something that could solve issues across all the architectures. Plus, I don't think anyone wants another four-month code review.

rochauha: @kupsch, @hainest -- I have moved AMDGPU codegen into `ast.C` with conditional compilation.

This hopefully shows how things will start looking further into development. For every internal AST that AMDGPU supports, there will be a `#ifdef`. For other ASTs, we either introduce an `#ifdef` with an assert saying that it isn't supported on AMDGPU, or we see some failure later down the codegen path.

A separate file would mean I could simply remove code for irrelevant ASTs to AMDGPU and simply have an assert. This would have been similar to how `inst-<arch>`, `codegen-<arch>`, etc files are today.

The separate `ast-amdgpu.C` was mostly a copy, and that's because I wanted to avoid preemptively cleaning it up too much, only to add a few bits and pieces back later. Since the separate file was only compiled for AMDGPU, there wouldn't be much to maintain for non-AMDGPU case on that file as it would be same as what it is now in `ast.C` i.e things breaking in combined `ast.C` and `ast-amdgpu.C` are equivalent to a "not implemented for AMDGPU" assertion that I would eventually add to `ast-amdgpu.C` to replace the AMDGPU-irrelevant portions of code.

hainest (approved):

dyninstAPI/src/ast-amdgpu.C
hainest: > I have moved AMDGPU codegen into ast.C with conditional compilation.

Thanks! All tests are passing, so I'm good to merge when @kupsch and @bbiiggppiigg are.

> A separate file would mean I could simply remove code for irrelevant ASTs to AMDGPU and simply have an assert. This would have been similar to how inst-<arch>, codegen-<arch>, etc files are today.

The difference is that `ast-amdgpu.C` is ~99% similar to `ast.C` whereas the inst/codegen/arch files are unique- even when they define the same functions. A great long time ago, we talked about how we could reconcile this with minimal conditional compilation, but we've ended up here. Of course, the real solution is making clean interfaces for the emitters and instrumenters, then making a class hierarchy from them. As I noted, I plan to extend Hsuan-Heng's work in #1970 once we get RISC-V codegen merged. That will just put to rest a lot of angst here.

rochauha:
@kupsch, @bbiiggppiigg -- ping

dyninstAPI/src/ast.C
```diff
@@ -1815,11 +1817,22 @@ bool AstOperandNode::generateCode_phase2(codeGen &gen, bool noCost,
    int len;
    BPatch_type *Type;
    switch (oType) {
+#if defined (DYNINST_CODEGEN_ARCH_AMDGPU_GFX908)
+   case operandType::Constant: {
```
bbiiggppiigg, line 1821: I have no issue other than the following

I don't know if it is better or worse to have 
1. a single case for operandType::Constant and split the logic into two with ifdefs
2. or two cases  for operandType::Constant  that are both guarded by ifdefs

@hainest @kupsch Any thoughts on this?

hainest: @bbiiggppiigg I think it's fine as-is. I've already started expanding your PR to make codegen architecture-independent to put an end to this whole bothersome mess.

PR #1988, opened 2025-09-09
Convert docs to rst

kwryankrattiger:
This is an attempt to to revive the efforts of #1653 

The big difference in this PR is using the XML generator from Doxygen to automatically update the API docs allowing the API docs to live next to the code as well as on the RTD page.

CC: @hainest


kwryankrattiger:
Closing, doxygen/breathe approach not desired by project.

https://github.com/dyninst/dyninst/pull/1990#issuecomment-3272328395

PR #1989, opened 2025-09-09
Create SECURITY.md

kwryankrattiger:


SECURITY.md
```diff
@@ -0,0 +1,25 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
```
hainest, line 5: Releases are tags on the master branch, so it's not possible to backport patches.

SECURITY.md
```diff
@@ -0,0 +1,25 @@
...
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+You can report a vulnerability using GitHub's private reporting
+feature:
+
+1. Go to [github.com/dyninst/dyninst/security](https://github.com/dyninst/dyninst/security).
```
hainest, line 20: It looks like that needs some configuration before it will work. In particular, the "Report a Vulnerability" isn't enabled for the repo. I don't have any objections to turning it on.

hainest (commented):
In the six years I've worked on Dyninst, I can't recall anyone submitting a change that was explicitly denoted as a security vulnerability. There have definitely been cases where code was fixed that could have led to a vulnerability (e.g., DoS). For the most part, we've been internally dilligent about removing as many standard vulnerability bugs (e.g., buffer overflow, UB, etc.) as possible.

That said, I welcome setting up explicit vulnerability reporting.

SECURITY.md
kwryankrattiger: Has anyone had a chance to look at this recently? Alternatively it could just be an email, but the github feature lets you reuse existing ACLs for the project once the button is clicked.

SECURITY.md
kwryankrattiger: Yeah this is just a placeholder. It could also simply be "we will only address issues related to the latest release" sort of thing.

hainest:
Replaced by #2360

PR #1990, opened 2025-09-09
Add doxygen

kwryankrattiger:
Incremental step to automating generation of API docs consistent with code base inline comments.

Extracted from #1988

CC: @hainest @kupsch


hainest:
We don't want to use Doxygen at all. The existing Doxygen comments are wholly incomplete and even outright wrong. That was a huge part of #1653.

kwryankrattiger:
Ah this had me confused then, my understanding was there needed to be a way to force them into sync. Will close this and the other PR then.

PR #1991, opened 2025-09-10
Add SoftwareException instruction category

bbiiggppiigg:
Rename isAbort to SoftwareExcpetion, and add the corresponding instructions that fits this category.
Also remove isAbort/isSoftwareException implementation for AMDGPU in IA_IAPI.

### DIFF
--- a/instructionAPI/src/InstructionCategories.C
+++ b/instructionAPI/src/InstructionCategories.C
@@ -122,7 +122,9 @@ namespace Dyninst { namespace InstructionAPI {
       case e_ud2:
       case aarch64_op_brk:
       case aarch64_op_hlt:
-        return c_AbortInsn;
+      case aarch64_op_wfe_hint:
+      case aarch64_op_wfi_hint:
+        return c_SoftwareExceptionInsn;
       default:
       	return c_NoCategory;
     }
--- a/instructionAPI/src/InstructionCategories.C
+++ b/instructionAPI/src/InstructionCategories.C
@@ -117,7 +117,8 @@ namespace Dyninst { namespace InstructionAPI {
       case e_syscall:
       case e_int:
         return c_SyscallInsn;
-
+      case e_hlt:
+        return c_AbortInsn;
       default:
       	return c_NoCategory;
     }
--- a/instructionAPI/src/InstructionCategories.C
+++ b/instructionAPI/src/InstructionCategories.C
@@ -118,6 +118,10 @@ namespace Dyninst { namespace InstructionAPI {
       case e_int:
         return c_SyscallInsn;
       case e_hlt:
+      case e_int3:
+      case e_ud2:
+      case aarch64_op_brk:
+      case aarch64_op_hlt:
         return c_AbortInsn;
       default:
       	return c_NoCategory;
--- a/dyninstAPI/src/Relocation/CFG/RelocBlock.C
+++ b/dyninstAPI/src/Relocation/CFG/RelocBlock.C
@@ -329,6 +329,14 @@ void RelocBlock::createCFWidget() {
    else {
       cfWidget_ = CFWidget::create(origAddr_);
    }
+   // FuncExit instrumentation will insert instrumentation before
+   // the last instruction of the block, thus we need to prevent a
+   // dummy CFWidget from being added as the last instruction of 
+   // the block. 
+   // TODO: Maybe this should be done for all endblocks that 
+   // doesn't end with CF instruction.
+   if(insn.isAbort())
+      return;
    elements_.push_back(cfWidget_);
 }
 
--- a/dyninstAPI/src/Relocation/CFG/RelocBlock.C
+++ b/dyninstAPI/src/Relocation/CFG/RelocBlock.C
@@ -335,7 +335,7 @@ void RelocBlock::createCFWidget() {
    // the block. 
    // TODO: Maybe this should be done for all endblocks that 
    // doesn't end with CF instruction.
-   if(insn.isAbort())
+   if(insn.isSoftwareException())
       return;
    elements_.push_back(cfWidget_);
 }
--- a/dyninstAPI/src/Relocation/CFG/RelocBlock.C
+++ b/dyninstAPI/src/Relocation/CFG/RelocBlock.C
@@ -329,14 +329,6 @@ void RelocBlock::createCFWidget() {
    else {
       cfWidget_ = CFWidget::create(origAddr_);
    }
-   // FuncExit instrumentation will insert instrumentation before
-   // the last instruction of the block, thus we need to prevent a
-   // dummy CFWidget from being added as the last instruction of 
-   // the block. 
-   // TODO: Maybe this should be done for all endblocks that 
-   // doesn't end with CF instruction.
-   if(insn.isSoftwareException())
-      return;
    elements_.push_back(cfWidget_);
 }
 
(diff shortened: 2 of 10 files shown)


bbiiggppiigg:
@hainest @kupsch @rochauha 
The solution turns out to be even simpler than I originally proposed.
Please let me know if you see any problem.

instructionAPI/src/InstructionCategories.C
```diff
@@ -117,7 +117,8 @@ namespace Dyninst { namespace InstructionAPI {
       case e_syscall:
       case e_int:
         return c_SyscallInsn;
-
+      case e_hlt:
+        return c_AbortInsn;
```
kupsch: To be consistent with `IA_IAPI::isAbort()`, I'd update this to include other instructions, and add the aarch64 instructions:

```suggestion
      case e_hlt:
      case e_int3:
      case e_ud2:
      case aarch63_op_brk:
      case aarch64_op_hlt:
        return c_AbortInsn;
```

Then you could change `IA_IAPI::isAbort()` to just call `isAbort()`.

bbiiggppiigg: Suggestion impmented.

bbiiggppiigg:
@hainest @kupsch
 I'm seeing CI failure related to the set up of ubuntu 24.10 runners, which I don't think is caused by any changes in this PR.

instructionAPI/src/InstructionCategories.C
```diff
@@ -117,7 +117,12 @@ namespace Dyninst { namespace InstructionAPI {
       case e_syscall:
       case e_int:
         return c_SyscallInsn;
-
+      case e_hlt:
```
hainest, line 120: We don't currently support it, but the PowerPC `stop` instruction should be added to this list. We should make an issue for that.

Jim and I talked about this yesterday, but `int3` doesn't have the same semantics as `hlt`, so it's unclear why it's in this list. `ud2` and its friends `ud0` and `ud1` are also not quite the same as `htl` because they just raise an exception that the OS has to handle. It _could_ be the case that the OS stops running your program, but `SIGILL` can be caught, so the user code would really determine what happens.

Taken together, I _think_ what `isAbort` is really trying to say is that "this instruction will unconditionally halt the current instruction execution stream, but could be resumed later." If that's the case, then it should be renamed. There are also a litany of additional instructions that should be added here (ppc has a bunch of thread-control instructions like `wait`). I'm fine leaving these updates to a separate PR.

hainest:
> @hainest @kupsch I'm seeing CI failure related to the set up of ubuntu 24.10 runners, which I don't think is caused by any changes in this PR.

That's weird, but definitely not a result of this PR. I'm guessing since 25.10 is out, they've marked 24.10 as not LTS. I'll try rebuilding the containers and see what happens.

hainest:
One other item: we really need a way of testing this; at least for x86. I was thinking we could have a mutator that inserts a call to `printf` into a mutatee that is just

```c++
int main() {
  __asm__ volatile ("hlt");
}
```

bbiiggppiigg:
> One other item: we really need a way of testing this; at least for x86. I was thinking we could have a mutator that inserts a call to `printf` into a mutatee that is just
> 
> ```c++
> int main() {
>   __asm__ volatile ("hlt");
> }
> ```

Do you mean to add this test to testsuite?
I've tested this locally with inline asm, and also there is default call to hlt in _start.

```asm
0000000000400420 <_start>:
  400420: f3 0f 1e fa                   endbr64
  400424: 31 ed                         xorl    %ebp, %ebp
  400426: 49 89 d1                      movq    %rdx, %r9
  400429: 5e                            popq    %rsi
  40042a: 48 89 e2                      movq    %rsp, %rdx
  40042d: 48 83 e4 f0                   andq    $-0x10, %rsp
  400431: 50                            pushq   %rax
  400432: 54                            pushq   %rsp
  400433: 49 c7 c0 90 05 40 00          movq    $0x400590, %r8          # imm = 0x400590
  40043a: 48 c7 c1 20 05 40 00          movq    $0x400520, %rcx         # imm = 0x400520
  400441: 48 c7 c7 0e 05 40 00          movq    $0x40050e, %rdi         # imm = 0x40050E
  400448: ff 15 a2 0b 20 00             callq   *0x200ba2(%rip)         # 0x600ff0 <__libc_start_main@@GLIBC_2.2.5+0x600ff0>
  40044e: f4                            hlt
```

instructionAPI/src/InstructionCategories.C
bbiiggppiigg: For instrumentation purpose, the important things is that whether the instruction after hlt/abort could ever be executed. If there is any chance that the code after hlt/abort could not be executed, the instrumentation needs to happen before such instruction.

Conditionality of the block ending instruction seems to be a separate issue. I believe the Instrumentor logic in dyninstAPI/Relocation/Transform handles the case for conditional return (in powerPC?) as a separate case from unconditional return (ret/hlt/s_endpgm).

hainest:
> > One other item: we really need a way of testing this; at least for x86. I was thinking we could have a mutator that inserts a call to `printf` into a mutatee that is just
> > ```c++
> > int main() {
> >   __asm__ volatile ("hlt");
> > }
> > ```
> 
> Do you mean to add this test to testsuite?

Oh no. Adding tests to the testsuite is a project far in the future. This would be something to add as a [regression](https://github.com/dyninst/dyninst/tree/master/tests/regression) test.

>I've tested this locally with inline asm

We need something that can be run in CI.

>also there is default call to hlt in _start.

But we need to test instrumentation insertion, and `_start` isn't a great place for that; though I suppose we could. It's likely easier to just make a tiny mutatee.

instructionAPI/src/InstructionCategories.C
bbiiggppiigg: > We don't currently support it, but the PowerPC `stop` instruction should be added to this list. We should make an issue for that.
> 
> Jim and I talked about this yesterday, but `int3` doesn't have the same semantics as `hlt`, so it's unclear why it's in this list. `ud2` and its friends `ud0` and `ud1` are also not quite the same as `htl` because they just raise an exception that the OS has to handle. It _could_ be the case that the OS stops running your program, but `SIGILL` can be caught, so the user code would really determine what happens.
> 
> Taken together, I _think_ what `isAbort` is really trying to say is that "this instruction will unconditionally halt the current instruction execution stream, but could be resumed later." If that's the case, then it should be renamed. There are also a litany of additional instructions that should be added here (ppc has a bunch of thread-control instructions like `wait`). I'm fine leaving these updates to a separate PR.

I think as long as there is a chance that the origin block ending instruction could cause the instruction that follows it to not be executed, the instrumentation should happen before that instruction.

So what is your suggestion for this PR?
Should I remove the int3, ud2?
Should I add arch64_op_wfi / wfe?
For PowerPC, I don't see any like power_op_stop.
And should I wait for this CI to pass for 24.10.

hainest: > I think as long as there is a chance that the origin block ending instruction could cause the instruction that follows it to not be executed, the instrumentation should happen before that instruction.

Agreed. But we should give `isAbort` a better name. Maybe `isTerminal`?

> So what is your suggestion for this PR?

1. Figure out a better name for `isAbort`
2. Add CI tests

> Should I remove the int3, ud2?

_I_ would argue we should remove `int3` and add `ud0/1`. But I want @kupsch to weigh in.

> Should I add arch64_op_wfi / wfe?

TBD. Let's talk with Jim.

> For PowerPC, I don't see any like power_op_stop.

We don't support it.

> And should I wait for this CI to pass for 24.10.

No. I'll fix that later.

PR #1994, opened 2025-09-16
Fix FuncExit instrumentation for software exceptions

bbiiggppiigg:
Change the logic for handling ending instruction of RelocBlock.

For FuncExit instrumentaiton, Dyninst will automatically insert instruction before the last instruction of the exit Block.  
The problem is that, for instructions that is not known to have cotnrol flow, (software exception instructions, for example), Dyninst will insert a dummy CFWidget, thus the instrumentation was inserted between the abort instruction and the dummy CFWidget.

This PR prevent the dummy CFWidget from being inserted if we know the last instruction is a software exception.

This is the old #1991


bbiiggppiigg:
@kupsch @hainest 
I believe the PR is ready for review now.

kupsch (approved):

PR #1995, opened 2025-09-18
Docker: remove Ubuntu 24. Add Ubuntu 25, Fedora 42, 43

hainest:
I have no idea why clang-16 has no package for Ubuntu 25. I haven't checked for a ppa, but we still have it in Fedora 37.

Ubuntu 25.10 is the same set of compilers, but has glibc 2.42. I didn't think it was really worthwhile to add it since Fedora 43 is effectively that same case.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -261,13 +261,15 @@ Currently, this is only set up to be run manually. In the future, it can be run 
 
 2. Add another entry to the `configs` dictionary in the config [script](scripts/compiler_configs.py) following the existing naming convention. There are several workflows that check for names of the format `<OS>-<version>`.
 
-4. Add any special handling (e.g., building dependencies from source) needed in **both** `docker/Dockerfile.<OS>` and the `build-base` job in `.github/workflows/refresh-containers.yaml`.
+3. Add any special handling (e.g., building dependencies from source) needed in **both** `docker/Dockerfile.<OS>` and the `build-base` job in `.github/workflows/refresh-containers.yaml`.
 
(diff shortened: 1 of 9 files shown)


hainest:
It looks like I mistakenly disabled the workflow for testing C++ standards versions back in October of last year (#1793). Yikes.

.github/workflows/README.md
```diff
@@ -261,27 +261,13 @@ Currently, this is only set up to be run manually. In the future, it can be run
 
 2. Add another entry to the `configs` dictionary in the config [script](scripts/compiler_configs.py) following the existing naming convention. There are several workflows that check for names of the format `<OS>-<version>`.
 
-3. Create a new base and dev container. For example to add ubuntu-X,
```
kupsch, line 264: Although not necessary (the markdown formatter doesn't care), to avoid confusion change the item number 4-7 to 3-6.  The same issue with the numbered list on line 236, and while you're there the unordered list item on line 237 should get indented one more space to format properly.

hainest: I swear I know how to count to 10...

hainest:
The spack build workflow is failing because libiconv-1.17 has a compilation issue with gcc-15 on Fedora 43. I'll see if I can fix the spack package for it. For now, we can ignore it.

hainest:
I have patches posted for the spack build failures (https://github.com/spack/spack-packages/pull/1631, https://github.com/spack/spack-packages/pull/1616), so they should be fixed soon. I'll merge now since the pr-tests workflow is now working.

PR #1996, opened 2025-09-18
Remove backend for unused BPatch statistics

hainest:
This should have been part of fbdf09f329
(https://github.com/dyninst/dyninst/pull/1910).


bbiiggppiigg (approved):
The stats collected by setting 
DYNINST_STATS_INST
seems useful.

But since there is no call to init_stats (maybe removed in some previous commit), I guess people don't really use it.

hainest:
> The stats collected by setting DYNINST_STATS_INST seems useful.
> 
> But since there is no call to init_stats (maybe removed in some previous commit), I guess people don't really use it.

As mentioned in the commit message for #1910, the call to `init_stats` was removed back in 2005. I don't know why, though. This is a frequent pattern I've seen where the top-level usage is removed, but all of the backend code is left to rot. All without any explanation, of course.

> The stats collected by setting DYNINST_STATS_INST seems useful.

It's not immediately obvious to me that only reporting stats is of general use. Does it matter that you inserted 3 instances? I would rather know _what_ those three instances were. Maybe that's just me.

PR #1998, opened 2025-09-18
Enable regression testing of known binary files

hainest:
Regression tests can now be created for binaries stored in https://github.com/dyninst/test_binaries. For example, a simple `Symtab::openFile` test would be:

```c++
#include "Symtab.h"

#include <cstdlib>
#include <iostream>

int main(int argc, char** argv) {
  if(argc != 2) {
    std::cerr << "Usage: " << argv[0] << " file\n";
    return EXIT_FAILURE;
  }

  namespace st = Dyninst::SymtabAPI;
  st::Symtab* s{};
  if(!st::Symtab::openFile(s, argv[1])) {
    std::cerr << "Unable to open file '" << argv[1] << "'\n";
    return EXIT_FAILURE;
  }
  return EXIT_SUCCESS;
}
```

```cmake
add_executable(parseapi_test1 parseapi_test1.cpp)
target_link_libraries(parseapi_test1 PRIVATE symtabAPI)
target_compile_options(parseapi_test1 PRIVATE ${SUPPORTED_CXX_WARNING_FLAGS})
add_test(NAME parseapi_test1_test COMMAND parseapi_test1 "${DYNINST_TEST_BINARIES_DIR}/parseAPI/test1")
set_tests_properties(parseapi_test1_test PROPERTIES LABELS "regression" )
```

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -115,7 +115,7 @@ Clang-specific libraries are installed automatically. For example, `compiler: 'c
 
 ## Workflows
 
-### Regression tests aginst known binaries ([workflows/binary-tests](workflows/binary-tests.yaml))
+### Regression tests against known binaries ([workflows/binary-tests](workflows/binary-tests.yaml))
 
 *Runs all tests, including those that require access to the dyninst/binary_tests repo*
 


hainest:
@kupsch @bbiiggppiigg We'll want to have some real test here before merging. This PR just sets the groundwork for pulling in the binary-tests repo.

kupsch (approved):

bbiiggppiigg:
> @kupsch @bbiiggppiigg We'll want to have some real test here before merging. This PR just sets the groundwork for pulling in the binary-tests repo.

How should the binaries be organized in that repository?
Should it match the examples that are using them?
Or should it be organized in architecture?

I think grouping based on architecture makes more sense since that allows different examples to make use of the same architectures.

hainest:
> > @kupsch @bbiiggppiigg We'll want to have some real test here before merging. This PR just sets the groundwork for pulling in the binary-tests repo.
> 
> How should the binaries be organized in that repository? Should it match the examples that are using them? Or should it be organized in architecture?
> 
> I think grouping based on architecture makes more sense since that allows different examples to make use of the same architectures.

A superlative question that I really should have enumerated before posting...

We definitely want to organize by architecture at the top level. From there, it's a bit fuzzy. If we organize by toolkit (e.g., symtabapi, parseapi, etc.), then we run into issues of a single binary that demonstrates issues in multiple toolkits, but exists in only one directly. We could organize by Dyninst GitHub issue, but that precludes adding binaries that don't have associated issues (maybe we just force creating an issue before adding a binary?). I don't think there is a rigid structure that will capture a wide range of use cases. Naming conventions for the files could also be an issue. For example, if we have multiple `libfoo.so` that demonstrate different issues in a single toolkit, how do we represent that? No easy answers here.

bbiiggppiigg:
> superlative

I'm thinking that maybe we can just have, under the top-level architecture, an auto-increment ID as its filename, alongside a documentation on the properties of the binary (where did we get this binary, any special property of the binary, and maybe which example/issue the bianry is asociated it with or to be test for).

hainest:
> > superlative
> 
> I'm thinking that maybe we can just have, under the top-level architecture, an auto-increment ID as its filename, 

That wouldn't work for an issue that requires multiple DSOs where one links to another. I admit I've not seen such a case, but I don't think we should preclude it.

>alongside a documentation on the properties of the binary (where did we get this binary, any special property of the binary, and maybe which example/issue the bianry is asociated it with or to be test for).

Definitely agree.

hainest:
I've successfully run this with a dummy test. It should be ready to use for others.

PR #2006, opened 2025-09-24
[AMDGPU] Add emitter implementation for gfx908

rochauha:
Emitter portion of #1987.

### DIFF
--- a/dyninstAPI/src/amdgpu-gfx908-details.C
+++ b/dyninstAPI/src/amdgpu-gfx908-details.C
@@ -69,14 +69,14 @@ void emitSop1(unsigned opcode, Register dest, Register src0, bool hasLiteral, ui
   setDstSop1(dest, newRawInst);
   setSrc0Sop1(src0, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
 
   if (hasLiteral) {
     append_memory_as(rawInstBuffer, literal);
   }
 
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SOP1 END ===
 
@@ -117,18 +117,18 @@ void emitSop2(unsigned opcode, Register dest, Register src0, Register src1, code
   setSrc1Sop2(src1, newRawInst);
   setSrc0Sop2(src0, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 
 void emitSop2WithSrc1Literal(unsigned opcode, Register dest, Register src0, uint32_t src1Literal,
                              codeGen &gen) {
   emitSop2(opcode, dest, src0, /* src1 = */ 255, gen);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, src1Literal);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SOP2 END ===
 
@@ -166,9 +166,9 @@ void emitSopC(unsigned opcode, Register src0, Register src1, codeGen &gen) {
   setSrc1SopC(src1, newRawInst);
   setSrc0SopC(src0, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SOPC END ===
 
@@ -207,9 +207,9 @@ void emitSopK(unsigned opcode, Register dest, int16_t simm16, codeGen &gen) {
   setDstSopK(dest, newRawInst);
   setSImm16SopK(simm16, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SOPK END ===
 
@@ -242,9 +242,9 @@ void emitSopP(unsigned opcode, int16_t simm16, codeGen &gen) {
   setOpcodeSopP(opcode, newRawInst);
   setSImm16SopP(simm16, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SOPP END ===
 
@@ -335,9 +335,9 @@ void emitSmem(unsigned opcode, uint64_t sdata, uint64_t sbase, uint64_t offset, 
   setR4Smem(newRawInst);
   setOffsetSmem(offset, newRawInst);
 
-  auto *rawInstBuffer = gen.cur_ptr();
+  auto rawInstBuffer = gen.cur_ptr();
   append_memory_as(rawInstBuffer, newRawInst);
-  gen.update((codeBuf_t *)rawInstBuffer);
+  gen.update(rawInstBuffer);
 }
 // === SMEM END ===
 
--- a/dyninstAPI/src/codegen.h
+++ b/dyninstAPI/src/codegen.h
@@ -156,6 +156,8 @@ class codeGen {
     // based on a new pointer
     void update(codeBuf_t *ptr);
 
+    void update(void *ptr) { update(reinterpret_cast<codeBuf_t*>(ptr)); }
+
     // Set the offset at a particular location.
     void setIndex(codeBufIndex_t offset);
 


rochauha:
Rebased on master.

rochauha:
@hainest - suggestions on how to test this?

hainest:
> @hainest - suggestions on how to test this?

Let me take a look at some of the test code I have in another branch, and I'll get back to you. If you don't hear from me by tomorrow afternoon, ping me here or send me an email.

hainest:
@rochauha I have an idea for how to do this. Could you send me the scripts you are using to do the manual comparison you mentioned in the meeting ?

rochauha:
> @rochauha I have an idea for how to do this. Could you send me the scripts you are using to do the manual comparison you mentioned in the meeting ?

I didn't use scripts. Early on, I worked on the emitter (well, a "dummy" emitter) for AMDGPU out-of-tree because rest of the components required to build for AMDGPU were not in place. I will send you an email with the details.

bbiiggppiigg:
Jim and I  were discussing whether we want to generate some of the header information (such as OP code for each encoding famility) from the XML files, and put them maybe in commom/src/arch-amdgpu.h.

hainest:
It turns out you can't test the emitter classes without the changes to the instrumentation classes- in particular `registerSpace`. That's reasonable since you need a set of registers to emit.

@kupsch @bbiiggppiigg I'm fine merging this and adding testing when extracting the requisite parts out of #1987.

bbiiggppiigg:
> It turns out you can't test the emitter classes without the changes to the instrumentation classes- in particular `registerSpace`. That's reasonable since you need a set of registers to emit.
> 
> @kupsch @bbiiggppiigg I'm fine merging this and adding testing when extracting the requisite parts out of #1987.

I'm fine with this as well.
We need this PR to rebased first?

hainest:
>We need this PR to rebased first?

No, there are no conflicts. It can be rebased-and-squashed.

rochauha:
@kupsch, any comments?

dyninstAPI/src/emit-amdgpu.C
```diff
@@ -0,0 +1,573 @@
...
+    assert(false && "size can only be 1, 2, or 4");
+  }
+
+  emitSmem(storeOpcode, source, (base >> 1), (uint64_t)offset, gen);
+}
+
+void EmitterAmdgpuGfx908::emitStoreShared(Register /* source */, const image_variable * /* var */,
+                                          bool /* is_local */, int /* size */,
+                                          codeGen & /* gen */) {
+  printf("emitStoreShared not implemented yet\n");
```
kupsch: Make all the `printf`'s that indicate an unimplemented function an assert with an error message for now using the idiom `assert(!"MESSAGE")`:
```suggestion
  assert(!"emitStoreShared not implemented yet\n");
```

dyninstAPI/src/emit-amdgpu.C
```diff
@@ -0,0 +1,573 @@
...
+}
+
+bool EmitterAmdgpuGfx908::emitMoveRegToReg(Register src, Register dest, codeGen &gen) {
+
+  emitSop1(S_MOV_B32, dest, src, /*hasLiteral =*/false, /*literal =*/0, gen);
+  return 0;
```
kupsch: For all bool returning functions use the constant `false` for a boolean value not `0`:
```suggestion
  return false;
```

rochauha:
@kupsch -- I've addressed your previous comments. Let me know if you have more comments.

dyninstAPI/src/emit-amdgpu.C
```diff
@@ -0,0 +1,573 @@
...
+    loadOpcode = S_LOAD_DWORDX4;
+    break;
+  case 8:
+    loadOpcode = S_LOAD_DWORDX8;
+    break;
+  case 16:
+    loadOpcode = S_LOAD_DWORDX16;
+    break;
+  default:
+    assert(false && "size can only be 1, 2, 4, 8 or 16");
```
bbiiggppiigg: try assert(!"your error string"), which Jim suggested.

rochauha: Done across all files in this codebase.

dyninstAPI/src/emit-amdgpu.C
```diff
@@ -0,0 +1,573 @@
...
+
+void EmitterAmdgpuGfx908::emitLoad(Register /* dest */, Address /* addr */, int /* size */,
+                                   codeGen & /* gen */) {
+  assert(!"emitLoad not implemented yet");
+}
+
+void EmitterAmdgpuGfx908::emitLoadConst(Register dest, Address imm, codeGen &gen) {
+  // Caller must ensure that dest is even; and dest, dest+1 are available.
+
+  assert(sizeof(Address) == 8); // must be a 64-bit address
```
hainest, line 215: This check is used in a lot of places throughout Dyninst, but it doesn't really indicate that an address is logically 64 bits. It should also really be a compile-time `static_assert`, but that's a separate issue. In general, should amdgpu codegen be disabled on i386? I wouldn't imagine there's a 32-bit ROCM. We don't need to change this here and now, but it's something to think about in the long-term.

rochauha:
Rebased on master.

dyninstAPI/src/emit-amdgpu.C
bbiiggppiigg: > This check is used in a lot of places throughout Dyninst, but it doesn't really indicate that an address is logically 64 bits. It should also really be a compile-time `static_assert`, but that's a separate issue. In general, should amdgpu codegen be disabled on i386? I wouldn't imagine there's a 32-bit ROCM. We don't need to change this here and now, but it's something to think about in the long-term.

I think this is becase we are generating AMDGPU code, which leverages 64 bit addresses.

rochauha: Yes. We have a 64-bit address and 32-bit registers on AMDGPU.

After that check, the lower and upper halves of the address are moved into a register pair.

hainest: Yes, I am aware of what this check is attempting to do. My point is that 1) it doesn't do that, and 2) it isn't logically correct.

The C++ standard does not guarantee that `unsigned long long` (the type of `Address`) can hold any more space than `unsigned char`. By extension, it doesn't require that an `unsigned char` can't be 64 bits. This makes the assert useless in determining whether or not `imm` actually represents a 64-bit value.

But what about

`emitLoadConst(reg, 0x1234, gen);`

Is the immediate a 64-, 32-, or 16-bit value? Is it an address? Would it be useful or required to change the instruction encoding depending on the intended operand size? None of these questions is answered by `assert`.

Further, it is logically incoherent to use the `sizeof` a Dyninst-level type to determine if it can be used to map to specific widths in hardware. This is especially true as we move to make code generation targets be independent of the host.

---

What the check here is doing is making sure that it's reasonable to perform the logical shifts. An explicit conversion to a `uint64_t` would guarantee that (by the C++ standard). I'm not asking that this change be made. It's a much wider problem throughout Dyninst that the size of builtins are tied to ISA/hardware operand widths that can't (and shouldn't) be solved here.

rochauha:
@kupsch, do you have any other comments?

dyninstAPI/src/emit-amdgpu.C
rochauha: I see what you mean.

I use `emitMovLiteral` to move a 32-bit literal into a register. By the signature, I think `emitLoadConst` should move an address into a register / register pair depending on the architecture.

I think the emitter interfaces should also be streamlined and cleaned up in the future.

rochauha:
Rebased on master, and made changes suggested by @kupsch.

dyninstAPI/src/amdgpu-gfx908-details.cpp
```diff
@@ -0,0 +1,344 @@
...
+              codeGen &gen) {
+  uint32_t newRawInst = 0xFFFFFFFF;
+  setEncodingSop1(newRawInst);
+  setFixedBitsSop1(newRawInst);
+  setOpcodeSop1(opcode, newRawInst);
+  setDstSop1(dest, newRawInst);
+  setSrc0Sop1(src0, newRawInst);
+
+  void *rawInstBuffer = gen.cur_ptr();
+  append_memory_as<void, uint32_t>(rawInstBuffer, newRawInst);
```
hainest: A (mostly) unimportant nit: you don't need to explicitly specify function template arguments unless you intend to override template type deduction.

In this case,

`append_memory_as<void, uint32_t>(rawInstBuffer, newRawInst);`

is the same as

`append_memory_as(rawInstBuffer, newRawInst);`

rochauha: Got it. I will update the PR.

dyninstAPI/src/amdgpu-gfx908-details.C
kupsch, line 1: In Dyninst C++ files use the `.C` extension instead of `.cpp` outside the test directory. Change this filename  to be consistent.

dyninstAPI/src/amdgpu-gfx908-details.cpp
```diff
@@ -0,0 +1,344 @@
...
+  setSrc0Sop1(src0, newRawInst);
+
+  void *rawInstBuffer = gen.cur_ptr();
+  append_memory_as(rawInstBuffer, newRawInst);
+  gen.update((codeBuf_t *)rawInstBuffer);
+
+  if (hasLiteral) {
+    append_memory_as(rawInstBuffer, literal);
+    gen.update((codeBuf_t *)rawInstBuffer);
+  }
```
kupsch: I think you can get rid of the casts, and only do one `gen.update`:

```suggestion
  auto rawInstBuffer = gen.cur_ptr();
  append_memory_as(rawInstBuffer, newRawInst);

  if (hasLiteral) {
    append_memory_as(rawInstBuffer, literal);
  }
  
  gen.update((codeBuf_t *)rawInstBuffer);
```

Do the same everywhere else this pattern appear.

dyninstAPI/src/amdgpu-gfx908-details.C
```diff
@@ -0,0 +1,344 @@
...
+
+void setOpcodeSop1(uint32_t value, uint32_t &rawInst) {
+  uint32_t mask = Mask_Sop1_Opcode;
+  rawInst = (rawInst & ~mask) | ((value << 8) & mask);
+}
+
+void setSrc0Sop1(uint32_t value, uint32_t &rawInst) {
+  uint32_t mask = Mask_Sop1_Src0;
+  rawInst = (rawInst & ~mask) | (value & mask);
+}
```
kupsch, line 61: There helper functions could be declared in an unnamed namespace, not in the header file, and be declared inline.  I'm OK with it as is, @hainest, what do you think.

rochauha:
Resolved few comments by @kupsch. Also rebased on master.

dyninstAPI/src/amdgpu-gfx908-details.C
```diff
@@ -0,0 +1,344 @@
...
+void emitSop1(unsigned opcode, Register dest, Register src0, bool hasLiteral, uint32_t literal,
+              codeGen &gen) {
+  uint32_t newRawInst = 0xFFFFFFFF;
+  setEncodingSop1(newRawInst);
+  setFixedBitsSop1(newRawInst);
+  setOpcodeSop1(opcode, newRawInst);
+  setDstSop1(dest, newRawInst);
+  setSrc0Sop1(src0, newRawInst);
+
+  auto *rawInstBuffer = gen.cur_ptr();
```
kupsch: No need the pointer in the type here so remove the `*` (fix identical code elsewhere):

```suggestion
  auto rawInstBuffer = gen.cur_ptr();
```

dyninstAPI/src/amdgpu-gfx908-details.C
```diff
@@ -0,0 +1,344 @@
...
+  append_memory_as(rawInstBuffer, newRawInst);
+
+  if (hasLiteral) {
+    append_memory_as(rawInstBuffer, literal);
+  }
+
+  gen.update((codeBuf_t *)rawInstBuffer);
```
kupsch: Let's get rid of the cast, by adding an overload of `codeGen::update` so it can take a `codeBuf_t*` parameter  (dyninstAPI/src/codegen.h after line 157):

```
    void update(void *ptr)  {update(reinterpret_cast<codeBuf_t*>(ptr));}
```

Then the cast in this line and others can be removed:
```suggestion
  gen.update(rawInstBuffer);
```

rochauha: Do you suggest doing it here or having another PR for this change?

kupsch: here.

rochauha:
Resolved comments by @kupsch.

dyninstAPI/src/amdgpu-gfx908-details.C
rochauha: There is only one instance where I need to use an internal function. In #1987, see `codegen-amdgpu.C`, line 109. In the function`insnCodeGen::modifyJcc`, `setSImm16SopP` is used to modify the immediate value in a conditional branch.

I would have preferred creating a new conditional branch instruction entirely, but that would require reading the opcode and would expose more details from `amdgpu-gfx908-details`.

PR #2007, opened 2025-09-25
GitHub CI: add builds for all codegen architectures

hainest:
Currently, GitHub CI only runs on x86_64, so we can't test all possible combinations of host and codegen architectures.

I'd like to merge this before #2006, so we can confirm it builds in CI.

PR #2008, opened 2025-09-25
Fix default implementation in emitElfStatic-stub

bbiiggppiigg:
Fix some of the errors in #2007.
Probably no longer need emitElfStatic-amdgpu in #1987

### DIFF
--- a/symtabAPI/src/emitElfStatic-stub.C
+++ b/symtabAPI/src/emitElfStatic-stub.C
@@ -90,7 +90,7 @@ bool emitElfStatic::isGOTRelocation(unsigned long) {
     return false;
 }
 
-Offset emitElfStatic::getGOTSize(LinkMap &) {
+Offset emitElfStatic::getGOTSize(Symtab *, LinkMap &, Offset &) {
     assert(!EMIT_STATIC_ASSERT);
     return 0;
 }
@@ -100,7 +100,7 @@ Offset emitElfStatic::getGOTAlign(LinkMap &) {
     return 0;
 }
 
-void emitElfStatic::buildGOT(LinkMap &) {
+void emitElfStatic::buildGOT(Symtab *, LinkMap &) {
     assert(!EMIT_STATIC_ASSERT);
 }
 
@@ -142,3 +142,20 @@ bool emitElfStatic::createNewDtorRegion(LinkMap &) {
 void emitElfStatic::getExcludedSymbolNames(set<string> &) {
     assert(!EMIT_STATIC_ASSERT);
 }
+
+
+Offset emitElfStatic::allocStubRegions(LinkMap &, Offset ) {
+    assert(!EMIT_STATIC_ASSERT);
+    return 0;
+}
+
+bool emitElfStatic::updateTOC(Symtab *, LinkMap &, Offset ) {
+    assert(!EMIT_STATIC_ASSERT);
+    return false;
+}
+bool emitElfUtils::updateRelocation(Symtab *, relocationEntry &, int) {
+    assert(!EMIT_STATIC_ASSERT);
+    return false;
+}
+
+


hainest:
I am curious why this is even needed. Shouldn't we require an implementation for _every_ architecture we support for doing static rewriting, even if we don't support rewriting statically-linked binaries (e.g., ARM)? The asserts are an old design choice that kind of informs the user that Dyninst doesn't support an action, but I guess it's ok? That's certainly a bigger change.

bbiiggppiigg:
> I am curious why this is even needed. Shouldn't we require an implementation for _every_ architecture we support for doing static rewriting, even if we don't support rewriting statically-linked binaries (e.g., ARM)? The asserts are an old design choice that kind of informs the user that Dyninst doesn't support an action, but I guess it's ok? That's certainly a bigger change.

I guess this is the default implementation for architectures that we haven't or don't know how to support yet.
If they are not called, then we don't have a problem, otherwise the user needs to implement those functions that asserts.

kupsch (approved):

hainest:
> > I am curious why this is even needed. Shouldn't we require an implementation for _every_ architecture we support for doing static rewriting, even if we don't support rewriting statically-linked binaries (e.g., ARM)? The asserts are an old design choice that kind of informs the user that Dyninst doesn't support an action, but I guess it's ok? That's certainly a bigger change.
> 
> I guess this is the default implementation for architectures that we haven't or don't know how to support yet.

Right, but my argument is that we should require the implementation of a `emitElfStatic-ARCH.C` for every architecture we support rewriting. It's ok not to have a real implementation (again, ARM). At the very least, the assert messages should be specialized. The stub should never be used.

>If they are not called, then we don't have a problem, otherwise the user needs to implement those functions that asserts.

And without a more uniform approach to error handling in Dyninst at large and Symtab in particular, that's probably the best we can do for now.

bbiiggppiigg:
> > > I am curious why this is even needed. Shouldn't we require an implementation for _every_ architecture we support for doing static rewriting, even if we don't support rewriting statically-linked binaries (e.g., ARM)? The asserts are an old design choice that kind of informs the user that Dyninst doesn't support an action, but I guess it's ok? That's certainly a bigger change.
> > 
> > 
> > I guess this is the default implementation for architectures that we haven't or don't know how to support yet.
> 
> Right, but my argument is that we should require the implementation of a `emitElfStatic-ARCH.C` for every architecture we support rewriting. It's ok not to have a real implementation (again, ARM). At the very least, the assert messages should be specialized. The stub should never be used.
> 
> > If they are not called, then we don't have a problem, otherwise the user needs to implement those functions that asserts.
> 
> And without a more uniform approach to error handling in Dyninst at large and Symtab in particular, that's probably the best we can do for now.

If we add these implementations for all 3 AMD GPU architectures, we can either have 6 dedicated files doing absolutely nothing (both the emitElfStatic-<arch> and some relocatio related file), or just have a default implementation that covers the base case so we can compile at the very least. 

I was thinking more like the IA_IAPI where we only implement needed functionality where we only implement something if it is actually needed. When we eventually turn these into base class and inheritence, I think the stub basically becomes the base class with empty implementation.

That being said, if you still think it is better that we have dedicated files with potentially empty implementation for newly added architecture, I'll just close this PR.

hainest:
> > > > I am curious why this is even needed. Shouldn't we require an implementation for _every_ architecture we support for doing static rewriting, even if we don't support rewriting statically-linked binaries (e.g., ARM)? The asserts are an old design choice that kind of informs the user that Dyninst doesn't support an action, but I guess it's ok? That's certainly a bigger change.
> > > 
> > > 
> > > I guess this is the default implementation for architectures that we haven't or don't know how to support yet.
> > 
> > 
> > Right, but my argument is that we should require the implementation of a `emitElfStatic-ARCH.C` for every architecture we support rewriting. It's ok not to have a real implementation (again, ARM). At the very least, the assert messages should be specialized. The stub should never be used.
> > > If they are not called, then we don't have a problem, otherwise the user needs to implement those functions that asserts.
> > 
> > 
> > And without a more uniform approach to error handling in Dyninst at large and Symtab in particular, that's probably the best we can do for now.
> 
> If we add these implementations for all 3 AMD GPU architectures, we can either have 6 dedicated files doing absolutely nothing (both the emitElfStatic- and some relocatio related file), or just have a default implementation that covers the base case so we can compile at the very least.
> 
> I was thinking more like the IA_IAPI where we only implement needed functionality where we only implement something if it is actually needed. When we eventually turn these into base class and inheritence, I think the stub basically becomes the base class with empty implementation.
> 
> That being said, if you still think it is better that we have dedicated files with potentially empty implementation for newly added architecture, I'll just close this PR.

Nah. I think it would be easier to move this to a set of classes than deal with this.

PR #2009, opened 2025-09-25
Fix string format warnings in emitElfStatic-ppc64

hainest:
These were found by building on an x86 host with
-DDYNINST_CODEGEN_ARCH=ppc64le.


kupsch (approved):

bbiiggppiigg:
Really wish we could upgrade to use std::print in C++23.

hainest:
> Really wish we could upgrade to use std::print in C++23.

Or at least use Boost.Format.

PR #2010, opened 2025-09-25
Fix unused function warning in x86-emit.C with i386 codegen on x86_64 host

hainest:
This was found by building on an x86 host with
-DDYNINST_CODEGEN_ARCH=i386.


bbiiggppiigg (approved):

kupsch:
The current title is confusing.  How about `Fix unused function warning with i386 codegen on x86_64 host`.

hainest:
> The current title is confusing. How about `Fix unused function warning with i386 codegen on x86_64 host`.

Heh. No idea why I wrong x86_64 instead of i386. I'll update it.

PR #2013, opened 2025-09-27
Fix cast-align warnings in emitElfStatic

hainest:
These were found when compiling with clang-21.

### DIFF
--- a/symtabAPI/src/emitElfStatic-ppc64.C
+++ b/symtabAPI/src/emitElfStatic-ppc64.C
@@ -50,6 +50,7 @@
 #include "debug.h"
 #include "common/src/arch-power.h"
 #include "elf.h"
+#include "unaligned_memory_access.h"
 
 using namespace Dyninst;
 using namespace Dyninst::SymtabAPI;
@@ -73,6 +74,13 @@ unsigned int setBits(unsigned int target, unsigned int pos, unsigned int len, un
 
     return target;
 }
+
+// The size of the return type of 'setBits' is used to determine the number of bytes
+// written to specific ELF fields. It's imperative they match the sizes of the types
+// provided by elfutils.
+static_assert(sizeof(decltype(setBits(0,0,0,0))) == sizeof(Elf32_Word), "setBits return type is incompatible with Elf32_Word");
+static_assert(sizeof(decltype(setBits(0,0,0,0))) == sizeof(Elf64_Word), "setBits return type is incompatible with Elf64_Word");
+
 
 #if defined(os_freebsd)
 #define R_X86_64_JUMP_SLOT R_X86_64_JMP_SLOT
@@ -349,11 +357,11 @@ bool emitElfStatic::archSpecificRelocation(Symtab* targetSymtab, Symtab* srcSymt
 
 	if (relocation_length == 64) {
 		char *td = (targetData + dest - (dest%8));
-		uint64_t target = relocation;
+		auto const target = static_cast<uint64_t>(relocation);
         	memcpy(td, &target, 2*sizeof(Elf64_Word));
 	} else {
 		char *td = (targetData + dest - (dest%4));
-	        unsigned int target = *((unsigned int *) td);
+	        auto target = Dyninst::read_memory_as<Elf64_Word>(td);
 	        target = setBits(target, relocation_pos, relocation_length, relocation);
         	memcpy(td, &target, sizeof(Elf64_Word));
 	}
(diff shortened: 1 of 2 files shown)


symtabAPI/src/emitElfStatic-ppc64.C
```diff
@@ -349,13 +350,13 @@ bool emitElfStatic::archSpecificRelocation(Symtab* targetSymtab, Symtab* srcSymt
 
 	if (relocation_length == 64) {
 		char *td = (targetData + dest - (dest%8));
-		uint64_t target = relocation;
-        	memcpy(td, &target, 2*sizeof(Elf64_Word));
+		auto const target = static_cast<uint64_t>(relocation);
```
bbiiggppiigg, line 360: Should we use Elf64_LWord here or uint32_t  on line 357 for consistency?

hainest: Probably. For now, I just wanted to keep the types as they were. I did just notice that I forgot to fix a couple items. I also want to add some `static_assert`s to make sure the type sizes are consistent since `setBits` traffics in `unsigned int` instead of `Elf64_Word`.

hainest:
@bbiiggppiigg Done with fixes. Ready for you, if/when you have time.

symtabAPI/src/emitElfStatic-ppc64.C
```diff
@@ -690,13 +695,12 @@ default:
...
 	}
-    if (branch_pred >= 0) {
-	unsigned int *td = (unsigned int *) targetData;
-	unsigned int target;
-	target = td[dest/4];
-	target = setBits(target, 10, 1, branch_pred);
-        memcpy(&td[dest/4], &target, sizeof(Elf32_Word));
-    } 
+  if (branch_pred >= 0) {
+    auto *td = &targetData[dest/4];
```
bbiiggppiigg: targetData was of type char *.
The address of targetData[dest/4] is targetData + (dst/4) * sizeof (char);
In the origina code, td was a pointer of unsigned int *;
Won't address of  td[dest/4] be targetData + (dst/4) * sizeof(unsigned int)?

symtabAPI/src/emitElfStatic-ppc64.C
```diff
@@ -670,17 +678,14 @@ default:
                        static_cast<unsigned long>(targetData[dest+3]));
 
   if (rel.getRelType() == R_PPC_REL24) {
-	unsigned int *td = (unsigned int *) targetData;
-	unsigned int target;
-	target = td[dest/4];
+	auto target = Dyninst::read_memory_as<Elf32_Word>(&targetData[dest/4]);
```
bbiiggppiigg: Same question as the one below.

symtabAPI/src/emitElfStatic-ppc64.C
```diff
@@ -670,17 +678,14 @@ default:
...
 	target = setBits(target, relocation_pos, relocation_length, relocation);
-        memcpy(&targetData[dest], &target, sizeof(Elf32_Word));
+	memcpy(&targetData[dest], &target, sizeof(target));
 	} else {
-	unsigned int *td = (unsigned int *) targetData;
-	unsigned int target;
-	target = td[dest/4];
+	auto *td = &targetData[dest/4];
```
bbiiggppiigg: Same question as the one below.

symtabAPI/src/emitElfStatic-ppc64.C
hainest: Derp, yes. Good catch.

hainest:
@bbiiggppiigg I backed out the unneeded changes to fix the calculations. It's now building cleanly with gcc-15 and clang-21 with `-DDYNINST_CODEGEN_ARCH=ppc64le` on an x86_64 host.

symtabAPI/src/emitElfStatic-ppc64.C
```diff
@@ -1091,7 +1099,7 @@ bool emitElfStatic::handleInterModuleSpecialCase(Symtab * /*target*/,
   // or create it. 
   //
 
-  unsigned int *td = (unsigned int *)(data + dest - (dest % 4));
```
kupsch, line 1094: The `(dest % 4)` seems sketchy; I would guess that `(dest % 4)` is always 0, otherwise this is not reading the instruction where `dest` points, but 1-3 bytes before.  It does force it to be 4-byte aligned.

hainest: This whole function is only ever used for `R_PPC64_REL24` which is defined as `(S + A – P) >> 2` where `S` represents the value of the symbol whose index resides in the relocation entry, `A` represents the addend used to compute the value of the relocatable field, and `P` represents the place (section offset or address) of the storage unit being relocated (computed using `r_offset`). Tracing through the calls to get here, [`dest`](https://github.com/dyninst/dyninst/blob/master/symtabAPI/src/emitElfStatic.C#L1234) is set to `S + r_offset` which would be `S + P`. Fun.

I'm not really sure what the person was trying to accomplish with `dest % 4`. I guess they misread the `>> 2`? No idea.

PR #2017, opened 2025-09-29
Add tests for x86_64 emitter

hainest:
@rochauha It's a little rough, but this should also work for doing amdgpu testing.

### DIFF
--- a/tests/unit/dyninstAPI/emitter/x86_64.cpp
+++ b/tests/unit/dyninstAPI/emitter/x86_64.cpp
@@ -14,7 +14,11 @@ int main() {
 
   BinaryEdit bin_edit{};
 
+  // The largest code sequence I've seen is 23 bytes, but there's no
+  // technical limitation on how much code could be generated for any
+  // operation. 128 bytes should be plenty.
   constexpr auto buffer_size = 128;
+
   codeGen gen(buffer_size);
   gen.setAddrSpace(&bin_edit);
   gen.setRegisterSpace(rs);


hainest:
@kupsch I was surprised that the x86_64 emitter _always_ uses a REX prefix when supported, even when you explicitly ask for a 32-bit register. For example,

```c++
emitter->emitLoadConst(REGNUM_EAX, Dyninst::Address{0x12345678}, gen);
```

gives `mov rax, 0x12345678`. It's not really a big deal, but I thought it was weird. Naturally, the problem is that `REGNUM_EAX` and `REGNUM_RAX` are both represented internally as just their ISA register id (0) instead of a proper object with propreties. :shrug:

hainest:
I ported these tests for the i386 emitter, but there is something strange going on with the virtual register stuff. I'll work on it and add those tests in a separate PR.

tests/unit/dyninstAPI/emitter/x86_64.cpp
```diff
@@ -0,0 +1,301 @@
...
+  using Dyninst::verify_emitter;
+  using Dyninst::emitter_buffer_t;
+
+  constexpr auto arch = Dyninst::Arch_x86_64;
+  auto const size = Dyninst::getArchAddressWidth(arch);
+  auto *rs = registerSpace::getRegisterSpace(size);
+
+  BinaryEdit bin_edit{};
+
+  constexpr auto buffer_size = 128;
```
bbiiggppiigg, line 20: How do you get this size, maybe worth a comment?

hainest: I just kind of guessed. The largest code sequence I saw was 23 bytes, but I haven't tested them all yet. I would _hope_ that 128 bytes would be big enough. I'll add a comment to that effect.

kupsch (approved):
comment has typo, otherwise looks good

PR #2019, opened 2025-09-29
Temporary fix for host-codegen separation on Power and AArch64

rochauha:
The `getMaxBranch` function is defined in inst-<arch> files for x86, power and aarch64. It is used by both host and codegen for power and aarch64. When host and codegen arch are different (i.e building for AMDGPU), the definition will be missing for the host side. This PR is a crude fix to this issue until we figure out a cleaner solution.

### DIFF
--- a/dyninstAPI/src/linux-aarch64.C
+++ b/dyninstAPI/src/linux-aarch64.C
@@ -45,6 +45,21 @@
 #include "dyninstAPI/src/function.h"
 #include "common/src/linuxHeaders.h"
 
+
+// FIXME: HOST+CODEGEN
+// This is a temporary fix as getMaxBranch is used for both host and codegen
+// This file is compiled when host architecture is AArch64. If codegen isn't for AArch64, we define the following:
+#if !defined(DYNINST_CODEGEN_ARCH_AARCH64)
+
+#define MAX_BRANCH_OFFSET      0x07ffffff  // 128MB  Used for B
+
+Address getMaxBranch() {
+    return MAX_BRANCH_OFFSET;
+}
+
+#endif
+
+
 #define DLOPEN_MODE (RTLD_NOW | RTLD_GLOBAL)
 
 const char DL_OPEN_FUNC_EXPORTED[] = "dlopen";
--- a/dyninstAPI/src/linux-power.C
+++ b/dyninstAPI/src/linux-power.C
@@ -45,6 +45,21 @@
 #include "dyninstAPI/src/function.h"
 #include "common/src/linuxHeaders.h"
 
+
+// FIXME: HOST+CODEGEN
+// This is a temporary fix as getMaxBranch is used for both host and codegen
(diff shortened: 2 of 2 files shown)


rochauha:
@hainest @kupsch @bbiiggppiigg -- This PR is ready for review.

bbiiggppiigg (approved):
Consider removing the blank lines

hainest:
Could you rebase on master so it can run through the codegen cross-build workflow?

rochauha:
Rebased on master.

bbiiggppiigg:
> This is causing DSO rewriting on ARM to fail.
> 
> [2019_arm_regressions.log](https://github.com/user-attachments/files/22648144/2019_arm_regressions.log)

Do you have any insight on what could be the likely cause? In theory all this is doing is to make sure the constants are conditionally included correctly, so unless the compiler complains I would assume we were doing the right thing.

hainest:
> > This is causing DSO rewriting on ARM to fail.
> > [2019_arm_regressions.log](https://github.com/user-attachments/files/22648144/2019_arm_regressions.log)
> 
> Do you have any insight on what could be the likely cause? In theory all this is doing is to make sure the constants are conditionally included correctly, so unless the compiler complains I would assume we were doing the right thing.

No, I'm not sure what happened. I agree it's weird since this seems like a straightforward change. The test suite's logs aren't helpful for crashes- only failures. You'd have to attach a debugger to even see where it crashed.

bbiiggppiigg:
> > > This is causing DSO rewriting on ARM to fail.
> > > [2019_arm_regressions.log](https://github.com/user-attachments/files/22648144/2019_arm_regressions.log)
> > 
> > 
> > Do you have any insight on what could be the likely cause? In theory all this is doing is to make sure the constants are conditionally included correctly, so unless the compiler complains I would assume we were doing the right thing.
> 
> No, I'm not sure what happened. I agree it's weird since this seems like a straightforward change. The test suite's logs aren't helpful for crashes- only failures. You'd have to attach a debugger to even see where it crashed.

Since we have a log, this means dyninst builds fine (with aarch64 as both host and codegen arch)?

hainest:
> Since we have a log, this means dyninst builds fine (with aarch64 as both host and codegen arch)?

Yup. It builds and the unit/integration tests pass, but the test suite tests that do static rewriting of DSOs are broken.

hainest:
This wasn't the cause of the test failures (see #2022). If you could rebase, then I'll test again.

bbiiggppiigg:
@rochauha

rochauha:
Rebased on master.

hainest:
> Rebased on master.

Thanks! I'll re-run the test suite.

bbiiggppiigg (approved):

hainest (approved):
No regressions on 17 machines. Apologies for the previous noise.

PR #2021, opened 2025-09-30
Fix useless absolute value of unsigned values (Wabsolute-value)

hainest:
This was found using clang-20.

---

This PR fixes the compiler warnings, but they have illuminated concerning uses of `Dyninst::Address`. It is defined to be an `unsigned long`, so subtraction to find an offset is not logically correct. You can see that in some places an `Address` getting cast to some signed type. Bizarrely, `Dyninst::Offset` is also an `unsigned long`. This means there is no uniform way to represent actual relative offsets. For example,

```c++
void EmitterAARCH64::emitStoreShared(Register source, const image_variable *var, bool is_local, int size, codeGen &gen) {
    // create or retrieve jump slot
    Address addr;
    int stackSize = 0;
    if(!is_local) {
        addr = getInterModuleVarAddr(var, gen);
    }
    else {
        addr = (Address)var->getOffset();    // <----- Is it an Offset or an Address?
    }

    // load register with address from jump slot
    Register baseReg = gen.rs()->getScratchRegister(gen, true);
    assert(baseReg != Null_Register && "cannot get a scratch register");

    emitMovePCToReg(baseReg, gen);
    Address varOffset = addr - gen.currAddr() + 4;    // <-------  This is definitely broken

    if(!is_local) {
      ...
    } else {
        std::vector<Register> exclude;
        exclude.push_back(baseReg);
        // mov offset to a reg
        auto addReg = insnCodeGen::moveValueToReg(gen, varOffset, &exclude);
        // add/sub offset to baseReg
        insnCodeGen::generateAddSubShifted(gen,

                (signed long long) varOffset>0?insnCodeGen::Add:insnCodeGen::Sub,
                ^^^^^ worrying cast

                0, 0, addReg, baseReg, baseReg, true);
```

This business of addresses, offsets, and relative offsets may need to get addressed sooner rather than later. It is not difficult to create some actual arithmetic types that have the proper relationships. Proliferating those types will definitely be a huge headache.

### DIFF
--- a/dyninstAPI/src/Relocation/Widgets/PCWidget-aarch64.C
+++ b/dyninstAPI/src/Relocation/Widgets/PCWidget-aarch64.C
@@ -106,7 +106,7 @@ bool IPPatch::apply(codeGen &gen, CodeBuffer *) {
   // Load the offset into a scratch register
   std::vector<Register> exclude;
   exclude.push_back(30 /* LR */);
-  Register scratchReg = insnCodeGen::moveValueToReg(gen, labs(RAOffset), &exclude);
+  Register scratchReg = insnCodeGen::moveValueToReg(gen, RAOffset, &exclude);
   // Put the original RA into LR
   insnCodeGen::generateAddSubShifted(gen, RAOffset>0?insnCodeGen::Add:insnCodeGen::Sub,
           0, 0, scratchReg, 30, 30, true);
--- a/dyninstAPI/src/inst-aarch64.C
+++ b/dyninstAPI/src/inst-aarch64.C
@@ -1096,7 +1096,7 @@ bool EmitterAARCH64::emitLoadRelative(Register dest, Address offset, Register ba
         std::vector<Register> exclude;
         exclude.push_back(baseReg);
         // mov sOffset to a reg
-        auto addReg = insnCodeGen::moveValueToReg(gen, labs(offset), &exclude);
+        auto addReg = insnCodeGen::moveValueToReg(gen, labs(sOffset), &exclude);
         // add/sub sOffset to baseReg
         insnCodeGen::generateAddSubShifted(gen,
                 sOffset>0?insnCodeGen::Add:insnCodeGen::Sub,
@@ -1280,7 +1280,7 @@ void EmitterAARCH64::emitLoadShared(opCode op, Register dest, const image_variab
         } else {
             std::vector<Register> exclude;
             exclude.push_back(baseReg);
-            auto addReg = insnCodeGen::moveValueToReg(gen, labs(varOffset), &exclude);
+            auto addReg = insnCodeGen::moveValueToReg(gen, varOffset, &exclude);
             insnCodeGen::generateAddSubShifted(gen,
                     (signed long long) varOffset>0?insnCodeGen::Add:insnCodeGen::Sub,
                     0, 0, addReg, baseReg, baseReg, true);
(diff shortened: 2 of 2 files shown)


kupsch (approved):

bbiiggppiigg:
> This was found using clang-20.
> 
> This PR fixes the compiler warnings, but they have illuminated concerning uses of `Dyninst::Address`. It is defined to be an `unsigned long`, so subtraction to find an offset is not logically correct. You can see that in some places an `Address` getting cast to some signed type. Bizarrely, `Dyninst::Offset` is also an `unsigned long`. This means there is no uniform way to represent actual relative offsets. For example,
> 
> ```c++
> void EmitterAARCH64::emitStoreShared(Register source, const image_variable *var, bool is_local, int size, codeGen &gen) {
>     // create or retrieve jump slot
>     Address addr;
>     int stackSize = 0;
>     if(!is_local) {
>         addr = getInterModuleVarAddr(var, gen);
>     }
>     else {
>         addr = (Address)var->getOffset();    // <----- Is it an Offset or an Address?
>     }
> 
>     // load register with address from jump slot
>     Register baseReg = gen.rs()->getScratchRegister(gen, true);
>     assert(baseReg != Null_Register && "cannot get a scratch register");
> 
>     emitMovePCToReg(baseReg, gen);
>     Address varOffset = addr - gen.currAddr() + 4;    // <-------  This is definitely broken
> 
>     if(!is_local) {
>       ...
>     } else {
>         std::vector<Register> exclude;
>         exclude.push_back(baseReg);
>         // mov offset to a reg
>         auto addReg = insnCodeGen::moveValueToReg(gen, varOffset, &exclude);
>         // add/sub offset to baseReg
>         insnCodeGen::generateAddSubShifted(gen,
> 
>                 (signed long long) varOffset>0?insnCodeGen::Add:insnCodeGen::Sub,
>                 ^^^^^ worrying cast
> 
>                 0, 0, addReg, baseReg, baseReg, true);
> ```
> 
> This business of addresses, offsets, and relative offsets may need to get addressed sooner rather than later. It is not difficult to create some actual arithmetic types that have the proper relationships. Proliferating those types will definitely be a huge headache.

We actually brought this in the group meeting with Bart. What we probably need is an relative offset type that is signed, in addition to the offset type, which is file offset, which can still be unsigned.

woodard:
We have actually been discussing a related issue in the DWARF for GPU's meetings. A DWARF location when applied to memory is (address space, byte offset, bit offset). Locations are different than the generic values previously used to DWARF. Many times generic values will be converted to locations by implicit conversion. 
Addresses and therefore offsets are by and large are unsigned but a new operator DW_OP_offset can take an signed value. I pointed out the trickiness with mixing signed and unsigned arithmetic but DW_OP_offset needs to go both ways. This means that the code must do something like this or else bad things can happen:
```
if( offset >= 0)
  address +=offset;
else
  address -=abs(offset);
```
GDB has more or less gotten away with not having to worry about such things because internally it uses infinite precision numbers.

There is a currently huge debate with another DWARF6 operator called DW_OP_overlay and the question is should it allow a negative overlay offsets.  I'm of the opinion that "no it should not".

Anyway, the point is DWARF has long played fast and loose with generic types, signed and unsigned numbers and addresses but we are doing our best to tighten it up in DWARF6. There are always going to be relative offsets which are negative and you must handle those but unless you are a bit more careful about the signedness of types, you may start to run into problems with DWARF6.

Anyway, the point is DWARF has played fast

hainest:
>Many times generic values will be converted to locations by implicit conversion. Addresses and therefore offsets are by and large are unsigned but a new operator DW_OP_offset can take an signed value. I pointed out the trickiness with mixing signed and unsigned arithmetic but DW_OP_offset needs to go both ways. This means that the code must do something like this or else bad things can happen:
> 
> ```
> if( offset >= 0)
>   address +=offset;
> else
>   address -=abs(offset);
> ```

This is a general issue with LSB128. Dealing with `DW_AT_encoding` both in size and sign is a real wart for user code. Dyninst would be well-served by having concrete, domain-specific types to represent addresses and offsets that can handle both the logical and ISA details so users can avoid writing those kinds of checks.

PR #2022, opened 2025-10-01
Fix offset calculations in EmitterAARCH64

hainest:
'varOffset' is defined as

Address varOffset = addr - gen.currAddr() + 4;

When 'addr' is less than 'gen.currAddr()', this calculation underflows.
Since 'labs' takes a 'long int', the underflowed 'unsigned long' (that is
what 'Address' really is) gets converted to a signed value. Removing
this conversion and the subsequent absolute value caused 'varOffset'
to be treated as a large, positive number.

This was broken by 9b6dd2aa6b.

### DIFF
--- a/dyninstAPI/src/Relocation/Widgets/PCWidget-aarch64.C
+++ b/dyninstAPI/src/Relocation/Widgets/PCWidget-aarch64.C
@@ -106,7 +106,7 @@ bool IPPatch::apply(codeGen &gen, CodeBuffer *) {
   // Load the offset into a scratch register
   std::vector<Register> exclude;
   exclude.push_back(30 /* LR */);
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
Whatt about
https://github.com/dyninst/dyninst/blob/31b490ec698a5eb898ee067b7651e6356fe99e05/dyninstAPI/src/inst-aarch64.C#L1079

hainest:
> Whatt about
> 
> https://github.com/dyninst/dyninst/blob/31b490ec698a5eb898ee067b7651e6356fe99e05/dyninstAPI/src/inst-aarch64.C#L1079

That one is fine since `sOffset` is a `signed long long` already.

hainest:
No regressions on 18 machines; included 3 ARM.

PR #2023, opened 2025-10-03
Remove unused AST Widget

bbiiggppiigg:


hainest:
It looks like this was added by 91246b9ca in 2011, but never used. Could you add that note to the commit message?

bbiiggppiigg:
> It looks like this was added by [91246b9](https://github.com/dyninst/dyninst/commit/91246b9ca539f5b27ca3aeb96d7c70b680e991ce) in 2011, but never used. Could you add that note to the commit message?

I also tracked down to this commit, but it seem like it is simply a rename of ASTAtom.
The actual remove of usage might be even before this.

hainest:
> > It looks like this was added by [91246b9](https://github.com/dyninst/dyninst/commit/91246b9ca539f5b27ca3aeb96d7c70b680e991ce) in 2011, but never used. Could you add that note to the commit message?
> 
> I also tracked down to this commit, but it seem like it is simply a rename of ASTAtom. The actual remove of usage might be even before this.

That was the oldest commit that git pickaxe came up with `git log --oneline -S ASTWidget`, so I think that's when it was created. There are only three commits to `ASTWidget.h` before that which happened before the rename.

hainest (approved):

PR #2024, opened 2025-10-06
GitHubCI: patch build error in launchmon

hainest:
This should be reverted once https://github.com/LLNL/LaunchMON/pull/65 is merged.

### DIFF
--- a/.github/workflows/consumers.yaml
+++ b/.github/workflows/consumers.yaml
@@ -253,6 +253,8 @@ jobs:
           cd /
           git clone --depth=1 https://github.com/llnl/launchmon.git
           cd launchmon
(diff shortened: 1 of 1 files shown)


hainest:
It's now building correctly on Fedora-43 and Ubuntu-25.10 with gcc-15.

bbiiggppiigg (approved):

bbiiggppiigg:
@hainest which part of launchmon is using dyninst? I tried to grep dyninst in their source code repo but I can't find anything.

hainest:
> @hainest which part of launchmon is using dyninst? I tried to grep dyninst in their source code repo but I can't find anything.

launchmon doesn't use Dyninst. STAT uses both Dyninst and launchmon, so we have to build it (as well as GraphViz, GraphLib, and MRNet).

PR #2025, opened 2025-10-06
GitHubCI: cleanup and fix extrae build

hainest:
1. Use their new bootstrap script
2. Remove ubuntu version check (we no longer use 24.10)


hainest:
extrae is now building correctly on Fedora-43 and Ubuntu-25.10 with gcc-1~5~3 (it still doesn't build with 14+).

PR #2027, opened 2025-10-06
Sort snippets at points

rochauha:
A stable sort on snippets at every points such that prologues < regular snippets < epilogues for each point.


rochauha:
@hainest, @bbiiggppiigg, @kupsch -- this PR is ready for review.

bbiiggppiigg:
> @hainest, @bbiiggppiigg, @kupsch -- this PR is ready for review.

I don't think this is complete. 
Do you not need to modify pushFront and pushBack?

bbiiggppiigg (changes requested):
Missing files.

rochauha:
Added updated files.

patchAPI/h/Point.h
```diff
@@ -157,6 +157,11 @@ Location(PatchFunction *f, PatchBlock *b, Dyninst::Address a, InstructionAPI::In
 // Used in PointType definition
 #define type_val(seq) (0x00000001u << seq)
 
+enum SnippetType {
```
hainest: The C++ standard doesn't require enumerators to be ordered. Since they are being used for sorting, they should get explicit values. Ideally, this would be a [scoped enum](https://en.cppreference.com/w/cpp/language/enum.html#Scoped_enumerations), but that would be a pain to use for sorting. The next best thing is to put an anonymous enum in a namespace. The enumerators would then be referenced using a scoped name instead of a bareword. For example, `SnippetType::PROLOGUE` instead of `PROLOGUE`.

```suggestion
namespace SnippetType {
  enum {
    PROLOGUE = 0,
    REGULAR = 1,
    EPILOGUE = 2
  };
}
```

I would also add a comment saying that the order of these values is important since it determines the ordering of the snippets. That way, the next person will know.

dyninstAPI/src/baseTramp.C
```diff
@@ -299,6 +299,11 @@ bool baseTramp::generateCodeInlined(codeGen &gen,
    std::vector<AstNodePtr> miniTramps;
 
    if (point_) {
+      // Stable sort snippets so that Prologues < Regular snippets < Epilogues.
```
hainest: This comment is too literal- it just tells me what `stable_sort` does. I would say something along the lines of "Order snippets by their type with prologues before regular snippets and epilogues at the end without changing the order of snippets of the same type."

hainest (changes requested):

patchAPI/h/Point.h
kupsch: It does require them be order in increasing order, unless you explicitly define an enumerator's value. WIthout explicit definition the first enumerator is always `0`, and other enumerators without an explicit definition have the value `<PREVIOUS_ENUMERATOR> + 1`.

I don't think it would be a pain for sorting if you used a scoped enum `enum class SnippetType` (without explicit values). Comparison of a scoped enum with another is allowed (arithmetic and conversions are not).

patchAPI/h/PatchCommon.h
```diff
@@ -35,7 +35,7 @@
 #include <string.h>
 #include <utility>
 #include <vector>
-#include <list>
+#include <deque>
```
kupsch, line 38: @hainest, what it be better to usa a std::vector as it is more efficient in time and space than a std::deque if the number of elements is small (I would guess that the typical number of snippets would be single digits).

patchAPI/h/PatchCommon.h
```diff
@@ -64,7 +64,7 @@ typedef PointSet::iterator PointIter;
 class Instance;
 typedef boost::shared_ptr<Instance> InstancePtr;
 typedef std::set<InstancePtr> InstanceSet;
-typedef std::list<InstancePtr> InstanceList;
+typedef std::deque<InstancePtr> InstanceList;
```
kupsch: For new types, prefer `using` to define a type alias instead of `typedef`:
```suggestion
using  InstanceList = std::deque<InstancePtr>;
```

patchAPI/h/Point.h
hainest: > It does require them be order in increasing order, unless you explicitly define an enumerator's value. WIthout explicit definition the first enumerator is always `0`, and other enumerators without an explicit definition have the value `<PREVIOUS_ENUMERATOR> + 1`.

That's what I thought, but I went back to check. Somehow, I missed the last sentence in the second clause that says exactly that. Derp.

> I don't think it would be a pain for sorting if you used a scoped enum `enum class SnippetType` (without explicit values). Comparison of a scoped enum with another is allowed (arithmetic and conversions are not).

Interesting. I didn't know there was a builtin to allow that. I assumed that since they aren't implicitly convertible to their underlying type/value, they couldn't be compared for inequality. I re-checked the standard, but couldn't find a definitive answer. gcc and clang seem to work, so TIL.

---

@rochauha Given this, the best would be to convert it to a scoped enum.

```c++
enum struct SnippetType {
```

patchAPI/h/PatchCommon.h
hainest: I recommended the deque because it has less cognitive load than manually shuffling a vector. It might not be as fast in the absolute sense, but random access and insertion at either end is constant time. That's better than a list.

rochauha:
Addressed comments by @hainest and @kupsch.

hainest:
> change title to be descriptive of the actual purpose: `Add support for pro/epllogue snippet (ordering) types`. Also add body something like: `The snippet type is used to ensure snippets on the same point execute in the order: PROLOGUE, REGULAR, EPILOGUE. Within each type they are executed in the order they added.`
> 
> Wait to see if Tim has any comments.

@rochauha You can change the commit title and comments in github when you squash-and-merge. That way, you don't have to wait for the CI pipeline to run.

PR #2028, opened 2025-10-06
GitHubCI: build MRNet serially

hainest:
It sometimes fails when built in parallel.


hainest:
It's building fine now.

bbiiggppiigg (approved):

PR #2029, opened 2025-10-07
GitHubCI: disable HPCToolkit on Fedora

hainest:
spack fails to bootstrap on Fedora 43. This can be re-enabled once https://github.com/spack/spack/issues/51410 is resolved.

### DIFF
--- a/.github/workflows/consumers.yaml
+++ b/.github/workflows/consumers.yaml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         os: ${{ fromJson(needs.get-oses.outputs.latest) }}
+         os: [ 'ubuntu-25.04' ]
     permissions:
       packages: read
     container:


hainest:
Consumers workflow is now completing successfully.

.github/workflows/consumers.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         os: ${{ fromJson(needs.get-oses.outputs.latest) }}
+         os: [ 'ubuntu-25.04' ]
```
bbiiggppiigg, line 27: I guess github workers only have fedora and ubuntu, so once youo fixed it to ubuntu, it is disable for fedora?
That being said, is there any reason to use a hard-coded version of ubuntu?

hainest: `needs.get-oses.outputs.latest` currently expands to `[ 'ubuntu-25.04', 'fedora-43' ]`, so I just put in the ubuntu version manually.

bbiiggppiigg (approved):

PR #1999, opened 2025-09-19
[BPatch] Cleanup BPatch_point

rochauha:
This PR renames llPoint() to getPoint() and makes other getPoint interfaces public. 

Appropriate getPoint() is selected by the compiler based on constness.

This is currently required for #1987.

PR #2030, opened 2025-10-09
[AMDGPU] Insert prologue and epilogue in instrumented kernels

rochauha:
Instrumented kernels need to be setup to access and write back instrumentation variables. 

When user inserts a snippet at a point, and if that point belongs to a kernel, we also insert a prologue at the entry point and an epilogue at the exit point of that kernel. The prologue and epilogue are inserted only once.

### DIFF
--- a/external/amdgpu/AMDHSAKernelDescriptor.h
+++ b/external/amdgpu/AMDHSAKernelDescriptor.h
@@ -1,3 +1,11 @@
+// Apart from this note, this file is a copy of AMDHSAKernelDescriptor.h from:
+//
+// repo: https://github.com/llvm/llvm-project.git
+// tag:  llvmorg-17.0.3
+// hash: 888437e1b600
+// path: llvm/include/llvm/Support/AMDHSAKernelDescriptor.h
+// date: 2023-10-11 16:23:56Z
+//
 //===--- AMDHSAKernelDescriptor.h -----------------------------*- C++ -*---===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
@@ -9,6 +17,13 @@
 /// \file
 /// AMDHSA kernel descriptor definitions. For more information, visit
 /// https://llvm.org/docs/AMDGPUUsage.html#kernel-descriptor
+///
+/// \warning
+/// Any changes to this file should also be audited for corresponding changes
+/// needed in both the assembler and disassembler, namely:
+/// * AMDGPUAsmPrinter.{cpp,h}
+/// * AMDGPUTargetStreamer.{cpp,h}
+/// * AMDGPUDisassembler.{cpp,h}
 //
 //===----------------------------------------------------------------------===//
 


rochauha:
@hainest @kupsch @bbiiggppiigg  -- this PR is ready for review.

dyninstAPI/src/BPatch_addressSpace.C
```diff
@@ -838,6 +838,63 @@ BPatchSnippetHandle *BPatch_addressSpace::insertSnippet(const BPatch_snippet &ex
...
 
+#if defined(DYNINST_CODEGEN_ARCH_AMDGPU_GFX908)
+void BPatch_addressSpace::insertPrologueAtPoints(AmdgpuPrologueSnippet& snippet, std::vector<BPatch_point *>& points) {
+  for (size_t i = 0; i < points.size(); ++i) {
+    instPoint *iPoint = static_cast<instPoint *>(points[i]->getPoint(BPatch_callBefore));
+    iPoint->pushBack(snippet.ast_wrapper, SnippetType::PROLOGUE);
+  }
+}
+
+void BPatch_addressSpace::insertPrologueIfKernel(BPatch_function *function) {
```
bbiiggppiigg: I don't know if this would be better or worse, but maybe we can make kernelInfo a map that is indexed by the MangleName.
This way, the logic of this function would be
if(kernelInfo.find(function->getMangledName() == kernelInfo.end) return;
insertPrologueAtPoints.

rochauha: This is more efficient, of course.

If possible, I want to refrain from optimizing this right now. Currently, the kernelInfos are simply parsed into a vector. For using maps, the kernelInfos will need to be set up as maps during/after parsing.

The whole kernelInfo business is temporary until we move the kernel descriptor parsing/modification into SymtabAPI (or elsewhere in Dyninst). Once that is done, I hope it becomes straightforward to check whether a function is a kernel (see if it has a corresponding `.kd` symbol).

rochauha:
Abstracted details out of BPatch_addressSpace using pimpl. Let me know if you have further comments @hainest, @kupsch, @bbiiggppiigg.

dyninstAPI/src/amdgpu-internal-impl.h
```diff
@@ -0,0 +1,64 @@
...
+};
+
+struct AmdgpuInternalImpl {
+  void insertPrologueAtPoints(AmdgpuPrologueSnippet &snippet, std::vector<BPatch_point *> &points);
+  void insertEpilogueAtPoints(AmdgpuEpilogueSnippet &snippet, std::vector<BPatch_point *> &points);
+
+  void insertPrologueIfKernel(BPatch_function *function);
+  void insertEpilogueIfKernel(BPatch_function *function);
+
+  static std::unordered_set<BPatch_function *> instrumentedFunctions;
```
bbiiggppiigg: why are these static?

rochauha: Reading kernelInfos and writing instrumented kernel names (later on, other details as well) must happen in `BPatch_binaryEdit.C`. These members need to be static to be able to access them. There is only one instance of `AmdgpuInternalImpl` in `BPatch_addressSpace`.

hainest:
I made a quick proof-of-concept. It is untested and filled with terrible names and interfaces. I think it's a better way forward with minimal code changes but will allow easy expansion to platform-agnostic codegen.

https://github.com/hainest/dyninst/commit/8690eb5155419664e2b3955173117b70a4129781

rochauha:
> `BPatch_addressSpace::amdgpuImpl` is never initialized. There is no way this code was tested. This also doesn't implement the pImpl idiom. It does, however, suggest that making a customization point in `insertSnippet` might be a better approach. I'll put together some thoughts.

I agree. I haven't tested this. Testing it requires some incompatible changes to #1987

rochauha:
> I made a quick proof-of-concept. It is untested and filled with terrible names and interfaces. I think it's a better way forward with minimal code changes but will allow easy expansion to platform-agnostic codegen.
> 
> [hainest@8690eb5](https://github.com/hainest/dyninst/commit/8690eb5155419664e2b3955173117b70a4129781)

Thanks @hainest. I will take a look.

rochauha:
> I made a quick proof-of-concept. It is untested and filled with terrible names and interfaces. I think it's a better way forward with minimal code changes but will allow easy expansion to platform-agnostic codegen.
> 
> [hainest@8690eb5](https://github.com/hainest/dyninst/commit/8690eb5155419664e2b3955173117b70a4129781)

@hainest, this solution won't work because `getPoint` is private. Making `snippetHandler` a friend also won't do because friend property isn't inherited.

One way would be to make `getPoint` public. Note that there is a `BPatch_point::llPoint` that is public, which is exactly`getPoint` but we need `getPoint(BPatch_callWhen when)`. See #1999 .

rochauha:
This PR is ready for review.

external/amdgpu/AMDGPUEFlags.h
```diff
@@ -0,0 +1,149 @@
+//===----------------------------------------------------------------------===//
+//
+// This file contains a copy of the e_flags struct definition from
+// llvm/include/llvm/BinaryFormat/ELF.h
+//
+// This file is also under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
kupsch, line 14: For code that is taken from an external source in addition to the license and the file path, also include the git repo path, tag, hash, date of hash/tag, and line(s). Also get the data from a commit that has release tag as these are less likely to contain errors:

```suggestion
// This file contains a copy of the e_flags struct definition from
//
// repo: https://github.com/llvm/llvm-project.git
// tag:  llvmorg-21.1.4
// hash: 222fc11f2b8f
// path: llvm/include/llvm/BinaryFormat/ELF.h
// date: 2025-10-21 08:14:55Z
// line: 766-918
//
// This file is also under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

`TZ=UTC git --date=iso` can be used to get the date replace ` +0000` with `Z` to indicate UTC time.

dyninstAPI/src/AmdgpuEpilogue.C
```diff
@@ -0,0 +1,47 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include "AmdgpuEpilogue.h"
+#include "emit-amdgpu.h"
+
+// Similar approach to prologue
+bool AmdgpuEpilogue::generate(Dyninst::PatchAPI::Point * /* point */, Dyninst::Buffer &buffer) {
+  codeGen gen(20);
```
hainest, line 37: How do we know that 20 bytes is enough?

dyninstAPI/src/AmdgpuEpilogue.C
```diff
@@ -0,0 +1,47 @@
...
+  EmitterAmdgpuGfx908 emitter;
+
+  emitter.emitScalarDataCacheWriteback(gen);
+
+  uint8_t *epilogue = (uint8_t *)gen.start_ptr();
```
hainest: I strongly recommend against C-style casts in C++ because they don't do just one thing ([ref](https://en.cppreference.com/w/cpp/language/explicit_cast.html#Explanation)). Here, it's not such an issue since the cast target is a primitive type. But it's a good idea to just break the habit of using them. A `reinterpret_cast` would be better.

dyninstAPI/src/AmdgpuEpilogue.h
```diff
@@ -0,0 +1,64 @@
...
+
+// The epilogue writes back the contents of scalar data cache to corresponding global memory.
+// PatchAPI snippet is used to insert the following epilogue:
+//
+//  s_dcache_wb
+//
+class AmdgpuEpilogue : public Dyninst::PatchAPI::Snippet {
+public:
+  AmdgpuEpilogue() = default;
+  ~AmdgpuEpilogue() = default;
```
hainest: You don't have to `default` these. The compiler will create them for you, if it needs to.

dyninstAPI/src/AmdgpuKernelDescriptor.C
```diff
@@ -0,0 +1,1011 @@
...
+
+bool AmdgpuKernelDescriptor::isGfx11() const {
+  return amdgpuMach == EF_AMDGPU_MACH_AMDGCN_GFX1100 ||
+         amdgpuMach == EF_AMDGPU_MACH_AMDGCN_GFX1103 ||
+         (amdgpuMach >= EF_AMDGPU_MACH_AMDGCN_GFX1101 &&
+          amdgpuMach >= EF_AMDGPU_MACH_AMDGCN_GFX1102);
+}
+
+AmdgpuKernelDescriptor::AmdgpuKernelDescriptor(uint8_t *kdBytes, size_t kdSize, unsigned amdgpuMachine) {
+  assert(kdSize == sizeof(kdRepr));
```
hainest: Since `kdRepr` isn't a parameter, it might be clearer to check `sizeof(kernel_descriptor_t)` since that's what is read from `kdBytes`.

dyninstAPI/src/AmdgpuKernelDescriptor.C
```diff
@@ -0,0 +1,1011 @@
...
+uint32_t AmdgpuKernelDescriptor::getCOMPUTE_PGM_RSRC1_FloatDenormMode32() const {
+  const uint32_t fourByteBuffer = kdRepr.compute_pgm_rsrc1;
+  return GET_VALUE(COMPUTE_PGM_RSRC1_FLOAT_DENORM_MODE_32);
+}
+
+void AmdgpuKernelDescriptor::setCOMPUTE_PGM_RSRC1_FloatDenormMode32(uint32_t value) {
+  uint32_t fourByteBuffer = kdRepr.compute_pgm_rsrc1;
+  fourByteBuffer = CLEAR_BITS(COMPUTE_PGM_RSRC1_FLOAT_DENORM_MODE_32);
+  assert(CHECK_WIDTH(COMPUTE_PGM_RSRC1_FLOAT_DENORM_MODE_32) &&
+         "value contains more bits than specificied");
```
hainest:
```suggestion
         "value contains more bits than specified");
```

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -0,0 +1,139 @@
...
+  // The kernel descriptor symbols have global visibility. So Dyninst will see them as global
+  // variables.
+
+  std::string kdName = f->getMangledName() + std::string(".kd");
+
+  BPatch_module *m = f->getModule();
+  BPatch_variableExpr* kernelDescriptorVariable = m->findVariable(kdName.c_str());
+
+  if (kernelDescriptorVariable) {
+      // FIXME : getSize is broken. Returns 4 instead of 64.
```
hainest, line 68: Is this something we should report to LLVM?

dyninstAPI/src/AmdgpuPrologue.C
```diff
@@ -0,0 +1,50 @@
...
+
+#include "AmdgpuPrologue.h"
+#include "emit-amdgpu.h"
+
+bool AmdgpuPrologue::generate(Dyninst::PatchAPI::Point * /* point */, Dyninst::Buffer &buffer) {
+  // To avoid any code duplication or refactoring right now, we use a 'codeGen'
+  // object to generate the code and copy what we get there into the
+  // 'Dyninst::Buffer' object passed here.
+
+  codeGen gen(20);
```
hainest, line 40: Same as above.

dyninstAPI/src/AmdgpuPrologue.h
```diff
@@ -0,0 +1,76 @@
...
+//
+// s_load_dwordx2 s[94:95], s[4:5], 0xabc
+// s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
+//
+class AmdgpuPrologue : public Dyninst::PatchAPI::Snippet {
+public:
+  AmdgpuPrologue(Dyninst::Register dest, Dyninst::Register base, unsigned offset)
+      : dest_(dest), base_(base), offset_(offset) {}
+
+  ~AmdgpuPrologue() = default;
```
hainest: You don't have to provide a dtor here.

dyninstAPI/src/AmdgpuEpilogue.C
rochauha: The epilogue is a `s_dcache_wb` instruction which is 8 bytes only.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: This is a Dyninst bug. I wanted to do the following:

```c++
if (kernelDescriptorVariable && kernelDescriptorVariable->getSize() == sizeof kernel_descriptor_t) {
// ...
}
```
But the size check doesn't work as `getSize()` returns 4. The variable's symbol corresponds to an object with global visibility. Here the symbol has size 64 bytes as per ABI and also as per readelf. So `BPatch_variableExpr::getSize()` must also return 64 instead of 4.

dyninstAPI/src/AmdgpuPrologue.C
rochauha: Prologue is a total of 12 bytes.

dyninstAPI/src/AmdgpuEpilogue.C
bbiiggppiigg: > I strongly recommend against C-style casts in C++ because they don't do just one thing ([ref](https://en.cppreference.com/w/cpp/language/explicit_cast.html#Explanation)). Here, it's not such an issue since the cast target is a primitive type. But it's a good idea to just break the habit of using them. A `reinterpret_cast` would be better.

Since Bart raised an issue about reinterpret_cast potentially being unportable, I started looking at static and reintprest_cast. Is there any reason why we don't want to use static_cast in this case?

hainest: > Since Bart raised an issue about reinterpret_cast potentially being unportable,

I've not heard of portability issues with any of the casts. The main issue with `reinterpret_cast` is that people try to use it to convert between types in a way that invokes UB. That's why Jim created the utilities in `unaligned_memory_access.h`.

> Is there any reason why we don't want to use static_cast in this case?

I prefer reinterpret_cast for pointers because it's my mental model for them, but a static_cast would be perfectly fine.

bbiiggppiigg: > I've not heard of portability issues with any of the casts. The main issue with `reinterpret_cast` is that people try to use it to convert between types in a way that invokes UB. That's why Jim created the utilities in `unaligned_memory_access.h`.
> 

This is what Bart said in the email
"The result of a reinterpret_cast cannot safely be used for anything other than being cast back to its original type. Other uses are, at best, nonportable."

I guess he found it from here, probaby by leaded by a google search.
https://learn.microsoft.com/en-us/cpp/cpp/reinterpret-cast-operator?view=msvc-170

hainest: It's a little more subtle than that article portrays. Yes, _many_ uses of `reinterpret_cast` break strict aliasing rules; that's why it has such a bad reputation. However, I tend to only use it for pointer-to-pointer conversions like here since those are guaranteed to work for [type-accessible](https://en.cppreference.com/w/cpp/language/reinterpret_cast.html#Type_accessibility) types, and primitive types and base-to-derived casts are type-accessible.

dyninstAPI/src/AmdgpuPointHandler.C
hainest: Oh, sorry. I completely misread where that error was coming from.

dyninstAPI/src/AmdgpuEpilogue.C
hainest: Then we should make the calculation here based on `sizeof(s_dcache_wb)` just in case it ever changes in size. Or at least add a comment to that effect.

dyninstAPI/src/AmdgpuEpilogue.C
kupsch: I think reinterpret_cast is appropriate.  Since this is casting a `void*` to a `uint8_t*` this is by definition not portable, unless you can guarantee alignment. Best to eliminate the casts altogether and use the `unaligned_memory_access.h` functions.

hainest: Actually, since we're just bit-blasting from `gen` to `buffer`, we could just use [`Buffer::copy`](https://github.com/dyninst/dyninst/blob/master/common/h/Buffer.h#L63). It's implementation leaves much to be desired, but it at least alleviates the need for casting here.

```c++
buffer.copy(gen.start_ptr(), gen.used());
```

dyninstAPI/src/AmdgpuEpilogue.C
rochauha: Instructions of a particular kind are grouped into encodings.  Based on this scheme, `s_dcache_wb` instruction is a `SMEM` instruction. `SMEM` instructions are 8 bytes long.

I have added a comment here and in the prologue about sizes.

rochauha:
Addressed comments and rebased on master.

dyninstAPI/src/AmdgpuKernelDescriptor.C
hainest: There are a bunch more of these typos.

rochauha:
Fixed typos. Updated `AMDHSAKernelDescriptor.h` along with adding a note.

hainest:
@rochauha You can go ahead and close this one since it's part of #1987 now.

PR #2033, opened 2025-10-14
Remove makeSizeType from InstructionDecoderImpl interface

hainest:
It's only used on x86, so move it there.

@wxrdnx Are you using this for RISC-V?


hainest:

PR #2034, opened 2025-10-16
Checking for a null pointer

ajanicijamd:
Sometimes a function returns a null pointer, which causes a segfault.

### DIFF
--- a/dyninstAPI/src/BPatch_edge.C
+++ b/dyninstAPI/src/BPatch_edge.C
@@ -117,6 +117,9 @@ BPatch_point *BPatch_edge::getPoint()
       BPatch_flowGraph *cfg = getFlowGraph();
       instPoint *ip = instPoint::edge(cfg->ll_func(),
                                       edge);
+      if (!ip) {
+          return point;
+      }
       AddressSpace *as = cfg->getllAddSpace();
       assert(as);
       


hainest:
The value of `ip` is ultimately determined in [`PatchAPI::PatchMgr::findPoint`](https://github.com/dyninst/dyninst/blob/v13.0.0/patchAPI/src/PatchMgr.C#L104). There are a few ways that it can return null. It would be nice to narrow it down to get a better root cause.

Do you have a reproducer for this? If not, could you do a bit more debugging?

For the case that `ip` is null, could you:

1. Check if `cfg->ll_func()` is null
2. Print out `cfg->ll_func()->format()`

Thanks!

bbiiggppiigg:
> Sometimes a function returns a null pointer, which causes a segfault.

Do you have reproducer for your problem?
While your fix probably is good on its own, the fact that a BPatch_edge does not have a corresponding point suggest there might be a bug somwhere else in the creation of this BPatch_edge/ during parsing.

hainest:
>the fact that a BPatch_edge does not have a corresponding point suggest there might be a bug somwhere else in the creation of this BPatch_edge/ during parsing.

Having an `instPoint` isn't a requirement for the class, but really should be. From [`BPatch_edge::BPatch_edge`](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/BPatch_edge.C#L79):

```c++
   // point is set when this edge is instrumented. instAddr is set
   // when either this edge or its conditional buddy is instrumented
   point = NULL; 
```

Ugh. That comment isn't even true. It's set if and only if `getPoint` is called. What was extra confusing to me is that [`instPoint::edge`](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/instPoint.C#L88) requests the creation of a point if one doesn't exist for the current edge. That ultimately never gets done because `PatchMgr::findPoint` does a bunch of checking that can cause a point not to be created. Naturally, there's no logging in that function...

bbiiggppiigg:
> Ugh. That comment isn't even true. It's set if and only if `getPoint` is called. What was extra confusing to me is that [`instPoint::edge`](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/instPoint.C#L88) requests the creation of a point if one doesn't exist for the current edge. That ultimately never gets done because `PatchMgr::findPoint` does a bunch of checking that can cause a point not to be created. Naturally, there's no logging in that function...

Location::EdgeInstance is created with trusted set to false.
So it will go through the PatchMgr::verify to check if both source block of the target block of the edge is a valid/known block (in loc.func.blocks()).

If it is able to pass the verification, it should go to PatchFunction::findPoint, which should either find or create the point.
So either there is no function associated with the edge, or either edge's src or target basic block is invalid.

ajanicijamd:
I cannot easily provide a reproducer for this, because our use case is a ctest test in https://github.com/ROCm/rocm-systems. But I will spend some more time debugging to try to narrow down the root cause of the problem.

PR #2037, opened 2025-10-23
Always decode instruction operands

hainest:
In every place that an instruction decoder is used in Dyninst, some property of the instruction is immediately used that requires decoding the operands. Delaying this process from the initial decode offers no benefit, but introduces overhead in the `Instruction` class. Capstone  doesn't provide a mechanism to delay decoding operands, so this behavior will eventually go away even without these changes.

### DIFF
--- a/instructionAPI/src/AMDGPU/gfx908/InstructionDecoder-amdgpu-gfx908.h
+++ b/instructionAPI/src/AMDGPU/gfx908/InstructionDecoder-amdgpu-gfx908.h
@@ -64,8 +64,6 @@ namespace Dyninst {
 
             virtual void setMode(bool)  { }
 
-            bool decodeOperands(const amdgpu_gfx908_insn_entry & insn_entry);
-
             static const std::array<std::string, 16> condNames;
             static MachRegister sysRegMap(unsigned int);
             static const char* bitfieldInsnAliasMap(entryID);
--- a/instructionAPI/src/AMDGPU/gfx90a/InstructionDecoder-amdgpu-gfx90a.h
+++ b/instructionAPI/src/AMDGPU/gfx90a/InstructionDecoder-amdgpu-gfx90a.h
@@ -64,8 +64,6 @@ namespace Dyninst {
 
             virtual void setMode(bool)  { }
 
-            bool decodeOperands(const amdgpu_gfx90a_insn_entry & insn_entry);
-
             static const std::array<std::string, 16> condNames;
             static MachRegister sysRegMap(unsigned int);
             static const char* bitfieldInsnAliasMap(entryID);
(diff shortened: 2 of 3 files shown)


hainest:
@bbiiggppiigg I noticed the AMDGPU decoders all declare `bool decodeOperands(const amdgpu_gfx908_insn_entry & insn_entry);`, but I couldn't find any implementation for them. Is that something that should be removed?

hainest:

bbiiggppiigg:
> @bbiiggppiigg I noticed the AMDGPU decoders all declare `bool decodeOperands(const amdgpu_gfx908_insn_entry & insn_entry);`, but I couldn't find any implementation for them. Is that something that should be removed?

Yes they probably should.
Since we are getting a new XML drop soon in theory, I can clean it up when that happens. 
Or if you want you can remove it in any PR you see fit.

bbiiggppiigg:
> In every place that an instruction decoder is used in Dyninst, some property of the instruction is immediately used that requires decoding the operands. Delaying this process from the initial decode offers no benefit, but introduces overhead in the `Instruction` class. Capstone doesn't provide a mechanism to delay decoding operands, so this behavior will eventually go away even without these changes.

I'm guessing it is that some people just want the instruction length initially, unless it is a branch instruction or something that requires you to look at the operands to determine branch target, to avoid the potentially expensive operand parsing.

That being said, for AMDGPU we always need to decode the operands to figure out what the length of an instructiton is, so I'm glad this is cleaned up i guess.

hainest:
> Didn't see any apparent breakage on AMDGPU architectures as well.

Thanks for checking. We _really_ need to get some kind of simple test for AMDGPU going. I would be fine with just literally doing `SymtabAPI::openFile`. Do you have a small binary we could add in the test-binaries repo for this?

hainest:
> > @bbiiggppiigg I noticed the AMDGPU decoders all declare `bool decodeOperands(const amdgpu_gfx908_insn_entry & insn_entry);`, but I couldn't find any implementation for them. Is that something that should be removed?
> 
> Yes they probably should. Since we are getting a new XML drop soon in theory, I can clean it up when that happens. Or if you want you can remove it in any PR you see fit.

I can remove them here since I'm touching all of those `decode`-like interfaces.

bbiiggppiigg:
> I can remove them here since I'm touching all of those `decode`-like interfaces.

Sure, please go ahead.

By the way, I don't think I have the rights to add binaries to the test_binaries repo.

hainest:
> By the way, I don't think I have the rights to add binaries to the test_binaries repo.

Oops. You should have write access now.

hainest:
> > I can remove them here since I'm touching all of those `decode`-like interfaces.
> 
> Sure, please go ahead.

Updated.

kupsch (approved):

PR #2039, opened 2025-10-23
Deprecate Expression::getChildren

hainest:
Having in-out parameters makes callsites clunky- particularly for range-base for loops.

This needs https://github.com/dyninst/testsuite/pull/267 and https://github.com/dyninst/examples/pull/65 to pass CI.

### DIFF
--- a/dataflowAPI/src/stackanalysis.C
+++ b/dataflowAPI/src/stackanalysis.C
@@ -1151,10 +1151,8 @@ void StackAnalysis::handleXor(Instruction insn, Block *block,
    operands[0].getWriteSet(writtenSet);
    operands[1].getReadSet(readSet);
 
-   std::vector<Expression::Ptr> children0;
-   std::vector<Expression::Ptr> children1;
-   operands[0].getValue()->getChildren(children0);
-   operands[1].getValue()->getChildren(children1);
+   auto children0 = operands[0].getValue()->getSubexpressions();
+   auto children1 = operands[1].getValue()->getSubexpressions();
 
    if (readSet.size() == 1 && writtenSet.size() == 1 &&
       (*readSet.begin())->getID() == (*writtenSet.begin())->getID() &&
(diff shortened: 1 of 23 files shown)


hainest:
@bbiiggppiigg @kupsch This has circular dependencies on the two linked issues. I'm not sure if it would be best to merge this or those first. I would prefer to merge the other two first, but I'm fine with either.

dataflowAPI/src/stackanalysis.C
```diff
@@ -1743,7 +1737,7 @@ void StackAnalysis::handleLEA(Instruction insn,
             Expression::Ptr scaleIndexExpr = children[1];
             STACKANALYSIS_ASSERT(dynamic_cast<BinaryFunction*>(scaleIndexExpr.get()));
             children.clear();
```
kupsch: The `children.clear();` before an assignment can be removed here and below

hainest: Updated.

hainest: I just noticed I forgot to deprecate in BinaryFunction and Dereference. I updated them.

instructionAPI/h/Expression.h
```diff
@@ -86,7 +87,10 @@ namespace Dyninst { namespace InstructionAPI {
     virtual bool bind(Expression* expr, const Result& value);
     virtual void apply(Visitor*) {}
 
-    virtual void getChildren(std::vector<Expression::Ptr>& children) const = 0;
+    DYNINST_DEPRECATED("Use getSubexpressions()")
+    virtual void getChildren(std::vector<Expression::Ptr>&) const {}
```
kupsch: You can make this non-virtual, implement here in the base and remove the implementation in the derived classes as their implementations are the same.

kupsch:
Merging this before the 2 linked issues if probably best.

instructionAPI/h/Expression.h
hainest: Agreed. Updated.

instructionAPI/h/BinaryFunction.h
```diff
@@ -32,6 +32,7 @@
 #define BINARYFUNCTION_H
 
 #include "ArchSpecificFormatters.h"
+#include "compiler_annotations.h"
```
kupsch: No longer needed
```suggestion
```

hainest: Bah. No matter how many times I look over these diffs, I always miss something.

PR #2040, opened 2025-10-25
Clean up InstructinAPI::InstructionDecoderImpl

hainest:


instructionAPI/src/InstructionDecoderImpl.h
```diff
@@ -88,4 +85,4 @@ class InstructionDecoderImpl
 }
 }
 
-#endif //!defined(INSTRUCTION_DECODER_IMPL_H)
```
bbiiggppiigg, line 91: Do we not want to keep this kind of comment?

hainest: Macro tag comments are generally useless; particularly for header include guards. I know some folks like them and clang-format inserts them by default, but they just don't add any value.

PR #2041, opened 2025-10-25
Import register uniqueness tests

hainest:
These were previously in the 'external-tests' repo. They should be here so they are actually unit tests.

The name [cache](https://github.com/dyninst/dyninst/blob/master/common/src/registers/MachRegister.C#L32) in `MachRegister` doesn't handle aliases, so the alias handling here is very ham-fisted. I should fix the `MachRegister` class at some point.

### DIFF
--- a/tests/unit/common/registers/uniqueness.cpp
+++ b/tests/unit/common/registers/uniqueness.cpp
@@ -46,7 +46,6 @@ bool check(Dyninst::Architecture arch) {
   auto success = true;
 
   std::unordered_map<unsigned int, std::string> vals;
-  int i = 0;
   for(auto r : regs) {
     auto res = vals.insert({r.val(), r.name()});
     auto itr = res.first;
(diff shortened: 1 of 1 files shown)


tests/unit/common/registers/uniqueness.cpp
```diff
@@ -0,0 +1,231 @@
...
+  int i = 0;
+  for(auto r : regs) {
+    auto res = vals.insert({r.val(), r.name()});
+    auto itr = res.first;
+    auto inserted = res.second;
+    if(!inserted && !known_alias(r.name())) {
+      auto val = itr->first;
+      auto name = itr->second;
+      std::cerr << "alias: 0x" << std::hex << val
+                << "  " << r.name() << "  " << std::dec << i << "\n";
```
bbiiggppiigg: Just wandering, what information does this i variable provides?

hainest: Oops. That's some debug code. I'll remove it.

bbiiggppiigg (approved):

PR #2042, opened 2025-10-25
Deprecate Expression::getUses

hainest:
Having in-out parameters makes callsites clunky- particularly for range-base for loops. The name is also unclear as to what it actually returns.

### DIFF
--- a/instructionAPI/h/BinaryFunction.h
+++ b/instructionAPI/h/BinaryFunction.h
@@ -32,7 +32,6 @@
 #define BINARYFUNCTION_H
 
 #include "ArchSpecificFormatters.h"
-#include "compiler_annotations.h"
 #include "Expression.h"
 #include "Register.h"
 #include "Result.h"
--- a/instructionAPI/h/Dereference.h
+++ b/instructionAPI/h/Dereference.h
@@ -33,7 +33,6 @@
 
 #include "ArchSpecificFormatters.h"
 #include "BinaryFunction.h"
-#include "compiler_annotations.h"
 #include "Expression.h"
 #include "Immediate.h"
 #include "Operand.h"
--- a/instructionAPI/h/MultiRegister.h
+++ b/instructionAPI/h/MultiRegister.h
@@ -32,7 +32,6 @@
 #define MULTIREGISTER_H
 
 #include "Architecture.h"
-#include "compiler_annotations.h"
 #include "Expression.h"
 #include "Register.h"
 #include "registers/MachRegister.h"
--- a/instructionAPI/h/Ternary.h
+++ b/instructionAPI/h/Ternary.h
@@ -32,7 +32,6 @@
 #define TERNARY_H
 
 #include "Architecture.h"
-#include "compiler_annotations.h"
 #include "Expression.h"
 
 #include <map>


dataflowAPI/src/AbslocInterface.C
```diff
@@ -58,22 +59,8 @@ void AbsRegionConverter::convertAll(InstructionAPI::Expression::Ptr expr,
 				    ParseAPI::Function *func,
                                     ParseAPI::Block *block,
 				    std::vector<AbsRegion> &regions) {
-  // If we're a memory dereference, then convert us and all
-  // used registers.
-  if (auto deref = boost::dynamic_pointer_cast<Dereference>(expr)) {
```
bbiiggppiigg, line 63: What about the dereferences? Is this handled anywhere?

hainest: `getUsedRegisters` visits all of the subnodes of an AST, so we don't need separate cases here.

bbiiggppiigg: > `getUsedRegisters` visits all of the subnodes of an AST, so we don't need separate cases here.

But your reg_visitor does nothing for Dereferences
https://github.com/dyninst/dyninst/pull/2042/files#diff-533da7fe8853102091a1053d631daeb4ea63a8c7683cf2ae028bc70f0e2f01cdR60
Are you certain your statement is correct?

hainest: With the exception of `MultiRegisterAST`, the AST classes do a depth-first traversal of their subtrees. For example, [`Dereference`](https://github.com/dyninst/dyninst/blob/v13.0.0/instructionAPI/h/Dereference.h#L226) visits its effective address. I'd have to look up why we didn't have `MultiRegisterAST` follow that same convention. It might have just been an oversight.

Here's a complete example.

```c++
#include "Architecture.h"

#include "Expression.h"
#include "BinaryFunction.h"
#include "Dereference.h"
#include "Immediate.h"
#include "MultiRegister.h"
#include "Register.h"
#include "Ternary.h"
#include "Visitor.h"

#include <iostream>
#include <cstdint>

struct reg_visitor : di::Visitor {
  void visit(di::RegisterAST* r) {
    std::cout << r->format() << ", ";
  }
  void visit(di::BinaryFunction*) {}
  void visit(di::Dereference*) {}
  void visit(di::Immediate*) {}
  void visit(di::MultiRegisterAST *mr) {
    for(auto r : mr->getRegs()) {
      r->apply(this);
    }
  }
};


int main() {
  // cmpxchg dword ptr [rbp + 0x12345678], esi
  uint8_t opcode[7] = {0x0f, 0xb1, 0xb5, 0x78, 0x56, 0x34, 0x12};
  
  auto arch = Dyninst::Arch_x86_64;
  di::InstructionDecoder d(opcode, 7, arch);
  auto insn = d.decode();
  
  std::cout << insn.format() << "\n";

  reg_visitor rv;  

  for(auto const& op : insn.getAllOperands()) {
    std::cout << "Visiting " << op.format(arch) << "\n  ";
    op.getValue()->apply(&rv);
    std::cout << "\n";
  }
}
```

gives

```
cmpxchg %eax,0x12345678(%rbp)
Visiting 0x12345678(%rbp)
  RBP, 
Visiting %esi
  ESI, 
Visiting %eax
  EAX, 
```

I just noticed some of the implicit registers are missing (it read/writes eflags). I'll double-check that for my Capstone branch.

Double-trouble: I just noticed the instruction printer is wrong. It thinks `eax` is the second operand instead of `esi`. That's neat.

bbiiggppiigg: > With the exception of `MultiRegisterAST`, the AST classes do a depth-first traversal of their subtrees. For example, [`Dereference`](https://github.com/dyninst/dyninst/blob/v13.0.0/instructionAPI/h/Dereference.h#L226) visits its effective address. I'd have to look up why we didn't have `MultiRegisterAST` follow that same convention. It might have just been an oversight.

Yeah I guess its just that I wasn't familiar with the visitor design pattern at the time so I implemented the traversal of the child elements in the wrong place.

hainest: > > With the exception of `MultiRegisterAST`, the AST classes do a depth-first traversal of their subtrees. For example, [`Dereference`](https://github.com/dyninst/dyninst/blob/v13.0.0/instructionAPI/h/Dereference.h#L226) visits its effective address. I'd have to look up why we didn't have `MultiRegisterAST` follow that same convention. It might have just been an oversight.
> 
> Yeah I guess its just that I wasn't familiar with the visitor design pattern at the time so I implemented the traversal of the child elements in the wrong place.

There's not a fixed rule about how the visitor gets applied. The original authors just did it that way, so we probably should have followed that. We could change it since `MultiRegisterAST` was added after the last release, so there's likely not any usercode handlling it. In fact, I think we should change the `Visitor` to not be an abstract class, but that's a different conversation.

kupsch: We can talk about MultiRegisterAST at the upcoming meeting

instructionAPI/src/MultiRegister.C
```diff
@@ -58,12 +58,6 @@ namespace Dyninst { namespace InstructionAPI {
       : Expression(inputRegASTs[0]->getID(), inputRegASTs.size()), m_Regs{std::move(inputRegASTs)} {
   }
 
-  void MultiRegisterAST::getUses(set<Expression::Ptr>& uses) {
-    for(const auto& m_Reg : m_Regs) {
```
bbiiggppiigg, line 62: The loop needs to be moved to accept for this change to be functionally equivalent.

hainest: The visitor in `getUsedRegisters` loops over the registers, so it should be the same. Yes?

`Expression::getUsed` -> `getUsedRegisters` -> loop over `MultiRegisterAST::getRegs()`.

Previously,

```c++
void foo(Expression::Ptr expr) {
  expr->getUses(...);  // Invokes MultiRegisterAST::getUses, loops over getRegs()
}
foo(MultiRegisterAST::Ptr{});
```

would now be


```c++
void foo(Expression::Ptr expr) {
  expr->getUses(...);  // Invokes Expression::getUses -> getUsedRegisters -> loop over getRegs()
}
foo(MultiRegisterAST::Ptr{});
```

bbiiggppiigg: Yeah I forgot u implemented in ur vistor.
Nvm.

hainest: No worries. I appreciate the careful review.

hainest:

PR #2044, opened 2025-10-30
Use depth-first traversal in MultiRegisterAST::visit

hainest:
This makes it consistent with the other AST types. This _could_ affect
user code. However, MultiRegisterAST hasn't been part of a release yet,
so a change of behavior is acceptable since API breakages are allowed
on major releases.

Summary of effects on current Visitor implementations in Dyninst:

getUsedRegisters
  updated

StateEvalVisitor
  no change in behavior

ExpressionConversionVisitor
  multireg isn't supported

thunkVisitor
  no change in behavior

detectToppedLoc
  There could be a change of behavior here since the registers of a
  multireg will now be seen before the multireg, itself. It's possible
  the RegisterAST visit method could do work that it wouldn't have
  before this change.

BPatch_memoryAccessAdapter
  multireg isn't supported

leaSimplifyVisitor
  This is only for x86 where the parts of a multireg can't be used as
  and index or base register for addressing, so there's no change in
  behavior.

ASTFactory
  multireg isn't supported.


bbiiggppiigg (approved):

kupsch (approved):

hainest:

PR #2045, opened 2025-10-30
Add unit tests for x86 call instructions

hainest:
This is a very limited subset of the possible 'call' instructions and mainly serves as a means of getting the decoder unit test framework into Dyninst.

@wxrdnx @rochauha Is this framework suitable for your testing needs?

### DIFF
--- a/tests/integration/InstructionAPI/decoder/memory_tests.h
+++ b/tests/integration/InstructionAPI/decoder/memory_tests.h
@@ -33,7 +33,6 @@
 
 #include "Instruction.h"
 #include "register_tests.h"
-#include "registers/register_set.h"
 
 namespace Dyninst { namespace InstructionAPI {
 


common/h/registers/register_set.h
```diff
@@ -0,0 +1,154 @@
...
+    bool is_empty() const {
+      return regs.empty();
+    }
+
+    // Union (in-place)
+    void operator|=(register_set const& rhs) {
+      for(auto r : rhs) {
+        regs.insert(r);
+      }
+    }
```
kupsch: The standard signature for compound assignment operators would be to return the type `T&` instead of `void` (here and the others below):
 ```suggestion
    register_set& operator|=(register_set const& rhs) {
      for(auto r : rhs) {
        regs.insert(r);
      }
      return *this;
    }
```

kupsch (approved):
Probably best to change the compound assignment return type, otherwise looks good.

common/h/registers/register_set.h
hainest: Yeah, that's true. I never think about using the result, but it is what the STL types do. Updated.

PR #2046, opened 2025-10-31
RISC-V Instruction API

wxrdnx:
- [x] Migrate old Capstone code
- [x] Add unit tests
- [x] Capstone Github Workflow

### DIFF
--- a/tests/integration/InstructionAPI/decoder/riscv/CMakeLists.txt
+++ b/tests/integration/InstructionAPI/decoder/riscv/CMakeLists.txt
@@ -14,61 +14,61 @@ add_executable(rv64i_base rv64i_base.cpp)
 target_compile_options(rv64i_base PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(rv64i_base PRIVATE instructionAPI cft_tests memory_tests
                                          register_tests opcode_tests)
-add_test(NAME instructionAPI_rv64i_base COMMAND rv64i_base)
-set_tests_properties(instructionAPI_rv64i_base PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_rv64i_base_riscv64 COMMAND rv64i_base)
+set_tests_properties(instructionAPI_rv64i_base_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(zicsr_ext zicsr_ext.cpp)
 target_compile_options(zicsr_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(zicsr_ext PRIVATE instructionAPI cft_tests memory_tests
                                         register_tests opcode_tests)
-add_test(NAME instructionAPI_zicsr_ext COMMAND zicsr_ext)
-set_tests_properties(instructionAPI_zicsr_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_zicsr_ext_riscv64 COMMAND zicsr_ext)
+set_tests_properties(instructionAPI_zicsr_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(m_ext m_ext.cpp)
 target_compile_options(m_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(m_ext PRIVATE instructionAPI cft_tests memory_tests register_tests
                                     opcode_tests)
-add_test(NAME instructionAPI_m_ext COMMAND m_ext)
-set_tests_properties(instructionAPI_m_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_m_ext_riscv64 COMMAND m_ext)
+set_tests_properties(instructionAPI_m_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(a_ext a_ext.cpp)
 target_compile_options(a_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(a_ext PRIVATE instructionAPI cft_tests memory_tests register_tests
                                     opcode_tests)
-add_test(NAME instructionAPI_a_ext COMMAND a_ext)
-set_tests_properties(instructionAPI_a_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_a_ext_riscv64 COMMAND a_ext)
+set_tests_properties(instructionAPI_a_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(f_ext f_ext.cpp)
 target_compile_options(f_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(f_ext PRIVATE instructionAPI cft_tests memory_tests register_tests
                                     opcode_tests)
-add_test(NAME instructionAPI_f_ext COMMAND f_ext)
-set_tests_properties(instructionAPI_f_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_f_ext_riscv64 COMMAND f_ext)
+set_tests_properties(instructionAPI_f_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(d_ext d_ext.cpp)
 target_compile_options(d_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(d_ext PRIVATE instructionAPI cft_tests memory_tests register_tests
                                     opcode_tests)
-add_test(NAME instructionAPI_d_ext COMMAND d_ext)
-set_tests_properties(instructionAPI_d_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_d_ext_riscv64 COMMAND d_ext)
+set_tests_properties(instructionAPI_d_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(c_ext c_ext.cpp)
 target_compile_options(c_ext PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(c_ext PRIVATE instructionAPI cft_tests memory_tests register_tests
                                     opcode_tests)
-add_test(NAME instructionAPI_c_ext COMMAND c_ext)
-set_tests_properties(instructionAPI_c_ext PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_c_ext_riscv64 COMMAND c_ext)
+set_tests_properties(instructionAPI_c_ext_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(branches branches.cpp)
 target_compile_options(branches PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(branches PRIVATE instructionAPI cft_tests memory_tests
                                        register_tests opcode_tests)
-add_test(NAME instructionAPI_branches COMMAND branches)
-set_tests_properties(instructionAPI_branches PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_branches_riscv64 COMMAND branches)
+set_tests_properties(instructionAPI_branches_riscv64 PROPERTIES LABELS "integration")
 
 add_executable(pseudo pseudo.cpp)
 target_compile_options(pseudo PRIVATE ${RISCV_INST_TESTS_CXX_WARNING_FLAGS})
 target_link_libraries(pseudo PRIVATE instructionAPI cft_tests memory_tests register_tests
                                      opcode_tests)
-add_test(NAME instructionAPI_pseudo COMMAND pseudo)
-set_tests_properties(instructionAPI_pseudo PROPERTIES LABELS "integration")
+add_test(NAME instructionAPI_pseudo_riscv64 COMMAND pseudo)
+set_tests_properties(instructionAPI_pseudo_riscv64 PROPERTIES LABELS "integration")


wxrdnx:
@hainest @ikitayama 
I think this PR can be reviewed.
Thanks!

ikitayama:
Sorry I am gone for sometime thus unable to do anything helpful

wxrdnx:
> Sorry I am gone for sometime thus unable to do anything helpful

Oops, I tagged the wrong person. Sorry.
I meant to tag @kupsch

instructionAPI/src/InstructionDecoderImpl.C
```diff
@@ -32,10 +32,12 @@
 #include "InstructionDecoder-x86.h"
 #include "InstructionDecoder-power.h"
 #include "InstructionDecoder-aarch64.h"
+#include "decoder/riscv/decoder.h"
```
kupsch, line 36: If capstone isn't installed this fails, so it needs to be guarded:

```suggestion
#ifdef DYNINST_ENABLE_CAPSTONE
#include "decoder/riscv/decoder.h"
#endif
```

wxrdnx:
Rebased on master (#2055)

common/src/arch-riscv64.h
```diff
@@ -0,0 +1,137 @@
...
+
+#ifndef _ARCH_RISCV64_H
+#define _ARCH_RISCV64_H
+
+#include <stdint.h>
+
+// In RISC-V, instruction length can be a multiple of 2 bytes
+
+// Standard RISC-V instructions (RVI, RVA, RVM, RVF, RVD, ...) are 4 bytes
+typedef uint32_t rvInsn_t;
```
hainest:
```suggestion
using rvInsn_t = uint32_t;
```
Prefer `using` aliases over `typedef`s.

instructionAPI/h/Instruction.h
```diff
@@ -174,11 +197,19 @@ namespace Dyninst { namespace InstructionAPI {
     void updateSize(const unsigned int new_size, const unsigned char * raw);
 
     void addSuccessor(Expression::Ptr e, bool isCall, bool isIndirect, bool isConditional, bool isFallthrough,
-                      bool isImplicit = false) const;
+                      bool isImplicit = false, bool appendOp = true) const;
```
hainest: Convert to an overload:

```c++
void addSuccessor(CFT cft) {
  m_Succesors.push_back(std::move(cft));
}
```

instructionAPI/src/decoder/riscv/categories.h
```diff
@@ -0,0 +1,185 @@
...
+      categories.push_back(di::c_ReturnInsn);
+      break;
+    case RISCV_GRP_ISRV32:
+    case RISCV_GRP_ISRV64:
+      categories.push_back(di::c_VectorInsn);
+      break;
+    }
+  }
+
+  // Relative calls aren't branches
```
hainest, line 80: Is this true for riscv?

instructionAPI/src/decoder/type_conversion.h
```diff
@@ -0,0 +1,91 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
```
hainest, line 30: I forgot a header guard here. Let's add one.

hainest (changes requested):

instructionAPI/src/decoder/riscv/categories.h
wxrdnx: Yes. It depends on the link register.
If the link register is x0 (zero), it is a branch . If the link register isn't, then it is call.
This means that branches and calls are disjoint.
Multi-instructions calls and branchs are another story, but this is Parse API's realm, not Instruction API.

common/src/arch-riscv64.h
wxrdnx: Got it.
Changed to `using`.

instructionAPI/h/Instruction.h
wxrdnx: Done.

instructionAPI/src/decoder/type_conversion.h
wxrdnx: Added the header guard.

hainest:
@wxrdnx #2064 is needed to run the unit tests in CI. I'll finish my review once that is done.

common/src/arch-riscv64.h
```diff
@@ -0,0 +1,137 @@
...
+
+// The minimum instruction size is 2 bytes
+using rvMinInsn_t = rvcInsn_t ;
+constexpr int RISCV_MIN_INSN_SIZE = sizeof(rvMinInsn_t);
+
+// The maximum instruction size is 4 bytes
+using rvMaxInsn_t = rvInsn_t;
+constexpr int RISCV_MAX_INSN_SIZE = sizeof(rvMaxInsn_t);
+
+// The register size is 4 bytes
```
kupsch: I don't understand this comment. What register? RISC-V registers are 64-bits in RV64.

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,795 @@
...
+                            implicit_state{false, true});
+    if (!res.second) {
+      // Register already existed, so was read. Mark it written.
+      res.first->second.written = true;
+    }
+  }
+  return regs;
+}
+
+cs_riscv_op make_reg_op(int32_t reg, uint8_t access) {
```
kupsch: `op.access` has type enum `1cs_ac_type`. You can't assign an int to enum, so just use the correct type for the parameter here and in the declaration:

```suggestion
cs_riscv_op make_reg_op(int32_t reg, cs_ac_type access) {
```

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,795 @@
...
+  return is_cft(insn) || insn.getOperation().getID() == riscv64_op_auipc;
+}
+
+struct implicit_state final {
+  bool read, written;
+};
+
+std::map<riscv_reg, implicit_state> implicit_registers(cs_detail const *);
+cs_riscv_op make_reg_op(int32_t, uint8_t);
+cs_riscv_op make_imm_op(int64_t);
```
kupsch: include parameter names in declarations here (and elsewhere).

instructionAPI/h/Instruction.h
```diff
@@ -65,6 +65,7 @@ namespace Dyninst { namespace InstructionAPI {
     friend class InstructionDecoder_amdgpu_gfx908;
     friend class InstructionDecoder_amdgpu_gfx90a;
     friend class InstructionDecoder_amdgpu_gfx940;
+    friend class riscv_decoder;
```
kupsch: Why does `riscv_decoder` uses a different naming convention from all the other decoders?  Why not `InstructionDecoder_riscv`?

kupsch:
This builds with and without capstone, but when I open a RISC-V binary using examples's disassemble, it fails:

```
disassemble: /nobackup/kupsch/kupsch/dyninst/parseAPI/src/IA_IAPI.C:128: static Dyninst::InsnAdapter::IA_IAPI* Dyninst::InsnAdapter::IA_IAPI::makePlatformIA_IAPI(Dyninst::Architecture, Dyninst::InstructionAPI::InstructionDecoder, Dyninst::Address, Dyninst::ParseAPI::CodeObject*, Dyninst::ParseAPI::CodeRegion*, Dyninst::InstructionSource*, Dyninst::ParseAPI::Block*): Assertion `!"unimplemented architecture"' failed.
```

This is correct if Capstone is not enabled, but not if it is enabled. The function `IA_IAPI::makePlatformIA_IAPI` should have a case for RISCV, which it does not. Is the update to this file missing?  Are there other files missing from this PR?

wxrdnx:
This is odd.
It looks like the code fail in Parse API, but it should fail in Instruction API instead.
Somehow it passes through the Capstone conditional compiling check in Instruction API, where it shouldn't.

wxrdnx:
@kupsch 
Never mind, I think this behavior is correct.
the disassemble example in the `examples` directory uses Parse API, and RISC-V Parse API is not included in this PR.

common/src/arch-riscv64.h
```diff
@@ -0,0 +1,137 @@
...
+using rvReg_t = int32_t;
+constexpr int RISCV_REG_SIZE = sizeof(rvReg_t);
+
+// The immeidate size can be 12 or 20 bits
+// They can be fit in a 32 bit integer
```
kupsch:
```suggestion
// The immediate size can be 12 or 20 bits
// Both fit in a 32 bit integer
```

instructionAPI/src/ArchSpecificFormatters.C
```diff
@@ -143,6 +143,86 @@ std::string ArmFormatter::formatBinaryFunc(const std::string &left, const std::s
...
+    return "0x" + evalString;
+}
+
+std::string RiscvFormatter::formatRegister(const std::string &regName) const  {
+    std::string::size_type substr = regName.rfind(':');
+    std::string ret = regName;
+
+    if (substr != std::string::npos) {
+        std::string regStr = ret.substr(substr + 1, ret.length());
+        if (regStr[0] == 'x') {
```
kupsch: Since you're converting the register names to the ABI names, should the same be done for floating point registers: `f<N>  ->  fa<N> | ft<N> | fs<N>`?

instructionAPI/src/InstructionDecoderImpl.C
```diff
@@ -66,6 +70,13 @@ namespace Dyninst
                 case Arch_aarch32:
                 case Arch_aarch64:
                     return Ptr(new InstructionDecoder_aarch64(a));
+                case Arch_riscv64:
+#if defined(DYNINST_ENABLE_CAPSTONE)
+                    return Ptr(new riscv_decoder(a));
+#else
+                    assert(!"Capstone 6.0.0-Alpha or above is required for RISC-V instruction parsing.");
```
kupsch:
```suggestion
                    assert(!"Dyninst must be configured with -DDYNINST_ENABLE_CAPSTONE=ON to parse RISC-V instruction.");
```

instructionAPI/src/InstructionDecoderImpl.C
```diff
@@ -32,10 +32,14 @@
...
 #include "InstructionDecoder-power.h"
 #include "InstructionDecoder-aarch64.h"
+#ifdef DYNINST_ENABLE_CAPSTONE
+#include "decoder/riscv/decoder.h"
+#endif
 #include "AMDGPU/gfx908/InstructionDecoder-amdgpu-gfx908.h"
 #include "AMDGPU/gfx90a/InstructionDecoder-amdgpu-gfx90a.h"
 #include "AMDGPU/gfx940/InstructionDecoder-amdgpu-gfx940.h"
 #include "MultiRegister.h"
+#include "MultiRegister.h"
```
kupsch: No need to include this header twice.
```suggestion
```

hainest:
Please rebase. Sorry for the long delay. Finding the testing bugs was painful.

kupsch:
**Hold on rebasing.** I'm in the middle of reviewing this, and don't want my comments lost.

hainest:
> **Hold on rebasing.** I'm in the middle of reviewing this, and don't want my comments lost.

How about a merge commit to bring in master and resolve conflicts to get the test workflows running?

kupsch (commented):
Here are some of my comments, including a fix to make decoder.C compile.

instructionAPI/src/InstructionDecoderImpl.C
kupsch: Merging in master would be good.  Angus is traveling today.

instructionAPI/h/Instruction.h
hainest: He took that convention from my branch. When changing the directory structure from `instructionAPI/src` to `instructionAPI/src/decoder/x86`, I changed the filename to `decoder.C` and the class name to `x86_decoder`. I guess the previous authors were too influenced by Java and named their classes after the files. I'm happy to discuss a naming convention. Ultimately, the only place those names will ever be used is in `InstructionDecoderImpl::makeDecoderImpl` once `Operation` and `Instruction` get sensible interfaces.

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,795 @@
...
+
+struct eflags_t final {
+  Dyninst::MachRegister reg;
+  implicit_state state;
+};
```
kupsch: Unused?

```suggestion
```

wxrdnx: Yes, it is unused.
Removed

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,795 @@
...
+  constexpr bool is_fallthrough = true;
+
+  auto const pc = makeRegisterExpression(MachRegister::getPC(this->m_Arch));
+  auto const reg_type = size_to_type_signed(RISCV_REG_SIZE);
+  auto const imm_type = size_to_type_signed(RISCV_IMM_SIZE);
+  switch (insn.getOperation().getID()) {
+  case riscv64_op_jal: {
+    MachRegister link_reg = riscv::translate_register(
+        static_cast<riscv_reg>(operands[0].reg), this->mode);
+    auto const rd = makeRegisterExpression(link_reg);
```
kupsch: Unused:
```suggestion
```

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,795 @@
...
+    auto const target = makeAddExpression(std::move(pc), imm, imm_type);
+    const bool is_call_insn = link_reg != riscv64::zero;
+    const Instruction::CFT cft(std::move(target), is_call_insn, !is_indirect, !is_conditional, !is_fallthrough);
+    insn.addSuccessor(std::move(cft));
+    break;
+  }
+  case riscv64_op_jalr: {
+    MachRegister link_reg = riscv::translate_register(
+        static_cast<riscv_reg>(operands[0].reg), this->mode);
+    auto rd = makeRegisterExpression(link_reg);
```
kupsch: Unused:
```suggestion
```

wxrdnx: Removed

instructionAPI/src/decoder/riscv/decoder.C
wxrdnx: Removed

common/src/arch-riscv64.h
wxrdnx: Comments are updated.

instructionAPI/src/decoder/riscv/decoder.C
wxrdnx: Fixed.

instructionAPI/src/decoder/riscv/decoder.C
wxrdnx: Fixed

common/src/arch-riscv64.h
wxrdnx: Comments changed.

instructionAPI/src/InstructionDecoderImpl.C
wxrdnx: Changed.

instructionAPI/src/ArchSpecificFormatters.C
wxrdnx: Fixed

hainest:
@wxrdnx Could you `git merge master` to resolve the conflicts and let the latest CI workflows run?

wxrdnx:
@hainest 
No problem.
I am currently enhancing the unit test
I will resolve the conflicts once they are done

wxrdnx:
> @wxrdnx Could you `git merge master` to resolve the conflicts and let the latest CI workflows run?

@hainest 
Done

tests/integration/InstructionAPI/decoder/riscv/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+include_guard(GLOBAL)
+
+add_executable(rv64i_base rv64i_base.cpp)
+target_compile_options(rv64i_base PRIVATE ${SUPPORTED_CXX_WARNING_FLAGS})
+target_link_libraries(rv64i_base PRIVATE instructionAPI cft_tests memory_tests
+                                         register_tests)
+add_test(NAME instructionAPI_rv64i_base COMMAND rv64i_base)
+set_tests_properties(instructionAPI_rv64i_base PROPERTIES LABELS "integration")
```
kupsch: @hainest, I think these test names would be better with `riscv64` in them `instructionAPI_riscv64_rv64i_base`.

Should we come up with a consistent naming convention for tests to put in `tests/README.md`?

I'd propose `<area>_<test>` and `<area>_<arch>_<test>` (or `<area>_<test>_<arch>` as this is the current convention):

- area (lowercase, we can make a list, subareas separated with `_`)
- arch (lowercase, we can make a list)
- test (lower case in general; sub-tests/words separated by `_`; camel case, uppercase OK if appropriate)

instructionAPI/h/Instruction.h
wxrdnx: @hainest @kupsch 
So, should I change the name to `InstructionDecoder_riscv`, or keep it as `riscv_decoder`?

wxrdnx:
@kupsch 
I updated all unit tests to include all possible RISC-V registers.

wxrdnx:
@hainest @kupsch 
The code that convert compressed instructions to base instructions are all done.

wxrdnx:
@kupsch 
The segfault is solved.

instructionAPI/h/Instruction.h
```diff
@@ -71,20 +72,48 @@ namespace Dyninst { namespace InstructionAPI {
...
+            return true;
+          }
+        }
+        return false;
+      }
+    };
+
     DYNINST_EXPORT Instruction(Operation what, size_t size, const unsigned char* raw,
                                Dyninst::Architecture arch);
+    Instruction(Operation what, Operation encoded_what, size_t size, const unsigned char* raw,
```
kupsch: Export this so users of the library can use this method:
```suggestion
    DYNINST_EXPORT Instruction(Operation what, Operation encoded_what, size_t size, const unsigned char* raw,
```

instructionAPI/h/Instruction.h
```diff
@@ -71,20 +72,48 @@ namespace Dyninst { namespace InstructionAPI {
...
+                Dyninst::Architecture arch);
     DYNINST_EXPORT Instruction();
 
     DYNINST_EXPORT Operation& getOperation();
     DYNINST_EXPORT const Operation& getOperation() const;
 
+    Operation& getEncodedOperation();
+    const Operation& getEncodedOperation() const;
```
kupsch: Export these also:
```suggestion
    DYNINST_EXPORT Operation& getEncodedOperation();
    DYNINST_EXPORT const Operation& getEncodedOperation() const;
```

instructionAPI/h/Instruction.h
```diff
@@ -71,20 +72,48 @@ namespace Dyninst { namespace InstructionAPI {
...
+
     DYNINST_EXPORT std::vector<Operand> getAllOperands() const;
     DYNINST_EXPORT std::vector<Operand> getExplicitOperands() const;
     DYNINST_EXPORT std::vector<Operand> getImplicitOperands() const;
 
     DYNINST_DEPRECATED("Use getallOperands()") DYNINST_EXPORT
     void getOperands(std::vector<Operand>& operands) const;
     DYNINST_EXPORT Operand getOperand(int index) const;
+    Operand getEncodedExplicitOperand(int index) const;
```
kupsch: Export this also:
```suggestion
    DYNINST_EXPORT Operand getEncodedExplicitOperand(int index) const;
```

instructionAPI/h/Instruction.h
```diff
@@ -71,20 +72,48 @@ namespace Dyninst { namespace InstructionAPI {
...
     DYNINST_EXPORT Instruction();
 
     DYNINST_EXPORT Operation& getOperation();
     DYNINST_EXPORT const Operation& getOperation() const;
 
+    Operation& getEncodedOperation();
+    const Operation& getEncodedOperation() const;
+
     DYNINST_EXPORT std::vector<Operand> getAllOperands() const;
     DYNINST_EXPORT std::vector<Operand> getExplicitOperands() const;
```
kupsch, line 111: We should also add a `DYNINST_EXPORT std::vector<Operand>getExplicitEncodedOperands() const` method

instructionAPI/src/Instruction.C
```diff
@@ -97,7 +98,14 @@ namespace Dyninst { namespace InstructionAPI {
...
                                           Dyninst::Architecture arch)
-      : m_InsnOp(what), m_Valid(is_valid_mnemonic(arch, what.getID())), m_size{static_cast<decltype(m_size)>(size)},
+      : m_InsnOp(what), m_EncodedInsnOp(what), m_Valid(is_valid_mnemonic(arch, what.getID())), m_size{static_cast<decltype(m_size)>(size)},
+        arch_decoded_from(arch), formatter(&ArchSpecificFormatter::getFormatter(arch)) {
+    copyRaw(size, raw);
+  }
+
+  Instruction::Instruction(Operation what, Operation encoded_what, size_t size, const unsigned char* raw,
+                                          Dyninst::Architecture arch)
+      : m_InsnOp(what), m_EncodedInsnOp(encoded_what), m_Valid(is_valid_mnemonic(arch, what.getID())), m_size{static_cast<decltype(m_size)>(size)},
```
kupsch: In both constructors put the m_size initializer on its own line to reduce the line length.

hainest:
@wxrdnx Please merge in master to get the latest CI fixes. I've noticed a few more pushes. Is this ready for re-review?

wxrdnx:
> @wxrdnx Please merge in master to get the latest CI fixes. I've noticed a few more pushes. Is this ready for re-review?

Merged.
Yes, it is ready for review. Thanks!

common/src/arch-riscv64.h
```diff
@@ -0,0 +1,138 @@
...
+constexpr int RISCV_MAX_INSN_SIZE = RISCV_INSN_SIZE;
+static_assert(sizeof(rvMaxInsn_t) == RISCV_MAX_INSN_SIZE, "rvMaxInsn_t size mismatch");
+
+// The bits in the register is 64 bits
+constexpr int RISCV_REG_SIZE = 8;
+// The immediates in instructions varies.
+// The maximum is 21 bits, which can be stored in a 32 bit integer
+constexpr int RISCV_IMM_SIZE = 4;
+
+// RISC-V general purpose registers values
```
kupsch: What are these register values?  Are they value used to encode the register in an instruction? Are they another enumeration or scalar RISC-V registers?  Are they the value that capstone uses? Please update the comment.

instructionAPI/src/decoder/riscv/categories.h
```diff
@@ -0,0 +1,166 @@
...
+
+  // Remove any duplicates
+  std::sort(categories.begin(), categories.end());
+  categories.erase(std::unique(categories.begin(), categories.end()),
+                   categories.end());
+
+  // None were found
+  if (!categories.size()) {
+    categories.push_back(di::c_NoCategory);
+  }
```
kupsch: I don't think this necessary. `getCategory` returns `c_NoCategory` if the vector is empty.

```suggestion
```

instructionAPI/src/decoder/riscv/decoder.h
```diff
@@ -0,0 +1,80 @@
...
+    cs_insn *insn{};
+  };
+
+private:
+  cs_mode mode{};
+  disassem disassembler{};
+
+public:
+  InstructionDecoder_riscv64(Dyninst::Architecture a);
+  InstructionDecoder_riscv64() = delete;
```
kupsch: Since there is a user defined construction declared, C++ will not create an implicit default constructor, so this is unnecessary:
```suggestion
```

tests/integration/InstructionAPI/decoder/riscv/c_ext.cpp
```diff
@@ -0,0 +1,383 @@
...
+ *  Chapter 27: "C" Extension for Compressed Instructions, Version 2.0
+ */
+
+namespace di = Dyninst::InstructionAPI;
+
+struct c_ext_tests {
+  std::vector<unsigned char> opcode;
+  di::register_rw_test regs;
+  di::mem_test mem;
+};
```
kupsch, line 26: @hainest, @wxrdnx,

A better name for the `opcode` member would be `rawBytes` (here and in the other tests). Should we add `opcode` and `encodedOpcode` that the instructions`operation->getID()`, so we also test that the instruction decode to the correct opcode?  Are there other characteristics, like the value returned from `format()` (this would make sure explicit operands are sane) that would be useful.

instructionAPI/src/decoder/riscv/opcode_xlat.C
```diff
@@ -0,0 +1,668 @@
...
+#include "opcode_xlat.h"
+
+namespace Dyninst {
+namespace InstructionAPI {
+namespace riscv {
+
+// clang-format off
+
+  // Use the entryID that Capstone has and ignore aliases found in entryIDs.h
+  entryID translate_encoded_opcode(riscv_insn id) {
```
kupsch: Is there a reason to have both `translate_encoded_opcode` and `translate_opcode`?  If not, just combine them.  A more descriptive name would be `capstone_opcode_to_entryID`

instructionAPI/src/decoder/riscv/opcode_xlat.C
```diff
@@ -0,0 +1,668 @@
...
+      case RISCV_INS_UNIMP: return "unimp";
+      case RISCV_INS_URET: return "uret";
+      case RISCV_INS_WFI: return "wfi";
+      case RISCV_INS_XOR: return "xor";
+      case RISCV_INS_XORI: return "xori";
+      default: return "INVALID";
+    }
+  }
+
+  std::string translate_mnemonic(riscv_insn id) {
```
kupsch: I think this would be better structures to have a `riscv_insn  get_capstone_uncompress_opcode(riscv_insn id)` function. Then combine `translate_mnemonic` and `translate_encoded_mnemonic` into one function `capstone_opcode_to_mnemonic` (that doesn't translate) and hopefully can use a capstone function instead of maintaining a switch statement. Then in decode you'd just call `capstone_opcode_to_mnemonic(get_capstone_uncompress_opcode(opcode))` to get the encode opcode.

instructionAPI/src/Instruction.C
wxrdnx: Done

instructionAPI/h/Instruction.h
wxrdnx: Done

common/src/arch-riscv64.h
wxrdnx: Comments are added.

These are raw register encoding used in RISC-V instruction encoding.
They are used to identify registers when decoding instructions.
They are also used in Codegen when emitting raw instructions.

instructionAPI/src/decoder/riscv/opcode_xlat.C
wxrdnx: ~`translate_opcode` will return the unencoded instruction opcode. For example, `c.jalr` become `jalr`.
`translate_encoded_opcode` will return its encoded opcode. For example, `c.jalr` remains `c.jalr`.~

wxrdnx: ~Capstone has no information on the unencoded opcode of a compressed instruction.
It treats compressed instructions as though it was a completely different insturction.
Because of this, we need to manually translate the compressed opcode to its corresponding uncompressed opcode.~

Update: @kupsch I removed  `translate_encoded_opcode`. I reimplemented it using `translate_encoded_opcode` as `translate_opcode(get_capstone_uncompress_opcode(encoded_opcode))`, as suggested in https://github.com/dyninst/dyninst/pull/2046#review-thread-or-comment-id-1690813615

instructionAPI/src/decoder/riscv/opcode_xlat.C
wxrdnx: Ok, will do.

kupsch: I meant `get_capstone_uncompressed_opcode` (missing `ed`)

wxrdnx: > I meant `get_capstone_uncompressed_opcode` (missing `ed`)

Okay. Changed to `get_capstone_uncompressed_opcode`.

wxrdnx: @kupsch All done. Ready to be reviewed.

instructionAPI/src/decoder/riscv/opcode_xlat.h
```diff
@@ -0,0 +1,49 @@
...
+#include "capstone/riscv.h"
+#include "entryIDs.h"
+
+namespace Dyninst {
+namespace InstructionAPI {
+namespace riscv {
+
+  riscv_insn get_capstone_uncompressed_opcode(riscv_insn id);
+  std::string capstone_opcode_to_mnemonic(riscv_insn id);
+  entryID translate_opcode(riscv_insn id);
```
kupsch: Let's use a descriptive function name here also:

```suggestion
  entryID capstone_opcode_to_entryid(riscv_insn id);
```

kupsch (commented):
one more small change. We still need to discuss the tests.

instructionAPI/src/decoder/riscv/decoder.C
```diff
@@ -0,0 +1,892 @@
...
+cs_riscv_op make_imm_op(int64_t imm) {
+  cs_riscv_op op{};
+  op.type = RISCV_OP_IMM;
+  op.imm = imm;
+  op.access = CS_AC_READ;
+  return op;
+}
+
+bool is_compressed(di::Instruction &insn) {
+  return insn.size() == 2;
```
kupsch, line 889: Since there is a constant for the RISC-V C instruction size, use it instead of the magic constant:
```suggestion
  return insn.size() == RISCVC_INSN_SIZE;
```

wxrdnx: done

tests/integration/InstructionAPI/decoder/riscv/c_ext.cpp
wxrdnx: @hainest 
Any thoughts on this?

hainest: > A better name for the opcode member would be rawBytes

Agreed.

> Should we add opcode and encodedOpcode that the instructionsoperation->getID(), so we also test that the instruction decode to the correct opcode?

Yes, this would be good to test.

> Are there other characteristics, like the value returned from format() (this would make sure explicit operands are sane) that would be useful.

I wouldn't test the `format()` string, but I was planning on adding the explicit registers as a separate item in `register_rw_test`.

---

I had considered having a generic `instruction_test`, but thought it would be too generic. It's probably most useful to have something like

```c++
struct instruction_test {
  std::vector<unsigned char> raw_bytes;
  entryID opcode;
  register_rw_test regs;
  memory_rw_test mem;
};
```

I think the `cft_test` is used infrequently enough, it doesn't really need to be part of that. Other things to include are all of the instruction properties like `isConditional`. I would add those at the test level (e.g., `branch_test`) instead of in something like `instruction_test`.

wxrdnx: @kupsch @hainest 
Ok, I will add `opcode` and `encodedOpcode` to the tester

wxrdnx: @kupsch @hainest 
Does this test looks good?
```cpp
#include "opcode_tests.h"

#include "Architecture.h"

#include <iostream>

namespace Dyninst { namespace InstructionAPI {

  bool verify(Instruction const &insn, opcode_test const &expected) {
    Operation actual_opcode = insn.getOperation();
    Operation actual_encoded_opcode = insn.getEncodedOperation();
    Operation expected_opcode = expected.opcode;
    Operation expected_encoded_opcode = expected.encoded_opcode;

    bool failed = false;

    if(actual_opcode.getID() != expected_opcode.getID()) {
      std::cerr << "Mismatched opcode\n";
      std::cerr << "Expected: " << expected_opcode.format() << ", Found: " << actual_opcode.format();
      std::cerr << '\n';
      failed = true;
    }

    if(actual_encoded_opcode.getID() != expected_encoded_opcode.getID()) {
      std::cerr << "Mismatched encoded opcode\n";
      std::cerr << "Expected: " << expected_encoded_opcode.format() << ", Found: " << actual_encoded_opcode.format();
      std::cerr << '\n';
      failed = true;
    }

    return !failed;
  }
}}
```

kupsch: I don't think that you want to construct an `Operation` object just to the get the `entryID`. Just store the two `entryID`s as part of the structure. The `Operation` isn't the opcode, so I'd add the `.getID()` method call so `actual_opcode` and `actual_encoded_opcode` are the opcode. Otherwise looks OK.

wxrdnx: @kupsch 
But then how do we print the error message out?
The Operation class has `format`, but entryID does not (it's enum).

wxrdnx: Ok I think I will use `entryID` instead, and write a translation function to print the error message out.

kupsch: You could include the `entryID`s and the mnemonic string in each test struct.

wxrdnx: Got it

wxrdnx: @kupsch @hainest 
All done.
Ready to be reviewed

wxrdnx:
@kupsch @hainest
I added instruction opcode check to integration test.
The new tests are ready to be reviewed.
Thanks.

tests/integration/InstructionAPI/decoder/riscv/a_ext.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  constexpr bool reads_memory = true;
+  constexpr bool writes_memory = true;
+
+  // clang-format off
+  return {
+    // --- Load-Reserved / Store-Conditional ---
+    { // lr.w t0, (s1)
+      {0xaf,0xa2,0x04,0x10},
+      di::opcode_test{riscv64_op_lr_w, riscv64_op_lr_w, "lr.w", "lr.w"},
```
kupsch: A lot of redundant code could be removed by adding a second constructor that had two parameters since for non-compressed instructions the entryID and mnemonic are identical for both the encoded and uncompressed values:

```suggestion
      di::opcode_test{riscv64_op_lr_w, "lr.w"},
```

wxrdnx: @kupsch 
All done

tests/opcode_tests.h
```diff
@@ -0,0 +1,59 @@
...
+  std::string opcode_mnemonic;
+  std::string encoded_opcode_mnemonic;
+  opcode_test(entryID op, entryID enc_op, const std::string &op_mnem,
+              const std::string &enc_op_mnem)
+      : opcode(op), encoded_opcode(enc_op), opcode_mnemonic(op_mnem),
+        encoded_opcode_mnemonic(enc_op_mnem) {}
+  // Constructor for the case where the opcode is the same as the encoded opcode
+  opcode_test(entryID op, const std::string &op_mnem)
+      : opcode(op), encoded_opcode(op), opcode_mnemonic(op_mnem),
+        encoded_opcode_mnemonic(op_mnem) {}
```
kupsch: You can use a delegating constructor here:
```suggestion
  opcode_test(entryID op, const std::string &op_mnem)
      : opcode_test(op, op, op_mnem, op_mnem) {}
```

wxrdnx:
There's a warning/error that the stack frame exceeds 20480 bytes (`-Wframe-larger-than`).
It is caused by constructing the test vector that is too big.
I'll try to fix it, possibly by splitting the test cases.

kupsch:
Let's see if we can figure out the cause of the large stack frame. Then we can fix it or increase the stackframe size allowed.

wxrdnx:
> Let's see if we can figure out the cause of the large stack frame. Then we can fix it or increase the stackframe size allowed.

I fixed it.
It's cause by constructing a big vector on stack in `make_tests64`.
My solution is to either split the test case into smaller ones, or remove test cases that looks like duplicate.

kupsch:
Yes, I could see the line of code that resulted in the warning. I still doesn't explain the underlying cause. A std::vector object should only take about 24 bytes or so (1 pointer to the base of element storage, 1 pointer to end of reserved, 1 pointer to end of allocated). All the rest of the storage should be on the heap. The compiler must be optimizing the construction using the stack instead of the heap (that is the unexplained underlying cause).

Any change to make the vector smaller in size just makes a future problem when we need to add more tests. Like I said it would be better to increase the allowed stack size now instead of eventually doing so in the future. This is not that difficult.

Are the tests structured better as they are now or as they were 2 commit earlier? If they were structured better 2 commits ago, we should revert and I'll fix the warning be increasing the allowed stack size for these tests.

wxrdnx:
> Yes, I could see the line of code that resulted in the warning. I still doesn't explain the underlying cause. A std::vector object should only take about 24 bytes or so (1 pointer to the base of element storage, 1 pointer to end of reserved, 1 pointer to end of allocated). All the rest of the storage should be on the heap. The compiler must be optimizing the construction using the stack instead of the heap (that is the unexplained underlying cause).

Ok. I'll try to take a look at the generated code to unearth the underlying reason.

> Any change to make the vector smaller in size just makes a future problem when we need to add more tests. Like I said it would be better to increase the allowed stack size now instead of eventually doing so in the future. This is not that difficult.

I think we can increase the stack size to something like 32768. I think that will be enough.

> Are the tests structured better as they are now or as they were 2 commit earlier? If they were structured better 2 commits ago, we should revert and I'll fix the warning be increasing the allowed stack size for these tests.

I think they are structured better now, except for the `c_ext`, which is split into two functions for testing. The ideal case is to use only one function.
I think we can revert the `c_ext` test case after increasing the stack size.

kupsch:
OK please revert, and I'll increase the frame limit. I manually ran the compiler multibuild (that normally runs once per week), and with your current changes 15 of the compiler/OS/optimization-level tests failed with probably 40+ tests still to run.

wxrdnx:
@kupsch Reverted.

wxrdnx:
@kupsch 
Here's what I found for the size of the test structure.
For each of the test struct
```c
struct m_ext_tests {    
  std::vector<unsigned char> rawBytes;    
  di::opcode_test opcodes;    
  di::register_rw_test regs;    
  di::mem_test mem;    
};
```
account for 360 bytes.

The breakdown is as follows:
- `rawBytes`: 24 bytes (a `std::vector<char>` is 24 bytes: 8 bytes for pointer to data on heap, 8 bytes for the size of the vector, 8 bytes for the capacity).
- `opcodes`: 72 bytes
  - `opcodes.opcode` and `opcodes.encoded_opcode`: 4 bytes each (basically an int)
  - `opcodes.opcode_mnemonic` and `opcodes.encoded_opcode_mnemonic`: 32 bytes each (The opcode string is smaller than 32 bytes, so internally C++ stores std::string directly as static char[32]).
- `regs`: 128 bytes
  - `regs.read` and `regs.write`: 64 bytes each (Dyninst::register_set is basically `std::unordered_set<Dyninst::MachRegister, detail::reg_hasher>`. I don't know how `unordered_set` is implemented in C++, it seems complicated)
- `mem`: 136 bytes
  - `mem.readsMemory` and `mem.writesMemory`: 8 bytes each (It's a boolean, and a boolean account for 8 bytes unfortunately)
  - `mem.regs`: 128 bytes, same as `regs`.

kupsch:
After 3.5 hours of running the PR and the compiler multibuild CI tests now pass.

hainest:
@wxrdnx @kupsch

Sorry, I've been working on spack stuff, so I'm a little behind. Why are we checking the mnemonic string? That's likely to introduce maintenance hazards on x86 between ATT and Intel syntax as we migrate to Capstone. I definitely agree we should check the opcode (I really should have done that in my original tests).

I would also recommend we update the `register_rw_test` to add a separate test for the explicit operands (i.e., `Instruction::getExplicitOperands`). Additionally checking the properties (e.g., `isConditional`) would be good, but definitely adds a lot of fields to the tests. Maybe that's not an issue.

---

cf the stack size expansion: we could just always build the tests in release mode. There's no particular reason to build them in the same mode as the rest of Dyninst since they are just using Dyninst. It would be akin to building the testsuite in debug mode because Dyninst was.

If it would be useful to have a quick chat, I'm free whenever.

hainest (approved):

PR #2047, opened 2025-11-03
Add access specifiers in Operand::format

hainest:
This is very useful for debugging.

### DIFF
--- a/instructionAPI/src/Operand.C
+++ b/instructionAPI/src/Operand.C
@@ -134,6 +134,21 @@ namespace Dyninst { namespace InstructionAPI {
       s.replace(0, 2, "0x0("); // fix-up ##X to indirection syntax 0x0(X)
       s += ')';
     }
+    if(this->isRead() || this->isWritten()) {
+      s += "{";
+      bool needs_comma = false;
+      if(this->isRead()) {
+        s += "r";
+        needs_comma = true;
+      }
+      if(this->isWritten()) {
+        if(needs_comma) {
+          s+= ", ";
+        }
+        s += "w";
+      }
+      s += "}";
+    }
     return s;
   }
 


instructionAPI/src/Operand.C
```diff
@@ -134,6 +134,21 @@ namespace Dyninst { namespace InstructionAPI {
       s.replace(0, 2, "0x0("); // fix-up ##X to indirection syntax 0x0(X)
       s += ')';
     }
+    if(this->isRead() || this->isWritten()) {
+      s += "{";
+      bool needs_comma = false;
```
bbiiggppiigg, line 139: I guess the goal is to generate either {r}, {w} or {r,w}, but maybe worth commenting.

bbiiggppiigg (dismissed):

bbiiggppiigg:
> This is very useful for debugging.

On second thoughts, I think these information should optionally printed, either through a different format functions with additional paramters or enabled when debug environment variables are set.

LIke one of the recent changes made to the disassemble example it to make its output resembles that of llvm-objdump, and this change will probably break it I suppose.

At the end of the day, I don't see why this thing is really needed, when you print operands one by one you can always access these information and add them.

If this is for the case when we format an entire instruction, then probably the Instruction::format needs to call a special version of Operand::format.

instructionAPI/src/Operand.C
hainest: Now that I think about it, there's not really a good reason to have a comma. I think `rw` would be clear to anyone who cared about that information.

hainest:
> > This is very useful for debugging.
> 
> On second thoughts, I think these information should optionally printed, either through a different format functions with additional paramters or enabled when debug environment variables are set.
> 
> LIke one of the recent changes made to the disassemble example it to make its output resembles that of llvm-objdump, and this change will probably break it I suppose.

Hmm. If someone were using Dyninst like a simple disassembler, I could see where this would interfere with that. We certainly do not want to imply that instruction formatting is guaranteed to follow any third-party convention. As it is, we only support AT&T syntax. We'll definitely switch to using Capstone's stringification once that is available. It is at least configurable.

I don't have a good workaround for this right now. I'll push these changes into the printers in the integration tests so the default disassembly isn't changed.

> 
> At the end of the day, I don't see why this thing is really needed, when you print operands one by one you can always access these information and add them.

Because then you wouldn't be able to just use `Instruction::format` directly.

> If this is for the case when we format an entire instruction, then probably the Instruction::format needs to call a special version of Operand::format.

That would require an interface on `Instruction` that controls information about how `Operand::format` works. That's just bad.

PR #2048, opened 2025-11-03
Remove InstructionDecoderImpl::setMode

hainest:
It's only used by x86, so move it there.

@wxrdnx This will affect #2046.


hainest:

PR #2051, opened 2025-11-04
make parse_attr lambdas noexcept

kupsch:
This was introduced in PR #1966.  The code as written works with C++ 11, but fails when compiling with C++ version 14 to 23, as std::function wants lambda stored in them to be noexcept.


kupsch:
@hainest this now passes the multicompiler CI

hainest (approved):
LGTM. Although I'm baffled why the libabigail tests are now passing. They were giving some ridiculous failures when I ran it against master this morning. :shrug:

PR #2052, opened 2025-11-05
Refactor Instruction raw representation

hainest:
The current implementation just makes things needlessly complicated.

### DIFF
--- a/instructionAPI/src/Instruction.C
+++ b/instructionAPI/src/Instruction.C
@@ -103,6 +103,8 @@ namespace Dyninst { namespace InstructionAPI {
   }
 
   void Instruction::copyRaw(size_t size, const unsigned char* raw) {
+    assert(size <= m_RawInsn.size() && "Requested size is larger than opcode buffer");
+
     auto last = std::copy_n(raw, size, m_RawInsn.data());
 
     // Zero-fill so 'operator<' and 'operator==' work correctly


instructionAPI/h/Instruction.h
```diff
@@ -147,21 +137,11 @@ namespace Dyninst { namespace InstructionAPI {
     DYNINST_EXPORT cftConstIter cft_end() const { return m_Successors.end(); }
 
     DYNINST_EXPORT bool operator<(const Instruction& rhs) const {
-      if(m_size < rhs.m_size)
-        return true;
-      if(m_size <= sizeof(m_RawInsn.small_insn)) {
-        return m_RawInsn.small_insn < rhs.m_RawInsn.small_insn;
-      }
-      return memcmp(m_RawInsn.large_insn, rhs.m_RawInsn.large_insn, m_size) < 0;
+      return this->m_RawInsn < rhs.m_RawInsn;
```
bbiiggppiigg: If this is doing lexicographical_compare
https://en.cppreference.com/w/cpp/container/vector/operator_cmp.html
Then to make the implementation equivalent we might need to check the size first?

bbiiggppiigg (approved):
LGTM

kupsch:
My only concern about using vector would be the overhead, otherwise this is a great improvement. I don't think that there is a small vector optimization like for strings. Right now the small instruction case (<=4 bytes on non-ppc, could be trivially made <=8 bytes with no further memory usage) takes 8 bytes for the raw_insn_T type + 4 bytes for m_size for 12 bytes, the vector takes 3*8 bytes (the pointer, the size and the capacity) + the allocated buffer size + overhead (for the small case this would be 4 bytes of memory allocation + probably 8 bytes (for length/bookkeeping)  for a total of 36 bytes. Also there would be an allocation and deallocation of the memory in the small case.

Do you think an additional best case 24 bytes per instruction (for 4 byte small size) or 28 bytes (for 8 byte small size), plus the additional new/delete for the buffer will affect performance much?  Once you get beyond the small instruction case the difference is negligible.

instructionAPI/h/Instruction.h
hainest: That's an interesting point. I wasn't worried about making it equivalent to the old one since ordering of instructions is a nonsensical idea.

@kupsch thoughts?

kupsch: I'd vote for **not** making it equivalent. ordering by size and then contents if the sizes are the same (using memcpy if large and integer compare is small) makes for a non-intuitive order.

bbiiggppiigg:
> My only concern about using vector would be the overhead, otherwise this is a great improvement. I don't think that there is a small vector optimization like for strings. Right now the small instruction case (<=4 bytes on non-ppc, could be trivially made <=8 bytes with no further memory usage) takes 8 bytes for the raw_insn_T type + 4 bytes for m_size for 12 bytes, the vector takes 3*8 bytes (the pointer, the size and the capacity) + the allocated buffer size + overhead (for the small case this would be 4 bytes of memory allocation + probably 8 bytes (for length/bookkeeping) for a total of 36 bytes. Also there would be an allocation and deallocation of the memory in the small case.
> 
> Do you think an additional best case 24 bytes per instruction (for 4 byte small size) or 28 bytes (for 8 byte small size), plus the additional new/delete for the buffer will affect performance much? Once you get beyond the small instruction case the difference is negligible.

Do you think
boost::small_vector would be of any help or no?
https://www.boost.org/doc/libs/1_68_0/doc/html/boost/container/small_vector.html

hainest:
> My only concern about using vector would be the overhead, otherwise this is a great improvement. I don't think that there is a small vector optimization like for strings. Right now the small instruction case (<=4 bytes on non-ppc, could be trivially made <=8 bytes with no further memory usage) takes 8 bytes for the raw_insn_T type + 4 bytes for m_size for 12 bytes, the vector takes 3*8 bytes (the pointer, the size and the capacity) + the allocated buffer size + overhead (for the small case this would be 4 bytes of memory allocation + probably 8 bytes (for length/bookkeeping) for a total of 36 bytes. Also there would be an allocation and deallocation of the memory in the small case.
> 
> Do you think an additional best case 24 bytes per instruction (for 4 byte small size) or 28 bytes (for 8 byte small size), plus the additional new/delete for the buffer will affect performance much? Once you get beyond the small instruction case the difference is negligible.

I had considered that. Perhaps a better solution would be `std::array<uint8_t, InstructionDecoder::maxInstructionLength>`? That would give a fixed overhead of 12 bytes on ppc. Although alignment is currently a multiple of at least 8 due to `large_insn` (not sure about `std::list` et al.), so it would really be an overhead of 8 bytes. I also want to remove `arch_decoded_from` since it's also stored in the `Operation`. That would give this PR a zero overhead. If we go the route of `std::array`, I'd rather move `maxInstructionLength` into `Instruction` since that value really belongs to an instruction rather than a decoder. Thoughts are welcome.

hainest:
> > My only concern about using vector would be the overhead, otherwise this is a great improvement. I don't think that there is a small vector optimization like for strings. Right now the small instruction case (<=4 bytes on non-ppc, could be trivially made <=8 bytes with no further memory usage) takes 8 bytes for the raw_insn_T type + 4 bytes for m_size for 12 bytes, the vector takes 3*8 bytes (the pointer, the size and the capacity) + the allocated buffer size + overhead (for the small case this would be 4 bytes of memory allocation + probably 8 bytes (for length/bookkeeping) for a total of 36 bytes. Also there would be an allocation and deallocation of the memory in the small case.
> > Do you think an additional best case 24 bytes per instruction (for 4 byte small size) or 28 bytes (for 8 byte small size), plus the additional new/delete for the buffer will affect performance much? Once you get beyond the small instruction case the difference is negligible.
> 
> Do you think boost::small_vector would be of any help or no? https://www.boost.org/doc/libs/1_68_0/doc/html/boost/container/small_vector.html

I think that would only gain in having no allocations since it will still need at least two pointers to support `push_back`. We should consider it, unless my [suggestion](https://github.com/dyninst/dyninst/pull/2052#issuecomment-3493140705) using `std::array` would work.

bbiiggppiigg:
>I also want to remove `arch_decoded_from` since it's also stored in the `Operation`. 

As long as the Instruction::getArch remains, and the Operation is properly copied when constructing a new Instruction, I think it should be fine?

hainest:
> > I also want to remove `arch_decoded_from` since it's also stored in the `Operation`.
> 
> As long as the Instruction::getArch remains, and the Operation is properly copied when constructing a new Instruction, I think it should be fine?

That would all be the same. There's just no need to store it in both places. Although I just noticed there isn't an `Operation::getArch`. I'll wait until #1646 since it won't be needed in `Operation` after that goes in.

kupsch:
how about changing maxInstructionLength to 15 (I think all architecture's instructions are less than 15 bytes), and make m_size be a uint8_t, then there is only 16 bytes for the array and the size. For `operator<` then becomes `(buf < rhs.buf) || (buf == rhs.buf && m_size < rhs.m_size)` (if we zero the unused elements), `lexicographic_compare` if in Boost (or C++17), or `comp = memcmp(buf, rhs.buf, std::min(m_size, rhs.m_size)); return comp < 0 || (comp == 0 && m_size < rhs.m_size)`.

hainest:
> how about changing maxInstructionLength to 15 (I think all architecture's instructions are less than 15 bytes), and make m_size be a uint8_t, then there is only 16 bytes for the array and the size.

That sounds reasonable to me.

> For `operator<` then becomes `(buf < rhs.buf) || (buf == rhs.buf && m_size < rhs.m_size)` (if we zero the unused elements), `lexicographic_compare` if in Boost (or C++17), or `comp = memcmp(buf, rhs.buf, std::min(m_size, rhs.m_size)); return comp < 0 || (comp == 0 && m_size < rhs.m_size)`.

I think it would be easier to just do zero-fill in the constructor and `updateSize` then use the provided comparison operator- except possibly with c++20 where they were replaced with the spaceship. I'll be sure to run the compiler matrix test on this PR.

instructionAPI/h/Instruction.h
```diff
@@ -149,21 +138,11 @@ namespace Dyninst { namespace InstructionAPI {
...
```
kupsch: `ff` should be less than `ff 00`, both of these have the same bytes in the array, so you need an extra clause to check for equality and then compare the sizes:

```suggestion
      return this->m_RawInsn < rhs.m_RawInsn || (this->m_size < rhs.m_size && this->m_RawInsn == rhs.m_RawInsn);
```

instructionAPI/h/Instruction.h
```diff
@@ -149,21 +138,11 @@ namespace Dyninst { namespace InstructionAPI {
...
     }
 
     DYNINST_EXPORT bool operator==(const Instruction& rhs) const {
-      if(m_size != rhs.m_size)
-        return false;
-      if(m_size <= sizeof(m_RawInsn.small_insn)) {
-        return m_RawInsn.small_insn == rhs.m_RawInsn.small_insn;
-      }
-      return memcmp(m_RawInsn.large_insn, rhs.m_RawInsn.large_insn, m_size) == 0;
+      return this->m_RawInsn == rhs.m_RawInsn;
```
kupsch: `ff` should be not be equal to `ff 00`, both of these have the same bytes in the array, so you need an extra clause to check if the sizes are equal:

```suggestion
      return this->m_size == rhs.m_size && this->m_RawInsn == rhs.m_RawInsn;
```

kupsch (approved):
`operator<` and `operator==` are not quite right, otherwise looks good.

instructionAPI/h/Instruction.h
hainest: You mean like you mentioned in your previous comment that I read and proceeded to immediately forget with my goldfishian memory?

hainest:
@kupsch After doing some experimenting, I found that changing `maxInstructionLength` didn't lead to a better layout- just changing the type of `m_size` to `uint8_t` did. I played around with changing the order of the member variables, but never got the compiler to put `arch_decoded_from` at an offset that allowed for collapsing `m_valid`, `m_RawInsn`, and `m_size` into something smaller than it is here. These changes reduced the total size by 8 bytes relative to the current implementation, removed memory allocations, and reduced complexity, so I'm calling that a win enough.

hainest:

instructionAPI/src/Instruction.C
```diff
@@ -97,26 +97,16 @@ namespace Dyninst { namespace InstructionAPI {
...
-    if(raw) {
-      m_size = size;
-      if(size <= sizeof(m_RawInsn.small_insn)) {
-        memcpy(&m_RawInsn.small_insn, raw, size);
-      } else {
-        m_RawInsn.large_insn = new unsigned char[size];
-        memcpy(m_RawInsn.large_insn, raw, size);
-      }
-    }
+    auto last = std::copy_n(raw, size, m_RawInsn.data());
```
bbiiggppiigg, line 108: Does m_size needs to be updated when copyRaw is called, if you fill the rest with 0?

bbiiggppiigg (changes requested):
LGTM except maybe the comment.

instructionAPI/src/Instruction.C
hainest: It's set in the constructors, but I forgot to update it when called from `updateSize`. Fixed.

bbiiggppiigg: I think it would be better to set it in copyRaw since that would prevent someone (Dyninst developer) from accidentally using copyRaw but forget to update m_size. But since copyRaw is private, maybe it is not that big of a deal?

hainest: > I think it would be better to set it in copyRaw since that would prevent someone (Dyninst developer) from accidentally using copyRaw but forget to update m_size. But since copyRaw is private, maybe it is not that big of a deal?

I tried to set up `m_size` as a class invariant since this function will go away once the decoders have a sane interface for constructing an Instruction. I'm still not entirely certain why the AMDGPU decoder needs to add bytes to the opcode mid-decode, but I'll leave that for another day.

bbiiggppiigg: > I tried to set up `m_size` as a class invariant since this function will go away once the decoders have a sane interface for constructing an Instruction. I'm still not entirely certain why the AMDGPU decoder needs to add bytes to the opcode mid-decode, but I'll leave that for another day.

AMDGPU instructions can use a 32 bit literal constant, but whether or not it is using the constant can only be determined after you decode its operand.

hainest: > AMDGPU instructions can use a 32 bit literal constant, but whether or not it is using the constant can only be determined after you decode its operand.

I see. Does that mean that every instruction has to be speculatively decoded?

bbiiggppiigg: > I see. Does that mean that every instruction has to be speculatively decoded?

Yes, that is what is done right now, and then during the operand decode, we set a flag if we see the need to use a constant.
Then a post-processing step checks if the flag is set, and if so update the instruction size and copy in the extra 32 bit.

This approach has the advantage of able to fill in details of the instructions along the way.
Arugably "cleaner" approaches would be
1. To hold the details of the instruction in temp variables, and create an Instruction object only after its operands are all parsed
2. Create an new Instruction object from a temporary Instruction object, copy everything but update the raw bytes and size.

Since we are suppose to get a new XML-ISA drop sooner or later, this is actually a good time to discuss if we could have done it in a better way, so I can update the implementation when the new drop arrives.

instructionAPI/src/Instruction.C
```diff
@@ -97,81 +97,26 @@ namespace Dyninst { namespace InstructionAPI {
...
```
kupsch, line 108: should we add an assert here, that `size <= m_RawInsn.size()` so we detect that the maximum instruction size has grown?

hainest: Added.

hainest:
No test regressions on 18 machines. All CI workflows are running and passing with spot-checks to ensure ctests is actually executing. This PR is finally ready.

PR #2053, opened 2025-11-06
Add Capstone Workflow

wxrdnx:
This pull request adds Capstone as a third-party dependency for Dyninst, including files related to CMake and Docker environments, separating files from #2046 .

### DIFF
--- a/docker/build_capstone.sh
+++ b/docker/build_capstone.sh
@@ -12,7 +12,7 @@ done
 
 version=$(grep capstone dependencies.versions | awk '{split($0,a,":"); print a[2]}')
 
-git clone https://github.com/capstone-engine/capstone --branch=${version}
+git clone --branch=${version} https://github.com/capstone-engine/capstone
 cd capstone
 mkdir build
 cd build


cmake/tpls/DyninstCapstone.cmake
```diff
@@ -0,0 +1,69 @@
...
+  set(capstone_DIR ${Capstone_ROOT_DIR})
+  mark_as_advanced(capstone_DIR)
+  set(_find_path_args NO_CMAKE_SYSTEM_PATH NO_SYSTEM_ENVIRONMENT_PATH)
+endif()
+
+find_package(capstone REQUIRED ${_find_path_args})
+
+# Only major versions of Capstone are compatible, but we can use anything newer than
+# the required minimum version.
+if(capstone_VERSION VERSION_LESS ${_min_version})
```
hainest, line 43: The [docs](https://cmake.org/cmake/help/latest/command/if.html#version-less) says

```
Component-wise integer version number comparison (version format is
major[.minor[.patch[.tweak]]],  omitted components are treated as zero). Any non-integer
version component or non-integer trailing part of a version component effectively truncates
the string at that point.
```

Did this work for you?

PR #2056, opened 2025-11-07
Refactor tests

hainest:
This is a purely cosmetic change that shouldn't affect any outstanding PR.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -26,7 +26,7 @@ inputs:
   extra-libs:
     required: false
   tests:
-    default: "ALL"
+    default: ""
     required: false
 
 runs:
--- a/.github/workflows/libabigail.yaml
+++ b/.github/workflows/libabigail.yaml
@@ -85,7 +85,6 @@ jobs:
           compiler: ${{ matrix.compiler }}
           src-dir: "${GITHUB_WORKSPACE}/src"
           install-dir: "${GITHUB_WORKSPACE}/install"
-          tests: "OFF"
 
       - name: Build Previous Dyninst
         if: ${{ matrix.compiler != 'gcc' }}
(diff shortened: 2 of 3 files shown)


tests/unit/instructionAPI/syscall-x86.cpp
```diff
@@ -110,31 +110,12 @@ bool run_64()  {
...
-  // Convention for CTest
-  constexpr int PASS =  0;
-  constexpr int FAIL =  1;
+int main() {
+  bool ok = run_32();
 
-  if(argc != 2) {
-    usage(argv[0]);
-    return FAIL;
+  if(!run_64()) {
```
bbiiggppiigg: So this means even if we fail run_32, we still going to run run_64 anyway?

hainest: Yeah. That's how the other tests I've written subsequently work. There's no need to fail early.

tests/unit/instructionAPI/syscall-x86.cpp
```diff
@@ -27,18 +27,22 @@ std::array<const unsigned char, num_bytes> buffer = {{
 
 template <int N>
 bool run(di::InstructionDecoder dec, std::array<bool, N> const& answers) {
+  bool failed = false;
```
bbiiggppiigg, line 44: Maybe change to default failure to true and only set it to true when expected = actual? But not important.

hainest: I'm not sure I understand. It defaults to false, unless a test fails. I did notice that I forgot to fix the early return in the decode step.

```c++
    if(!insn.isValid()) {
      std::cerr << "Decode failed for test " << (i+1) << "\n";
      return false;
    }
```

I'll update that.

PR #2061, opened 2025-11-11
Increase minimum Capstone version to 6.0.0-Alpha5

hainest:
The 'capstone::capstone' target was removed just before the
6.0.0-Alpha5. By making this the minimum, we can guaratee we get 1) a
version of Capstone built with PIC that is needed for linking into
libinstructionAPI, and 2) the needed RISC-V stuff.

@kupsch @wxrdnx
I would prefer to use libcapstone.a, but the provided targets don't export the `POSITION_INDEPENDENT_CODE` property[1], so there isn't a way to check if it's linkable into a DSO. For now, we'll just require their shared library.

[1]
```cmake
get_target_property(_isipic capstone::capstone_static INTERFACE_POSITION_INDEPENDENT_CODE)
get_target_property(_ispic capstone::capstone_static POSITION_INDEPENDENT_CODE)
```

### DIFF
--- a/dataflowAPI/rose/x86_64InstructionSemantics.h
+++ b/dataflowAPI/rose/x86_64InstructionSemantics.h
@@ -710,12 +710,19 @@ struct X86_64InstructionSemantics {
                     }
                     case 4: {
                         // xchg eax,eax is always a nop
-                        SgAsmx86RegisterReferenceExpression* rre0 = isSgAsmx86RegisterReferenceExpression(operands[0]);
-                        X86GeneralPurposeRegister reg0 = (X86GeneralPurposeRegister)(rre0->get_register_number());
-                        SgAsmx86RegisterReferenceExpression* rre1 = isSgAsmx86RegisterReferenceExpression(operands[1]);
-                        X86GeneralPurposeRegister reg1 = (X86GeneralPurposeRegister)(rre1->get_register_number());
-                        if(reg0 == x86_gpr_ax && reg1 == x86_gpr_ax)
+                        bool const is_eax_nop = [&operands]() {
+                          SgAsmx86RegisterReferenceExpression* rre0 = isSgAsmx86RegisterReferenceExpression(operands[0]);
+                          if(!rre0) return false;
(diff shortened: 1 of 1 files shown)


wxrdnx:
lgtm

docker/build_capstone.sh
```diff
@@ -16,7 +16,9 @@ git clone --branch=${version} https://github.com/capstone-engine/capstone
 cd capstone
 mkdir build
 cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=OFF
+
+# Capstone renamed their build args, so use both until we get a stable 6.0 release from them
+cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCAPSTONE_BUILD_SHARED_LIBS=ON -DBUILD_SHARED_LIBS=ON
```
kupsch: I think you don't need `BUILD_SHARED_LIBS=ON` as `CAPSTONE_BUILD_SHARED_LIBS` adds `SHARED` to the library (see capstone's CMakeFile.txt:824:

```suggestion
cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCAPSTONE_BUILD_SHARED_LIBS=ON
```

hainest: > I think you don't need `BUILD_SHARED_LIBS=ON` as `CAPSTONE_BUILD_SHARED_LIBS` adds `SHARED` to the library (see capstone's CMakeFile.txt:824:

`CAPSTONE_BUILD_SHARED_LIBS` isn't available in the 6.0.0-Alpha1 tag. Actually, @wxrdnx is there a reason we can't just use whatever the HEAD of 'next' is here? That would simplify things.

kupsch: How about making the minimum 6.0.0-Alpha5?

wxrdnx: Yeah, we can make the minimum 6.0.0-Alpha5
There's no reason not to do that

hainest: I made the minimum 6.0.0-Alpha5 and required `capstone::capstone_shared` to make sure we get a PIC target to link against.

PR #2062, opened 2025-11-12
Fix detection of conditional jump in slicing

bbiiggppiigg:
Currently, in backward slicing, the detection of the conditional jump assumes both 
COND_TAKEN and COND_NOT_TAKEN are present.

However, under current parsing implementation, it could be the case where only 
one of the edges are present when parsing the indirect control flow in the fallthrough block.

As a result, when backward slicing on the jump table index, the slicer would miss the fact that 
predecessor block ends with a conditional jump instruction, and fail to compute the bound.

Paritally Fixes #2050

PR #2063, opened 2025-11-12
Fix semantic for move 32 bit register

bbiiggppiigg:


kupsch (approved):

hainest (approved):
Tested this with #2062; no regressions on 18 machines.

PR #2067, opened 2025-11-14
Update DYNINST_ENABLE_TESTS option

hainest:
It now takes several possible arguments rather than just being turned on/off.

### DIFF
--- a/cmake/DyninstOptions.cmake
+++ b/cmake/DyninstOptions.cmake
@@ -51,7 +51,7 @@ set(_dyninst_test_types "ALL REGESSION INTEGRATION UNIT")
 mark_as_advanced(_dyninst_test_types)
 
 set(DYNINST_ENABLE_TESTS
-    "NONE"
+    "OFF"
     CACHE STRING "Build tests (${_dyninst_test_types})")
 set_property(CACHE DYNINST_ENABLE_TESTS PROPERTY STRINGS ${_dyninst_test_types})
 


kupsch:
We should also validate that DYNINST_ENABLE_TESTS against typos (all the cmake _false_ values are filtered out earlier):

In tests/CMakeLists.txt after the line

```cmake
string(TOUPPER ${DYNINST_ENABLE_TESTS} _test_flag)
```

add

```cmake
get_property(_valid_tests CACHE DYNINST_ENABLE_TESTS PROPERTY STRINGS)
if(NOT _test_flag IN_LIST _valid_tests)
    message(FATAL_ERROR "Invalid DYNINST_ENABLE_TYPE (${_test_flag}), must be one of ${_valid_tests}")
endif()
```

hainest:
> We should also validate that DYNINST_ENABLE_TESTS against typos (all the cmake _false_ values are filtered out earlier):
> 
> In tests/CMakeLists.txt after the line
> 
> ```cmake
> string(TOUPPER ${DYNINST_ENABLE_TESTS} _test_flag)
> ```
> 
> add
> 
> ```cmake
> get_property(_valid_tests CACHE DYNINST_ENABLE_TESTS PROPERTY STRINGS)
> if(NOT _test_flag IN_LIST _valid_tests)
>     message(FATAL_ERROR "Invalid DYNINST_ENABLE_TYPE (${_test_flag}), must be one of ${_valid_tests}")
> endif()
> ```

Yeah, I noticed that. I'd like to keep the `-DDYNINST_ENABLE_TESTS=<truthy>` to enable unit tests like it currently exists. I'll see what I can do.

cmake/DyninstOptions.cmake
```diff
@@ -47,7 +47,14 @@ mark_as_advanced(DYNINST_CXXSTDLIB)
 option(DYNINST_FORCE_RUNPATH "Require the use of RUNPATH instead of compiler's default"
        OFF)
 
-option(DYNINST_ENABLE_TESTS "Build tests" OFF)
+set(_dyninst_test_types "ALL REGESSION INTEGRATION UNIT")
+mark_as_advanced(_dyninst_test_types)
+
+set(DYNINST_ENABLE_TESTS
+    "NONE"
```
kupsch: Either 

1. `NONE` should get added to `_dyninst_test_types` and handled in `tests/CMakeLists.txt`
2. `NONE` here should be changed to one of the cmake false values (`OFF`?)

kupsch (approved):
Other than the `NONE` problem, it looks good

cmake/DyninstOptions.cmake
hainest: Oops. Good catch. I defaulted to 'OFF'.

kupsch (approved):

PR #2068, opened 2025-11-14
GitHub CI: limit where known binary tests are run

hainest:
There's no need to run those tests when doing build tests. Running the unit, integration, and regression tests is needed to validate the build is usable.


bbiiggppiigg (approved):

kupsch (approved):

hainest:
Good news! This also fixes the libabigail check failures. If software always behaved like this, it would have been outlawed in the 1940s.

hainest:
All CI workflows are passing with these changes.

PR #2073, opened 2025-11-20
Prevent process create failure when threads are spawned in constructor functions

kcossett-amd:
I was debugging an issue for `rocprofiler-systems` involving programs that use `libomptarget`. After some investigation, I was able to narrow down the issue to Dyninst crashing in `processCreate` when threads are spawned in constructor functions (`__attribute__((constructor))`).

**Minimal Reproducer**: Consider the program below:
```
#include <iostream>
#include <pthread.h>
#include <atomic>
#include <unistd.h>

std::atomic<bool> g_keep_running{true};

void* busy_thread(void* arg) {
    while (g_keep_running.load()) {}
    return nullptr;
}

// Spawn threads before main() using constructor attribute. Priority of 101 is not necessary here.
__attribute__((constructor(101)))
void spawn_threads_in_constructor() {
    std::cout << "[Constructor] Spawning thread before main()\n";
    
    pthread_t thread1;
    pthread_create(&thread1, nullptr, busy_thread, (void*)1);
    pthread_detach(thread1);
    g_keep_running.store(false);
}

int main() {
    std::cout << "[Main] Program started\n";
    std::cout << "[Main] Program ending\n";
    return 0;
}
```
And some other program that uses Dyninst (call it dyntest.cpp). In this case, I know the problem occurs in `processCreate`, so we can use:
```
int main(int argc, char* argv[]) {
    BPatch bpatch;
 
    const char* target = argv[1];
    const char* target_argv[] = {target, NULL};
    
    cout << "[MYDYNINSTPROGRAM] Creating process: " << target << endl;
    BPatch_process* app = bpatch.processCreate(target, target_argv);
    std::cerr << "[MYDYNINSTPROGRAM] Process created: " << app << endl;
   
    return 0;
}
```
**Problem 1**

When I pass in the first program, compiled using `amdclang++ program.cpp -o program`, into dyntest.cpp using `./dyntest ./program`, there is a *chance* (it depends on when the helper thread exits) that the following failure will be observed:

```
/usr/include/boost/smart_ptr/shared_ptr.hpp:784: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = const Dyninst::ProcControlAPI::Process; typename boost::detail::sp_member_access<T>::type = const Dyninst::ProcControlAPI::Process*]: Assertion `px != 0' failed.
```

After some debugging and print statements, the failure is due to Dyninst failing to remove the breakpoint at main entry as the thread spawned in the constructor function has its user state as running. 
```
[DEBUG-PRINT] Breakpoint handler setting user states - Event sync type: sync_thread, Address: 0x201c50

[handler.C:2289-U] - Triggering callback for event 'Breakpoint'
[handler.C:2294-U] - Triggering callback #0 for event 'Breakpoint'
[140737348376960]pcEventMuxer.C[388]: removing breakpoint at main
[DEBUG-PRINT] process.C - hasRunningThread() - Thread 2732299/2732299: user=stopped, handler=stopped, generator=stopped
[DEBUG-PRINT] process.C - hasRunningThread() - Thread 2732299/2732314: user=RUNNING, handler=RUNNING, generator=RUNNING
[process.C:7078-U] - Error: User attempted to remove breakpoint on running process
[140737348376960]dynProcess.C[1052]: failed to remove breakpoint at main entry: 0x201c50
/usr/include/boost/smart_ptr/shared_ptr.hpp:784: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = const Dyninst::ProcControlAPI::Process; typename boost::detail::sp_member_access<T>::type = const Dyninst::ProcControlAPI::Process*]: Assertion `px != 0' failed.
/home/rocm/dyninst/dyninstAPI_RT/src/RTlinux.c:434: dyninstTrapHandler: Assertion `hdr' failed.
```

To fix this, I suggest making the main breakpoint force a `sync_process`, that way all threads stop and the breakpoint can be removed.

**Problem 2**

Applying the fix for problem 1 allows the main breakpoint to be removed, but another issue arises. It is possible for the following error to occur:

```
/home/rocm/DyninstTest/multiple-thread-fix/proccontrol/src/event.C:538: virtual Dyninst::ProcControlAPI::Thread::const_ptr Dyninst::ProcControlAPI::EventNewLWP::getNewThread() const: Assertion `thr' failed.
```

From what I understand, an `LWPCreate` is put into the mailbox for Dyninst API to process. HOWEVER, if the helper thread exits before Dyninst can process this event, it will attempt to dereference a null pointer (see https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/pcEventHandler.C#L356).

To fix this, I suggest to check if the non-main `thr` in `EventNewLWP::getLWP()` is null and if it is an `LWPCreate`. If so, we return `Thread::const_ptr()` so that in `PCEventHandler::handleThreadCreate`, we can check for this and ignore the `LWPCreate` event as the thread has most likely exited.

We also need to handle the `LWPDestroy`. For this, I just return `true` in `PCEventHandler::handleThreadDestroy` if `bbproc == NULL` and the event is `LWPDestroy`. 

I made this fix only work for Linux because I am unsure if this would work on Windows.

**Fix Results**: 

With these fixes in, the program will not crash anymore in `processCreate`. Furthermore, LLVM OpenMP offload programs will also work as well (and so will `hsa` ones that call `hsa_init()` in the constructor function). 

**Note**: If you want a reproducer that fails more often, I recommend using some basic OpenMP program and compiling it so that it uses LLVM's `libomptarget` (it calls `hsa_init()` in its constructor assuming amdgpu). In their code, the constructor function should spawn 3 helper threads and has a constructor priority of 101.

### DIFF
--- a/dyninstAPI/src/dynProcess.C
+++ b/dyninstAPI/src/dynProcess.C
@@ -963,8 +963,10 @@ bool PCProcess::insertBreakpointAtMain() {
     }
     Dyninst::Address addr = main_function_->addr();
 
-    // Create the breakpoint
+    // Create the breakpoint and mark it
     mainBrkPt_ = Breakpoint::newBreakpoint();
+    static const uintptr_t MAIN_BREAKPOINT_TAG = 0xDEB19001;
+    mainBrkPt_->setData((void*)MAIN_BREAKPOINT_TAG);
     if( !pcProc_->addBreakpoint(addr, mainBrkPt_) ) {
         startup_printf("%s[%d]: failed to insert a breakpoint at main entry: 0x%lx\n",
                 FILE__, __LINE__, addr);
--- a/dyninstAPI/src/pcEventHandler.C
+++ b/dyninstAPI/src/pcEventHandler.C
@@ -353,6 +353,15 @@ bool PCEventHandler::handleThreadCreate(EventNewThread::const_ptr ev, PCProcess 
   proccontrol_printf("%s[%d]: entering handleThreadCreate for %d/%d\n",
 		     FILE__, __LINE__, evProc->getPid(), ev->getLWP());
 
+    // Check if the thread still exists
+#if defined(os_linux) || defined(os_freebsd)
+    if (!ev->getNewThread()) {
+        proccontrol_printf("%s[%d]: thread %d/%d not found, it may have already exited. Ignoring thread create event.\n",
+            FILE__, __LINE__, evProc->getPid(), ev->getLWP());
+            return true;
+    }
+#endif
+
     if( !ev->getNewThread()->haveUserThreadInfo() ) {
         proccontrol_printf("%s[%d]: no user thread info for thread %d/%d, postponing thread create\n",
                 FILE__, __LINE__, evProc->getPid(), ev->getLWP());
@@ -407,6 +416,13 @@ bool PCEventHandler::handleThreadDestroy(EventThreadDestroy::const_ptr ev, PCPro
 			 FILE__, __LINE__, evProc->getPid(), ev->getThread()->getLWP()); 
         BPatch_process *bpproc = BPatch::bpatch->getProcessByPid(evProc->getPid());
         if( bpproc == NULL ) {
+#if defined(os_linux) || defined(os_freebsd)
+            if (ev->getEventType().code() == EventType::LWPDestroy) {
+                proccontrol_printf("%s[%d]: failed to locate process %d, for corresponding LWPDestroy event. It may have already exited. Ignoring event.\n",
+                    FILE__, __LINE__, evProc->getPid());
+                return true;
+            }
+#endif
             proccontrol_printf("%s[%d]: failed to locate BPatch_process for process %d\n",
                     FILE__, __LINE__, evProc->getPid());
             return false;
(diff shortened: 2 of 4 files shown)


hainest:
I've done very little work in proccontrol, but I'll try to take a look at this. On my first brief pass, my main concern is using a magic number to indicate "this is the main breakpoint".

kcossett-amd:
Consider it a placeholder.

I was originally going to put a string, but that would have required an ugly cast to avoid warnings, although its not too bad. Perhaps some sort of preprocessor var or constexpr could be put in place, but I'm not familiar with the coding conventions used in this repo (or perhaps there is already an identifier for main that I missed).

kcossett-amd:
Hi @hainest  , has there been any updates on your side regarding this PR?

hainest:
Sorry this fell to the side. The development focus over the last couple of months shifted to getting AMD gpu rewriting working. These changes certainly fix the symptom of the crash, but the larger issue is that proccontrol doesn't understand that threads can be created before `main` is entered. That is potentially a much larger problem to solve. There is currently only one full-time dev working on Dyninst and he's also helping three grad students (I'm here voluntarily and have even more limited time). This is further complicated by a lack of experience working with proccontrol. I'll see if I can meet with the other dev this week and make a plan for how we can move forward here to help you get your work finished while minimizing any possible one-off changes to Dyninst.

kcossett-amd:
Handled under #2131

Issue #2050, opened 2025-11-04
bad slicing and non-determinism in parse


mwkrentel:
I'm seeing a lot of non-determinism in `parse()` in Dyninst master.
Dyninst 13.0.0 had a fair amount of non-det in parse, but now master
is hopeless to get two runs to exactly agree 100%.

For example, running `parse()` on the `hpcstruct` binary 50 times, with
master I got 50 unique answers (as verified by md5sum).  13.0.0 would
have a consensus answer about 70-80% of the time with clusters of
other answers.

I've traced the problem in large part to inconsistent indirect jump table
analysis.  That is, there will be an indirect jump instruction (`jmp *%rdx`)
with a data race for the set of outgoing edges.

This is kinda serious, especially for `hpcstruct`.  Bogus edges in the
CFG produce bad and misleading output for hpcstruct.  There are
distorted loops, code that appears in multiple functions, plus it
makes it near impossible to keep a reference answer to test correctness.

The hpcstruct binary (as input to hpcstruct) is a decent stress test
for this because it pulls in code from a variety of sources: C++ maps,
Boost, XED, Dyninst, Intel TBB, OpenMP, etc.  It seems XED is
particularly good at causing problems.

Here are five examples.  I ran hpcstruct (including parse() with
dynininst openmp) at 24 threads analyzing itself on x86_64 and turned
on some knobs to dump the CFG (functions, blocks, edges, linemap and
inline sequences) as computed by Symtab and ParseAPI.  Then I looked
at the diffs of the debug output.

Note: I also tried this serially.  It was deterministic at one thread.
This suggests the problem is a data race instead of an algorithmic
mistake, which I guess is good news.

I've put the binary (`STRUCT`) along with objdump and a few runs on ufront:
```
/home/krentel/Dyninst-slicing
-rw-r--r-- 1 krentel users  28789586 Nov  3 10:58 Objdump
-rwxr-xr-x 1 krentel users  15404976 Nov  3 10:58 STRUCT
-rw-r--r-- 1 krentel users 141577782 Nov  3 10:58 out-2
-rw-r--r-- 1 krentel users 141369516 Nov  3 16:26 out-5
-rw-r--r-- 1 krentel users 141600114 Nov  3 10:58 out-ref
```

`out-ref` is the "reference" output that I got running hpcstruct
serially.  It's not guaranteed to be correct, but usually makes more
sense than the other runs where they differ.

[...]

PR #2074, opened 2025-11-21
Fix jump detection in instruction adapters

bbiiggppiigg:
Certain if predicates are missing a not, introduced in #1813 .

This fix the regression reported in #2050.


hainest:
Ufdah. It is _deeply_ concerning that this mistake didn't get caught in testing.

kupsch (approved):

bbiiggppiigg:
@hainest I had made a comment in #1813 about a regression introduced in IA_power.C 
that might or might not have been cleaned up in #1929 
Can you verify that it is indeed cleaned up?

hainest (approved):
Let's change the title to something like "Fix jump detection in PowerPC, x86".

PR #2077, opened 2025-12-01
GitHub CI: force test directory

hainest:
When run under GitHub's CI, ctest doesn't find the build directory specified by
'--test-dir' _and_ it doesn't raise an error even though '--no-tests=error'.
This fixes both.


hainest:
This is actually running the unit tests now. It was running them when I did my previous set of changes, as well, so I don't know what that's worth.

kupsch (approved):

PR #2078, opened 2025-12-12
Pin GHA dependencies by hash

williamjallen:
Hash-pinned dependencies are an important requirement enforced by the OpenSSF scorecard tool: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

This commit pins all GitHub Actions workflows to a commit hash.

### DIFF
--- a/.github/workflows/binary-tests.yaml
+++ b/.github/workflows/binary-tests.yaml
@@ -38,7 +38,7 @@ jobs:
       latest: ${{ steps.all.outputs.latest }}
     steps:
       - name: Checkout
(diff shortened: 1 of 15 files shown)


hainest:
Will the GitHub action dependabot still pick these up for automatic updating?

williamjallen:
@hainest Yes, Dependabot will continue updating as usual, but will reference the commit hash for each version instead of the version tag.  Dependabot will also update the version comment.  You can see examples in a bunch of repos linked here: https://github.com/dependabot/dependabot-core/issues/4691

PR #2079, opened 2025-12-15
Add SAIL to ROSE parser

wxrdnx:
- Refactored the SAIL to ROSE parser. 
  - Split the original Perl script into various modules.
  - Separate internal AST node construction, node adjustment, and code printing.
  - Add dispatcher of SAIL functions to ROSE functions to improve extensibility
- Updated the SAIL parser to v0.20 (the current latest version) to support sail-riscv v0.9.
- Add documentation of how to use the parser.

### DIFF
--- a/dataflowAPI/sail/ASTBuildFunc.pm
+++ b/dataflowAPI/sail/ASTBuildFunc.pm
@@ -734,6 +734,7 @@ our @FN_DISPATCH = (
 
             my ($id_ast) = @$exps_ast;
             my $id_node = $self->do_exp( $id_ast, $curr_set );
+            my $id = $id_node->get_value();
 
             if ( grep( /^$id$/, @ParserConfig::supported_riscv_exts ) ) {
                 return ASTNode->new_bool_node(1);
--- a/dataflowAPI/sail/ParserConfig.pm
+++ b/dataflowAPI/sail/ParserConfig.pm
@@ -26,7 +26,7 @@ our @EXPORT_OK = qw(
 );
 
 # Instruction subsets supported currently
-our %supported_riscv_subsets = (
+our @supported_riscv_subsets = (
     "UTYPE",   "BTYPE",    "ITYPE",      "JAL",
     "JALR",    "SHIFTIOP", "RTYPE",      "LOAD",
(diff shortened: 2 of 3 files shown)


hainest:
My big concern here is attribution. From the README, it's clear this came from someone else's work. I would prefer not to import other people's code into Dyninst unless it's strictly necessary (e.g., the LLVM stuff for gfx908 kernels). If that's not possible or is a big pain, then we should definitely provide detailed information on where this code came from. Any code you wrote should get the standard Dyninst license notice.

I'd also recommend putting this in `external/sail` to make it clear it's not strictly part of Dyninst.

wxrdnx:
> My big concern here is attribution. From the README, it's clear this came from someone else's work. I would prefer not to import other people's code into Dyninst unless it's strictly necessary (e.g., the LLVM stuff for gfx908 kernels). If that's not possible or is a big pain, then we should definitely provide detailed information on where this code came from.
> 
> I'd also recommend putting this in `external/sail` to make it clear it's not strictly part of Dyninst.

The SAIL -> JSON part is a modification of the existing SAIL compiler. The original SAIL compiler is of course not my work; I extended it by adding a JSON backend. However, I believe the modified SAIL compiler is not imported into Dyninst. I believe I only provide a reference to the SAIL repository in the README, along with instructions on how to use the JSON backend.

The JSON -> C++ ROSE on the other hand is written in Perl by me, with no dependency on other code. These are all the `.pl` files and `.pm` files in the `sail` directory.

The RISC-V SAIL definition is also not written by me. They are formal semantic specifications of RISC-V instructions. However, I believe I do not import them directly; I just wrote some documentation on how to run the parser with these semantics.

> Any code you wrote should get the standard Dyninst license notice.

Got it. I'll add them.

wxrdnx:
@kupsch @hainest 
I moved the code to this repo:
<https://github.com/wxrdnx/dyninst-sail-to-rose>

New changes
- Comments removed in JSON
- Remove `Set::Scalar` and `File::Slurp` Perl module.
- Update README.md with more detailed commands
- Add Dyninst header to the Perl script

PR #2080, opened 2025-12-16
add vex encoding suppport for vpbrodcast family

bbiiggppiigg:


hainest (changes requested):
I forgot to ask for vmovd, but let's add some tests for this. I'd recommend putting them in a new test; something like `tests/integration/InstructionAPI/decoder/x86/sse.cpp`.

bbiiggppiigg:
will be replaced by  zydis

Issue #2016, opened 2025-09-27
Incorrectly handling of ELF RELA relocations in GOT


hainest:
The addend is inconsistently accounted for when rewriting statically-linked ELF files. This probably also applies to dynamically-linked ones because there are some [commented-out](https://github.com/dyninst/dyninst/blob/v13.0.0/symtabAPI/src/emitElf.C#L2078) offset adjustments involving the addend.

In [`emitElfStatic::createLinkMap`](https://github.com/dyninst/dyninst/blob/v13.0.0/symtabAPI/src/emitElfStatic.C#L612), GOT relocations are saved in `LinkMap::gotSymbols` with the addend explicitly set to 0. This is an unnecessary duplication of the information in `LinkMap::gotSymbolTable`, but that's a separate issue. When the GOT is being built (cf. `buildGOT`), the addend is **not** accounted for on [x86](https://github.com/dyninst/dyninst/blob/v13.0.0/symtabAPI/src/emitElfStatic-x86.C#L541), but **is** accounted for on [ppc64](https://github.com/dyninst/dyninst/blob/v13.0.0/symtabAPI/src/emitElfStatic-ppc64.C#L951). Note that the x86 code uses `gotSymbols` whereas the ppc64 code uses `gotSymbolTable`.

May solve #762, #1193

PR #2082, opened 2025-12-22
Add RISC-V Parse API and Dataflow API

wxrdnx:
This PR contains RISC-V Parse API and Dataflow API.
The RISC-V ROSE semantics are updated to v0.9.

### DIFF
--- a/dataflowAPI/rose/semantics/DispatcherRiscv64.C
+++ b/dataflowAPI/rose/semantics/DispatcherRiscv64.C
@@ -1485,7 +1485,7 @@ DispatcherRiscv64::RoundTowardsZero(const BaseSemantics::SValuePtr &expr) {
 }
 
 BaseSemantics::SValuePtr
-DispatcherRiscv64::CountTrailingZeroBits(const BaseSemantics::SValuePtr &expr) {
+DispatcherRiscv64::CountTrailingSignBits(const BaseSemantics::SValuePtr &expr) {
   size_t len = expr->get_width();
 
   for (int idx = 0; idx < len; idx++)


dataflowAPI/rose/semantics/BaseSemantics2.h
```diff
@@ -1712,6 +1712,11 @@ namespace rose {
                      *  number zero. The begin_bit and end_bit values must be valid for the width of @p a. */
                     virtual SValuePtr extract(const SValuePtr &a, uint64_t begin_bit, uint64_t end_bit) = 0;
 
+                    /** Extracts bits from a value.  The specified bits from begin (inclusive) through end (exclusive) are copied into
+                     *  the low-order bits of the return value (other bits in the return value are cleared). The least significant bit is
+                     *  number zero. The begin and end values must be valid for the width of @p a. */
+                    virtual SValuePtr extract(const SValuePtr &a, const SValuePtr &begin, const SValuePtr &end) = 0;
```
hainest, line 1718: This file comes from ROSE (it's now called [`RiscOperators.h`](https://github.com/rose-compiler/rose/blob/weekly/src/Rose/BinaryAnalysis/InstructionSemantics/BaseSemantics/RiscOperators.h)). Any modifications here will make it very difficult, if not impossible, to migrate toward using ROSE like a normal dependency. Is there any possibility that this functionality could be provided without modifying the ROSE classes?

wxrdnx: Ok this is kind of annoying. The reason why I introduce an overloaded version is that the original `extract` accepts begin and end as fixed integers like that:

```cpp
extract(const BaseSemantics::SValuePtr &a_, uint64_t begin, uint64_t end)
```

There's really no reason why `begin` and `end` should be a fixed integer. In fact, it can be rather arbitrary `SValuePtr` expressions as long as we're confident that `begin` and `end` lies in the valid range (which is `[0..64)` ).

---

My other solution will be adding a custom extract function to `DispatcherRISCV64`. I am not sure if that will work, though. I will give it a try.

external/rose/riscv64InstructionEnum.h
```diff
@@ -171,49 +171,6 @@ enum Riscv64InstructionKind {
...
-    rose_riscv64_op_c_sdsp,
-    rose_riscv64_op_c_slli,
-    rose_riscv64_op_c_srai,
-    rose_riscv64_op_c_srli,
-    rose_riscv64_op_c_sub,
-    rose_riscv64_op_c_subw,
-    rose_riscv64_op_c_sw,
-    rose_riscv64_op_c_swsp,
-    rose_riscv64_op_c_unimp,
-    rose_riscv64_op_c_xor,
```
hainest, line 216: Do these not have corresponding entries in `common/h/registers/riscv64_regs.h`?

instructionAPI/h/Instruction.h
```diff
@@ -162,6 +162,7 @@ namespace Dyninst { namespace InstructionAPI {
     DYNINST_EXPORT bool isVector() const { return getCategory() == c_VectorInsn; }
     DYNINST_EXPORT bool isGPUKernelExit() const { return getCategory() == c_GPUKernelExitInsn; }
     DYNINST_EXPORT bool isSoftwareException() const { return isGPUKernelExit() || getCategory() == c_SoftwareExceptionInsn; }
+    DYNINST_EXPORT bool isConditionalBranch() const { return isBranch() && checked_category(c_ConditionalInsn); }
```
hainest: Let's change this to `isConditional`. There are some instructions that are conditional, but not branches. For example, x86's `cmov`.

parseAPI/h/CFGModifier.h
```diff
@@ -118,7 +118,7 @@ class InsertedRegion : public CodeRegion {
    DYNINST_EXPORT Address offset() const { return base_; }
    DYNINST_EXPORT Address length() const { return size_; }
    DYNINST_EXPORT unsigned int getAddressWidth() const {
-      if (arch_ == Arch_ppc64 || arch_ == Arch_x86_64 || arch_ == Arch_aarch64) return 8;
+      if (arch_ == Arch_ppc64 || arch_ == Arch_x86_64 || arch_ == Arch_aarch64 || arch_ == Arch_riscv64) return 8;
```
hainest, line 121: After this PR is merged, code like this needs to be replaced with a call to [`getArchAddressWidth`](https://github.com/dyninst/dyninst/blob/master/common/h/Architecture.h#L55). There are undoubtedly other places that should use `getArchAddressWidth`.

parseAPI/h/CodeSource.h
```diff
@@ -128,6 +128,16 @@ class DYNINST_EXPORT CodeSource : public Dyninst::InstructionSource {
      */
     mutable std::map<Address, std::string> _linkage;
 
+    /*
+     * .rela.dyn external linkage table. Used in RISC-V
+     */
+    mutable std::map<Address, std::pair<std::string, Address> > _reladyn_linkage;
```
hainest: It looks like this is only used in `SymtabCodeSource::init_linkage()`. If that's the case, it doesn't need to be `mutable`.

parseAPI/h/CodeSource.h
```diff
@@ -183,6 +193,9 @@ class DYNINST_EXPORT CodeSource : public Dyninst::InstructionSource {
     virtual Address loadAddress() const { return 0; }
 
     std::map< Address, std::string > & linkage() const { return _linkage; }
+    std::map< Address, std::pair<std::string, Address> > & reladyn_linkage() const { return _reladyn_linkage; }
```
hainest:
```suggestion
    const std::map< Address, std::pair<std::string, Address> > & reladyn_linkage() const { return _reladyn_linkage; }
```

Returning a reference to a non-const object from a `const` member is a great way to get into trouble.

parseAPI/h/CodeSource.h
```diff
@@ -183,6 +193,9 @@ class DYNINST_EXPORT CodeSource : public Dyninst::InstructionSource {
...
+    std::map< Address, std::string> & symtab_linkage() const { return _symtab_linkage; }
```
hainest: Same as above.

parseAPI/src/BoundFactData.C
```diff
@@ -602,7 +602,8 @@ bool BoundFact::ConditionalJumpBound(Instruction insn, EdgeTypeEnum type) {
 		break;
 	    }
 	    case e_jae:
-	    case e_jnb_jae_j: {
```
hainest, line 605: I don't see any obvious issues with your changes here, but this function is fraught with problems.

1. Why is the x86 subtract instruction checked outside of the `EdgeTypeEnum` checks? (`if (pred.id == e_sub)`).
2. Should we add a check there for the riscv subtract instruction?

Items that should probably be turned into a github issue:

1. The instruction mnemonics checked here have to be the same as the ones checked by `IsConditionalJump` is `BoundFactCalculator.C`. This PR correctly uses the `Instruction` properties (`isBranch` et al.) which means changes in the Instruction class need to be reflected here. That's just a maintenance mess.
2. The `default` case for the `type == COND_NOT_TAKEN` branch asserts, but the one for `type == COND_TAKEN` doesn't.
3. What about subtract instructions for ppc and arm?

parseAPI/src/CodeSource.C
hainest, line 252: This function replicates information from `isAligned` in each `common/src/arch-ARCH.h` (except for riscv).

@hainest This should be moved into `Architecture.h`.

hainest (changes requested):

external/rose/riscv64InstructionEnum.h
wxrdnx: Sorry, I'm not sure what you mean.

These are compressed instructions, and they are converted to base instructions in instructionAPI, so ROSE don't need to deal with them.

While `common/h/registers/riscv64_regs.h` contains RISC-V registers.

instructionAPI/h/Instruction.h
wxrdnx: Done

parseAPI/h/CFGModifier.h
wxrdnx: Got it.

parseAPI/h/CodeSource.h
wxrdnx: `mutable` removed.

parseAPI/h/CodeSource.h
wxrdnx: Fixed

parseAPI/src/BoundFactData.C
hainest: Derp. My brain went on holiday. I meant the entryIDs.

Do you have thoughts on my first two points about the subtraction operations?

wxrdnx: > 1. Why is the x86 subtract instruction checked outside of the EdgeTypeEnum checks? (if (pred.id == e_sub)).

I'm not sure what the original rationale is.
My bet is that it's some sort of a hack that's x86/x64 specific.
There's a lot of these hacks in jump table analysis in Dyninst, and I'd prefer not to touch the logic as long as it works on x86/x64.

> 2. Should we add a check there for the riscv subtract instruction?

I don't think so. Works fine without it.

external/rose/riscv64InstructionEnum.h
wxrdnx: Any comments on this?

We already spent quite a lot of time discussing compressed instruction encoding.
Since compressed instructions are decoded back to their base instruction counterpart in InstructionAPI, they no longer require special handling in DataflowAPI.

hainest: Sorry. I agree with decompressing the instructions to regular ones and then using those- even if it means a rewritten binary will have a different set of instructions.

parseAPI/src/BoundFactData.C
hainest: > > 1. Why is the x86 subtract instruction checked outside of the EdgeTypeEnum checks? (if (pred.id == e_sub)).
> 
> I'm not sure what the original rationale is. My bet is that it's some sort of a hack that's x86/x64 specific. There's a lot of these hacks in jump table analysis in Dyninst, and I'd prefer not to touch the logic as long as it works on x86/x64.

It's undoubtedly a hack of some sort. Let's not perpetuate it for any other architecture. Naturally, table jump analysis really only works on x86...

> > 2. Should we add a check there for the riscv subtract instruction?
> 
> I don't think so. Works fine without it.

:+1:

hainest (commented):
Could you run `clang-format` on the `DispatchRiscv64.*` files? I know you copy/pasted from the existing ones, but they are so hard to read. Also, could you remove any commented-out code or add an additional comment around them explaining why they can't yet be implemented?

parseAPI/h/CodeSource.h
```diff
@@ -128,6 +128,16 @@ class DYNINST_EXPORT CodeSource : public Dyninst::InstructionSource {
...
 
+    /*
+     * .rela.dyn external linkage table. Used in RISC-V
+     */
+    std::map<Address, std::pair<std::string, Address> > _reladyn_linkage;
+
+    /*
+     * .symtab external linkage table. Used in RISC-V
+     */
+    mutable std::map<Address, std::string> _symtab_linkage;
```
hainest: Sorry, missed this one. This one doesn't need to be mutable, either.

parseAPI/src/IA_riscv64.C
```diff
@@ -0,0 +1,475 @@
...
+    return false;
+}
+
+bool IA_riscv64::isThunk() const
+{
+    return false;
+}
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
```
hainest: Is this needed?

parseAPI/src/IA_riscv64.C
```diff
@@ -0,0 +1,475 @@
...
+{
+    return false;
+}
+
+bool IA_riscv64::isReturn(Dyninst::ParseAPI::Function *func, Dyninst::ParseAPI::Block* /*currBlk*/) const
+{
+    Instruction insn = curInsn();
+    entryID eid = insn.getOperation().getID();
+
+    if (eid == riscv64_op_jalr) {
```
hainest, line 283: Is `jalr` always in `RISCV_GRP_RET` (cf `decode_categories`) so that this check agrees with the result of `Instruction::isBranch`?

I recall we chatted before about adding `Instruction::isNonABIReturn`, but now I'm thinking maybe that shouldn't be there since it's not a property that can be determined statically from just the instruction and its operands. By having it be an instruction property, a user may get the impression that it accurately reflects *how* the instruction is being used- even though we can't guarantee that without parsing the function that contains it. Having it be part of the instruction adapter makes sense since the parser has all the data needed to make that determination. There's a similar issue on PowerPC with similar uses of branch instructions as function returns that aren't really represented in instructionAPI yet (see #1929).

@kupsch Should we re-hash this?

parseAPI/src/IA_riscv64.C
```diff
@@ -0,0 +1,475 @@
...
+
+bool IA_riscv64::isMultiInsnJump(Address *target, Function *context, Block *currBlk) const {
+    Instruction insn = curInsn();
+
+    // Basic Instruction Matching
+    // This is used to prevent calling slicing all the time
+    // Slicing is not as efficient as instruction pattern matching
+    entryID curInsnID = insn.getOperation().getID();
+
+    if (curInsnID == riscv64_op_jalr) {
```
hainest, line 358: Same as above, but for `isMultiInsnBranch`.

parseAPI/src/IA_riscv64.h
```diff
@@ -0,0 +1,75 @@
...
+#include "dataflowAPI/h/Absloc.h"
+#include "dataflowAPI/h/SymEval.h"
+#include "dataflowAPI/h/slicing.h"
+#include "IA_IAPI.h"
+
+namespace Dyninst {
+namespace InsnAdapter {
+
+
+class IA_riscv64 : public IA_IAPI {
```
hainest, line 41: For a bit of modernization, could you remove the `virtual` and add `override` to the member functions?

parseAPI/src/IndirectAnalyzer.C
```diff
@@ -331,7 +331,7 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
         JumpTableReadVisitor jtrv(index, v, cs, block->region(), isZeroExtend, memoryReadSize);
         jumpTargetExpr->accept(&jtrv);
        if (jtrv.valid && cs->isCode(jtrv.targetAddress)) {
-            if (cs->getArch() == Arch_x86_64 && FindJunkInstruction(jtrv.targetAddress)) {
+            if ((cs->getArch() == Arch_x86_64 || cs->getArch() == Arch_riscv64) && FindJunkInstruction(jtrv.targetAddress)) {
```
hainest, line 334: The junk instruction detection is part of the Windows-only defensive mode. Does this need to be checked for riscv?

wxrdnx:
@hainest 
No problem

parseAPI/src/SymtabCodeSource.C
```diff
@@ -611,6 +611,26 @@ SymtabCodeSource::init_linkage()
...
         _linkage[(*fbtit).target_addr()] = (*fbtit).name(); 
     }
+    // Other relocation entries used by RISC-V
+    if (getArch() == Arch_riscv64) {
+        SymtabAPI::Region * relaDyn = NULL;
+        if (_symtab->findRegion(relaDyn, ".rela.dyn")) {
+            std::vector<SymtabAPI::relocationEntry> &relocs = relaDyn->getRelocations();
+            for (auto re_it = relocs.begin(); re_it != relocs.end(); ++re_it) {
+                SymtabAPI::relocationEntry &r = *re_it;
+                _reladyn_linkage[r.rel_addr()] = std::make_pair(r.name(), r.target_addr());
```
hainest: 1. Is the name '.rela.dyn' given by the riscv elf spec, or is it just what gcc uses?
2. Why store the region name? It doesn't appear to be used anywhere.
3. We should probably store the addend, even though it will almost always be zero. The elf emitter does a bad job of handling rela section rewriting (#2016), so it would be good to have that data here.

hainest:
> @hainest No problem

Thanks! I'll take a look at those files tomorrow.

parseAPI/src/IA_riscv64.C
wxrdnx: Looks like copy pasta thing.
Removed.

parseAPI/h/CodeSource.h
wxrdnx: Removed `mutable`

parseAPI/src/IA_riscv64.h
wxrdnx: Sure

parseAPI/src/IndirectAnalyzer.C
wxrdnx: Yes. I'm not sure what Windows-only defensive mode means, but I'm fairly certain this is a general defensive mechanism to protect the case when jump table index range is predicted wider than the actual index range. I've previously mentioned that there's a lot of hacks in jump table index analysis, and this is one of them.

I specifically recall running into a particular kind of jump table on both x86 and RISC-V where this line of code need to be called to make the jump table index analysis correct.

In fact, if this happens on x86 and RISC-V, I believe this also happens on PowerPC and ARM. Given that, what do you think about removing the architecture check entirely?

```diff
- if (cs->getArch() == Arch_x86_64 && FindJunkInstruction(jtrv.targetAddress)) {
+ if (FindJunkInstruction(jtrv.targetAddress))
```

parseAPI/src/SymtabCodeSource.C
wxrdnx: Ok, so for context this is a G++-specific implementation hellhole that need to be addressed carefully.

Normally, GCC generates PLT calls in the same way as on other architectures. However, in some cases where the address of a specific function is known at link time (e.g., a member function of a class, since a member function is defined in the same ELF), RISC-V G++ _optimizes away_ the PLT call. Instead, it loads the function address directly from `.rela.dyn` section and calls it directly. This appears to be an optimization to avoid PLT call overhead. This happens A LOT if you compile a C++ binary using RISC-V version of G++.

Therefore, to correctly obtain the address of the PLT call, we must parse and record relevant information from `.rela.dyn`.

So for the question:

> 1. Is the name '.rela.dyn' given by the riscv elf spec, or is it just what gcc uses?

This is ELF specific across architectures. As mentioned above, RISC-V need to use the information from this section. For other architectures, they are unused.

> 2. Why store the region name? It doesn't appear to be used anywhere.

For debugging purpose.

> 3. We should probably store the addend, even though it will almost always be zero. The elf emitter does a bad job of handling rela section rewriting (https://github.com/dyninst/dyninst/issues/2016), so it would be good to have that data here.

Agree, I think I'll just replace the region name with addend.

parseAPI/src/IA_riscv64.C
wxrdnx: I think this is the same problem, but the PR confuses me.

For PR #1929 , the patch looks like removing the instruction return analysis in ParseAPI, adding static analysis on InstructionAPI, and using the fallthrough property to determine whether the instruction is a return or not _statically_.

Also, see below

parseAPI/src/IA_riscv64.C
wxrdnx: I sort of agree that `Instruction::isNonABICall`, `Instruction::isNonABIReturn`,`Instruction::isMultiInsnCall`, and `Instruction::isMultiInsnJump` can reside in instruction adapter.

The main issue is that `IA_IAPI` calls the `Instruction::isBranch` everywhere in the code, so if we move the logic into the instruction adapter, we'll need to modify some core implementation in ParseAPI:

- Change every `Instruction::isBranch` to `IA_<arch>::isBranch`.
- Change every `Instruction::isCall` to `IA_<arch>::isCall`
- Change every `Instruction::isReturn` to `IA_<arch>::isReturn`

Here, `IA_<arch>::isBranch` simply returns `Insn::isBranch` if the arch has no multi-instruction jumps/calls or non-abi-compliant jumps/calls (e.g. x86).

We'll probably need another PR for this.

Does this proposal looks good to you?

parseAPI/src/IA_riscv64.C
hainest: > the patch looks like removing the instruction return analysis in ParseAPI

The analysis there was for ancient power (probably Power2 or something) that no longer applies to Power8/9/10, so it got ripped out.

> adding static analysis on InstructionAPI, and using the fallthrough property to determine whether the instruction is a return or not statically.

Correct. This is what I'd suggest we do for `jalr` since we just need to check if `zero` is `x0`, the immediate is zero, and `reg` is `ra`. This ensures that `Instruction::isReturn` is correctly populated when not using the InstructionAdapter (e.g., when decoding a buffer like in the tests).

We can't do that for the non-abi case since it needs more context. Hence, I'd recommend we don't make it a property of the Instruction.

This function could then be implemented as

```c++
if (insn.isReturn()) {
  return true;
}

// Do the non-abi case
```

parseAPI/src/IndirectAnalyzer.C
hainest: Let's chat about this. @kupsch do you have any time this week?

parseAPI/src/SymtabCodeSource.C
hainest: > Ok, so for context this is a G++-specific implementation hellhole that need to be addressed carefully.
> 
> Normally, GCC generates PLT calls in the same way as on other architectures. However, in some cases where the address of a specific function is known at link time (e.g., a member function of a class, since a member function is defined in the same ELF), RISC-V G++ _optimizes away_ the PLT call. Instead, it loads the function address directly from `.rela.dyn` section and calls it directly. This appears to be an optimization to avoid PLT call overhead. This happens A LOT if you compile a C++ binary using RISC-V version of G++.
> 
> Therefore, to correctly obtain the address of the PLT call, we must parse and record relevant information from `.rela.dyn`.

Oh, interesting. This seems like the kind of thing LTO would do to improve its IPO or do devirtualization. This might be something we should consider for other architectures. But that's for another day (probably).

> 
> So for the question:
> 
> > 1. Is the name '.rela.dyn' given by the riscv elf spec, or is it just what gcc uses?
> 
> This is ELF specific across architectures. As mentioned above, RISC-V need to use the information from this section. For other architectures, they are unused.

It's the name I've seen on x86 and power, but I was curious if it was a convention or a requirement. We really should analyze this section for all architectures.

> 
> > 2. Why store the region name? It doesn't appear to be used anywhere.
> 
> For debugging purpose.
> 
> > 3. We should probably store the addend, even though it will almost always be zero. The elf emitter does a bad job of handling rela section rewriting ([Incorrectly handling of ELF RELA relocations in GOT #2016](https://github.com/dyninst/dyninst/issues/2016)), so it would be good to have that data here.
> 
> Agree, I think I'll just replace the region name with addend.

Sounds good.

dataflowAPI/rose/semantics/DispatcherRiscv64.h
```diff
@@ -0,0 +1,292 @@
...
+           const RegisterDictionary *regs = NULL) {
+    return DispatcherRiscv64Ptr(new DispatcherRiscv64(ops, addrWidth, regs));
+  }
+
+  /** Virtual constructor. */
+  virtual BaseSemantics::DispatcherPtr
+  create(const BaseSemantics::RiscOperatorsPtr &ops, size_t addrWidth = 0,
+         const RegisterDictionary *regs = NULL) const {
+    if (0 == addrWidth) {
+      addrWidth = 64;
```
hainest:
```suggestion
      addrWidth = Dyninst::getArchAddressWidth(Dyninst::Arch_riscv64);
```

parseAPI/src/IA_riscv64.C
hainest: Let's chat about this, as well.

hainest:
Bah, sorry! I forgot to check if #2099 would introduce conflicts here. Fortunately, it's just some whitespace in SymEval.C.

parseAPI/src/IndirectAnalyzer.C
hainest: This is fine as-is. We'll clean up `FindJunkInstruction` later.

instructionAPI/h/Instruction.h
```diff
@@ -162,11 +162,11 @@ namespace Dyninst { namespace InstructionAPI {
     DYNINST_EXPORT bool isVector() const { return getCategory() == c_VectorInsn; }
     DYNINST_EXPORT bool isGPUKernelExit() const { return getCategory() == c_GPUKernelExitInsn; }
     DYNINST_EXPORT bool isSoftwareException() const { return isGPUKernelExit() || getCategory() == c_SoftwareExceptionInsn; }
+    DYNINST_EXPORT bool isConditional() const { return checked_category(c_ConditionalInsn); }
 
-    DYNINST_EXPORT bool isMultiInsnCall() const { return isCall() && getOperation().isMultiInsnCall; }
-    DYNINST_EXPORT bool isMultiInsnBranch() const { return isBranch() && getOperation().isMultiInsnBranch; }
-    DYNINST_EXPORT bool isNonABICall() const { return isCall() && getOperation().isNonABICall; }
-    DYNINST_EXPORT bool isNonABIReturn() const { return isReturn() && getOperation().isNonABIReturn; }
+    DYNINST_EXPORT void forceSetReturn() const;
```
hainest:
```suggestion
    DYNINST_EXPORT void forceReturn();
```

instructionAPI/h/Instruction.h
```diff
@@ -162,11 +162,11 @@ namespace Dyninst { namespace InstructionAPI {
...
+
+    DYNINST_EXPORT void forceSetCall() const;
```
hainest:
```suggestion
    DYNINST_EXPORT void forceCall();
```

parseAPI/src/Block.C
```diff
@@ -274,6 +274,15 @@ Block::getInsns(Insns &insns) const {
     insns[off] = insn;
     off += insn.size();
   }
+  Instruction lastInsn = insns[last()];
```
hainest:
```suggestion
  Instruction& lastInsn = insns[last()];
```

It needs to be a reference. Otherwise, it's a copy.

parseAPI/src/IA_IAPI.C
```diff
@@ -561,6 +568,22 @@ void IA_IAPI::getNewEdges(std::vector<std::pair< Address, EdgeTypeEnum> >& outEd
             }
         }
 
+        if (isMultiInsnJump(&target, context, currBlk)) {
+            CodeSource *cs = currBlk->obj()->cs();
+            std::map<Dyninst::Address, std::string> symtab_entries = cs->symtab_linkage();
```
hainest:
```suggestion
            const std::map<Dyninst::Address, std::string>& symtab_entries = cs->symtab_linkage();
```

hainest:
The CI jobs are timing out, so I cancelled the run. Were you able to run the tests locally?

wxrdnx:
> The CI jobs are timing out, so I cancelled the run. Were you able to run the tests locally?

@hainest 
Yes.

parseAPI/src/BoundFactCalculator.C
```diff
@@ -351,6 +354,28 @@ void BoundFactsCalculator::CalcTransferFunction(Node::Ptr curNode, BoundFact *ne
 	newFact->SetPredicate(node->assign(), se.ExpandAssignment(node->assign()) );
 	return;
     }
+
+    if (node->assign()->block()->obj()->cs()->getArch() == Arch_riscv64 &&
+            node->assign() &&
```
hainest:
```suggestion
```

This check is useless since `node->assign()` is used unchecked in the previous expression. I'd say swap them or remove the `node->assign()` check.

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+ * the / operator) to match how integer division is normally performed in C/C++.
+ * Of course, this would be invalid for floating point division but that is not
+ * currently supported. */
+BaseSemantics::SValuePtr
+DispatcherRiscv64::RoundTowardsZero(const BaseSemantics::SValuePtr &expr) {
+  return expr;
+}
+
+BaseSemantics::SValuePtr
+DispatcherRiscv64::CountTrailingSignBits(const BaseSemantics::SValuePtr &expr) {
```
kupsch: The function name doesn't make sense. I think you mean zero instead of sign:
```suggestion
DispatcherRiscv64::CountTrailingZeroBits(const BaseSemantics::SValuePtr &expr) {
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+  //   |a| = (a ^ sign) - sign
+  //   result = (mul(|a|, b) ^ sign) - sign
+
+  BaseSemantics::SValuePtr sign = operators->shiftRightArithmetic(
+      a, operators->number_(XLENBITS, XLENBITS - 1));
+  BaseSemantics::SValuePtr absA =
+      operators->subtract(operators->xor_(a, sign), sign);
+  BaseSemantics::SValuePtr result = operators->subtract(
+      operators->xor_(operators->unsignedMultiply(absA, b), sign), sign);
+  return result;
```
kupsch, line 1632: Instead of using the magic number XLENBITS, let's get the value from the argument `a`.

The multiply returns a value whose width is the sum of the argument widths, so the value needs to be limited to half the value; otherwise the two's complement does not work correctly.

```suggestion
  // Symbolically, the result of signed(a) * unsigned(b) is
  //   unsigned_multiply(|a|, b) with proper signednes added
  // Thus, the following sequence is equivalent to signed(a) * unsigned(b):
  //   sign = a >> (XLEN-1)                    // -1 (all ones) or 0 (all zeros)
  //   |a| = (a ^ sign) - sign                 // two's complement if negative
  //   result = (umul(|a|, b) ^ sign) - sign   // two's complement umul if a was negative

  auto numBits = a->get_width();
  auto sign = operators->shiftRightArithmetic(
      a, operators->number_(numBits, numBits - 1));
  // two's complement if a is negative
  auto absA = operators->subtract(operators->xor_(a, sign), sign);
  // lower numBits of unsigned multipy
  auto m = operators->extract(operators->unsignedMultiply(absA, b), 0, numBits);
  // two's complement if a is negative
  auto result = operators->subtract(operators->xor_(m, sign), sign);

  return result;
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+    count = operators->add(count, bit);
+  }
+
+  return count;
+}
+
+BaseSemantics::SValuePtr
+DispatcherRiscv64::Brev8(const BaseSemantics::SValuePtr &expr) {
+  size_t len = expr->get_width();
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
  BaseSemantics::SValuePtr result = operators->number_(len, 0);
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+      return operators->number_(expr->get_width(), len - 1 - idx);
+
+  return operators->number_(expr->get_width(), len);
+}
+
+BaseSemantics::SValuePtr
+DispatcherRiscv64::CarrylessMultiply(const BaseSemantics::SValuePtr &a,
+                                     const BaseSemantics::SValuePtr &b) {
+  size_t len = a->get_width();
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
  BaseSemantics::SValuePtr result = operators->number_(len, 0);
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+  }
+
+  return result;
+}
+
+BaseSemantics::SValuePtr
+DispatcherRiscv64::Rev8(const BaseSemantics::SValuePtr &expr) {
+  size_t len = expr->get_width();
+  size_t num_bytes = len / 8;
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
  BaseSemantics::SValuePtr result = operators->number_(len, 0);
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+                                            const BaseSemantics::SValuePtr &b) {
+  size_t len = a->get_width();
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
+
+  // iterate from most significant bit down to least
+  for (int i = len - 1; i >= 0; i--) {
+    BaseSemantics::SValuePtr b_bit = operators->extract(b, i, i + 1);
+    BaseSemantics::SValuePtr term = operators->ite(
+        b_bit, operators->shiftLeft(a, operators->number_(len, i)),
+        operators->number_(0, len));
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
        operators->number_(len, 0));
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
+
+  for (size_t i = 0; i < len; i++) {
+    // extract bit i from b
+    BaseSemantics::SValuePtr b_bit = operators->extract(b, i, i + 1);
+
+    // if b_bit == 1, XOR (a << i) into result; else XOR 0
+    BaseSemantics::SValuePtr term = operators->ite(
+        b_bit, operators->shiftLeft(a, operators->number_(len, i)),
+        operators->number_(0, len));
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
        operators->number_(len, 0));
```

dataflowAPI/rose/semantics/DispatcherRiscv64.C
```diff
@@ -0,0 +1,1634 @@
...
+  }
+
+  return result;
+}
+
+BaseSemantics::SValuePtr
+DispatcherRiscv64::CarrylessMultiplyReverse(const BaseSemantics::SValuePtr &a,
+                                            const BaseSemantics::SValuePtr &b) {
+  size_t len = a->get_width();
+  BaseSemantics::SValuePtr result = operators->number_(0, len);
```
kupsch: I think you want a len bit number with the value 0, not a 0 bit number:


```suggestion
  BaseSemantics::SValuePtr result = operators->number_(len, 0);
```

wxrdnx:
@kupsch 
I've finished the cleanup.
Let me know if I still need to fix anything.
Thanks.

hainest:
@wxrdnx It looks like there are a couple of rebase errors. The `Instruction::isConditional` and the declaration for `DispatcherRiscv64::CountTrailingZeroBits` are missing.

wxrdnx:
@hainest 
I'll fix them later (probably after this Wednesday)

hainest:
@wxrdnx Feel free to merge when you're ready.

hainest:
@wxrdnx @kupsch @bbiiggppiigg I need some of these changes for the work I'm doing with the decoders. Does anyone object to merging?

wxrdnx:
> @wxrdnx Feel free to merge when you're ready.

Alright, merged

PR #2086, opened 2025-12-29
Add RISC-V Codegen

wxrdnx:
This PR contains RISC-V Codegen.
WIP:
- [x] Migrate codegen buffer to raw pointer
- [x] Improve RISC-V codegen constant name
- [x] For `codegen-riscv64.C`: return generated instruction size instead of boolean.
- [x] handle all cases where scratch register is null

To discuss:

- [ ] Should init_array and fini_array be its own PR?

### DIFF
--- a/common/src/arch-riscv64.C
+++ b/common/src/arch-riscv64.C
@@ -304,7 +304,7 @@ unsigned instruction::getCondBranchOp() const {
 }
 
 unsigned instruction::getCondBranchReg1() const {
-  if (isCondBranch()) {
+  if (!isCondBranch()) {
     return false;
   }
   if (isCompressed()) {
@@ -316,7 +316,7 @@ unsigned instruction::getCondBranchReg1() const {
 }
 
 unsigned instruction::getCondBranchReg2() const {
-  if (isCondBranch()) {
+  if (!isCondBranch()) {
     return false;
   }
   if (isCompressed()) {
@@ -360,14 +360,14 @@ bool instruction::isAuipc() const {
 }
 
 Dyninst::Address instruction::getAuipcOffset() const {
-  if (isAuipc()) {
+  if (!isAuipc()) {
     return -1;
   }
   return static_cast<Dyninst::Address>(insn_.raw & AUIPC_IMM_MASK);
 }
 
 unsigned instruction::getAuipcReg() const {
-  if (isAuipc()) {
+  if (!isAuipc()) {
     return -1;
   }
   return ((insn_.raw & AUIPC_REG_MASK) >> AUIPC_REG_SHIFT);


common/src/]
```diff
@@ -0,0 +1,524 @@
+/*
```
Comment (hainest) [inline common/src/]]: I think this filename is a typo.

common/src/]
```diff
@@ -0,0 +1,524 @@
...
+// The bits in the register is 64 bits
+constexpr int RISCV_REG_SIZE = 8;
+// The immediates in instructions varies.
+// The maximum is 21 bits, which can be stored in a 32 bit integer
+constexpr int RISCV_IMM_SIZE = 4;
+
+// Raw register encoding used in RISC-V instruction encoding
+// These values identify registers when decoding instructions
+// They are also used to emit register encoding during codegen
+constexpr int32_t GPR_X0     = 0;
```
Comment (hainest) [inline common/src/]]: I know all the other architectures do it this way, but I would suggest using a scoped enum for all these constants. Not a requirement.

common/src/]
```diff
@@ -0,0 +1,524 @@
...
+constexpr int64_t JTYPE_IMM_MAX = 0x1000000; // 21 bits signed (not 20 because imm is shifted 1 bits left)
+constexpr int64_t JTYPE_IMM_MIN = -0x1000000; // 21 bits signed (not 20 because imm is shifted 1 bits left)
+constexpr int64_t STYPE_IMM_MAX  = 0xfff;
+constexpr int64_t STYPE_IMM_MIN  = 0xfff;
+constexpr int64_t UTYPE_IMM_MAX  = 0x80000000LL;  // 32 bits signed
+constexpr int64_t UTYPE_IMM_MIN  = -0x80000000LL; // 32 bits signed
+
+// RISC-V immediates in jump/branch instructions are scrambled:
+// The following arrays store the corresponding indices to reorder the immediates back
+const std::vector<int> JAL_REORDER   = {12, 13, 14, 15, 16, 17, 18, 19, 11, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 20};
```
Comment (hainest) [inline common/src/]]: To avoid allocations, you could use a `std::initializer_list` for these. You can get that by default by using just the braced init list `const auto JAL_REORDER = {...};`. Sadly, they can't be `constexpr` until we switch to c++17 (that was added in c++14). :(

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,359 @@
...
+        return true;
+    } else if (isUncondBranch()) {
+        return true;
+    }
+   
+    return 0;
+}
+
+bool instruction::isCall() const {
+    assert(isUncondBranch());
```
hainest: Asserting here seems a little strict. I would recommend changing the assert to a check with return false. The same with the other property-checking functions.

dyninstAPI/src/Relocation/Widgets/CFWidget-riscv64.C
```diff
@@ -0,0 +1,227 @@
...
+    insnCodeGen::generate(gen, mod_insn);
+    buffer.addPIC(gen, tracker(trace));
+    return true;
+}
+
+bool CFPatch::apply(codeGen &gen, CodeBuffer *buf) {
+    if (isPLT(gen)) {
+        relocation_cerr << "\t\t\t isPLT..." << endl;
+        if (!applyPLT(gen, buf)) {
+            relocation_cerr << "PLT special case handling in PPC64" << endl;
```
hainest: copy/pasta.

dyninstAPI/src/Relocation/Springboard.C
```diff
@@ -242,6 +242,7 @@ SpringboardBuilder::generateSpringboard(std::list<codeGen> &springboards,
 
    // Arbitrarily select the first function containing this springboard, since only one can win. 
    generateBranch(r.from, r.destinations.begin()->second, tmpGen);
+
```
hainest: Let's remove this incidental whitespace change.

hainest (commented):
I've not looked over everything, but I have a few items. Could you add some codegen tests like in `tests/unit/dyninstAPI/emitter/x86_64.cpp`?

common/src/]
Comment (wxrdnx) [reply common/src/]]: Oh, I accidentally commit junk files.
Removed

common/src/arch-riscv64.C
wxrdnx: Okay, fixed.

dyninstAPI/src/Relocation/Widgets/CFWidget-riscv64.C
wxrdnx: Fixed

dyninstAPI/src/Relocation/Springboard.C
wxrdnx: Removed.

wxrdnx:
> I've not looked over everything, but I have a few items. Could you add some codegen tests like in tests/unit/dyninstAPI/emitter/x86_64.cpp?

Will do

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+  int32_t result = insn_.raw & ATOMIC_OP_MASK;
+  return result == ATOMIC_OP_SC;
+}
+
+bool instruction::isAuipc() const {
+  return (insn_.raw & AUIPC_INSN_MASK) == AUIPC_INSN;
+}
+
+Dyninst::Address instruction::getAuipcOffset() const {
+  if (isAuipc()) {
```
wsxarcher: Should be negated (?)
Also below

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+  if (isAtomicMemOp()) {
+    return true;
+  }
+  int32_t result = insn_.raw & ATOMIC_OP_MASK;
+  return result == ATOMIC_OP_LR;
+}
+
+bool instruction::isAtomicStore() const {
+  if (isAtomicMemOp()) {
+    return true;
```
wsxarcher, line 352: isAtomicStore seems to be true also when it's a load

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+  if (isCompressed()) {
+    return CBRANCH_REG1_ADD +
+           ((insn_.raw & CBRANCH_REG1_MASK) >> CBRANCH_REG1_SHIFT);
+  } else {
+    return ((insn_.raw & BRANCH_REG1_MASK) >> BRANCH_REG1_SHIFT);
+  }
+}
+
+unsigned instruction::getCondBranchReg2() const {
+  if (isCondBranch()) {
```
wsxarcher: Should this be negated?

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+    } else {
+      return BRANCH_COND_NE;
+    }
+  } else {
+    return ((insn_.raw & BRANCH_COND_MASK) >> BRANCH_COND_SHIFT);
+  }
+}
+
+unsigned instruction::getCondBranchReg1() const {
+  if (isCondBranch()) {
```
wsxarcher: Should this be negated?

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+    return false;
+  }
+  if (isCompressed()) {
+    return ((insn_.raw & CJR_REG_MASK) >> CJR_REG_SHIFT);
+  } else {
+    return ((insn_.raw & JALR_REG_MASK) >> JALR_REG_SHIFT);
+  }
+}
+
+Dyninst::Address instruction::getBranchOffset() const {
```
wsxarcher, line 237: The offset on branches is signed, so if this function extract a negative number it will be interpreted incorrectly as a 64bit number.

For example c.jal -2 immediate would be 111111111110 which would currently be seen as 4094 but it should be sign extended to to be -2 also for the 64 bit return value, so for this one signExtend should be called with pos 12 (it start from 1st not 0) if I'm not mistaken (the msb of this immediate)

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+
+bool instruction::isAuipc() const {
+  return (insn_.raw & AUIPC_INSN_MASK) == AUIPC_INSN;
+}
+
+Dyninst::Address instruction::getAuipcOffset() const {
+  if (isAuipc()) {
+    return -1;
+  }
+  return static_cast<Dyninst::Address>(insn_.raw & AUIPC_IMM_MASK);
```
wsxarcher, line 366: Also getAuipcOffset should return a signed integer not an address
Example auipc x0, -4 should work correctly

common/src/arch-riscv64.C
wxrdnx: Yes, should be negated. Thanks for catching that!
Previously, it was `assert(isCondBranch())`, and I was asked to replace it with if statements, but somehow the `!` got slipped through. My fault.
Fixed.

common/src/arch-riscv64.C
wxrdnx: Same.
Fixed.

common/src/arch-riscv64.C
wxrdnx: Same. Fixed.

common/src/arch-riscv64.C
wxrdnx: My reason for using `Address` is consistency. If you take a look at the function arguments of `emit-*`, you can see that the offsets are pretty much of type `Address`. That's not my decision! Had the argument be `Offset`, I would 100% use the type `Offset` as opposed to `Address`.

I don't mind changing it to a signed type like `Offset`, though. If you decided that changing the type to `Offset` makes more sense, please let me know. Just be careful that this will require type conversion to be handled correctly elsewhere.

common/src/arch-riscv64.C
wxrdnx: Maybe our interpretation is different

I wrote these functions to differentiate between `lr`, `sc`, and `amo` functions:
- `isAtomicMemOp` -> returns `amo` instructions
- `isAtomicLoad` -> return `lr` instructions
- `isAtomicStore` -> return `sc` instructions

`amo` insns  are load + store, `lr` insns are load, `sc` inns are store.

Let me know if this needs to be changed.

common/src/arch-riscv64.C
wxrdnx: Same.
See my earlier reply.

common/src/arch-riscv64.C
wsxarcher: In my opinion Offest would definitely make more sense as branches are indeed signed.

But the main point is loss of information, even if Address is returned, the value is wrong because the sign bit is not extended, so in case of a negative branch the current value does not have any sense at all, and if anyone is using this value they would need to know from which branch instruction is coming to know which sign bit position to extend. 

If you want to keep this function unchanged and consistent like you mentioned and still be usable by others we should have an additional one called `getBranchOffsetSignExtended` that does the calculation for you:

Example (Not tested)

```
Offset instruction::getBranchOffsetSignExtended() const {
    if (!isBranchOffset()) {
       return -1;
    }
    int signBit  = 0;
    if (isUncondBranch() && (isCompressed() || isBranchReg())
        signBit = 11;
    else if (isUncondBranch())
        signBit = 20;
    else if (isCondBranch() && isCompressed())
        signBit = 8;
    else if (isCondBranch())
        signBit = 12;
    return (Offset) signExtend(getBranchOffset(), signBit+1);
}
```

common/src/arch-riscv64.C
wsxarcher: In this case sign extension happens automatically from int to long because the auipc immediate is 32 bit so there is no sign to extend.
It would definitely be better to return an Offset but it's not necessary as a cast to it would suffice, so you can ignore this one.

common/src/arch-riscv64.h
```diff
@@ -130,11 +145,320 @@ constexpr int32_t GPR_T3     = 28;
...
+  // `flushInsnBuffer` to flush the instruction buffer into the 2-byte code
+  // buffer `code_buff` short-by-short.
+
+  instructUnion insn_;
+
+public:
+  instruction() : insn_() {}
+  instruction(rvMaxInsn_t raw) { insn_.raw = raw; }
+  instruction(const void *ptr) { insn_ = *((const instructUnion *)ptr); }
+  instruction(const void *ptr, bool) : instruction(ptr) {}
```
wsxarcher, line 404: is the bool to indicates if you want to parse a compressed instruction? if that's the case it should ignore the upper parts

```c
instruction(const void *ptr, bool compressed) {
    insn_ = *((const instructUnion *)ptr); 
    if (compressed) insn_ &= 0xFFFF;
}
```

common/src/arch-riscv64.C
```diff
@@ -0,0 +1,376 @@
...
+bool instruction::isCall() const {
+  if (!isUncondBranch()) {
+    return false;
+  }
+  if (isCompressed()) {
+    int32_t result = insn_.raw & CJUMP_INSN_MASK;
+    if (result == CJAL_INSN) {
+      return true;
+    }
+    result = insn_.raw & CJR_INSN;
```
wsxarcher, line 199: This should be CJR_INSN_MASK.
Also the mask is wrong because it currently fails thinking c.mv are a call so CJR_INSN_MASK needs to be changed to 0xf07f to be sure rs2 is 0

hainest:
@wxrdnx Would you mind if I resolve the branch conflicts here? I don't want to mess up your local branch, if you're doing active work.

hainest:
In `doNotOverflow` (what is now `can_encode_directly`, #2214), always returns `true` which I don't think is correct because there are several instructions that take an immediate, but they require 6, 12 or 32 bits. From the manual,

```
2.4.1. Integer Register-Immediate Instructions

ADDI adds the sign-extended 12-bit immediate to register rs1.

SLTI (set less than immediate) places the value 1 in register rd if register rs1 is less than the sign-
extended immediate when both are treated as signed numbers,...

ANDI, ORI, XORI are logical operations that perform bitwise AND, OR, and XOR on register rs1 and the
sign-extended 12-bit immediate...
```

There are more in 2.4.2, 2.4.3, and 27.5.2.

bigtrak:
Tim, why can't this be done with architecture specific dispatch
Instead of having all the archs having  embedded details in a random file?

Even if it is a static in an architecture class... and also as you pointed
Out, it depends on  a particular instruction being used, which the
Per-arch code can isolate and correct for each arch or sub-arch.

Also, for example, the amd gpu "crash because this doesn't work" idiom is
Removed with this change.

Bolo
________________________________

hainest:
I think you meant to respond to #2214. This PR is just for RISC-V.

> Tim, why can't this be done with architecture specific dispatch

It can and likely will get merged into something resembling what are now the Emitter classes. Right now, I'm just trying to pull things apart to find out what the actual interfaces are in order to make progress toward architecture-independent codegen.

> Instead of having all the archs having embedded details in a random file?

`operandAST` is the only place this is used, so I put it there.

> Also, for example, the amd gpu "crash because this doesn't work" idiom is Removed with this change.

I mentioned that in the commit notes.

PR #2087, opened 2026-01-01
Fix removal of bad jump table edges

### LINKED ISSUE #2081: Failing to write instrumented shared library
**Intention**
Instrument library functions using BPatch_callBefore and BPatch_callAfter

**Describe the bug**
Segmentation fault when writing with writeFile, after inserting the snippets.

**To Reproduce**
Made  a copy of the examples here:
https://github.com/jordialcaraz/examples/tree/master

Clone the repository
Build the examples
Go to codeCoverage folder:
And execute the modified example:
`./code_coverage -s lib_to_instrument/libtorch_python.so libtorch_python.so`

**Expected behavior**
I would expect if to build without crashing.

System: x86_64

Installed Dyninst with `spack install dyninst`

```
  spack find -v -d dyninst
  -- linux-ubuntu22.04-cascadelake / %c,cxx=gcc@11.4.0 ------------
  dyninst@13.0.0~ipo+openmp~stat_dysect~static build_system=cmake build_type=Release generator=make
      boost@1.89.0+atomic~charconv+chrono~clanglibcpp~cobalt+container~context~contract~conversion~coroutine+date_time~debug+exception~fiber+filesystem~graph~graph_parallel~icu~iostreams~json~locale~log~math~mpi~mqtt5+multithreaded~nowide~numpy~pic~program_options~python+random~regex~serialization+shared~signals2~singlethreaded~stacktrace+system~taggedlayout~test+thread+timer~type_erasure~url~versionedlayout~wave build_system=generic cxxstd=11 patches:=a440f96 visibility=hidden
      cmake@3.31.9~doc+ncurses+ownlibs~qtgui build_system=generic build_type=Release
          curl@8.15.0~gssapi~ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs:=shared,static tls:=openssl
              nghttp2@1.67.1 build_system=autotools
              openssl@3.6.0~docs+shared build_system=generic certs=mozilla
                  ca-certificates-mozilla@2025-08-12 build_system=generic
              perl@5.42.0+cpanm+opcode+open+shared+threads build_system=generic
                  berkeley-db@18.1.40+cxx~docs+stl build_system=autotools patches:=26090f4,b231fcc
                  gdbm@1.25 build_system=autotools
                      readline@8.3 build_system=autotools patches:=21f0a03
          ncurses@6.5-20250705~symlinks+termlib abi=none build_system=autotools patches:=7a351bc
          zlib-ng@2.2.4+compat+new_strategies+opt+pic+shared build_system=autotools
      compiler-wrapper@1.0 build_system=generic
      elfutils@0.193~debuginfod+exeprefix+nls build_system=autotools
          bzip2@1.0.8~debug~pic+shared build_system=generic
              diffutils@3.12 build_system=autotools
          gettext@0.23.1+bzip2+curses+git~libunistring+libxml2+pic+shared+tar+xz build_system=autotools
              libxml2@2.13.5~http+pic~python+shared build_system=autotools
              tar@1.35 build_system=autotools zip=pigz
                  pigz@2.8 build_system=makefile
          libiconv@1.18 build_system=autotools libs:=shared,static
          m4@1.4.20+sigsegv build_system=autotools
              libsigsegv@2.14 build_system=autotools
          pkgconf@2.5.1 build_system=autotools
          xz@5.6.3~pic build_system=autotools libs:=shared,static
          zstd@1.5.7+programs build_system=makefile compression:=none libs:=shared,static
      gcc@11.4.0~binutils+bootstrap~graphite~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++,fortran'
      gcc-runtime@11.4.0 build_system=generic
      glibc@2.35 build_system=autotools
      gmake@4.4.1~guile build_system=generic
      intel-tbb@2022.3.0~ipo+shared+tm build_system=cmake build_type=Release cxxstd=default generator=make
          hwloc@2.12.2~cairo~cuda~gl~level_zero~libudev+libxml2~nvml~opencl+pci~rocm build_system=autotools libs:=shared,static
              libpciaccess@0.17 build_system=autotools
                  util-macros@1.20.1 build_system=autotools
      libiberty@2.41+pic build_system=autotools
  
```

**Additional context**
We want to instrument shared libraries that are installed with python -m pip install

Seems to only throw the segmentation fault due to line 218 of code_coverage.C
`
BPatchSnippetHandle* handle_b = mutateeAddressSpace->insertSnippet(exitTrace, *funcExit, BPatch_callAfter, BPatch_lastSnippet);`

The error is not shown if that line is removed, but I would like to instrument both.

bbiiggppiigg:
When a basic block is pointed by more than one jump table entry, Dyninst considers one of the jump table entry overrun and trim the edge from the jump table, and transitively remove the outgoing edge of the shared block. This behavior causes inconsistent view of CFG between patchAPI and parseAPI, where a relocation block ending with a control flow widget can have no control flow edge because they are removed.

This PR changes this behavior so the transitive edge removal will no longer trigger unless the block has no other incoming edges.

Required to fix #2081, but not the full story.

### DIFF
--- a/parseAPI/src/Parser.C
+++ b/parseAPI/src/Parser.C
@@ -1185,9 +1185,7 @@ Parser::delete_bogus_edges(Edge* e)
 
     // If the target block has other incoming edges,
     // then we do not remove the block at this point.
-    // It is possible that all incoming edges are bogus
-    // indirect edges, then the last removed edge will
-    // trigger the deletion of the block.
+// If there are other incoming edges, do not propagate.
     if (cur->sources().size() > 0) return;
 
     // If an indirect edge points a function entry,


bbiiggppiigg:
Ran testsuite on corriander. test_thread_1 crashed once, but didn't crash when I rerun it.

bbiiggppiigg:
@hainest @kupsch
I think the PR is ready for review.

hainest:
I think we should carefully analyze these comments from `finalize_jump_tables()`:

```c
During jump table analysis, we do not use any information about other jump tables.

Here, we know all the jump table starts and we assume that "no jump 
tables share any entries". Therefore, if a jump table overruns into
another jump table, we trim the overrun entries.
```

It makes me think these assumptions may be implicitly used beyond just `delete_bogus_edges`.

bbiiggppiigg:
> It makes me think these assumptions may be implicitly used beyond just `delete_bogus_edges`.

I think these comments are describing exactly what was being done in delete_bogus_edges.
All it is saying is that,

1. Jump tables are first parsed independently (therefore `no using any information about other jump tables`). 

2. During the finalization step, it trim overrun entries based on the assumption that no entries should overlap.

If you want we can schedule a meeting to discuss this.

bbiiggppiigg:
@hainest @kupsch 

Can you guys rereview this PR?
Now the transitive removal is only triggered when the block has no other incoming edge (instead of no other non-indirect incoming edge) .

parseAPI/src/Parser.C
```diff
@@ -1191,7 +1191,7 @@ Parser::delete_bogus_blocks(Edge* e)
     Block::edgelist sources;
     cur->copy_sources(sources);
     for (auto eit = sources.begin(); eit != sources.end(); ++eit)
-        if ((*eit)->type() != INDIRECT && (*eit)->src() != e->src()) 
+        if ((*eit)->src() != e->src()) 
             return;
```
kupsch: We just want to check if there are any source edges remaining. These 5 lines can be written more clearly; just check the blocks `sources` no need for sources variable, copying edges or the loop:
```cpp
if (sources().size() > 0)
  return;
```
I don't see how a block with no edges actually gets deleted. @bbiiggppiigg, @hainest We can discuss if that would be useful. If block don't need to be deletes then these changes are sufficient.

bbiiggppiigg: I was thinking that there might be duplicate edges from the same source, which may ore may note be true.
If there are duplicate edges, we need to check if all the remaining edges came from the same source as the one we want to remove, if there are no duplicates, then your change should suffice.

I probably need to check which case it is.

parseAPI/src/Parser.C
```diff
@@ -1206,7 +1206,7 @@ Parser::delete_bogus_blocks(Edge* e)
     Block::edgelist targets;
     cur->copy_targets(targets);
     for (auto eit = targets.begin(); eit != targets.end(); ++eit) {
-        delete_bogus_blocks(*eit);
+        delete_bogus_edges(*eit);
```
hainest, line 1203: I may be misremembering, but I thought only the edge between the table overrun and the target needs to be removed. If the target block is valid (i.e., someone else points to it), then we don't want to continue removing any edges to/from it. If the edge from the table overrun is the only one pointing to the block, then it has to be a bad block and it should get deleted. Although I'm not sure how that's accomplished here, either. I have to be missing something.

bbiiggppiigg: So the for loop in on line 1193 checks if it is the case that there any other edges pointing to this block, if so, return.
Then the for loop on line 1208 says, well it is the case that there are no other edges pointing to this block, and this block is bogus, we need to remove this block (and whatever target block/edges this block is pointing to  has, even though we are not actually removing the blocks because removing blocks can be hard).

parseAPI/src/Parser.C
```diff
@@ -1188,11 +1188,7 @@ Parser::delete_bogus_blocks(Edge* e)
     // It is possible that all incoming edges are bogus
     // indirect edges, then the last removed edge will
     // trigger the deletion of the block.
```
kupsch:
```suggestion
// If there are other incoming edges, do not propagate.
```

kupsch:
A better title would be `Fix removal of bad jump table edges`

PR #2088, opened 2026-01-07
Cleanup Symbol class

kupsch:
- remove Symbol::SymbolTags (not used)
- remove Symbol::Ptr (not used, never useful)
- made getMangleName method inline
- fix typo in docs:  ST_SETION -> ST_SECTION

### DIFF
--- a/symtabAPI/h/Symbol.h
+++ b/symtabAPI/h/Symbol.h
@@ -211,34 +211,34 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
 
    private:
 
-   Module*       module_;
-   SymbolType    type_;
-   int           internal_type_;
-   SymbolLinkage linkage_;
-   SymbolVisibility visibility_;
-   Offset        offset_;
-   Offset        ptr_offset_;  // Function descriptor offset.  Not available on all platforms.
-   Offset        localTOC_;
-   Region*       region_;
-   Symbol* 	 referring_;
-   unsigned      size_;  // size of this symbol. This is NOT available on all platforms.
+   Module*       module_{nullptr};
+   SymbolType    type_{ST_NOTYPE};
+   int           internal_type_{0};
+   SymbolLinkage linkage_{SL_UNKNOWN};
+   SymbolVisibility visibility_{SV_UNKNOWN};
+   Offset        offset_{0};
+   Offset        ptr_offset_{0};  // Function descriptor offset.  Not available on all platforms.
+   Offset        localTOC_{0};
+   Region*       region_{nullptr};
+   Symbol* 	 referring_{nullptr};
+   unsigned      size_{0};  // size of this symbol. This is NOT available on all platforms.
 
-   bool          isDynamic_;
-   bool          isAbsolute_;
-   bool          isDebug_;
+   bool          isDynamic_{false};
+   bool          isAbsolute_{false};
+   bool          isDebug_{false};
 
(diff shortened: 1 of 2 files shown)


symtabAPI/h/Symbol.h
```diff
@@ -176,8 +139,8 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
    /***********************************************************
      Name Output Functions
     ***********************************************************/		
-   std::string      getMangledName () const;
-   std::string	 getPrettyName() const;
+   std::string      getMangledName () const { return mangledName_; }
```
hainest, line 142: While we're in here, should we change the return type to `std::string const&`? As a future cleanup, we should probably add member variables for prettyName and typedName so we can cache them instead of calling the demangler every time their getters get called.

kupsch: My only concern about caching the pretty and typed names is the storage to store them.  We've already have issues reported about the memory usage of symbol names, so I think that will require additional study.

Without changing all three of them to have a member `getMangledName` has a different return type than `getPrettyName` and `getTypedName` which causes issues with the `Aggregate` class as we now need 2 different `Aggregate::name_iter` types that need to used for return types in `Aggregate` and 5 classes in dyninstAPI. If users use `auto` on the iterator methods there will be no issue, but if not this will break code. Given this would you suggest making this changing the return type, or wait until further clean up (that would maybe add caching member variables)?

bbiiggppiigg (approved):

symtabAPI/h/Symbol.h
hainest: > My only concern about caching the pretty and typed names is the storage to store them. We've already have issues reported about the memory usage of symbol names, so I think that will require additional study.

Oh, interesting. I would expect the reallocation for each call to be worse, but that would need to be tested.

> Without changing all three of them to have a member `getMangledName` has a different return type than `getPrettyName` and `getTypedName` which causes issues with the `Aggregate` class as we now need 2 different `Aggregate::name_iter` types that need to used for return types in `Aggregate` and 5 classes in dyninstAPI. If users use `auto` on the iterator methods there will be no issue, but if not this will break code. Given this would you suggest making this changing the return type, or wait until further clean up (that would maybe add caching member variables)?

Let's ignore it for now. We can add this to the infinitely-delayed todo list.

kupsch:
@hainest, use default member initialization to cleanup the constructors. Can you glance as these changes.
The testsuite passes successfully.

symtabAPI/h/Symbol.h
```diff
@@ -211,34 +211,34 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
...
+   SymbolType    type_{ST_NOTYPE};
+   int           internal_type_{0};
+   SymbolLinkage linkage_{SL_UNKNOWN};
+   SymbolVisibility visibility_{SV_UNKNOWN};
+   Offset        offset_{0};
+   Offset        ptr_offset_{0};  // Function descriptor offset.  Not available on all platforms.
+   Offset        localTOC_{0};
+   Region*       region_{nullptr};
+   Symbol* 	 referring_{nullptr};
+   unsigned      size_{0};  // size of this symbol. This is NOT available on all platforms.
```
hainest, line 224: Likely irrelevant nit:

```suggestion
   unsigned      size_{0u};  // size of this symbol. This is NOT available on all platforms.
```

PR #2090, opened 2026-01-09
Remove ARM ISA spec files

hainest:
These should have been removed by c9ccbb8e2 in 2024.

These were used to build the entryIDs for ARM. The importer script was replaced with one based on Capstone, so these should have been removed with it.


bbiiggppiigg:
Should these be removed only when the capstone branch is merged?

hainest:
> Should these be removed only when the capstone branch is merged?

They were only used by `instructionAPI/aarch64_pseudocode_extractor.py` to import the `entryID`s once. That importer was replaced in c9ccbb8e25 by one based on Capstone. These files should have been removed in that change.

Issue #2043, opened 2025-10-27
parseapi misses noreturn function


mwkrentel:
We found that `parse()` in Dyninst misses that `__stack_chk_fail()`
should be a noreturn function.

For now, I only have the bug in a fairly large binary (40 meg), but at
least the part that fails is small.

We build `hpcstruct` as part of HPCToolkit in an Ubuntu 24.04
container with gcc 13.3.0.

The problem occurs in the function `crypto_compute_hash_string()` at
the call instruction at address `0x1dac77`
```
00000000001dac10 <crypto_compute_hash_string>:
...
crypto_compute_hash_string():
/workspaces/hpctoolkit/builddir-O2/../src/common/lean/crypto-hash.c:180
  1dac4f:       48 8b 54 24 18          mov    0x18(%rsp),%rdx
  1dac54:       64 48 2b 14 25 28 00    sub    %fs:0x28,%rdx
  1dac5b:       00 00 
  1dac5d:       75 18                   jne    1dac77 <crypto_compute_hash_string+0x67>
  1dac5f:       48 83 c4 28             add    $0x28,%rsp
  1dac63:       5b                      pop    %rbx
  1dac64:       5d                      pop    %rbp
  1dac65:       c3                      ret
  1dac66:       66 2e 0f 1f 84 00 00    cs nopw 0x0(%rax,%rax,1)
  1dac6d:       00 00 00 
/workspaces/hpctoolkit/builddir-O2/../src/common/lean/crypto-hash.c:169
  1dac70:       b8 ff ff ff ff          mov    $0xffffffff,%eax
  1dac75:       eb d8                   jmp    1dac4f <crypto_compute_hash_string+0x3f>
/workspaces/hpctoolkit/builddir-O2/../src/common/lean/crypto-hash.c:180
  1dac77:       e8 a4 7e f6 ff          call   142b20 <__stack_chk_fail@plt>
  1dac7c:       0f 1f 40 00             nopl   0x0(%rax)

00000000001dac80 <hpctoolkit_demangle>:
hpctoolkit_demangle():
/workspaces/hpctoolkit/builddir-O2/../src/common/lean/demangle.c:34
  1dac80:       f3 0f 1e fa             endbr64
/workspaces/hpctoolkit/builddir-O2/../src/common/lean/demangle.c:35
  1dac84:       48 85 ff                test   %rdi,%rdi
  1dac87:       74 47                   je     1dacd0 <hpctoolkit_demangle+0x50>
  ...
```

[...]

PR #2093, opened 2026-01-09
cold clone symbols are not fncts, now type CODE

kupsch:
A symbol with a cold clone suffix is not a function entry point, but is part of the function without the suffix.  The new symbol type "ST_CODE" represents this type of symbol.

Without this change, cold function symbols are incorrectly treated as real functions.

### DIFF
--- a/symtabAPI/h/Symbol.h
+++ b/symtabAPI/h/Symbol.h
@@ -90,6 +90,7 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
       ST_TLS,
       ST_DELETED,
       ST_INDIRECT,
+      ST_CODE,
       ST_NOTYPE
    };
 
@@ -142,6 +143,13 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
    std::string      getMangledName () const { return mangledName_; }
    std::string      getPrettyName() const;
    std::string      getTypedName() const;
+
+   std::string      getCloneSuffix() const  {return getCloneSuffix(getMangledName());}
+   std::string      getVersionSuffix() const  {return getVersionSuffix(getMangledName());}
+   bool             isColdClone() const  {return isColdClone(getMangledName());}
+   static std::string   getCloneSuffix(const std::string& s);
+   static std::string   getVersionSuffix(const std::string& s);
+   static bool          isColdClone(const std::string& s);
 
    Module *getModule() const { return module_; } 
    Symtab *getSymtab() const;
--- a/symtabAPI/src/Symbol.C
+++ b/symtabAPI/src/Symbol.C
@@ -71,6 +71,51 @@ DYNINST_EXPORT string Symbol::getTypedName() const
   return P_cplus_demangle(mangledName_, true);
 }
 
+// Return a pair containing the clone suffix and the version suffix of s.
+// If either suffix is not present it is indicated by an empty string.
+// The clone suffix starts with the first '.' before the version suffix, and
+// ends with at the version suffix or the end of the s.
+// The version suffix starts with the first '@' and ends with the end of s.
+static std::pair<std::string, std::string> GetCloneAndVersionSuffix(const std::string& s)
+{
+    auto nameEnd = s.find_first_of(".@");           // find version or clone start
+    if (nameEnd == std::string::npos)  {
+	nameEnd = s.size();
+    }
+    auto cloneEnd = s.find_first_of("@", nameEnd);  // find version start
+    if (cloneEnd == std::string::npos)  {
+	cloneEnd = s.size();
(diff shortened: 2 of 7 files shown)


kupsch:
In my testing this fixes non-determinism reported by HPCToolkit in issue #2043. With this fix their test case runs deterministically. Mark is going to test to verify.

symtabAPI/h/Symbol.h
```diff
@@ -143,6 +144,13 @@ class DYNINST_EXPORT Symbol : public AnnotatableSparse
    std::string      getPrettyName() const;
    std::string      getTypedName() const;
 
+   std::string      getCloneSuffix() const  {return getCloneSuffix(getMangledName());}
+   std::string      getVersionSuffix() const  {return getVersionSuffix(getMangledName());}
+   bool             isColdClone() const  {return isColdClone(getMangledName());}
+   static std::string   getCloneSuffix(const std::string& s);
+   static std::string   getVersionSuffix(const std::string& s);
+   static bool          isColdClone(const std::string& s);
```
hainest, line 152: I don't see definitions for these static member funcs. Did I miss them?

symtabAPI/src/Symbol.C
```diff
@@ -71,6 +71,51 @@ DYNINST_EXPORT string Symbol::getTypedName() const
...
 }
 
+// Return a pair containing the clone suffix and the version suffix of s.
+// If either suffix is not present it is indicated by an empty string.
+// The clone suffix starts with the first '.' before the version suffix, and
+// ends with at the version suffix or the end of the s.
+// The version suffix starts with the first '@' and ends with the end of s.
+static std::pair<std::string, std::string> GetCloneAndVersionSuffix(const std::string& s)
+{
+    auto nameEnd = s.find_first_of(".@");           // find version or clone start
```
hainest, line 81: I can't find an official spec, but the linux kernel folks just look for '.cold' in the [string](https://lkml.org/lkml/2025/11/20/1676). I'm not sure the llvm folks include '@'.

symtabAPI/h/Symbol.h
kupsch: The static method definitions are in Symbol.C. They are present so they can be used before a Symbol is created.

symtabAPI/src/Symbol.C
kupsch: I also could not find any spec, but from compiler source code, binutils, and various mailing lists & web pages, the general format is (NAME matches the re `[^.@]*`):

- Just NAME without any `.` is a non-clone name.
- A name with `.` is a clone name with NAME as the base followed by clone suffix(es). There may be multiple clone suffixes concatenated together, where the `.` begins the clone(s) and goes to the end of the string or an `@` (which begins dynamic linking version string). The individual clone suffixes have the form:
  - `\.\d+` - old format just a number (I don't think these are used anymore)
  - `\.[^.@\d]+(\.\d+)*` - a clone type name optionally followed by a sequence of dotted numbers.  This allows several instances of a particular clone type (such as specialization based on hardware capabilities), and enables different optimizations that generate clones to have unique names. This allows names like `NAME.isra.0.cold` which is a `.cold` clone of NAME's `.isra.0` clone (the first ISRA optimized version of NAME).

Since the clones are typically artifacts of optimization, you need to build with optimizations or OMP to see them. 

The `@` begins the dynamic linking version string suffix and is always the last part of a symbol.

symtabAPI/h/Symbol.h
hainest: Derp. I concluded that those definitions were for the member functions, even though I saw the definitions of those right above here. My brain just failed to make the connection.

symtabAPI/src/Symbol.C
hainest: Ah, I didn't think about versioned cold symbols. Could you add all/some of this description to the docs?

hainest:
@kupsch 

I was finally able to get clang to emit cold symbols using `-mllvm -hot-cold-split=true`. When building Dyninst, it did use `STT_FUNC` for them, so the current check should work fine. I'll make a PR to add this flag when building with clang. Annoyingly, the linker complains when it gets this flag, so I might need to do some refactoring first.

symtabAPI/src/Symbol.C
kupsch: I updated the docs and the comment in symbolDemange.c.

I also updated the `elfSymType` function to convert the Symbol type `ST_CODE` back to the ELF symbol type `STT_TYPE`.

hainest:
@kupsch Could you rebase this on master to pick up the changes to the tests?

kupsch:
@hainest done.

PR #2094, opened 2026-01-11
Remove AST::hasCycle

hainest:
It should have been removed by e3dadf3fe in 2014.


hainest:
> Since this PR also touches AST.C
> 
> I would not merge until 1987 is done.
> 
> In case that he mess up the rebase or whatsoever.

Agreed.

PR #2095, opened 2026-01-11
Remove invalid DataflowAPI AST types

hainest:
These were all added ca 2010, but never implemented.


hainest:
The consumers CI workflow is passing, so it doesn't seem anyone is using them.

hainest:

PR #2096, opened 2026-01-13
Extract address alignment check into Architecture

hainest:
This can't be merged until after #2082 and riscv check is added.

@bbiiggppiigg Do the amd gpus have alignment requirements?


bbiiggppiigg:
I don't think so.

hainest:

PR #2097, opened 2026-01-15
Fix handling of CMAKE_C_FLAGS and CMAKE_CXX_FLAGS

hainest:
CMAKE_<LANG>_FLAGS were improperly expanded, so didn't handle cases with spaces.

This happens when passing multiple values to `CMAKE_<LANG>_FLAGS`. For example, using `cmake -DCMAKE_C_FLAGS='-Wall -Wextra'` would ultimately produce `gcc '-Wall -Wextra'`. With this change, the compiler arguments are correctly expanded to `gcc -Wall -Wextra`.


cmake/DyninstOptimization.cmake
```diff
@@ -102,10 +102,13 @@ endif()
 #   ${CMAKE_<LANG>_FLAGS} ${CMAKE_<LANG>_FLAGS_<BUILD>} <options> ${CMAKE_<LANG>_FLAGS}
 #
 string(TOUPPER ${CMAKE_BUILD_TYPE} _build_type)
+string(REPLACE " " ";" _c_flags "${CMAKE_C_FLAGS}")
+string(REPLACE " " ";" _cxx_flags "${CMAKE_CXX_FLAGS}")
 
-set(DYNINST_C_FLAGS_${_build_type} ${DYNINST_FLAGS_${_build_type}} ${CMAKE_C_FLAGS})
+set(DYNINST_C_FLAGS_${_build_type} ${DYNINST_FLAGS_${_build_type}} ${CMAKE_C_FLAGS}  ${_c_flags})
```
bbiiggppiigg: So does this mean that both versions of the value of CMAKE_C_FLAGS 
(One where space got replaced with semicolon and the original)
will be set to (DYNINST_C_FLAGS_${_build_type}?

hainest: Bah, no. That's a bug. I thought I fixed that.

hainest: @bbiiggppiigg It's fixed now. Thanks for the careful review.

bbiiggppiigg (approved):

hainest:

PR #2099, opened 2026-01-15
Refactor dataflowAPI AST types

hainest:
This is a big cleanup of the AST types to make their declarations explicit (instead of wandering through macros), make their relationships clearer, and simplify the headers to reduce inclusion confusion. I didn't update the docs to change where the Variable/Constant/Rose/Stack/Bottom types are declared. I can, if folks think that would be useful.

Notable changes:

1. The return type of Variable/Constant/ROSEOperation/Bottom::format() changed from `const std::string` to `std::string` to allow NRVO/moving. This is an ABI-break, but ultimately doesn't affect users since they can't safely inherit from these classes.
2. The default constructors are now public. This allows for using `boost::make_shared` and removes at least one allocation when constructing a shared pointer in the AST's `create` method.
3. Dynamic casts through references have been converted to pointers to avoid `std::bad_cast` in `isStrictEqual`.

    ```c++
    const name &other(dynamic_cast<const name&>(rhs);
    ...
    ```
    
    to 
    
    ```c++
    auto const *ptr = dynamic_cast<RoseAST const *>(&rhs);
    if(!ptr) {
      return false;
    }
    ...
    ```

4. Converting constructors (e.g., `Variable(Dyninst::AbsRegion r)` are now `explicit`.

---

I'd like to make these user-visible ABI-breaking changes, but will leave this for another day:

1. Change the return type of `AST::format()` from `const std::string` to `std::string` to enable NRVO/move semantics.
2. Put `StackAST` in the `DataflowAPI` namespace

### DIFF
--- a/dataflowAPI/h/BottomAST.h
+++ b/dataflowAPI/h/BottomAST.h
@@ -55,7 +55,7 @@ namespace Dyninst { namespace DataflowAPI {
       return boost::make_shared<BottomAST>(t);
     }
 
-    const std::string format() const override {
+    std::string format() const override {
       std::stringstream ret;
       ret << "<" << std::boolalpha << t_ << ">";
       return ret.str();
--- a/dataflowAPI/h/ConstantAST.h
+++ b/dataflowAPI/h/ConstantAST.h
@@ -93,7 +93,7 @@ namespace Dyninst { namespace DataflowAPI {
       return boost::make_shared<ConstantAST>(std::move(t));
     }
 
-    const std::string format() const override {
+    std::string format() const override {
       std::stringstream ret;
       ret << "<" << t_ << ">";
       return ret.str();
--- a/dataflowAPI/h/VariableAST.h
+++ b/dataflowAPI/h/VariableAST.h
@@ -97,7 +97,7 @@ namespace Dyninst { namespace DataflowAPI {
       return boost::make_shared<VariableAST>(std::move(t));
     }
 
-    const std::string format() const override {
+    std::string format() const override {
       std::stringstream ret;
       ret << "<" << t_ << ">";
       return ret.str();
--- a/dataflowAPI/h/AST.h
+++ b/dataflowAPI/h/AST.h
@@ -85,7 +85,7 @@ namespace Dyninst {
       return *this == (*rhs);
     }
 
-    virtual const std::string format() const = 0;
+    virtual std::string format() const = 0;
 
     // Substitute every occurrence of `a` with `b` in AST `in`.
     // Returns a new AST.
--- a/dataflowAPI/h/RoseAST.h
+++ b/dataflowAPI/h/RoseAST.h
@@ -243,7 +243,7 @@ namespace Dyninst { namespace DataflowAPI {
       return boost::make_shared<RoseAST>(std::move(t), std::move(c));
     }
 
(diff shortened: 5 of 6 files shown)


hainest:

dataflowAPI/src/SymEval.C
```diff
@@ -589,7 +593,7 @@ bool SymEval::expandInsn(const Instruction &insn,
 }
 
 
-SymEval::Retval_t SymEval::process(SliceNode::Ptr ptr,
+SymEval::Retval_t process(SliceNode::Ptr ptr,
```
bbiiggppiigg: Does this one need to be static as well?

hainest: It's declared `static`:

```c++
static SymEval::Retval_t process(SliceNode::Ptr, Result_t &, std::set<Edge::Ptr> &);
```

The definition doesn't have to be explicity marked `static`.

dataflowAPI/h/BottomAST.h
```diff
@@ -0,0 +1,97 @@
...
+  private:
+    bool isStrictEqual(Dyninst::AST const &rhs) const override {
+      auto const *ptr = dynamic_cast<BottomAST const *>(&rhs);
+      if(!ptr) {
+        return false;
+      }
+      return t_ == ptr->val();
+    }
+
+    bool t_;
```
kupsch: The default ctor will leave this uninitialized; default member initialize to fix:
```suggestion
    bool t_{};
```

dataflowAPI/h/ConstantAST.h
```diff
@@ -0,0 +1,135 @@
...
+
+    bool operator==(const Constant &rhs) const {
+      return ((rhs.val == val) && (rhs.size == size));
+    }
+
+    bool operator<(const Constant &rhs) const {
+      if(val < rhs.val) {
+        return true;
+      }
+      if(size < rhs.size) {
```
kupsch, line 61: I would think that you'd only do the size comparison if the val's are equal so `(a < b)` is equivalent to `!(b < a) && !(b == a))`
```suggestion
      if(val > rhs.val) {
        return false;
      }
      if(size < rhs.size) {
```

dataflowAPI/h/VariableAST.h
```diff
@@ -0,0 +1,139 @@
...
+
+    explicit Variable(Dyninst::AbsRegion r) : reg(std::move(r)), addr(0UL) {}
+
+    Variable(Dyninst::AbsRegion r, Dyninst::Address a) : reg(std::move(r)), addr(a) {}
+
+    bool operator==(const Variable &rhs) const {
+      return ((rhs.addr == addr) && (rhs.reg == reg));
+    }
+
+    bool operator<(const Variable &rhs) const {
```
kupsch, line 60: Is this a broken comparison operator? `reg` seems like the major component and `addr` the minor. Should it compare `reg` first and only compare `addr` is `reg` is equal?

bbiiggppiigg (approved):

dataflowAPI/h/BottomAST.h
hainest: Fixed.

hainest:
There are no test regressions on the last set of changes. I'll make a note to look at the operators.

PR #2101, opened 2026-01-16
Remove AstMiniTrampNode::generateTramp

hainest:
Its usage was removed by 7fbd797d66 in 1997.

### DIFF
--- a/dyninstAPI/src/ast.C
+++ b/dyninstAPI/src/ast.C
@@ -537,40 +537,6 @@ AstNode::~AstNode() {
     //printf("at ~AstNode()  count=%d\n", referenceCount);
 }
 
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
@hainest 
Is this class AstMiniTrampNode  even used?
It seems like it is only used when someone call miniTrampNode, but I don't see where it is called.

hainest:
> @hainest Is this class AstMiniTrampNode even used? It seems like it is only used when someone call miniTrampNode, but I don't see where it is called.

Yeah, it's not used anymore. The original commit I cited was for a different version of the function in a long-removed class. The usage of `AstMiniTrampNode` was removed in 2011. I'll close this and make a new PR. Thanks for checking!

PR #2104, opened 2026-01-16
Remove AstLabelNode

hainest:
It was added by 20b5e65d718c5 in 2006, but never used.


bbiiggppiigg (approved):

kupsch (approved):

hainest:
No build failures on x86, arm, or ppc. I didn't run the full testing battery because there's no way this class could be used.

PR #2105, opened 2026-01-17
Remove findTags

hainest:
Its definition was commented out by f7b76cd496 in 1995 and its usage was removed by db8cb862cd9 in 1997.


hainest:

PR #2106, opened 2026-01-17
Remove initDefaultPointFrequencyTable

hainest:
Its last usage was removed by 5dabf3ca8b1d in 2006.


bbiiggppiigg (approved):

hainest:
No build failures on x86, arm, or ppc. I didn't run the full testing battery because there's no way these could be used.

PR #2107, opened 2026-01-17
Remove dead instrumentation function declarations

hainest:


bbiiggppiigg (approved):

kupsch (approved):

hainest:
No build failures on x86, arm, or ppc. I didn't run the full testing battery because there's no way these could be used.

PR #2108, opened 2026-01-18
Remove int_thread::changeLWP

hainest:
This was only used for BlueGene and should have been removed by cccad8ee9 in 2020.


kupsch (approved):

hainest:
This is never used, so I'm not going to run the usual test suite battery.

PR #2109, opened 2026-01-18
Remove FreeBSD x86-specific files

hainest:
These were slowly removed over time, but bulk was removed by c7ed07ce3 in 2010.


bbiiggppiigg (approved):

hainest:
I don't have any FreeBSD machine to test this on, so I'm merging as-is.

PR #2110, opened 2026-01-18
[AMDGPU] Update the kernel descriptor at instrumentation time

rochauha:
There is a separate tool that was used to update the kernel descriptors. Since we moved the kernel descriptor code in master, that tool is no longer needed. This PR makes Dyninst rewrite the kernel descriptors too.

### DIFF
--- a/dyninstAPI/src/AmdgpuPointHandler.C
+++ b/dyninstAPI/src/AmdgpuPointHandler.C
@@ -178,7 +178,7 @@ void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function)
 }
 
 static constexpr uint32_t roundUpTo8(uint32_t x) {
-  return ((x + 7) >> 3) << 3;
+  return ((x + 7) / 8) * 8;
 }
 
 void AmdgpuGfx908PointHandler::maximizeSgprAllocationIfKernel(BPatch_function *function) {


rochauha:
@bbiiggppiigg, @kupsch, @hainest, this PR is ready for review.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -171,6 +171,29 @@ void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function)
...
+  bool success = kdVariable->readValue(&rawKd, sizeof rawKd);
+  assert(success);
+
+  AmdgpuKernelDescriptor kd(rawKd, this->eflag);
+
+  // This math comes from LLVM AMDGPUUsage documentation
+  uint32_t newValue = 2 * ((112 / 16) - 1); // Set SGPRs to 112 (max value).
+  kd.setCOMPUTE_PGM_RSRC1_GranulatedWavefrontSgprCount(newValue);
+  uint32_t kernargSize = kd.getKernargSize();
+  while (kernargSize % 8) {
```
hainest: There is a nice bit-twiddle for finding the next greater multiple of 8: https://www.geeksforgeeks.org/dsa/round-to-next-greater-multiple-of-8. That's likely less expensive than mod+loop. Definitely want to make a clear comment for it or refactor it into a function with a good name.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -171,6 +171,29 @@ void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function)
   insertEpilogueAtPoints(epilogueSnippet, exitPoints);
 }
 
+void AmdgpuGfx908PointHandler::maximizeSgprAllocationIfKernel(BPatch_function *function) {
+  BPatch_variableExpr *kdVariable = getKernelDescriptorVariable(function);
+  assert(kdVariable);
+
+  llvm::amdhsa::kernel_descriptor_t rawKd;
+  bool success = kdVariable->readValue(&rawKd, sizeof rawKd);
+  assert(success);
```
hainest: I'd recomend adding some logging here for a little bit better debugging experience.

```c++
if(!success) {
  inst_printf("Unable to read kernel descriptor for '%'\n", kdVariable->getName());
  assert(0);
}
```

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -171,6 +171,29 @@ void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function)
...
+  uint32_t newValue = 2 * ((112 / 16) - 1); // Set SGPRs to 112 (max value).
+  kd.setCOMPUTE_PGM_RSRC1_GranulatedWavefrontSgprCount(newValue);
+  uint32_t kernargSize = kd.getKernargSize();
+  while (kernargSize % 8) {
+    ++kernargSize;
+  }
+  kd.setKernargSize(kernargSize + 8);
+
+  // We have modified the kernel descriptor. Now overwrite the original one with it.
+  kdVariable->writeValue(kd.getRawPtr(), sizeof(rawKd), true);
```
hainest: The default value for `bool saveWorld` in [`writeValue`](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/h/BPatch_snippet.h#L436) is `false`. Yet, that variable isn't used in the [definition](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/BPatch_snippet.C#L1357). Since we're passing `true` here, is there some intended behavior that's not actually used by `BPatch_variableExpr`?

rochauha: If no argument is passed, it becomes ambiguous for the compiler to choose between `bool writeValue(const void *src, bool saveWorld=false)` and `bool writeValue(const void *src, int len,bool saveWorld=false)`. 

This leads to a compile error on gcc 8.5.0

rochauha:
Addressed comments by @hainest

dyninstAPI/src/AmdgpuKernelDescriptor.C
```diff
@@ -122,17 +122,15 @@ void AmdgpuKernelDescriptor::setKernelCodeEntryByteOffset(int64_t value) {
 }
 
 #define GET_VALUE(MASK) ((fourByteBuffer & MASK) >> (MASK##_SHIFT))
-#define GET_ITH_BIT_AFTER(MASK, i)                                                                 \
-  (((fourByteBuffer & (1 << (MASK##_WIDTH) + i - 1)) << (MASK##_SHIFT)) != 0)
 
-static inline uint32_t insertField32(uint32_t op, uint32_t value, uint32_t mask, uint32_t shift) {
-  assert(value >> shift && "value contains more bits than specified");
+static inline uint32_t insertField32(uint32_t op, uint32_t value, uint32_t mask, uint32_t width, uint32_t shift) {
+  assert(value >> width == 0 && "value contains more bits than specified");
```
bbiiggppiigg: I know this works under operator precedence, but can you change this and the following one to use paranthesis for better readability 

((value >> width)  ==0 ) && ...

rochauha: Done.

rochauha:
Addressed comment by @bbiiggppiigg.

rochauha:
@hainest, @kupsch - any comments?

dyninstAPI/src/AmdgpuPointHandler.C
hainest: Oh, so you're having to work around some old hacked-up code that someone didn't put enough thought into designing. That still doesn't answer my question as to why you passed `true` when the default is `false`. I guess it doesn't matter?

rochauha: It doesn't matter. Do you want me to change it to false?

hainest: Yeah, that would make it consistent with the default. Could you also add a comment that says it's just there to fix an overload resolution bug?

rochauha:
@hainest, I have clarified use of writeValue and rebased on master.

hainest (approved):

rochauha:
@kupsch, any comments?

rochauha:
@kupsch, please let me know if you have any comments

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -171,6 +177,43 @@ void AmdgpuGfx908PointHandler::insertEpilogueIfKernel(BPatch_function *function)
   insertEpilogueAtPoints(epilogueSnippet, exitPoints);
 }
 
+static constexpr uint32_t roundUpTo8(uint32_t x) {
+  return ((x + 7) >> 3) << 3;
```
kupsch: Don't obfuscate divide and multiply by 8
```suggestion
  return ((x + 7) / 8) * 8;
```

rochauha: Done.

PR #2111, opened 2026-01-18
Use Dyninst::Register in CFWidget

hainest:


bbiiggppiigg:
@hainest 
Can you explain a bit on what this PR is doing?
Was it not already using Dyninst::Register?

hainest:
> @hainest Can you explain a bit on what this PR is doing? Was it not already using Dyninst::Register?

Based on the comment above the typedef, I'd say that `Dyninst::Register` was not standardized at the time CFWidget was designed. Ultimately, I have no idea why all registers are treated as containing only unsigned integral value. That seems nonsensical.

kupsch (approved):

hainest:

PR #2112, opened 2026-01-18
Introduce a richer Register type

rochauha:
During codegen, Dyninst models registers simply using an unsigned int that represents a register id.

For AMDGPU, we may need to allocate, track and free single registers, pairs or quads depending on the use case. Further, we also need to distinguish between scalar and vector registers.

This PR introduces a new Register type that is non-intrusive for other architectures.

### DIFF
--- a/common/src/dyn_register.h
+++ b/common/src/dyn_register.h
@@ -75,7 +75,7 @@ public:
   constexpr Register()
       : id(OperandRegId(static_cast<uint32_t>(-1))), kind(RegKind::UNKOWN_KIND), count(BlockSize(1)) {}
 
-  // This is only to make existing code work, and must go away in the future.
+  // Support existing Register usage that initializes with uint32_t.
   constexpr Register(uint32_t rawId)
       : id(OperandRegId(rawId)), kind(RegKind::SCALAR), count(BlockSize(1)) {}
 


hainest:
I would strongly discourage doing this. There are currently 8 completely separate register representations in Dyninst (see below). Adding another one expands the problem- especially one that only works on one architecture and in the codegen classes that already treat registers as integers (`Dyninst::Register`) and containing only integers (`Dyninst::RegValue`).

I definitely agree there needs to be a generic register type for codegen that actually encodes information, instead of being an opaque unsigned integral. I would recommend something like

```c++
struct register_t {
  Dyninst::Machregister reg;
  value_t val;
  MachRegNum reg_num;
};
```

By using `MachRegister`, you gain a huge amount of information that is used throughout Dyninst. For example, `enum RegKind` is `MachRegister::isVector` and friends. Notably, `MachRegister::isMatrix` is missing, but should be added since we support the ARMv9 Advanced Matrix Extension and will soon support Intel's AMX. And `enum RegUsage` is `MachRegister::isGeneralPurpose` et al. Naturally, this raises the question of `RegUsage::TEMPORARY`. I'm not sure how this would be used, but I'm assuming it's something similar to how [`registerSlot::initialLiveness_t`](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/registerSpace.h#L94) is used in [`getScratchRegister`](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/registerSpace.C#L356).

This raises two important questions:

1. How do you represent the value contained in a register (i.e., what is the definition of `value_t`). Currently, everything uses `Dyninst::RegValue` but I think that's only because rewriting floating-point registers was never implemented. For vector registers, this will clearly need to be a collection of values.
2. We need a uniform means of representing the machine-level registers number (here, `MachRegNum`). This is an ID assigned to each register by the ISA definition. I believe all supported architectures just use an int-like type for this. @wxrdnx Is this true for riscv64? However, future platforms might need something more complex. I'm not saying we need to solve that problem now, but we should at least allow for the possibility.
  
---

Current register representations:

```
MachRegister

common/src/dyn_register.h
  typedef unsigned int Register;
  typedef long long int RegValue;

dyninstAPI/src/Widgets/CFWidget.h (will be removed by #2111)
  typedef unsigned int Register;

src/registerSpace.h
  {power,aarch64Registers_t}Registers_t

rose/semantics/Registers.h
  RegisterDictionary

rose/semantics/RegisterDescriptor.h

src/ABI.h
  As a bitArray

src/RegisterMap.h

src/liveness.C
  regs{1,2,3}
```

rochauha:
This PR isn't ready for review. I was experimenting with the new Register type.

@hainest, thanks for the feedback. I skimmed through MachRegister. I think I might be able to leverage it instead of the new Register type.

I want to ensure existing code for liveness and codegen mostly works. The idea of a register block is only to allow allocating / freeing consecutive scalar or vector registers. On looking at a block, liveness analysis should look at individual bits corresponding to the registers in the block.

About temporary registers -- I don't think I will need it. But AMDGPU does have temporary registers `ttmp0` to `ttmp15`, so I thought I should model it in case we ever need it.

rochauha:
What is the use case for `value_t` or its equivalent? We don't ever know what value is held inside a register unless it is a constant that we actually move into the register.

hainest:
> What is the use case for `value_t` or its equivalent? We don't ever know what value is held inside a register unless it is a constant that we actually move into the register.

Yes, specifying an immediate is the most likely use case. Whether a `register_t` would need to carry around a possible value or not is a good question. In any case, we need a better type than the current `RegValue`.

wxrdnx:
> I believe all supported architectures just use an int-like type for this. @wxrdnx Is this true for riscv64?

Yes, that's true.

rochauha:
@hainest, `ABI.h` and `RegisterMap.h` don't contain register definitions. RegisterMap is mapping MachRegisters to a unique index in a bitvector (the bitArray) for liveness analysis (the dataflow values are represented as bitvectors for liveness). This is required.

I was investigating using MachRegister entirely instead of the old Register (unsigned int) and the new Register type here (union). I don't think MachRegister it is suitable for codegen yet.

Both our solutions introduce a new type to replace the current unsigned int representation of registers. I believe my change is less intrusive -- just this PR. This new Register type (union) works without breaking codegen for any architecture. AMDGPU will rely on the richer type for codegen. Other architectures still work as-is without additional conditional compilation.

Introducing a new struct as you mentioned will be an intrusive change for codegen because existing code uses a number for the register across the board, and the struct won't fit. It will also updating the `RegisterConversion`s, and the register allocation and tracking mechanism - which is a huge mess.

PR #2115 shows the changes required for AMDGPU. I haven't touched register allocation yet. It works because of backwards compatibility (even for AMDGPU).

I am quite time bound and it would be irresponsible on my end to pursue a major Dyninst refactor right now.

That said, I think the new Register type in this PR can be used as a transition step and we can then slowly refactor MachRegister so it can be used for codegen as well (my next comment discusses my use cases). MachRegister is a `uint32_t` wrapped inside a class. Register in this PR is a `uint32_t` as well. Once the refactor is complete, we evaluate it using `typedef MachRegister Register`, remove RegisterConversions and fix remaining codegen issues. Then we can transition to using MachRegister entirely, and remove dyn_register.h, RegisterConversions, etc.

rochauha:
Though MachRegister has the most amount of information, one currently can't create a complete MachRegister instance that is suitable for codegen. The implementation contains way too many magic numbers.

MachRegister is used for liveness and other analyses as analysis requires as much information as possible.

However, codegen is mostly concerned with using a _subset of registers_; and at times, I need to be able to exactly say which register I want to use. For example, the kernarg pointer is held in a register pair, which can be determined by looking at the kernel descriptor. After determining the pair, I then need to be able to say that I need to create and pass this register -- see [here](https://github.com/dyninst/dyninst/pull/2115/files#diff-d06f84df0f4004feeb2ad273b60e8be45b4a07757afcedb00747ebc3dd22855fR689). The emitter can then look at this and use that scalar register pair beginning at regId. I can also use assertions to verify that regId is even.

Using my new Register type, the implementation becomes simpler -- I can then extend my emitter to take in a register pair, register quad, work with vectors, scalars and a mix -- which is only required for AMDGPU right now. For example, just by looking at the operand triplets, the emitter can determine which instructions to generate (scalar = scalar op scalar, vector = vector op vector, vector = scalar op vector). A slightly different configuration if we have pairs. But everything can be determined by looking at the operand, no additional state needs to be maintained. The regIds for SGPRs and VGPRs can be the same (i.e v0 and s0 are both 0, v1 and s1 are both 1, etc). But the register kinds will be different and I can choose which instructions to use.

rochauha:
@hainest, @bbiiggppiigg, @kupsch - this PR is ready for review.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
-/* This needs to be an int since it is sometimes used to pass offsets
-   to the code generator (i.e. if-statement) - jkh 5/24/99 */
 
-/* a register number, e.g., [0..31]  */
+/* Dyninst has been using the following typedef for representing a register number, e.g., [0..31]
 typedef unsigned int Register;
+*/
+
+// Without making intrusive changes to existing CPU architectures, we want to introduce a richer
+// Register type to model different kinds of registers.
```
kupsch: These comments are useless in the source code. What would they mean to someone in the future? Delete.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+};
+
+constexpr uint32_t REG_ID_WIDTH = 12;
+constexpr uint32_t REG_KIND_WIDTH = 4;
+constexpr uint32_t REG_USAGE_WIDTH = 4;
+constexpr uint32_t REG_COUNT_WIDTH = 5;
+constexpr uint32_t REG_RESERVED_WIDTH = 7;
+// total = 32
+
+struct __attribute__((packed)) RegisterFields {
```
kupsch: why is there an attribute here? Also I'd prefer removal of blank lines in the struct. We need a better name for this also. You should add a constructor.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+  uint32_t usage : REG_USAGE_WIDTH;
+
+  uint32_t count : REG_COUNT_WIDTH;
+
+  uint32_t reserved : REG_RESERVED_WIDTH;
+};
+
+union Register {
+  RegisterFields details;
+  uint32_t raw;
```
kupsch: `value` would be a better name than `raw`.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+  constexpr Register(uint32_t val) : raw(val) {}
+
+  constexpr Register() : raw(static_cast<uint32_t>(-1)) {}
```
kupsch: add default value and the default ctor can go away:
```suggestion
  constexpr Register(uint32_t val = -1) : raw(val) {}
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+// Register type to model different kinds of registers.
+
+enum RegKind : uint32_t { SCALAR = 0, VECTOR = 1, MATRIX = 2, UNDEFINED_KIND = 3 };
+
+enum RegUsage : uint32_t {
+  GENERAL_PURPOSE = 0,
+  SPECIAL_PURPOSE = 1,
+  PREDICATE = 2,
+  UNDEFINED_USAGE = 3
+};
```
kupsch: I don't understand the purpose of these two enums. I understand SCALAR and VECTOR. The rest not so much.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+struct __attribute__((packed)) RegisterFields {
+  uint32_t id : REG_ID_WIDTH;
+
+  uint32_t kind : REG_KIND_WIDTH;
+
+  uint32_t usage : REG_USAGE_WIDTH;
+
+  uint32_t count : REG_COUNT_WIDTH;
+
+  uint32_t reserved : REG_RESERVED_WIDTH;
```
kupsch: unnecessary.  delete.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+  PREDICATE = 2,
+  UNDEFINED_USAGE = 3
+};
+
+constexpr uint32_t REG_ID_WIDTH = 12;
+constexpr uint32_t REG_KIND_WIDTH = 4;
+constexpr uint32_t REG_USAGE_WIDTH = 4;
+constexpr uint32_t REG_COUNT_WIDTH = 5;
+constexpr uint32_t REG_RESERVED_WIDTH = 7;
+// total = 32
```
kupsch: `REG_RESERVED_WIDTH` unnecessary and comment is unneeded. delete:

```suggestion
```

common/src/dyn_register.h
kupsch: REG_RESERVED_WIDTH unnecessary. delete:
```suggestion
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+  uint32_t kind : REG_KIND_WIDTH;
+
+  uint32_t usage : REG_USAGE_WIDTH;
+
+  uint32_t count : REG_COUNT_WIDTH;
+
+  uint32_t reserved : REG_RESERVED_WIDTH;
+};
+
+union Register {
```
kupsch: Register should be a class. This union should be a private member.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+  constexpr RegUsage getUsage() const { return static_cast<RegUsage>(details.usage); }
+
+  constexpr uint32_t getCount() const { return details.count; }
+
+  // A register block represents consecutive reqisters starting from id. Not all architectures
+  // need or support this.
+  //
+  // So count = 0 is an individual register.
+  // count >= 1 means "register block" with 1 or more registers.
+  constexpr bool isRegisterBlock() const { return details.count != 0; }
```
kupsch: Nope. An individual register has count 1. count 0 mean NO registers. Is this method useful? It's only use in this PR, delete:
```suggestion
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+  constexpr operator uint32_t() const { return raw; }
+
+  // If this is a register block, return individual registers in that block.
+  void getIndividualRegisters(std::vector<Register> &individualRegisters) const {
+    assert(this->isRegisterBlock() && "This must be a register block");
```
kupsch: This is an unnecessary restriction. delete:
```suggestion
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+// Along with the older Register type, which was just an unsigned int, use of RegValue and
+// Null_Register are also scattered around the codebase.
 
 /* register content 64-bit */
 typedef long long int RegValue;
 
 /* '255' */
-constexpr Register Null_Register{static_cast<unsigned int>(-1)};
+constexpr Register Null_Register(static_cast<unsigned int>(-1));
```
kupsch: The default ctor makes a Null_Register:
```suggestion
constexpr Register Null_Register();
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+} // namespace Dyninst
 
-}
+// Adding a hash function for the new Register type
+namespace std {
+template <> struct hash<Dyninst::Register> {
+  size_t operator()(const Dyninst::Register &reg) const noexcept {
+    return std::hash<uint32_t>{}(reg.raw);
```
kupsch: Since raw is going away, this does the same thing:
```suggestion
    return std::hash<uint32_t>{}(reg);
```

dyninstAPI/src/ast.C
```diff
@@ -3064,7 +3064,7 @@ void regTracker_t::debugPrint() {
 
     for (auto iter = tracker.begin(); iter != tracker.end(); ++iter) {
        fprintf(stderr, "AstNode %p: register %u, condition level %d\n",
-               (void*)iter->first, iter->second.keptRegister, iter->second.keptLevel);
+               (void*)iter->first, (iter->second.keptRegister).raw, iter->second.keptLevel);
```
kupsch: please don't add parentheses that do not clarify precedence. These are just noise here and elsewhere

```suggestion
               (void*)iter->first, iter->second.keptRegister.raw, iter->second.keptLevel);
```

dyninstAPI/src/emit-x86.C
```diff
@@ -1793,7 +1793,7 @@ Register EmitterAMD64::emitCall(opCode op, codeGen &gen, const std::vector<AstNo
    int frame_size = 0;
    for (int u = operands.size() - 1; u >= 0; u--) {
       Address unused = ADDR_NULL;
-      unsigned reg = Null_Register;
+      Register reg = Null_Register;
```
kupsch, line 1780: use auto:
```suggestion
      auto reg = Null_Register;
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+  Register(uint32_t id, RegKind kind, RegUsage usage, uint32_t count) {
+    assert(id >> REG_ID_WIDTH == 0 && "id is wider than than specified");
+    assert(count >> REG_COUNT_WIDTH == 0 && "count is wider than specified");
+    details.id = id;
+    details.kind = kind;
+    details.usage = usage;
+    details.count = count;
+    details.reserved = 0;
+  }
```
kupsch: Initialize details using member initialization.

common/src/dyn_register.h
```diff
@@ -31,19 +31,128 @@
...
+
+  constexpr uint32_t getId() const { return details.id; }
+
```
kupsch: Also add method for `getValue()` since `raw` is going away.

common/src/dyn_register.h
rochauha: Since you suggested making a constructor for the struct with bitfields, I used that instead.

dyninstAPI/src/emit-x86.C
rochauha: Done

dyninstAPI/src/ast.C
rochauha: Raw is no longer there, so I used a cast instead of this.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: I simiplfied it. I removed RegUsage entirely.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: The attribute was to ensure that the struct fits in 32 bits. Do you want me to removei t? 

I Removed the blank lines.

common/src/dyn_register.h
rochauha: Removed.

common/src/dyn_register.h
rochauha: Done.

rochauha:
@kupsch , I have addressed all your comments. Please let me know if you have any more.

common/src/dyn_register.h
rochauha: Removed.

common/src/dyn_register.h
rochauha: Done.

rochauha:
@bbiiggppiigg, @hainest, please let me know if you have any comments

common/src/dyn_register.h
kupsch: Yes, this is unnecessary. The struct should be 4 bytes without this.

dyninstAPI/src/ast.C
kupsch: Please change all the casts to `getValue()` method calls. I general casting is bad.

common/src/dyn_register.h
rochauha: Using the default ctor here confuses the compiler into thinking that Null_Register() is a function.

common/src/dyn_register.h
rochauha: Done.

dyninstAPI/src/ast.C
rochauha: Done.

rochauha:
Addressed new comments by @kupsch.

common/src/dyn_register.h
```diff
@@ -31,19 +31,106 @@
...
+// total = 32
+
+struct __attribute__((packed)) RegisterInfo {
+  uint32_t id : REG_ID_WIDTH;
+  uint32_t kind : REG_KIND_WIDTH;
+  uint32_t count : REG_COUNT_WIDTH;
+
+  constexpr RegisterInfo(uint32_t regId, uint32_t regKind, uint32_t numRegs)
+      : id(regId), kind(regKind), count(numRegs) {}
+};
```
kupsch: Make the ctor interface for `RegisterInfo` use types that are force correct values (count is limited to 255, but that is enough for now). Also use `RegKind` for `kind`'s type to eliminate casts. Then add `static_asserts` to detect problems at compile time as this will allow run-time asserts to go away:

```suggestion
constexpr uint32_t REG_COUNT_WIDTH = 14;  // width in API restricted to 8
  
struct RegisterInfo {
  uint16_t id : REG_ID_WIDTH;
  RegKind kind : REG_KIND_WIDTH;
  uint32_t count : REG_COUNT_WIDTH;
      
  constexpr RegisterInfo(uint16_t regId, RegKind regKind, uint8_t numRegs)
      : id(regId), kind(regKind), count(numRegs) {
    static_assert(REG_ID_WIDTH >= sizeof(regId), "sizeof numRegs >= field width");
    static_assert(REG_COUNT_WIDTH >= sizeof(numRegs), "sizeof numRegs >= field width");
  } 
};

static_assert(sizeof(RegisterInfo) == sizeof(uint32_t), "RegisterInfo size error");
static_assert(REG_ID_WIDTH + REG_KIND_WIDTH + REG_COUNT_WIDTH == 8 * sizeof(uint32_t), 
    "RegisterInfo field not 32 total bits");
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,106 @@
...
+  Register(uint32_t id, RegKind kind, uint32_t count) {
+    assert(id >> REG_ID_WIDTH == 0 && "id is wider than than specified");
+    assert(count >> REG_COUNT_WIDTH == 0 && "count is wider than specified");
+    assert(count != 0 && "count must be non-zero");
+
+    uint32_t kindNum = static_cast<uint32_t>(kind);
+    assert(kindNum >> REG_KIND_WIDTH == 0 && "kind is out of range");
+
+    info = RegisterInfo(id, kindNum, count);
+  }
```
kupsch: You should always member initialize instead of assigning in the body.

With `RegisterInfo` having a better interface, we can use brace initialization to get narrowing warnings at compile time, the asserts can be removed since the parameters are guaranteed to be in range and it can be `constexpr` since the asserts are gone:

```suggestion
  constexpr Register(uint16_t id, RegKind kind, uint8_t count) : info{id, kind, count}  {}
```

common/src/dyn_register.h
kupsch: Use brace initialization:
```suggestion
constexpr Register Null_Register{};
```

common/src/dyn_register.h
kupsch: If you use brace initialization for constructing a `Register` the compiler will warn if the argument sizes cause narrowing.

common/src/dyn_register.h
```diff
@@ -31,19 +31,106 @@
...
+#include <stdint.h>
+#include <vector>
+
 namespace Dyninst {
-/* This needs to be an int since it is sometimes used to pass offsets
-   to the code generator (i.e. if-statement) - jkh 5/24/99 */
 
-/* a register number, e.g., [0..31]  */
-typedef unsigned int Register;
+enum RegKind : uint32_t { SCALAR, VECTOR, SCALAR_PREDICATE, UNDEFINED_KIND };
```
hainest:
```suggestion
enum class RegKind : uint32_t { SCALAR, VECTOR, SCALAR_PREDICATE, UNDEFINED_KIND };
```

These are really generic names, so let's not pollute the global namespace with them.

common/src/dyn_register.h
```diff
@@ -31,19 +31,106 @@
...
+  constexpr RegKind getKind() const { return static_cast<RegKind>(info.kind); }
+
+  constexpr uint32_t getCount() const { return info.count; }
+
+  constexpr uint32_t getValue() const { return value; }
+
+  constexpr operator uint32_t() const { return value; }
+
+  // If this is a contiguous register block, return individual registers in that block.
+  void getIndividualRegisters(std::vector<Register> &individualRegisters) const {
```
hainest:
```suggestion
  std::vector<Register> getIndividualRegisters() const {
```

common/src/dyn_register.h
```diff
@@ -31,19 +31,106 @@
...
+    uint32_t baseId = this->getId();
+    uint32_t numRegs = this->getCount();
+    uint32_t lastId = baseId + numRegs - 1;
+    RegKind kind = this->getKind();
+
+    assert(numRegs > 1 && "This must be a register block");
+    for (uint32_t id = baseId; id <= lastId; ++id) {
+      individualRegisters.push_back(Register(id, kind, 1));
```
hainest:
```suggestion
      individualRegisters.emplace_back(id, kind, 1);
```

kupsch:
I think the enum is appropriate here as this is an efficient mechanism to categorize a set of value, there is no behavior where polymorphism would be useful on these types (at least not now).

Ronak's id, is the raw register id value.

If we don't care about a Register being 32-bits in size, then the data in `Register` can be: 

```
uint32_t value;
RegKind kind;
uint8_t count;
```

and the constructors become:
```
constexpr Register(uint32_t v = -1) : value{v}, kind{UNKNOWN_KIND}, count{1} {}

constexpr Register(uint32_t v, RegKind k, uint8_t c) : value{v}, kind{k}, count{c}  {}
```

The union goes away, `RegisterInfo` and bit field sizes can go away, `getId` goes away (use `getValue`), the UB goes away.

The == and != operators, and hash (use `boost::hash_combine`) need to use all the fields.

**Would this make address all issues? It works for me. Moving to a more generic register type that can be used everywhere in Dyninst would be better, but that can wait for another day.**

The other changes @hainest suggested should also be made.

hainest:
> I think the enum is appropriate here as this is an efficient mechanism to categorize a set of value, there is no behavior where polymorphism would be useful on these types (at least not now).

A type like `ScalarRegister` isn't intended to be used polymorphically (I wish C++ had strong typedefs). Using the enum internally is stil fine.

```c++
class Register {
...
protected:
  enum class RegKind ...;
};

struct final ScalarRegister : public Register {
  ScalarRegister(uint32_t id, uint32_t numRegs) : Register(id, RegKind::SCALAR, numRegs) {}
};
```
Another option is "factories".

```c++
class Register {
public:
  static Register scalar(...);
  static Register vector(...);
...
private:
  enum class RegKind ...;
};
```
Not my favorite.


> Ronak's id, is the raw register id value.

I know. I'd still recommend something like `MachineID` for clarity.

> If we don't care about a Register being 32-bits in size

Is there a strong reason to keep it that size? It's adding a lot of complexity.

rochauha:
> Is there a strong reason to keep it that size? It's adding a lot of complexity.

I don't have a strong reason for keeping `Register` a it 32-bit type. So the below is also fine by me as long as we can pass a `Register` by value, because existing code passes it around by value.

```c++
uint32_t value;
RegKind kind;
uint8_t count;
```

Existing code also assigns integers to `Register`s. So that will also need to be accounted for.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: Done.

rochauha:
Addressed some comments. Rebased on master as well.

@hainest, @kupsch, @bbiiggppiigg, please let me know if there are any more comments.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
```diff
@@ -31,19 +31,144 @@
...
+    return Register(machId, RegKind::VECTOR, blockSize);
+  }
+
+  // Default is an invalid register.
+  constexpr Register()
+      : id(MachineId(static_cast<uint32_t>(-1))), kind(RegKind::UNKOWN_KIND), count(BlockSize(1)) {}
+
+  // This is only to make existing code work, and must go away in the future.
+  constexpr Register(uint32_t rawId)
+      : id(MachineId(static_cast<uint32_t>(rawId))), kind(RegKind::SCALAR), count(BlockSize(1)) {}
```
hainest:
```suggestion
      : id(MachineId(rawId)), kind(RegKind::SCALAR), count(BlockSize(1)) {}
```

`rawID` is already a `uint32_t`, so the cast isn't needed.

common/src/dyn_register.h
```diff
@@ -31,19 +31,144 @@
...
+    std::vector<Register> individualRegisters;
+
+    uint32_t baseId = this->getId();
+    uint32_t numRegs = this->getCount();
+    uint32_t lastId = baseId + numRegs - 1;
+    RegKind regKind = this->kind;
+
+    assert(numRegs > 1 && "This must be a register block");
+    for (uint32_t idNum = baseId; idNum <= lastId; ++idNum) {
+      individualRegisters.emplace_back(Register(MachineId(idNum), regKind, BlockSize(1)));
```
hainest:
```suggestion
      individualRegisters.emplace_back(MachineId(idNum), regKind, BlockSize(1));
```

`emplace_back` takes the constructore arguments directly so that you aren't forced to create a temporary object like with `push_back`.

common/src/dyn_register.h
```diff
@@ -31,19 +31,144 @@
...
+    }
+    return individualRegisters;
+  }
+
+  // Required for hashmap lookup
+  constexpr bool operator==(const Register &other) const {
```
hainest, line 117: I might be overlooking something, but couldn't the `RegKind` compare reduce to `this->kind == rhs.kind`?

common/src/dyn_register.h
```diff
@@ -31,19 +31,144 @@
...
+            (this->isVector() && other.isVector()) ||
+            (this->isScalarPredicate() && other.isScalarPredicate()) ||
+            (this->isUnkownKind() && other.isUnkownKind()));
+  }
+
+  constexpr bool operator!=(const Register &other) const { return !(*this == other); }
+
+  // Required for compatibility with integers.
+  // TODO: This must go away in the future.
+  constexpr bool operator==(const int &other) const {
```
hainest:
```suggestion
  constexpr bool operator==(int other) const {
```

It's not technically wrong, but idiomatic C++ passes primitive types by value.

---

While I'm thinking about it, the following should work without this overload:

```c++
Register r{3};
if(r == 3) {
  ...
}
```

This invokes one builtin conversion (`int` -> `uint32_t`) and one user-defined conversion (`uint32_t` -> `Register`). If that works, we could remove these overloads to make things easier for ourselves.

dyninstAPI/src/emit-x86.C
```diff
@@ -860,22 +860,22 @@ static void emitRex(bool is_64, Register* r, Register* x, Register* b, codeGen &
     // returning since we account for it in the rex prefix
     
     // "R" register - extension to ModRM reg field
-    if (r && *r & 0x08) {
+    if (r && r->getId() & 0x08) {
```
hainest: The implicit conversion to `uint32_t` should be invoked here. Were you getting compiler errors? I'm fine with leaving them as being more explicit using `getID`. That will make it easier to find their usage as the `Register` class evolves.

hainest (changes requested):
Just a couple minor items.

common/src/dyn_register.h
```diff
@@ -31,19 +31,144 @@
...
+    return individualRegisters;
+  }
+
+  // Required for hashmap lookup
+  constexpr bool operator==(const Register &other) const {
+    return id.getId() == other.getId() && count.getCount() == other.getCount() &&
+           ((this->isScalar() && other.isScalar()) ||
+            (this->isVector() && other.isVector()) ||
+            (this->isScalarPredicate() && other.isScalarPredicate()) ||
+            (this->isUnkownKind() && other.isUnkownKind()));
```
kupsch:
```suggestion
           this->kind == other.kind;
```

common/src/dyn_register.h
kupsch: The problem is that there is `operator==(const Register &other)` and `operator==(const unsigned &other)`. The first is a match with a conversion of `int -> Register`, the second with `int -> unsigned` making it ambiguous.

dyninstAPI/src/emit-x86.C
rochauha: I just checked, the implicit conversion does work. This change has trickled when I was fixing the compiler errors on my original variant of the Register type.

I think keeping explicit makes it clear that we are using an object. At some point the implicit conversion must go away.

common/src/dyn_register.h
rochauha: I am now passing the argument by value. 

Also, removing the `int` variant of `==` causes compiler errors at least on x86_64.

common/src/dyn_register.h
rochauha: Fixed it.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
rochauha: Done.

rochauha:
Addressed comments by @hainest.

common/src/dyn_register.h
```diff
@@ -31,19 +31,142 @@
...
+    boost::hash_combine(seed, reg.getId());
+    boost::hash_combine(seed, reg.getCount());
+
+    uint8_t kindVal = 3; // for RegKind::UNKOWN_KIND
+    if (reg.isScalar())
+      kindVal = 0;
+    else if (reg.isVector())
+      kindVal = 1;
+    else if (reg.isScalarPredicate())
+      kindVal = 2;
```
kupsch: Just use the value of kind directly that is the state of this object. Right now at most one of these functions can return true, but if that is not the case then first one true wins is not good:
```suggestion
    uint8_t kindVal = static_cast<uint8_t>(reg.Kind());
```

kupsch: You will also need to add a Kind() method.

common/src/dyn_register.h
```diff
@@ -31,19 +31,142 @@
...
+
+    boost::hash_combine(seed, kindVal);
```
kupsch: Ignore previous comment `hash_combine` can hash an enum directly so


```suggestion
    boost::hash_combine(seed, reg.Kind());
```

dyninstAPI/src/emit-x86.C
kupsch: undo the `*r` to `r->regId()` it just add noise to this PR, and it is not done throughout the code base so it is misleading. The only old use of Register that needs to change is the `r &= ...`.

common/src/dyn_register.h
rochauha: I did it this way because @hainest originally suggested to not allow the interface to return the kind enums. Is the consensus that we should add a `getKind()` method?

hainest: > I did it this way because @hainest originally suggested to not allow the interface to return the kind enums. Is the consensus that we should add a `getKind()` method?

It would make this code a lot cleaner, so let's go ahead.

kupsch: Each interface has its uses. Right now property test methods map are using the enumeration as exclusive flags. For instance on other architectures there are two set of scalar registers: integer and floating point. So it would reasonable to have a INTEGER_SCALAR and FLOATING_POINT_SCALAR kind, but the isScalar would be true for both. A `Kind()` method is useful when code needs to do something for every kind; this way you can use a switch statement and get a warning when another kind is added, which you don't get when you have a chain of if/else statements (the generated code is also likely more efficient when the number of cases grows past a certain amount.

common/src/dyn_register.h
rochauha: Done.

dyninstAPI/src/emit-x86.C
rochauha: Done.

common/src/dyn_register.h
rochauha: Done.

common/src/dyn_register.h
```diff
@@ -31,19 +31,135 @@
...
+  // TODO: This must go away in the future.
+  constexpr bool operator==(int other) const {
+    return id.getId() == static_cast<uint32_t>(other);
+  }
+
+  constexpr bool operator!=(int other) const { return !(*this == other); }
+
+  constexpr bool operator==(unsigned other) const {
+    return id.getId() == static_cast<uint32_t>(other);
+  }
```
kupsch, line 135: `operator==` should not return true based only on `id`. `kind` and `count` must also be match. Implement them using `operator==(const Register &)`

```suggestion
  constexpr bool operator==(int other) const {
    return *this == Register(other);
  }

  constexpr bool operator!=(int other) const { return !(*this == other); }

  constexpr bool operator==(unsigned other) const {
    return *this == Register(other);
  }
```

rochauha: Done.

rochauha:
Addressed comments by @kupsch.

common/src/dyn_register.h
```diff
@@ -31,19 +31,135 @@
...
+  // If this is a contiguous register block, return individual registers in that block.
+  std::vector<Register> getIndividualRegisters() const {
+    std::vector<Register> individualRegisters;
+
+    uint32_t baseId = this->getId();
+    uint32_t numRegs = this->getCount();
+    uint32_t lastId = baseId + numRegs - 1;
+    RegKind regKind = this->kind;
+
+    assert(numRegs > 1 && "This must be a register block");
```
kupsch: Why? If numRegs is 1 return the register, if it is 0 return an empty vector. Remove.
```suggestion
```

rochauha: numRegs should never be 0.

Also, if we allow numRegs = 1, it becomes possible to call `getIndividualRegisters()`, and then call it on the individual registers too.

rochauha: I am thinking of a block as an entity that contains individual registers, so it must have more than 1 registers, otherwise it is a single register.

kupsch: This Register is not a block, it can be a block. Why does any of that matter? You're making it unnecessarily restrictive; semantically the behavior is correct. It is like saying that 0 should not be an argument to `+` or 1 should not be an argument to `*` because in my mind it is _obviously a user error_.

Let's say that I have a register and I don't know its kind, but I want to get the set of actual registers, now I need to test the kind and do one thing it count is >1 and another it is not. Having an assert is unexpected behavior and Dyninst doesn't need more asserts for error handing.

hainest: >This Register is not a block, it can be a block.

We're (ab)using the `Register` class for this. I really feel like we need a `RegisterBlock` that is just a collection of `Register`s and whatever other properties a block has. I think that can be safely deferred without too much pain.

rochauha: I removed the assert.

rochauha:
Removed the assert as per @kupsch's suggestion.

common/src/dyn_register.h
```diff
@@ -31,19 +31,133 @@
...
+
+  static Register makeVectorRegister(OperandRegId regId, BlockSize blockSize) {
+    return Register(regId, RegKind::VECTOR, blockSize);
+  }
+
+  // Default is an invalid register.
+  constexpr Register()
+      : id(OperandRegId(static_cast<uint32_t>(-1))), kind(RegKind::UNKOWN_KIND), count(BlockSize(1)) {}
+
+  // This is only to make existing code work, and must go away in the future.
```
kupsch: The future is a bigger rewrite:
```suggestion
  // Support existing Register usage that initializes with uint32_t.
```

rochauha:
@bbiiggppiigg - any comments?

hainest:
@rochauha Bah, sorry. I forgot to check if this had conflicts with #2124 before merging.

bbiiggppiigg (approved):

rochauha:
Rebased on master.

PR #2113, opened 2026-01-18
Remove dead code in EmitterIA32::emitStoreOrigRegister

hainest:
The assert was added by 4a4b40a24 in 2009 without explanation.

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -392,18 +392,9 @@ void EmitterIA32::emitLoadOrigRegister(Address register_num, Register dest, code
 
 }
 
-void EmitterIA32::emitStoreOrigRegister(Address register_num, Register src, codeGen &gen) {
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
@hainest 
I don't think any version of emitStoreOrigRegister is used in the code base.
Its assert(0) in aarch64, amdgpu, power, ia32.
It does something in amd64, but I don't see it is called anywhere.

So maybe we should just remove this function (from emitter as well) once and for all?

hainest:
> @hainest I don't think any version of emitStoreOrigRegister is used in the code base. Its assert(0) in aarch64, amdgpu, power, ia32. It does something in amd64, but I don't see it is called anywhere.
> 
> So maybe we should just remove this function (from emitter as well) once and for all?

Yeah, it should go away. I don't want to disturb Angus' work, so I'll make a note to remove it complete once he's done.

hainest:
This is clearly never used, so I'm not going to run the usual test suite battery.

PR #2114, opened 2026-01-18
Clean up dyninstAPI AstNode classes

hainest:
This removes a bunch of dead code and simplifies several implementations.

### DIFF
--- a/dyninstAPI/src/ast.C
+++ b/dyninstAPI/src/ast.C
@@ -365,7 +365,6 @@ AstVariableNode::AstVariableNode(vector<AstNodePtr>&ast_wrappers, vector<pair<Dy
     ranges_(ranges), index(0)
 {
    children = ast_wrappers;
-   vector<AstNodePtr>::iterator i;
    assert(!children.empty());
 }
 


hainest:

hainest:
No regressions on 18 machines after latest updates.

dyninstAPI/src/ast.C
```diff
@@ -396,15 +404,16 @@ AstSequenceNode::AstSequenceNode(std::vector<AstNodePtr > &sequence) :
...
 AstVariableNode::AstVariableNode(vector<AstNodePtr>&ast_wrappers, vector<pair<Dyninst::Offset, Dyninst::Offset> > *ranges) :
-    ast_wrappers_(ast_wrappers), ranges_(ranges), index(0)
+    ranges_(ranges), index(0)
 {
+   children = ast_wrappers;
    vector<AstNodePtr>::iterator i;
-   assert(!ast_wrappers_.empty());
+   assert(!children.empty());
 
    for (i = ast_wrappers.begin(); i != ast_wrappers.end(); i++) {
```
bbiiggppiigg: Should here be changed to children as well?

hainest: It could be. Since `children` is just a copy of `ast_wrappers` now. However, I left it this way to indicate that this constructor is explicitly modifying its in/out parameter (something you should _NEVER_ do in a constructor).

hainest:
@bbiiggppiigg @kupsch A thousand pardons. I just found a few more items that should go in here. I'll convert this to a draft until I'm done working on these classes.

hainest:

hainest:
Uh, gcc failed to find the unused variable. Has anyone else seen gcc not doing a good job with `-Wunused-variable`?

bbiiggppiigg:
> Uh, gcc failed to find the unused variable. Has anyone else seen gcc not doing a good job with `-Wunused-variable`?

What do you mean? Is there a specific variable that should trigger unused variable check but didn't?
Or did gcc report a problem with unused variable but didn't specify which?

kupsch:
The compiler does warns about an unused variable, only if the variable is not used:

1. no data read or written (members or the object)
2. address not taken
3. not passed as a parameter
4. no methods called

Number 4 is what is preventing the warning here, as the iterator has a constructor that gets called. This happens even for synthesized constructors and destructors if a member is member initialized or has has a constructor. So only variables that are scalars, or class/struct whose members are not initialized.

For classes that implement RAII, no warning is correct, but for the majority of classes getting a warning would be useful. An attribute for ctors like ` [[constitutes_use]]` would be useful on classes that implement RAII and then the warning could be emitted for all other classes that are not used other than a ctor/dtor.

dyninstAPI/src/ast.h
```diff
@@ -676,17 +521,13 @@ class AstSequenceNode : public AstNode {
...
     virtual BPatch_type	  *checkType(BPatch_function* func = NULL);
-    virtual bool accessesParam();
-    virtual bool canBeKept() const;
 
-    virtual void getChildren(std::vector<AstNodePtr> &children);
+    virtual bool canBeKept() const {
+      // Theoretically we could keep the entire thing, but... not sure
+      // that's a terrific idea. For now, don't keep a sequence node around.
+        return false;
+    }
```
kupsch, line 529: This is the same as the base class's catBeKept. Is this here just for the comment?

hainest:
> Number 4 is what is preventing the warning here, as the iterator has a constructor that gets called.

Oh, I always forget about constructors as having side effects (gross). clang must have some builtin rules about STL types since it identified this as unused.

dyninstAPI/src/ast.h
hainest: Yeah. I assumed the comment meant something to someone at some point, so I kept it.

PR #2115, opened 2026-01-20
[AMDGPU] Begin using the new Register type for codegen

rochauha:
Use the new Register type for AMDGPU codegen.

### DIFF
--- a/dyninstAPI/src/AmdgpuPointHandler.C
+++ b/dyninstAPI/src/AmdgpuPointHandler.C
@@ -103,11 +103,10 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
 }
 
 bool AmdgpuGfx908PointHandler::isScalarRegAvailable(Register reg, const AmdgpuKernelDescriptor &kd) const {
-  // The register pair we want available should not be used in the kernel at all.
-  // Since we max out register usage, it should also not be a s100-101. So maximum usable SGPR is s99.
-  //
-  // This check is coarse. We eventually want to move away from maxing out register
-  // usage and this check will also go away.
+  // Since we max out register usage, the maximum SPGR is 101. AMDGPU GFX908
+  // aliases the VCC (opcode register ids 106 and 107) to highest allocated
+  // register pair (100 and 101 in this case), so the actual maximum available
+  // is 99.
 
   assert(reg.isScalar() && "reg must be scalar");
 


rochauha:
@hainest, @kupsch, @bbiiggppiigg - this PR is ready for review.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,12 +89,12 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
   return (kd.getCOMPUTE_PGM_RSRC1_GranulatedWavefrontSgprCount() != getMaxGranulatedWavefrontSgprCount());
 }
 
-bool AmdgpuGfx908PointHandler::isRegPairAvailable(Register reg, BPatch_function *function) {
-  assert(reg % 2 == 0 && "reg must be even");
-  assert(reg <= 101 && "reg must be a valid SGPR");
+bool AmdgpuGfx908PointHandler::isRegPairAvailable(Register regPair, BPatch_function *function) {
+  vector<Register> individualRegs = regPair.getIndividualRegisters();
+  assert(individualRegs.size() == 2);
```
kupsch: Since Register can represent multiple registers or a single register, make this function support any Register and change the name accordingly `isRegAvailable`. Remove the assert. remove the variables for the pair of registers. Test each element of the vector. If any of them are dead this is false otherwise true. The null_register (count 0) is therefore always true.

rochauha: Done.

common/src/arch-amdgpu.C
```diff
@@ -32,6 +32,380 @@
 
 namespace NS_amdgpu {
 
+namespace RegisterConstants {
```
hainest, line 35: This huge pile of declarations directs mirror what we see in the `MachRegister` declarations. This is the kind of duplication I was hoping to avoid. It's too late for this now, but we should endeavor to work to minimize this kind of code.

dyninstAPI/src/amdgpu-gfx908-details.h
```diff
@@ -38,111 +38,8 @@ using Register = Dyninst::Register;
...
-  SGPR_95,
-  SGPR_96,
-  SGPR_97,
-  SGPR_98,
-  SGPR_99,
-  SGPR_100,
-  SGPR_101,
-}; // ScalarRegs
+constexpr uint32_t MIN_SGPR_ID = 0;
+constexpr uint32_t MAX_SGPR_ID = 101;
```
hainest: These are very fragile since they have to be coordinated with the declarations/definitions of the `Register` objects in `arch-amdgpu.{h,C}`. I would recommend moving these values over to `arch-amdgpu.h` just so it's closer to those declarations.

dyninstAPI/src/AmdgpuKernelDescriptor.C
```diff
@@ -665,21 +665,21 @@ bool AmdgpuKernelDescriptor::supportsArchitectedFlatScratch() const {
 
 #undef GET_VALUE
 
-unsigned AmdgpuKernelDescriptor::getKernargPtrRegister() {
-  unsigned kernargPtrReg = 0;
+Register AmdgpuKernelDescriptor::getKernargPtrRegisterPair() {
```
hainest, line 668: It would be really nice if `registerSpace` could provide a list of available registers and then functions like this could just throw away the unusable ones instead of being required to use the ISA representation through integers. The current code likely works (has it been tested? can those tests be included in this PR?), but is so fragile and operates at a "spooky action at a distance" relative to the `Register` class.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
-bool AmdgpuGfx908PointHandler::isRegPairAvailable(Register reg, BPatch_function *function) {
-  assert(reg % 2 == 0 && "reg must be even");
-  assert(reg <= 101 && "reg must be a valid SGPR");
-
-  MachRegister machReg = convertRegID(reg, Arch_amdgpu_gfx908);
-  MachRegister nextMachReg = convertRegID(reg + 1, Arch_amdgpu_gfx908);
-
-  bool isMachRegLive = false, isNextMachRegLive = false;
+bool AmdgpuGfx908PointHandler::isRegAvailable(Register regPair, BPatch_function *function) {
+  vector<Register> individualRegs = regPair.getIndividualRegisters();
```
hainest:
```suggestion
  vector<Register> const& individualRegs = regPair.getIndividualRegisters();
```

It's likely not a big deal, but might as well avoid a copy via lifetime extension.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
-      livenessAnalyzer.query(location, LivenessAnalyzer::Before, nextMachReg, isNextMachRegLive);
+  // Check for liveness of the entire block
+  bool isBlockLive = true;
+  for (const auto &individualReg : individualRegs) {
+    MachRegister machReg = convertRegID(individualReg, Arch_amdgpu_gfx908);
+    bool isMachRegLive;
+    bool success = livenessAnalyzer.query(location, LivenessAnalyzer::Before, machReg, isMachRegLive);
+
+    if (!success) {
+      std::cerr << "AmdgpuPointHandler : liveness query on " << function->getMangledName()
```
hainest: Please use `ast_cerr`.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
+                << " failed.\n"
+                << "exiting...\n";
+      exit(1);
```
hainest:
```suggestion
      assert(0);
```

For as much as I hate it, it's at least consistent with the rest of Dyninst.

---

Ideally, we'd have a fallback mechanism like the "allocate everything". But that might not be useful in the long run.

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -137,8 +137,8 @@ void AmdgpuGfx908PointHandler::insertPrologueIfKernel(BPatch_function *function)
     exit(1);
   }
 
-  int reg = 94;
-  if (!isRegPairAvailable(reg, function)) {
+  Register regPair = Register::makeScalarRegister(OperandRegId(94), BlockSize(2));
```
hainest, line 150: Aside from the difference in the block size, I'm curious how this logically differs from `RegisterConstants::s94`. More specifically, why do we have the latter if we really want the former?

dyninstAPI/src/ast.C
```diff
@@ -2637,18 +2639,26 @@ bool AstAtomicOperationStmtNode::generateCode_phase2(codeGen &gen, bool noCost,
...
-  // TODO : Remove all hardcoded registers.
-  emitter->emitMoveRegToReg(94, 88, gen);
-  emitter->emitMoveRegToReg(95, 89, gen);
-  emitter->emitAddConstantToRegPair(88, (Address)offset->getOValue(), gen);
+  // TODO : allocate addrRegPair as a SGPR block of size 2.
+  // TODO : baseRegPair should be the cached value for a particular kernel.
+  Register addrRegPair(OperandRegId(88), RegKind::SCALAR, BlockSize(2));
+  Register baseRegPair(OperandRegId(94), RegKind::SCALAR, BlockSize(2));
+
+  std::vector<Register> addrRegs = addrRegPair.getIndividualRegisters();
```
hainest: Same comment as above about copies.

dyninstAPI/src/emit-amdgpu.C
```diff
@@ -40,26 +40,49 @@ using namespace AmdgpuGfx908;
...
+bool EmitterAmdgpuGfx908::isValidSgpr(Register reg) const {
+  return reg.isScalar() &&
+         reg.getCount() == 1 && reg.getId() >= MIN_SGPR_ID &&
+         reg.getId() <= MAX_SGPR_ID;
+}
+
+bool EmitterAmdgpuGfx908::isValidSgprBlock(Register regBlock) const {
+  auto firstRegId = regBlock.getId();
+  auto numRegs = regBlock.getCount();
+  auto lastRegId = firstRegId + numRegs - 1;
```
hainest, line 58: Should probably check that the block has more than 1 register in it here instead of in the return statement so that any future code updates can't accidentally use a bad register id.

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -33,11 +33,12 @@
 #include "dyninstAPI/src/registerSpace.h"
 #include "arch-amdgpu.h"
 #include "emit-amdgpu.h"
+#include "dyn_register.h"
 
 /************************************* Register Space **************************************/
 
 void registerSpace::initialize32() {
-  using namespace NS_amdgpu;
+  // using namespace NS_amdgpu;
```
hainest:
```suggestion
```

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -48,33 +49,35 @@ void registerSpace::initialize32() {
   // use for instrumentation are considered. This might need work later on.
 
   // SGPRs
-  for (unsigned idx = sgpr0; idx <= sgpr101; idx++) {
+  for (unsigned id = 0; id <= 101; id++) {
```
hainest:
```suggestion
  for (unsigned id = MIN_SGPR_ID; id <= MAX_SGPR_ID; id++) {
```

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -48,33 +49,35 @@ void registerSpace::initialize32() {
...
-    registers.push_back(new registerSlot(flat_scratch_lo, "flat_scratch_hi", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(xnack_mask_lo, "xnack_mask_lo", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(xnack_mask_lo, "xnack_mask_hi", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(vcc_lo, "vcc_lo", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(vcc_lo, "vcc_hi", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(exec_lo, "exec_lo", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(exec_lo, "exec_hi", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::flat_scratch_lo, "flat_scratch_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::flat_scratch_hi, "flat_scratch_hi", true, registerSlot::liveAlways, registerSlot::SPR));
```
hainest, line 73: The previous was `flat_scratch_lo` with a name `flat_scratch_hi`. Did this fix a bug in the previous code?

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -48,33 +49,35 @@ void registerSpace::initialize32() {
...
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::xnack_mask_lo, "xnack_mask_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::xnack_mask_hi, "xnack_mask_hi", true, registerSlot::liveAlways, registerSlot::SPR));
```
hainest, line 75: Same.

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -48,33 +49,35 @@ void registerSpace::initialize32() {
...
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::vcc_lo, "vcc_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::vcc_hi, "vcc_hi", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::exec_lo, "exec_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::exec_hi, "exec_hi", true, registerSlot::liveAlways, registerSlot::SPR));
```
hainest, line 79: Same.

dyninstAPI/src/registerSpace.h
```diff
@@ -502,7 +502,7 @@ class registerSpace {
     static unsigned VGPR(Dyninst::Register x) { return x; }
     static unsigned AGPR(Dyninst::Register x) { return x; }
 
-    int framePointer() { return sgpr33; }
+    int framePointer() { return Register::makeScalarRegister(OperandRegId(33), BlockSize(1)); }
```
hainest:
```suggestion
    int framePointer() { return RegisterConstants::s33; }
```

dyninstAPI/src/AmdgpuKernelDescriptor.C
rochauha: The current code works in my experiments. The test can't be included in the PR because it requires an AMDGPU ELF object, which must be extracted from the final x86 executable binary.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: The s94 used here is actually a register block of size 2 from s94, i.e it is 2 registers s94, and s95. RegisterConstants::s94 is a single definition for s94 register

RegisterConstants exist for mapping registers to MachRegisters, which is what liveness uses.

Other architectures also do the same, but they just have a number or an enum value that maps to MachRegisters.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: Done.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: Done.

dyninstAPI/src/inst-amdgpu.C
rochauha: Done.

dyninstAPI/src/inst-amdgpu.C
rochauha: Done.

dyninstAPI/src/inst-amdgpu.C
rochauha: I didn't see a bug, but I noticed that I made a mistake, so I fixed it here.

dyninstAPI/src/registerSpace.h
rochauha: Done.

rochauha:
Addressed comments by @hainest.

dyninstAPI/src/AmdgpuPointHandler.C
hainest: Yes, I know all of that. My concern here is that the keys in the conversion map have count=1, but this has count=2. `Register::operator==` considers the count when checking for equality. This means that `Register <-> MachRegister` conversion will fail for cases like this.

rochauha: For converting to `MachRegister`, we need to get individual registers in a register block and then convert. `isRegAvailable` does this.

rochauha: We can add a check or assert in `convertRegID` to enforce the requirement of count = 1.

rochauha:
@hainest, @kupsch, @bbiiggppiigg - any further comments?

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
-              << " failed.\n"
-              << "exiting...\n";
-    exit(1);
+    isBlockLive &= isMachRegLive;
   }
 
-  // The register pair must be dead.
-  return !isMachRegLive && !isNextMachRegLive;
+  // The block must be dead
+  return !isBlockLive;
```
kupsch: No need for `isBlockLive`. Just exit loop on first dead, no need to check the others. The final comment is wrong.

```suggestion
    if (isMachRegLive)
      return false;      // live register, block unavailable
  }

  return true;           // all dead, block available
```

rochauha:
@kupsch, @bbiiggppiigg - any comments?

dyninstAPI/src/ast.C
```diff
@@ -2637,18 +2639,26 @@ bool AstAtomicOperationStmtNode::generateCode_phase2(codeGen &gen, bool noCost,
...
   assert(emitter);
 
-  // TODO : Remove all hardcoded registers.
-  emitter->emitMoveRegToReg(94, 88, gen);
-  emitter->emitMoveRegToReg(95, 89, gen);
-  emitter->emitAddConstantToRegPair(88, (Address)offset->getOValue(), gen);
+  // TODO : allocate addrRegPair as a SGPR block of size 2.
+  // TODO : baseRegPair should be the cached value for a particular kernel.
+  Register addrRegPair(OperandRegId(88), RegKind::SCALAR, BlockSize(2));
+  Register baseRegPair(OperandRegId(94), RegKind::SCALAR, BlockSize(2));
```
kupsch, line 2645: What's up with `88`? Why 88 and not 90 or 92, is is just a randomly selected register? Is is assumed to be available?

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
-
-  MachRegister machReg = convertRegID(reg, Arch_amdgpu_gfx908);
-  MachRegister nextMachReg = convertRegID(reg + 1, Arch_amdgpu_gfx908);
-
-  bool isMachRegLive = false, isNextMachRegLive = false;
+bool AmdgpuGfx908PointHandler::isRegAvailable(Register regPair, BPatch_function *function) {
+  vector<Register> const &individualRegs = regPair.getIndividualRegisters();
 
   ParseAPI::Location location(ParseAPI::convert(function));
   LivenessAnalyzer livenessAnalyzer(Arch_amdgpu_gfx908, /* width = */ 8);
```
kupsch, line 112: I don't think this function does what you think it does. It looks like you want this function to return true if `regPair` is unused in `function`. We'll ignore for now this does not check called functions, but that is still a problem.

What this does is check if `regPair` is available (not live) at the first instruction of this function. At the first instruction of the function every register is dead except parameters and register used for global implicit state like a stack pointer, or ABI defined callee saved registers. In your use case you are checking (s94, s95). This function will return available for every function unless (s94, s95) are read before written (an uninitialized value). It won't detect a write to s94 followed by other read/writes. Have you tested this code on a binary that uses (s94, s95)?

You need to check that reg is dead everywhere in function (and eventually recursively that all functions called have this property too). If you expanded the registers for the kernel you know this is true and no need to use liveness. Otherwise not live at the exit from function should work (you will need to test and think about this).

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -89,31 +89,31 @@ bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) c
...
 
-bool AmdgpuGfx908PointHandler::isRegPairAvailable(Register reg, BPatch_function *function) {
-  assert(reg % 2 == 0 && "reg must be even");
-  assert(reg <= 101 && "reg must be a valid SGPR");
-
-  MachRegister machReg = convertRegID(reg, Arch_amdgpu_gfx908);
-  MachRegister nextMachReg = convertRegID(reg + 1, Arch_amdgpu_gfx908);
-
-  bool isMachRegLive = false, isNextMachRegLive = false;
+bool AmdgpuGfx908PointHandler::isRegAvailable(Register regPair, BPatch_function *function) {
```
kupsch: Not a pair

```suggestion
bool AmdgpuGfx908PointHandler::isRegAvailable(Register reg, BPatch_function *function) {
```

dyninstAPI/src/codegen-amdgpu.C
```diff
@@ -53,8 +53,9 @@ void insnCodeGen::generateBranch(codeGen &gen, Dyninst::Address from, Dyninst::A
   if (wordOffset >= INT16_MIN && wordOffset <= INT16_MAX) {
     emitter->emitShortJump(wordOffset, gen);
   } else {
-    // TODO: Right now hardcoding s90. But this needs to use 2 pairs of dead registers.
-    emitter->emitLongJump(90, from, to, gen);
+    // TODO: Right now hardcoding s80. But this needs to use 2 pairs of dead registers.
+    Register regBlock = Register::makeScalarRegister(OperandRegId(80), BlockSize(4));
```
kupsch, line 57: Same comment as s88 above. I know this is temporary, but why the apparent random register usage. Is there a reason not use consecutive registers from the highest register down to avoid conflict with the compiler allocating registers from the lowest down.

dyninstAPI/src/inst-amdgpu.C
```diff
@@ -48,33 +48,35 @@ void registerSpace::initialize32() {
...
-    registers.push_back(new registerSlot(exec_lo, "exec_lo", true, registerSlot::liveAlways, registerSlot::SPR));
-    registers.push_back(new registerSlot(exec_lo, "exec_hi", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::flat_scratch_lo, "flat_scratch_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::flat_scratch_hi, "flat_scratch_hi", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::xnack_mask_lo, "xnack_mask_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::xnack_mask_hi, "xnack_mask_hi", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::vcc_lo, "vcc_lo", true, registerSlot::liveAlways, registerSlot::SPR));
+    registers.push_back(new registerSlot(NS_amdgpu::RegisterConstants::vcc_hi, "vcc_hi", true, registerSlot::liveAlways, registerSlot::SPR));
```
kupsch, line 77: how is aliasing of VCC to highest allocated SGPR handled?

dyninstAPI/src/ast.C
rochauha: Randomly selected and assumed available. The next PR which is very small, makes this go away as shown here : https://github.com/dyninst/dyninst/blob/ronak/regblock-alloc/dyninstAPI/src/ast.C#L2644

dyninstAPI/src/inst-amdgpu.C
rochauha: We don't need to handle it explicitly now.

We max out register usage regardless of how many we use for instrumentation. 

When we start using liveness for registers, we may never reach the high limit (s100, s101).

Another approach can be to explicitly disallow s100, s101 for GPR allocation.

dyninstAPI/src/codegen-amdgpu.C
rochauha: It isn't entirely random. I am trying to leave some gap between the hardcoded registers to avoid any errors if they overlap if I changed any number somewhere.

This is work in progress, the next PR will address all hardcoded registers except s94, s95. The branch is here : https://github.com/dyninst/dyninst/tree/ronak/regblock-alloc

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: I see what you mean.

Additionally, we also need to check that the register does not appear in the GEN or KILL sets of any block.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: Done.

rochauha:
Addressed comments by @kupsch.

The `isRegAvailable` portion should be for a later PR. This PR is about switching to using register blocks for AMDGPU instead of raw numbers.

dyninstAPI/src/AmdgpuPointHandler.C
rochauha: Done.

dyninstAPI/src/inst-amdgpu.C
kupsch: Please explain why you don't think you need to handle it now? What is the consequence of not handling it? is that a good outcome? You need to think about this more.

dyninstAPI/src/codegen-amdgpu.C
kupsch: Please don't do this in the future. Make named constants, use the constants not hard coded magic numbers, you should have learned by now. Don't leave unexplained gaps because of some future change you might introduce. The choices should be logical to someone reading the code or there should be a comment explaining why it is not.

dyninstAPI/src/inst-amdgpu.C
rochauha: I meant we don't need to handle it in this PR.

As I understand, the AMDGPU port now going through incremental development because we don't want a huge PR like PR #1987. So I can't address everything in a single PR.

We max out SGPR allocation for kernels when we rewrite the kernel descriptor. So vcc should map to s101, s102. The vcc aliasing can be easily handled if we disallow allocating s101, s102 for use in instrumentation. I can add the check in the next register block allocation PR.

dyninstAPI/src/codegen-amdgpu.C
rochauha: Got it.

rochauha:
@bbiiggppiigg, @kupsch - any further comments?

rochauha:
@kupsch, @hainest, @bbiiggppiigg, I have addressed comments regarding register availability. Please let me know if you have further comments.

hainest (approved):
LGTM

dyninstAPI/src/AmdgpuPointHandler.C
```diff
@@ -85,39 +85,46 @@ uint32_t AmdgpuGfx908PointHandler::getMaxGranulatedWavefrontSgprCount() const {
...
+
 bool AmdgpuGfx908PointHandler::canInstrument(const AmdgpuKernelDescriptor &kd) const {
   return (kd.getCOMPUTE_PGM_RSRC1_GranulatedWavefrontSgprCount() != getMaxGranulatedWavefrontSgprCount());
 }
 
-bool AmdgpuGfx908PointHandler::isRegAvailable(Register reg, BPatch_function *function) {
-  vector<Register> const &individualRegs = reg.getIndividualRegisters();
+bool AmdgpuGfx908PointHandler::isScalarRegAvailable(Register reg, const AmdgpuKernelDescriptor &kd) const {
+  // The register pair we want available should not be used in the kernel at all.
+  // Since we max out register usage, it should also not be a s100-101. So maximum usable SGPR is s99.
```
kupsch: Change this comment from

```
  // Since we max out register usage, it should also not be a s100-101. So maximum usable SGPR is s99.
```
to explain why it is two registers less:
```
  // Since we max out register usage, the maximum SPGR is 101.  AMD GPUs alias
  // the VCC (opcode register ids 106 and 107) to highest allocated register
  // pair (100 and 101 in this case), so the actual maximum available is 99.
```
Also in general for comments that are a block try to keep each line less than 80 characters so they won't wrap on smaller screens and github reviews. This isn't a hard rule if wider improves the readability such as for a table then it is OK, but that is not the case for plain sentences. No need to fix this everywhere, just something to keep in mind for the future.

rochauha: Done.

PR #2116, opened 2026-01-21
Remove AstNode::labelNode(std::string)

hainest:
This should have been removed by eac8cf2c8.

This should be merged after #2114.


hainest:
Merged into #2114.

PR #2119, opened 2026-01-23
Fix placement of newly created loadable sections

bbiiggppiigg:
Previously, Dyninst made the assumption that for the last PT_LOAD segment, the last section in it will have its addr + offset exactly matching the end of the segment (segment addr + offset).

However, we have seen binaries where such assumptions no longer holds.

This PR simply moves the logic of inserting the loadable sections right after the last section of the last data segment, to the last section, which should be permitted by the ELF standard.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -482,7 +482,6 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
         ".dynsym", /*".dynstr",*/ ".rela.dyn", ".rela.plt", ".dynamic", ".symtab"};
     std::unordered_map<string, pair<unsigned, unsigned>> dataLinkInfo;
 
-    bool createdLoadableSections = false;
     unsigned scncount;
     unsigned sectionNumber = 0;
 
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
This + #2087 Should fix #2081 .

hainest:
I ran this on one machine of each support architecture and didn't find any test regressions. I'd say it's safe to go ahead and clean up the unused variables.

bbiiggppiigg:
@hainest 

It seems that one of the boolean variable BSSExpandFlag is always false, and by removing checks on it I can remove a bunch of dead code as well, do you think that should be part of this PR as well?

hainest:
> It seems that one of the boolean variable BSSExpandFlag is always false, and by removing checks on it I can remove a bunch of dead code as well, do you think that should be part of this PR as well?

I think that would probably be better as a separate PR. It looks like its usage was removed when BGQ support was added (034a5316a).

bbiiggppiigg:
Rebased on master.

@hainest @kupsch  The PR should be ready for review now.

hainest:
I'm seeing regressions on three (two x86, one ppc) of the 18 machines I test on. I'll send you some details once I get a reproducer.

hainest:
The rewritten binary (binaries/rewritten_dyninst_cxx_group_test.dyn_g++_64_none_none_234974949) is missing several sections, including the dynamic section. The particular test that was failing is `./test_driver -64 -rewriter -g++ -none -dynamiclink -nonpic -test test5_1`.

[PR2119.tar.gz](https://github.com/user-attachments/files/24945878/PR2119.tar.gz)

hainest:
The duplicated conditional is definitely a bug, but I don't know what it was _supposed_ to be.

```c++
        if (isStaticBinary && (strcmp(name, ".rela.plt") == 0)) {
            string newName = ".o";
            newName.append(name, 2, strlen(name));
            renameSection(name, newName, false);
        }

        if (isStaticBinary && (strcmp(name, ".rela.plt") == 0)) {
            string newName = ".o";
            newName.append(name, 2, strlen(name));
            renameSection(name, newName, false);
            // Clear the PLT type; use PROGBITS
            newshdr->sh_type = SHT_PROGBITS;
        }
```

bbiiggppiigg:
> The duplicated conditional is definitely a bug, but I don't know what it was _supposed_ to be.
> 
> ```c++
>         if (isStaticBinary && (strcmp(name, ".rela.plt") == 0)) {
>             string newName = ".o";
>             newName.append(name, 2, strlen(name));
>             renameSection(name, newName, false);
>         }
> 
>         if (isStaticBinary && (strcmp(name, ".rela.plt") == 0)) {
>             string newName = ".o";
>             newName.append(name, 2, strlen(name));
>             renameSection(name, newName, false);
>             // Clear the PLT type; use PROGBITS
>             newshdr->sh_type = SHT_PROGBITS;
>         }
> ```

So the first if will be entered, .rela.plt will be renamed to .oela.plt.
The second if will still be entered since we didnt change the value of name.
The renameSection will be called but have no effect since the section is already renamed.
So the net effect is the same as keeping the second if I think.

hainest:
@bbiiggppiigg If/when you have a moment, could you rebase this so I can test it?

bbiiggppiigg:
> @bbiiggppiigg If/when you have a moment, could you rebase this so I can test it?

rebased on master.

hainest:
> > @bbiiggppiigg If/when you have a moment, could you rebase this so I can test it?
> 
> rebased on master.

Thanks! I'll put it in the queue.

hainest:
This has build errors. Go ahead and kill the CI jobs.

bbiiggppiigg:
@hainest My apologies. Should be fixed now.

hainest:
@bbiiggppiigg I think github is limiting our CI jobs, so I cancelled the old runs queued for this PR so that only the newest ones will run.

hainest:
Not surprisingly, I'm seeing the same errors as before. Maybe we should have a chat about how to proceed here? There is just so much going on in the emitter that it might be easier to do a heavy refactor than try to chase the bugs.

PR #2121, opened 2026-01-28
Remove i386 MMX detection for instrumentation

hainest:
This wouldn't compile for 32-bit Linux because 'xmmCapable' calls 'cpuidCall' which is only compiled on Windows. The Win64 version reads off the end of 'result', so that likely never worked. At this point, I'm quite confident that anyone rewriting Win32 binaries is likely doing so on i386 that supports MMX. This wouldn't work on non-x86 Win32 platforms because of the inline assembly.


kupsch (approved):

hainest:
No test regressions on 3 machines (one of each arch). This code was always unused, so I didn't run the full battery of tests.

PR #2122, opened 2026-01-28
Remove unused func_instance::setReturnValue(int)

hainest:
Its usage was removed by 7a737b6a09 in 2010.


hainest:
No regressions on 3 machines (one of each supported architecture). I didn't run the full battery since this code is unused.

PR #2123, opened 2026-01-28
Fix Block::remove_source/target and == operators

bbiiggppiigg:
Fix the == operator for parseAPI::Block and Edge
Make sure remove_source/remove_target remove potentially duplicated edges.

### DIFF
--- a/parseAPI/src/Block.C
+++ b/parseAPI/src/Block.C
@@ -285,9 +285,6 @@ Block::getInsn(Offset a) const {
 
 
 bool Block::operator==(const Block &rhs) const {
-    // prevent dead lock when compariing a block to itself
-    if(this == &rhs) return true;
-    // prevent deadlock by locking both blocks at the same time
     // All sinks are equal
     if(_start == std::numeric_limits<Address>::max()) {
         return rhs._start == _start;


parseAPI/src/Block.C
```diff
@@ -285,6 +285,7 @@ Block::getInsn(Offset a) const {
 
 
 bool Block::operator==(const Block &rhs) const {
+    if(this == &rhs) return true;
```
hainest: It might be good to add a comment here to say something like "Comparing a block to itself results in deadlock from the mutexes below".

hainest:
This is causing the testsuite to only run a few tests (~8) on some x86 machines at Frank. Have you checked this on coriander?

bbiiggppiigg:
> This is causing the testsuite to only run a few tests (~8) on some x86 machines at Frank. Have you checked this on coriander?

I haven't. I'm going to try this now.

bbiiggppiigg:
It seems like our new implementation of the removeTarget and removeSource is the problem.
When I revert those changes into  _trglist.erase(e) and _srclist.erase(e) everything seems to work.
I'm not sure what is done wrong.

bbiiggppiigg:
I believe the == operator on Edges used in removeTarget and removeSource is causing deadlocks.

bbiiggppiigg:
@hainest 
I belive this is now passing for most if not all the tests.
On one test, I get 

==============================

"pure virtual method called"
terminate called without an active exception
test2_8                    gcc    none 64  create   NA      dynamic PIC     CRASHED (Group Teardown)

hainest:
I'm not sure I like where this has gone. Adding a mutex and using recursive locks seems like we're needing to more work than I would expect. That we're also calling an unimplemented pure virtual function is a big red flag. Do you have a backtrace for the uncaught exception crash?

bbiiggppiigg:
@hainest 

> I'm not sure I like where this has gone. Adding a mutex and using recursive locks seems like we're needing to more work than I would expect. 

If more than on thread that triggers the Edge ==, which can trigger the Block ==, then deadlock can happen depending on the order of aquiring lock. The load on atomic source can also be a cause of the deadlock. I came up with this fix with the help of Gemini but I think the fix make sense.

> That we're also calling an unimplemented pure virtual function is a big red flag. Do you have a backtrace for the uncaught exception crash?

When I rerun the testsuite I see no failures at all. So I don't know If that is a bug introduced by this PR, or how can I reliably get a backtrace for it.

parseAPI/src/Block.C
```diff
@@ -285,21 +285,20 @@ Block::getInsn(Offset a) const {
...
 
 bool Block::operator==(const Block &rhs) const {
-    boost::lock_guard<const Block> g1(*this);
-    boost::lock_guard<const Block> g2(rhs);
+    // prevent dead lock when compariing a block to itself
+    if(this == &rhs) return true;
+    // prevent deadlock by locking both blocks at the same time
+    boost::lock(this->_mutex,rhs._mutex);
+    boost::lock_guard<boost::recursive_mutex> g1(this->_mutex,boost::adopt_lock);
+    boost::lock_guard<boost::recursive_mutex> g2(rhs._mutex,boost::adopt_lock);
```
kupsch: replace `lock` with `scoped_lock` it does the RAII making `lock_guard` unnecessary.

```suggestion
    boost::scoped_lock(this->_mutex,rhs._mutex);
```

bbiiggppiigg: > replace `lock` with `scoped_lock` it does the RAII making `lock_guard` unnecessary.

/home/wuxx1279/dyninsts/src/dyn2081-1/parseAPI/src/Block.C:291:12: error: 'scoped_lock' is not a member of 'boost'; did you mean 'shared_lock'?
  291 |     boost::scoped_lock(this->_mutex,rhs._mutex);

Does this mean I'm missing header ?
I tried including boost/thread/locks.hpp but it doesn't seem to help

kupsch: In C++ 17 or later there is std::scoped_lock. Well there used to be a scoped_lock in boost, that didn't do the same thing and doesn't seem to exist anymore, so ignore this.

hainest:
`_obj`, `_region`, and `_start` are class invariants. `_end` is only updated via `updateEnd`, but it's not guarded by a lock. Given all of this, I would suggest we remove the locks entirely.

>I came up with this fix with the help of Gemini but I think the fix make sense.

Unless and until a chat box gains the ability to say "I don't know", I strongly recommend leaving them in the garbage where they belong.

bbiiggppiigg:
> `_obj`, `_region`, and `_start` are class invariants. `_end` is only updated via `updateEnd`, but it's not guarded by a lock. Given all of this, I would suggest we remove the locks entirely.

Or maybe guarding updateEnd as well?
I felt like we should schedule a call over this PR

hainest:
> > `_obj`, `_region`, and `_start` are class invariants. `_end` is only updated via `updateEnd`, but it's not guarded by a lock. Given all of this, I would suggest we remove the locks entirely.
> 
> Or maybe guarding updateEnd as well?

I meant remove the locks in `operator==`. Although it is a good question as to whether `updateEnd` should be guarded.

> I felt like we should schedule a call over this PR

Surely. I'm free whenver.

hainest:
No regressions on 17 machines (I didn't run on coriander since you did).

PR #2124, opened 2026-01-29
Separate architecture-specific baseTramp implementations

hainest:
This is a first step to making `baseTramp` polymorphic. There is still code in shouldRegenBaseTramp that needs to be refactored, but that will have to wait until a real class hierarchy can be created.

### DIFF
--- a/dyninstAPI/src/trampolines/baseTramp-riscv64.h
+++ b/dyninstAPI/src/trampolines/baseTramp-riscv64.h
@@ -36,12 +36,12 @@
 
 class baseTramp_riscv64 final : public baseTramp {
 public:
-  bool generateSaves(codeGen &gen, registerSpace *) override {
+  bool generateSaves(codeGen &, registerSpace *) override {
     inst_printf("baseTramp codegen isn't implemented on RISCV64");
     return false;
   }
   
-  bool generateRestores(codeGen &gen, registerSpace *) override {
+  bool generateRestores(codeGen &, registerSpace *) override {
     inst_printf("baseTramp codegen isn't implemented on RISCV64");
     return false;
   }


hainest:

dyninstAPI/src/trampolines/baseTramp.C
kupsch: While you're here how about default member initializing the members and making the ctor default?

dyninstAPI/src/trampolines/baseTramp.C
```diff
@@ -76,8 +80,28 @@ baseTramp::~baseTramp()
...
+#ifdef DYNINST_CODEGEN_ARCH_X86_64
+  return new baseTramp_x86();
+#elif DYNINST_CODEGEN_ARCH_I386
+  return new baseTramp_x86();
+#elif DYNINST_CODEGEN_ARCH_AARCH64
+  return new baseTramp_aarch64();
+#elif DYNINST_CODEGEN_ARCH_POWER
+  return new baseTramp_ppc();
+#elif DYNINST_CODEGEN_ARCH_RISCV64
+  return new baseTramp();
```
kupsch: I would be good to put a comment here that this is temporary and RISCV64 baseTramps need to implemented when RISCV codegen is merged.

hainest: I just noticed that #2086 doesn't add a baseTramp class for riscv. I'll add a stub class here with error messages (asserts?...).

dyninstAPI/src/trampolines/baseTramp.C
```diff
@@ -77,7 +78,11 @@ baseTramp::~baseTramp()
 }
 
 baseTramp *baseTramp::create() {
-  return new baseTramp();
+#ifdef DYNINST_CODEGEN_ARCH_X86_64
```
bbiiggppiigg, line 85: Both case use the same baseTramp?

hainest: They do for now. Once these classes can be turned into an actual hiearchy, the x86_64 and i386 code will get split apart. They share a lot of code, so it will require a bit of work.

bbiiggppiigg (approved):

PR #2125, opened 2026-01-29
Split up dyninstAPI AST node classes

hainest:
This makes them much easier to reason about and to see where the architecture-specific parts are.

### DIFF
--- a/dyninstAPI/src/linux-aarch64.C
+++ b/dyninstAPI/src/linux-aarch64.C
@@ -46,8 +46,7 @@
 #include "common/src/linuxHeaders.h"
 
 using codeGenASTPtr = Dyninst::DyninstAPI::codeGenASTPtr;
-
-namespace NullNode = Dyninst::DyninstAPI::NullNode;
+using nullAST = Dyninst::DyninstAPI::nullAST;
 
 
 // FIXME: HOST+CODEGEN
@@ -110,7 +109,7 @@ bool PCProcess::getOPDFunctionAddr(Dyninst::Address &) {
 
 codeGenASTPtr PCProcess::createUnprotectStackAST() {
     // This is not necessary on power
-    return NullNode::create();
+    return nullAST::create();
 }
 
 bool Frame::setPC(Dyninst::Address newpc) {
--- a/dyninstAPI/src/linux-power.C
+++ b/dyninstAPI/src/linux-power.C
@@ -46,8 +46,7 @@
 #include "common/src/linuxHeaders.h"
 
 using codeGenASTPtr = Dyninst::DyninstAPI::codeGenASTPtr;
-
-namespace NullNode = Dyninst::DyninstAPI::NullNode;
+using nullAST = Dyninst::DyninstAPI::nullAST;
 
 
 // FIXME: HOST+CODEGEN
@@ -121,7 +120,7 @@ bool PCProcess::getOPDFunctionAddr(Dyninst::Address &addr) {
 
 codeGenASTPtr PCProcess::createUnprotectStackAST() {
     // This is not necessary on power
-    return NullNode::create();
+    return nullAST::create();
 }
 
 bool Frame::setPC(Dyninst::Address newpc) {


hainest:
No test regressions on 18 machines after rebase.

kupsch:
Should all the AST names be in a namespace (at least Dyninst)? The previous version was not in a namespace, so leaving it as wouldn't make things worse.

hainest:
> Should all the AST names be in a namespace (at least Dyninst)? The previous version was not in a namespace, so leaving it as wouldn't make things worse.

Currently, none of dyninstAPI is in a namespace. Since these are internal classes, I was thinking about creating `DyninstAPI` or `Dyninst::DyninstAPI` and then putting them in that. I don't have strong feelings about the name.

---

Speaking of bikeshedding names, I _really_ don't like the stack node names, but might be a discussion for another day. The namespaces and names of the helper functions I added are also not great. For example, `AstNullNode::create()` vs `StackInsertNode::canary()` is at least a bit more descriptive. Thoughts? @bbiiggppiigg?

I introduced the `AstStackNode` type to share the implementation of `allocateCanaryRegister` between `AstStackInsertNode` and `AstStackRemoveNode`. `AstStackNode` isn't used polymorphically, so I didn't make `AstStackGenericNode` inherit from it since it doesn't do canary registers. However, if someone ever did something like `void foo(AstStackNode *)`, it wouldn't work for `AstStackGenericNodes`. I'm not sure if that's something to worry about or not.

kupsch:
Using the global namespace isn't great. Dyninst::DyninstAPI sounds good.

hainest:
@kupsch I wrapped them in `Dyninst::DyninstAPI`. There are some ugly situations like in regTracker where there is a required forward declaration of `AstNode`, but I tried to make everything as uniform as I could.

dyninstAPI/src/ASTs/AstCallNode.h
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  inline AstNodePtr call(func_instance *func, std::vector<AstNodePtr> &args) {
+    if(!func) {
+      return {};
+    }
+    return boost::make_shared<AstCallNode>(func, args);
+  }
+
+  inline AstNodePtr replace(func_instance *func) {
```
bbiiggppiigg: What does this function do exactly?

dyninstAPI/src/ASTs/AstScrambleRegistersNode.h
```diff
@@ -0,0 +1,59 @@
...
+public:
+  bool usesAppRegister() const override {
+    return true;
+  }
+
+private:
+  bool generateCode_phase2(codeGen &gen, bool, Dyninst::Address &, Dyninst::Register &) override;
+};
+
+namespace RegisterNode {
```
bbiiggppiigg: The mismatch between namespace name and node name might make it hard for ppl to find the right namespace by node or find the right node by namespace.

dyninstAPI/src/ASTs/AstThreadIndexNode.h
```diff
@@ -0,0 +1,65 @@
...
+private:
+  bool generateCode_phase2(codeGen &, bool, Dyninst::Address &, Dyninst::Register &) override {
+    // No codegen needed
+    return true;
+  }
+
+  static AstThreadIndexNode node;
+};
+
+namespace ThreadNode {
```
bbiiggppiigg: Shouldn't this be called threadIndexNode?

dyninstAPI/src/ASTs/AstCallNode.C
```diff
@@ -141,20 +141,6 @@ BPatch_type *AstCallNode::checkType(BPatch_function *func) {
   return ret;
 }
 
-bool AstCallNode::canBeKept() const {
```
bbiiggppiigg: Is it correct to remove this?

dyninstAPI/src/ASTs/AstCallNode.h
hainest: That's probably a terrible name for it. I based it on the comment from the previous implementation:

```c++
static AstNodePtr funcCallNode(func_instance *func); // Special case for function call replacement.
```

dyninstAPI/src/ASTs/AstScrambleRegistersNode.h
hainest: Yeah, this has the same ugliness as `NullNode::create`. I'm not really sure what the best approach is. All of these AST classes have terrible names.

dyninstAPI/src/ASTs/AstThreadIndexNode.h
hainest: This was an attempt to avoid another `NullNode::create`.

dyninstAPI/src/ASTs/AstCallNode.C
hainest: It is. I left a comment in the commit note: https://github.com/dyninst/dyninst/pull/2125/changes/60837ccb77ac0a26f10270e01ae8c7e381ea3374.

dyninstAPI/src/ASTs/AstActualAddrNode.C
```diff
@@ -0,0 +1,33 @@
...
+}
+
+namespace AddressNode {
+
+  AstNodePtr actual() {
+    static auto node = boost::make_shared<AstActualAddrNode>();
+    return node;
+  }
+
+}
```
kupsch: Just curious as to why you choose to create a new namespace `NullNode` with a `create` function instead of `static AstNullNode::create`.

hainest: I wanted to move everything into uniform namespaces instead of having static `create`. I'm not satisfied with the names. I'll take a harder look at this to see if I can come up with a better convention.

hainest:
@kupsch @bbiiggppiigg 

I'm running the testsuite now, but I don't anticipate any further issues.

The latest updates:

1. The classes have been renamed from something like `AstOperatorNode` to `operatorAST` to reflect the naming conventions used in `InstructionAPI` and `DataflowAPI`.
2. They have a `Ptr` typedef like the other AST classes.
3. The 'creator' methods return a pointer to the concrete type instead of a generic `codeGenAST` so they can possibly be used more effectively in the future.

For the creator methods, I put them in the class when each creator returns an object of the class type but with a different internal configuration (e.g., `operatorAST::plus`), but kept them in a namespace when they are used to create a specific AST that has a different type (e.g., `AddressAST::actual`). I did this to minimize the number of overly-generic `create` members (like with `nullAST::create`) that don't help to explain how they are different. I suppose they could be split into `originalAddressAST::create`, if that would be more consistent. I don't have strong opinions.

I do not like the spelling of a lot of these AST types- in particular, `genericStackAST`. Maybe it doesn't matter.

PR #2126, opened 2026-01-30
Remove unused variables in emitElf

bbiiggppiigg:


hainest:
I think we can also remove `exceptions` in `driver`.

hainest:
And `dynsym_info` and `oldIndexNameMapping`.

---

Could you also remove the several C-style casts to `std::string`? I'm surprised they even work.

`renameSection((string) name, newName, false);` -> `renameSection(name, newName, false);`

symtabAPI/src/emitElf.C
```diff
@@ -853,7 +804,7 @@ void emitElf<ElfTypes>::createNewPhdrRegion(std::unordered_map<std::string, unsi
 
 
 template<class ElfTypes>
-void emitElf<ElfTypes>::fixPhdrs(unsigned &extraAlignSize) {
+void emitElf<ElfTypes>::fixPhdrs(unsigned &) {
```
kupsch: instead of just removing the parameter name, remove the parameter since it is not public:

```suggestion
void emitElf<ElfTypes>::fixPhdrs() {
```

kupsch (approved):
looks correct. one small change.

hainest:
No regressios on 18 machines, including the two that saw failures with #2119.

hainest (approved):

PR #2128, opened 2026-01-30
Fix preserving ELF .rel(a).plt sections

bbiiggppiigg:
Current implementation only clear the section header type for .rela.plt.
This PR handles the case for .rel.plt as well.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -602,19 +602,15 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
             renameSection(name, newName, false);
         }
 
-        if (isStaticBinary && (strcmp(name, ".rela.plt") == 0)) {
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
One failure on corriander.

pure virtual method called
terminate called without an active exception
test1_12                   g++    none 64  create   NA      dynamic PIC     CRASHED (Group Teardown)

hainest:
> One failure on corriander.
> 
> pure virtual method called terminate called without an active exception *** Child terminated abnormally via signal 6. test1_12 g++ none 64 create NA dynamic PIC CRASHED (Group Teardown)

Is it consistently failing?

bbiiggppiigg:
> > One failure on corriander.
> > pure virtual method called terminate called without an active exception *** Child terminated abnormally via signal 6. test1_12 g++ none 64 create NA dynamic PIC CRASHED (Group Teardown)
> 
> Is it consistently failing?

No. Second run only give me test_thread1 crash.

hainest:
Could you change the title to something like `Fix preserving ELF .rel(a).pl sections`?

PR #2130, opened 2026-01-31
Fix docker builds by adding missing libicu for peparse

wsxarcher:
Since 3 weeks ago libicu is a required dependency of peparse.
https://github.com/trailofbits/pe-parse/commit/eb6da5021c17935d3d8f80d118e0a7ef6b1f1264
Fix docker builds which currently is broken by adding the missing transative dependency.

### DIFF
--- a/docker/Dockerfile.fedora
+++ b/docker/Dockerfile.fedora
@@ -28,6 +28,7 @@ RUN yum update -y &&   \
       g++              \
       git              \
       glibc-static     \
+      libicu-devel     \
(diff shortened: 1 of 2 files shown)


hainest:
Thanks for the update! We don't see a lot of use for Dyninst on Windows binaries. If possible, could you describe a little bit about how you use Dyninst?

wsxarcher:
Not using PEs, but the Docker build is currently always broken because parsepe is always included

PR #2131, opened 2026-01-31
Add proccontrol support for threads spawned before reaching main

hainest:
Use a synchronized breakpoint for 'main'
    
    If a thread is spawned before reaching main (e.g., in a .ctor entry),
    proccontrol does not observe events from it and so assumes it is safe
    to remove the breakpoint at 'main'. It is unclear at this time why no
    events are observed. Particularly becuase there is a check that _should_
    catch this case:
    
      BPatch::processCreate
              V
      BPatch_process::BPatch_process
              V
      PCProcess::createProcess
              V
      Process::createProcess
              V
      int_process::createProcess
              V
      PCProcess::bootstrapProcess()
              V
      assert( pcProc_->allThreadsStopped() );

Ignore events for destroyed threads launched before main
    
    If a thread is launched and destroyed before main is reached, there
    could be events in the queue that belong to that dead thread. We should
    ignore them. This seems at odds with proccontrol's failure to detect
    that there are threads alive before main is reached, but there is a
    user-reported crash that is fixed by these changes.

@kcossett-amd This updated fix is working for my using gcc. Could you test this in your environment with amdclang?

### DIFF
--- a/dyninstAPI/src/pcEventHandler.C
+++ b/dyninstAPI/src/pcEventHandler.C
@@ -353,6 +353,15 @@ bool PCEventHandler::handleThreadCreate(EventNewThread::const_ptr ev, PCProcess 
   proccontrol_printf("%s[%d]: entering handleThreadCreate for %d/%d\n",
 		     FILE__, __LINE__, evProc->getPid(), ev->getLWP());
 
+    // Check if the thread still exists
+#if defined(os_linux) || defined(os_freebsd)
+    if (!ev->getNewThread()) {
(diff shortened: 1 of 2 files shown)


kcossett-amd:
Fix works in my environment with both `amdclang` and `amdflang`. Thank you

hainest:
@kcossett-amd Do you have a preferred email for co-authored commits?

kcossett-amd:
@hainest You can use Kian.Cossettini@amd.com

hainest:
No tests regressions on 18 machines.

bbiiggppiigg (approved):

PR #2133, opened 2026-02-01
Consolidate implementations of BinaryEdit::doStaticBinarySpecialCases

hainest:
Each platform was doing the same thing, so there's no reason for
separate implementations. The 'irel' case will be ignored pretty
much everywhere (see comments), even though it was previously
only on x86.


hainest:
Needs #2159.

hainest:

PR #2134, opened 2026-02-01
Consolidate mapped_object ctor/dtor lookup

hainest:
All of the architectures do the same thing, so there's no need for separate implementations. There's no need to assert for unimplemented architectures.

### DIFF
--- a/dyninstAPI/src/mapped_object.C
+++ b/dyninstAPI/src/mapped_object.C
@@ -2013,16 +2013,8 @@ string mapped_object::fileName() const {
   
 }
 
-func_instance *mapped_object::findGlobalConstructorFunc(const std::string &ctorHandler) {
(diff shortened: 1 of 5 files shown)


hainest:
No test regressions on 16 machines.

hainest:
> It seems like even the two functions are doing the exact samething, the only difference is what their callers decide to pass to it.
> 
> func_instance *mapped_object::findGlobalConstructorFunc(const std::string &ctorHandler) func_instance *mapped_object::findGlobalDestructorFunc(const std::string &dtorHandler)

I had considered that, but ultimately decided not to. On second glance, I think it would be better to merge them. I'll update.

hainest:
> From @bbiiggppiigg point, these could be consolidated to a `FindGlobalFunc` method. Otherwise looks good.

Agreed. Updated.

PR #2135, opened 2026-02-01
Remove unused OpenMP parsing in parse_func

hainest:
From the history of `image::analyzeImage()`, it looks like it was only ever used on aix and sparc.


hainest:
No regressions on 16 machines (two of the usual ones are down).

PR #2136, opened 2026-02-02
Remove unused parse_func_registers::specialPurposeRegisters

hainest:
It was added by 7c2938c70 in 2006, but never used.


hainest:
No regressions on 3 machines (one of each platform). I didn't run the full battery of tests since this code is never used.

PR #2137, opened 2026-02-02
Add a synchronized breakpoint type

hainest:
This can be used to ensure that all threads in a process are stopped before the breakpoint is executed.

I'm adding this separately from #2131 since it contains additional fixes.

### DIFF
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -8253,6 +8253,13 @@ Breakpoint::ptr Breakpoint::newHardwareBreakpoint(unsigned int mode,
   return newbp;
 }
 
+Breakpoint::ptr Breakpoint::newSynchronousBreakpoint() {
+  Breakpoint::ptr newbp = Breakpoint::ptr(new Breakpoint());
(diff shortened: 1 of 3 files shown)


hainest:
This builds fine on all supported platforms. There isn't any code using it yet, so there's no need running the usual battery of tests.

proccontrol/src/process.C
```diff
@@ -8253,6 +8253,13 @@ Breakpoint::ptr Breakpoint::newHardwareBreakpoint(unsigned int mode,
   return newbp;
 }
 
+Breakpoint::ptr Breakpoint::newSynchronousBreakpoint() {
+  Breakpoint::ptr newbp = Breakpoint::ptr(new Breakpoint());
```
kupsch, line 8257: make_shared?
```suggestion
  Breakpoint::ptr newbp = boost::make_shared<Breakpoint>();
```

kupsch (approved):
one comment otherwise looks good

proccontrol/src/process.C
hainest: I wanted to, but `Breakpoint`s default constructor is private.

PR #2138, opened 2026-02-02
Add clang flag to split hot/cold code

hainest:
gcc does this by default.


kupsch (approved):

hainest:
The compiler multibuild is passing.

PR #2139, opened 2026-02-03
GitHub CI: auto-cancel in-progress workflows

hainest:
Pushing changes to a PR's branch will relaunch all of the workflows. This change ensures that any workflows running against that PR are terminated to free resources and run with the updated branch.

Each workflow is given a group ID that is the workflow name and the PR number or ref. For example, 'Pull Request Tests' running for PR 17 will be named "pr-tests-17".


hainest:
I just did a quick test, and it seems to be working. Testing these CI files is a guessing game at best.

PR #2140, opened 2026-02-03
Make Dyninst::Register constructors noexcept

hainest:
For std::vector<T>::emplace_back, the standard notes

  If the move constructor of T is not noexcept and is not CopyInsertable
  into *this, vector will use the throwing move constructor.

gcc-15 and newer warn via '-Werror=noexcept' when the move_if_noexcept condition fails and recommends making the ctor noexcept. This is not a requirement, but the CI jobs build with -Werror.

### DIFF
--- a/common/src/dyn_register.h
+++ b/common/src/dyn_register.h
@@ -72,14 +72,14 @@ public:
   }
 
   // Default is an invalid register.
-  constexpr Register()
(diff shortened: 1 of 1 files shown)


hainest:
The compiler multibuild workflow is passing.

PR #2141, opened 2026-02-05
Remove parse_func::writesSPRs

hainest:
It was added by ad610d9a0c90 in 2006, but never implemented.


kupsch (approved):

hainest:
No build issues on arm and ppc. I didn't run the full battery of tests since this is unused code.

PR #2144, opened 2026-02-06
Fix CMake configuration errors when using system Boost library

jbelloncastro:
`common` library is missing Boost components, leaving some symbols unresolved at link time:
- `dyn_rwlock::lock_shared` uses `boost::mutex`, so it depends on `Boost::thread` library target.
- `Dyninst::filesystem` uses `boost::filesystem`, so it depends on `Boost::filesystem` library target.

`Dyninst::Boost_headers` target is not defined in some cases when Dyninst::Boost` is, producing errors at the end of configuration stage.

### DIFF
--- a/cmake/tpls/DyninstBoost.cmake
+++ b/cmake/tpls/DyninstBoost.cmake
@@ -53,7 +53,9 @@ if(NOT TARGET Dyninst::Boost)
   target_include_directories(Dyninst::Boost SYSTEM INTERFACE ${Boost_INCLUDE_DIRS})
   target_compile_definitions(Dyninst::Boost
                              INTERFACE BOOST_MULTI_INDEX_DISABLE_SERIALIZATION)
+endif()
 
+if(NOT TARGET Dyninst::Boost_headers)
   # Just the headers (effectively a simplified Boost::headers target)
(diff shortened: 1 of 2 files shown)


jbelloncastro:
Hi there
I found these problems when trying to build rocprofiler-systems with a system install of the boost libraries.
After the proposed changes were made the configuration errors and undefined reference linker errors were solved.

I can't really explain what is the reason behind the missing `Dyninst::Boost_headers` target, as it is clearly defined in the same file as other targets that were visible to CMake. However, the fix was simple enough and solved the issue for me.

hainest:
Could you post the link errors with the complete link line?

kupsch:
Using a simple test, I am unable to reproduce this on Ubuntu 24.04 using boost provided by the OS distribution. @jbelloncastro is it possible for you post a reproducer?

jbelloncastro:
I've tried to create a small standalone project linking against one of these libraries but I have not been able to get the linker error with it.

I investigated a bit deeper on the original project and it seems this happens because dyninst targets are not imported with `find_package` config modules. I can't really explain why the fix I am suggesting solves the issue but it might be an odd interaction with how this is done at the moment.

In summary, this seems to be another project's problem with importing the targets. I'll close the PR.

PR #2147, opened 2026-02-06
GitHub CI: don't autocancel container workflows

hainest:
'Refresh base and dev containers' depends on
'Build and Deploy Development Containers'. 'Purge old workflow runs' isn't in that dependency chain, but it could be added to a longer workflow in the future. It doesn't matter if it is run multiple times.


hainest:
I can now run the 'Refresh base and dev containers' workflow successfully.

PR #2148, opened 2026-02-07
Extract arch-specific register clobbering out of parse_func

hainest:
The logic used between aarch64 and power is effectively the same. However, i386 is different from x86_64 and both are different from power and aarch64. These should eventually be merged into the same functionality based at least in part (but not completely) on ABI calling conventions. It's also not clear that the x86 recursive parsing of 3 levels of calls is a useful optimization.

This also fixes a bug in `writesFPRs` that used the incorrected registers when parsing x86_64 code.

@wxrdnx This will effect #2086, but should be a quick fix.


hainest:

hainest:

PR #2149, opened 2026-02-09
Make Emitter::getInterModule(Func,Var)Addr overloadable

hainest:
This makes the implementations architecture-specific without having possible symbol collisions when all architectures are linked together (a necessity for having architecture-independent codegen).


hainest:

bbiiggppiigg (approved):

PR #2150, opened 2026-02-10
Correctly detect compressed instructions for operand formatting

hainest:
This was inadvertently broken by e0c12bc4ad in 2025.

### DIFF
--- a/tests/integration/InstructionAPI/decoder/x86/CMakeLists.txt
+++ b/tests/integration/InstructionAPI/decoder/x86/CMakeLists.txt
@@ -3,6 +3,6 @@ include_guard(GLOBAL)
 add_executable(call_tests call.cpp)
 target_compile_options(call_tests PRIVATE ${SUPPORTED_CXX_WARNING_FLAGS})
 target_link_libraries(call_tests PRIVATE instructionAPI cft_tests memory_tests
-                                         register_tests)
+                                         register_tests opcode_tests)
 add_test(NAME instructionAPI_call_tests COMMAND call_tests)
 set_tests_properties(instructionAPI_call_tests PROPERTIES LABELS "integration")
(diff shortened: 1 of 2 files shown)


hainest:

hainest:
@kupsch @bbiiggppiigg I'm not particularly happy with the hackish changes for x86. It would be really nice if `MachRegister::name()` returned just the name instead of it prefixed by the architecture (e.g., `rax` instead of `x86_64::rax`). I'm curious if that would be a breaking change...

bbiiggppiigg:
> @kupsch @bbiiggppiigg I'm not particularly happy with the hackish changes for x86. It would be really nice if `MachRegister::name()` returned just the name instead of it prefixed by the architecture (e.g., `rax` instead of `x86_64::rax`). I'm curious if that would be a breaking change...

Different architectures could potentially have registers of the same name. I think it is still kind of required to know which architecture it belongs to. 

Not sure if it is a good example, but for AMDGPU, gfx908, gfx90a and gfx940 all have scalar registers s0-s101, vector registers v0-v255. 

Of course if we add check to make sure each architecture is working with the registers belonging to them, maybe we can omit the architecture.

hainest:
> > @kupsch @bbiiggppiigg I'm not particularly happy with the hackish changes for x86. It would be really nice if `MachRegister::name()` returned just the name instead of it prefixed by the architecture (e.g., `rax` instead of `x86_64::rax`). I'm curious if that would be a breaking change...
> 
> Different architectures could potentially have registers of the same name. I think it is still kind of required to know which architecture it belongs to.

Yes, but it's not clear why that needs to be part of the MachRegister's name. `RegisterAST::format` removes it because it's not useful. I can't see how this would generally be useful for users. There is `Dyninst::getArchitectureName`, if folks need it. The question is if it would be a breaking change. It would certainly be a convenient one.

kupsch:
I don't see any good reason to have a `architecture::` prefix in the name string. We could add a `name_with_architecture` method if there is any need, but I don't see any obvious need for this.  Changing `DEF_REGISTER` and `DEF_REGISTER_ALIAS` be an easy start. Removing the code that checks and removes the prefix wouldn't be hard, then it's just testing.

I always thought it was strange. If you want to remove it, I'd say go for it.

PR #2156, opened 2026-02-25
Remove remnants of defunct parallel region parsing

hainest:
The OpenMP-specific parts were removed by 081f5af2d3bb6. The entire parallel region parsing code was disabled by 0cbe68409 in 2015 when the usage of `parse_func::addParRegion` was removed. No reason was given, but comments throughout the code suggest that this was only used on now-unsupported platforms like sparc and SunOS.


bbiiggppiigg (approved):

hainest:
No test regressions on 3 machines (one x86, arm, ppc). The full battery wasn't run since this is dead code.

PR #2157, opened 2026-02-25
Remove architecture prefix in MachRegister name

hainest:
Register names are no longer prefixed with the architecture (e.g., `x86::eax` is now `eax`). This makes instruction formatting and decode testing easier.

### DIFF
--- a/instructionAPI/src/ArchSpecificFormatters.C
+++ b/instructionAPI/src/ArchSpecificFormatters.C
@@ -383,13 +383,7 @@ std::string x86Formatter::formatImmediate(const std::string &evalString) const
 
 std::string x86Formatter::formatRegister(const std::string &regName) const
 {
-    std::string outReg = "%" + regName;
-
-    for(char &c : outReg) {
-      c = std::tolower(c);
-    }
-
-    return outReg;
+    return "%" + regName;
 }
 
 std::string x86Formatter::formatDeref(const std::string &addrString) const


hainest:
No test regressions on 3 machines (one of x86, arm, ppc). I didn't run the full battery since these aren't really functional changes.

instructionAPI/src/ArchSpecificFormatters.C
```diff
@@ -411,18 +389,11 @@ std::string x86Formatter::formatImmediate(const std::string &evalString) const
...
+    std::string outReg = "%" + regName;
 
-    auto regNameOffset = regName.find("::");
-    if (regNameOffset == std::string::npos)  {
-	regNameOffset = 0;	// no "::", copy whole string
-    }  else  {
-	regNameOffset += 2;	// skip "::"
+    for(char &c : outReg) {
+      c = std::tolower(c);
     }
```
kupsch: I think this is unnecessary and can go:

```suggestion
```

In their definitions the x86 registers are all lower case. I think we should just use the case from the definition.

kupsch (approved):
One small change I'd recommend, otherwise looks good.

instructionAPI/src/ArchSpecificFormatters.C
hainest: Yeah, I was going to delete that. Not sure why I didn't. Interestingly, arm converts them to uppercase. I'm not sure that's useful, either.

kupsch: all the arm assembly language examples seem to use lowercase, so converting to uppercase is probably not right and should just use the case as is. The only thing that might break would be if a string compare of a register name is done in Dyninst or tests.

hainest: I fixed those two. amdgpu was also doing an uppercase conversion, so I removed it. I'll re-run the test suite to double-check, but I don't anticipate any failures.

hainest: Unsurprisingly, removing the case modifications didn't cause any test failures. If they work for you, I'll merge.

kupsch: merge away

PR #2158, opened 2026-02-27
Clean up, refactor registerSpace classes

hainest:


hainest:
No test regressions on 3 machines (one each of x86, arm, ppc). There is no need to run the full battery since this is just code cleanup.

hainest:

hainest:
> What does SMF stands for ?

[Special Member Functions](https://en.wikipedia.org/wiki/Special_member_functions).

PR #2160, opened 2026-03-02
GitHubCI: patch build error in extrae

hainest:
https://github.com/bsc-performance-tools/extrae/pull/140


hainest:
This fixes the build issue with extrae, but there are other consumers failing so the workflow still isn't passing.

PR #2161, opened 2026-03-03
Fix AddressTranslateSysV::getAOut

kupsch:
- Clean up AuxvParser: ctor, getter functions, and file names
    
- Improve AuxvParser to get the number of program headers and the size 
  of each program header

- Fix AddressTranslateSysV::getAOut to eliminate assumptions on program
  header types and order; program headers can be any order.  The current
  could read past the end of the program headers or stop early failing
  to find the lowest or any PT_LOAD program header. Now loops over all
  program headers to correctly find the PT_PHDR and lower PT_LOAD
  virtual addresses.
---

Below is the original description of problem from @wsxarcher in pr #2151. This is a more general fix.

Fixes #2151.

When compiling a simple binary on RISC-V Ubuntu 24.04 on the default GCC and Clang, RISCV_ATTRIBUTES are before the PT_LOAD.

```
ubuntu@ubuntu:~$ readelf -l tll

Elf file type is DYN (Position-Independent Executable file)
Entry point 0x990
There are 10 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000000040 0x0000000000000040
                 0x0000000000000230 0x0000000000000230  R      0x8
  INTERP         0x0000000000000270 0x0000000000000270 0x0000000000000270
                 0x0000000000000021 0x0000000000000021  R      0x1
      [Requesting program interpreter: /lib/ld-linux-riscv64-lp64d.so.1]
  RISCV_ATTRIBUT 0x00000000000020d2 0x0000000000000000 0x0000000000000000
                 0x0000000000000053 0x0000000000000000  R      0x1
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000e88 0x0000000000000e88  R E    0x1000
  LOAD           0x0000000000001dd0 0x0000000000001dd0 0x0000000000001dd0
                 0x00000000000002b0 0x00000000000002e8  RW     0x1000
  DYNAMIC        0x0000000000001de8 0x0000000000001de8 0x0000000000001de8
                 0x00000000000001e0 0x00000000000001e0  RW     0x8
  NOTE           0x0000000000000294 0x0000000000000294 0x0000000000000294
                 0x0000000000000044 0x0000000000000044  R      0x4
  GNU_EH_FRAME   0x0000000000000dbc 0x0000000000000dbc 0x0000000000000dbc
                 0x000000000000002c 0x000000000000002c  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
  GNU_RELRO      0x0000000000001dd0 0x0000000000001dd0 0x0000000000001dd0
                 0x0000000000000230 0x0000000000000230  R      0x1
```

If this function find an unknown segment it will exit before setting the correct load_vaddr.

Segment name is define gated to support older libelf


hainest (approved):

The gymnastics of object management in createAuxvParser is completely unnecessary because the two places where it's called immediately delete the object. There is no sharing. That doesn't need to be cleaned up here.

bbiiggppiigg (approved):

Issue #1455, opened 2023-07-24
symtab doesn't understand dwarf5 inline functions


mwkrentel:
**Describe the bug**
It seems that clang 14 produces some inline dwarf info that symtabAPI
doesn't understand.  Note: this is the inline dwarf info computed by
`getContainingFunction()`, not the terminal line map info.

**To Reproduce**
Here is a short way to reproduce the problem.
This is on Ubuntu 22.04.

```
$ cat inline.c
int xx, yy;

__attribute__((always_inline))
static inline void inl() {
  xx += yy + 2;
}

int main() {
  inl();
  return 0;
}

$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ clang -g -O2 inline.c -o inline
```

Then, to see what symbtab produces for the inline info, I was using
hpcstruct with some debug info turned on.  Of course, there are other
ways to read this.  This is using Dyninst master as of July 20, 2023
and elfutils 0.189.

```
block:
stmt:  0x1130 (6)  l=5  f='/home/krentel/clang/inline.c'

0x1130  l=9  file:  Unknown file
0x1130  symtab:  inl
0x1130  demang:  inl
   inline:  l=9  f='Unknown file'  p='inl'

0x1130  l=9  file:  Unknown file
0x1130  symtab:  inl
0x1130  demang:  inl
stmt:  0x1136 (6)  l=5  f='/home/krentel/clang/inline.c'

0x1136  l=9  file:  Unknown file
0x1136  symtab:  inl
0x1136  demang:  inl
   inline:  l=9  f='Unknown file'  p='inl'

0x1136  l=9  file:  Unknown file
0x1136  symtab:  inl
0x1136  demang:  inl
stmt:  0x113c (2)  l=5  f='/home/krentel/clang/inline.c'
...
```

So, symtab gets the terminal line map correct, but the inline steps
return "Unknown file".

I suspect the problem is some newer DIE object that clang outputs but
symtab doesn't understand.

I didn't see any errors/warning from dyninst, but objdump 2.36.1 reports these errors.

```
$ objdump -Wi inline >/dev/null
objdump: Warning: Unrecognized form: 37
objdump: Warning: Unrecognized form: 37
objdump: Warning: Unrecognized form: 37
objdump: Warning: Unrecognized form: 27
[...]

PR #2164, opened 2026-03-04
Add regression test for DWARF5 inlined functions

hainest:
See https://github.com/dyninst/dyninst/issues/1455 for details.

Requires https://github.com/dyninst/test_binaries/pull/2/changes.


hainest:
The new tests are passing on 3 machines (x86, ppc, arm).

PR #2165, opened 2026-03-04
Clean up logic related to movePHdrFirst

bbiiggppiigg:
movePHdrFirst is set to true after the constructor, hence clean up the mess for ease of debugging.


hainest (approved):

PR #2166, opened 2026-03-05
Fix logic for finding last loadable section

### LINKED ISSUE #2081: Failing to write instrumented shared library

bbiiggppiigg:
The last loadable section of a binary is the last section of the last loadable segment, which is where dyninst inserts its newly created sections.

The current implementation asumes the last section will end on the boundary of the last segment, which does not hold for certain binaries.

This PR updates the logic to find the section with the highest address within the bound of the last loadable segment, which relaxes the assumption.

Replaces #2119 
Fixes #2081 in theory.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -376,9 +376,8 @@ typename emitElf<ElfTypes>::Elf_Off emitElf<ElfTypes>::findLastLoadableSec() {
         tmp++;
     }
     Elf_Scn *scn = NULL;
-    Elf_Shdr *shdr;
     while((scn = elf_nextscn(oldElf, scn))) {
-        shdr = ElfTypes::elf_getshdr(scn);
+        Elf_Shdr * shdr = ElfTypes::elf_getshdr(scn);
         Elf_Off secStart = shdr->sh_addr;
         if( lastDataSegStart <= secStart  && secStart < lastDataSegEnd && secStart > lastLoadableSecStart)
           lastLoadableSecStart = secStart;
@@ -512,7 +511,6 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
         secNames.push_back(name);
         newshdr->sh_name = secNameIndex;
         secNameIndex += strlen(name) + 1;
-        // For sections with vaddr > 0, adjust load addddress by library_adjust
         if (newshdr->sh_addr) {
             newshdr->sh_addr += library_adjust;
 
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -360,9 +360,6 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
 template<class ElfTypes>
 typename emitElf<ElfTypes>::Elf_Off emitElf<ElfTypes>::findLastLoadableSec() {
     Elf_Phdr *tmp = ElfTypes::elf_getphdr(oldElf);
-    // Find the offset of the start of the text & the data segment
-    // The first LOAD segment is the text & the second LOAD segment
-    // is the data
     Elf_Off lastDataSegStart=0, lastDataSegEnd=0, lastLoadableSecStart=0;
     for (unsigned i = 0; i < oldEhdr->e_phnum; i++) {
         if (tmp->p_type == PT_LOAD) {
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -78,8 +78,7 @@ emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, ObjectELF *ob
         olddynStrData(NULL), olddynStrSize(0),
         symTabData(NULL), dynsymData(NULL), dynData(NULL),
         phdrs_scn(NULL), verneednum(0), verdefnum(0),
-        newSegmentStart(0), firstNewLoadSec(NULL),
-        dataSegEnd(0), dynSegOff(0), dynSegAddr(0),
+        newSegmentStart(0), firstNewLoadSec(NULL), dynSegOff(0), dynSegAddr(0),
         phdrSegOff(0), phdrSegAddr(0), dynSegSize(0),
         secNameIndex(0), currEndOffset(0), currEndAddress(0),
         linkedStaticData(NULL), loadSecTotalSize(0),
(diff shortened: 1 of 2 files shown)


hainest:

I've not taken a deep look at this, but it looks ok superficially. I'll give it a thorough check tomorrow.

symtabAPI/src/emitElf.C
```diff
@@ -359,21 +358,32 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
 
 // Find the end of data/text segment
 template<class ElfTypes>
-void emitElf<ElfTypes>::findSegmentEnds() {
+typename emitElf<ElfTypes>::Elf_Off emitElf<ElfTypes>::findLastLoadableSec() {
     Elf_Phdr *tmp = ElfTypes::elf_getphdr(oldElf);
     // Find the offset of the start of the text & the data segment
     // The first LOAD segment is the text & the second LOAD segment
     // is the data
```
hainest: Is this assumption still encoded in these changes? As we discussed, it is important to not assume the ordering of segments.

symtabAPI/src/emitElf.C
```diff
@@ -359,21 +358,32 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
...
+                lastDataSegEnd = tmp->p_vaddr + tmp->p_memsz;
+                lastDataSegStart = tmp->p_vaddr;
+              }
         } else if (PT_TLS == tmp->p_type) {
             TLSExists = true;
         }
         tmp++;
     }
+    Elf_Scn *scn = NULL;
+    Elf_Shdr *shdr;
```
hainest:
```suggestion
```

symtabAPI/src/emitElf.C
```diff
@@ -359,21 +358,32 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
...
+    while((scn = elf_nextscn(oldElf, scn))) {
+        shdr = ElfTypes::elf_getshdr(scn);
```
hainest:
```suggestion
        Elf_Shdr *shdr = ElfTypes::elf_getshdr(scn);
```

symtabAPI/src/emitElf.C
```diff
@@ -503,7 +512,7 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
         secNames.push_back(name);
         newshdr->sh_name = secNameIndex;
         secNameIndex += strlen(name) + 1;
-
+        // For sections with vaddr > 0, adjust load addddress by library_adjust
```
hainest:
```suggestion
        // For sections with vaddr > 0, adjust load address by library_adjust
```

hainest (changes requested):
Overall, it looks good. We should add a regression test for libpytorch from 2081.

symtabAPI/src/emitElf.C
bbiiggppiigg: @hainest 
I don't think the assumptions were true even before this PR.
All this loop is trying to do is to find the last PT_LOAD segment.
If we want to be safe we can skip PT_LOAD segments that contains text or is readonly ?

hainest: >If we want to be safe we can skip PT_LOAD segments that contains text or is readonly ?

We shouldn't skip loadable segments if we want the last one. Otherwise, we might overwrite the skipped one.

symtabAPI/src/emitElf.C
```diff
@@ -359,21 +358,32 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
...
             TLSExists = true;
         }
         tmp++;
     }
+    Elf_Scn *scn = NULL;
+    Elf_Shdr *shdr;
+    while((scn = elf_nextscn(oldElf, scn))) {
+        shdr = ElfTypes::elf_getshdr(scn);
+        Elf_Off secStart = shdr->sh_addr;
+        if( lastDataSegStart <= secStart  && secStart < lastDataSegEnd && secStart > lastLoadableSecStart)
```
kupsch, line 379: I think you want max _n_ the `scn[n]->sh_addr` where the address is in the segment, not the one with the greatest value so the last clause should be eliminated. Sections can be any order, they don't have to be in order:

```suggestion
        if( lastDataSegStart <= secStart  && secStart < lastDataSegEnd)
```

symtabAPI/src/emitElf.C
kupsch: There is no defined order; just common practice. The currently code likely forces RWX or the program header and if not we could change the code to force RWX and emit a warning.

symtabAPI/src/emitElf.C
bbiiggppiigg: @kupsch 

The first two clause checks if the section sits in the address range of the last data segment,
the third clause make sure we get the one with the max address.

This is just mimicing the logic of the old code, which is also basing its calculation on the address.
Now I think it maybe this should be based on file offset?
We should maybe have a meeting about this ?

symtabAPI/src/emitElf.C
bbiiggppiigg: Based on the old comments, I'm assuming they want the last PT_LOAD so they insert after DATA and not TEXT, so if that is what they "actually" want to achieve we could skip TEXT.

PR #2196, opened 2026-04-04
Add explicit include of TBB version header

hainest:
The latest version (2022.3.0, released 29-OCT-2025) doesn't have the implicit include anymore.

### DIFF
--- a/cmake/tpls/DyninstTBB.cmake
+++ b/cmake/tpls/DyninstTBB.cmake
@@ -49,4 +49,13 @@ message(STATUS "Found TBB ${TBB_VERSION}")
 get_target_property(_tmp TBB::tbb INTERFACE_INCLUDE_DIRECTORIES)
 message(STATUS "TBB include directories: ${_tmp}")
 
+# Starting with 2021.2, the version macros were extracted into 'version.h'
+# To retain support for earlier versions, we conditionally include it
+foreach(_dir ${_tmp})
+  if(EXISTS "${_dir}/tbb/version.h")
+    target_compile_definitions(Dyninst::TBB INTERFACE "DYNINST_TBB_HAS_VERSION_H")
+    break()
+  endif()
+endforeach()
+
 unset(_find_path_args)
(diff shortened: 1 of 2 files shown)


hainest:

kupsch (approved):

fche:
One problem with this approach is that any client binary that #includes the dyninst headers, but doesn't do the same cmake DYNINST_TBB_HAS_VERSION_H check, will have that macro missing and thus the dyninst concurrent.h file not include tbb/version.h .

On a c++17 library, this could work more generally:

```
#if defined(__has_include)
#  if __has_include(<tbb/version.h>)
#    include <tbb/version.h>
#  endif
#endif
```

hainest:
> One problem with this approach is that any client binary that #includes the dyninst headers, but doesn't do the same cmake DYNINST_TBB_HAS_VERSION_H check, will have that macro missing and thus the dyninst concurrent.h file not include tbb/version.h .

Projects that don't consume Dyninst as a CMake project are pretty much on their own to get any compile definitions correct. I have briefly looked at generating a pkgcfg file, but not put much time into it. Is that something the RedHat folks would find useful?

>On a c++17 library, this could work more generally:

After the 14.0 release (no timeline on that), we'll require c++20 to use Dyninst so that will likely be the solution to use instead of the CMake checks.

fche:
Hi -

>    Projects that don't consume Dyninst as a CMake project are pretty much
>    on their own to get any compile definitions correct.

Ah, a cmake-based app client would inherit dyninst's TBB-related
configury macros?  (Do you also have the option of shipping the
resulting cmake configury build-time results as a separate config
header installed with "dyninst-devel", so a client could just include
the headers and dyninst headers would self-serve?)

>    I have briefly looked at generating a pkgcfg file, but not put
>    much time into it. Is that something the RedHat folks would find
>    useful?

Yeah, possibly, esp. if this build configuration macro stuff is all in
there (or referred to from there).

- FChE

PR #2205, opened 2026-04-10
[dyninstAPI] Implement `BPatch_mod`

kcossett-amd:
I was doing some testing involving instrumenting functions after N calls. I wanted to use `BPatch_mod` but was met with the `assert(0)`. 

`BPatch_mod` is now a first-class opcode (`modOp`) with architecture-specific code generation, following the same convention as `BPatch_divide`/`divOp`.

### x86 (IA32 + AMD64)

`DIV`/`IDIV` computes both quotient (RAX) and remainder (RDX) in a single instruction. `emitDiv` reads RAX; `emitMod` reads RDX instead, requiring zero additional instructions. `emitModImm` adds fast paths for constant divisors: 
1. `mov dest, 0` for modulo 1,
2. `AND dest, (N-1)` for unsigned power-of-2. 
3. Other constants fall back to `IDIV`, consistent with `emitDivImm`.

### AArch64

No remainder instruction. Matching GCC/Clang output: `SDIV`/`UDIV` + `MSUB`. A `generateMSub` helper is added. `MSUB` shares the same encoding as `generateMul`/MADD, with bit 15 (`o0`) selecting subtract instead of add.

### Power

Scalar modulo instructions exist on POWER9+ (`modsw`/`moduw`) but GCC avoids them due to [performance issues with certain inputs](https://gcc.gnu.org/pipermail/gcc-patches/2023-June/620622.html). Matching GCC: `divw` + `mullw` + `subf`, using the same XOFORM encoding macros as `divOp`.

### AMDGPU

Marked as not implemented.

### Testing

**Emitter byte-encoding tests** (`tests/unit/dyninstAPI/emitter/x86_64.cpp`): Two new cases verify exact machine code for `emitMod` and `emitModImm`, alongside existing `emitDiv`/`emitDivImm` tests. These passed locally.

I have also tested with some other local (and simple) examples on `x86_64`

### DIFF
--- a/dyninstAPI/src/ASTs/operatorAST.C
+++ b/dyninstAPI/src/ASTs/operatorAST.C
@@ -686,6 +686,7 @@ bool operatorAST::generateCode_phase2(codeGen &gen, bool noCost, Dyninst::Addres
     case xorOp:
     case timesOp:
     case divOp:
+    case modOp:
     case orOp:
     case andOp:
     case eqOp:
@@ -836,6 +837,7 @@ bool operatorAST::canBeKept() const {
     case xorOp:
     case timesOp:
     case divOp:
+    case modOp:
     case neOp:
     case noOp:
     case orOp:
--- a/dyninstAPI/src/ASTs/operatorAST.h
+++ b/dyninstAPI/src/ASTs/operatorAST.h
@@ -82,6 +82,9 @@ public:
   static Ptr minus(codeGenASTPtr lhs, codeGenASTPtr rhs) {
     return boost::make_shared<operatorAST>(minusOp, std::move(lhs), std::move(rhs));
   }
+  static Ptr mod(codeGenASTPtr lhs, codeGenASTPtr rhs) {
+    return boost::make_shared<operatorAST>(modOp, std::move(lhs), std::move(rhs));
+  }
   static Ptr ne(codeGenASTPtr lhs, codeGenASTPtr rhs) {
     return boost::make_shared<operatorAST>(neOp, std::move(lhs), std::move(rhs));
   }
--- a/dyninstAPI/src/BPatch_snippet.C
+++ b/dyninstAPI/src/BPatch_snippet.C
@@ -430,8 +430,7 @@ BPatch_arithExpr::BPatch_arithExpr(BPatch_binOp op,
                         ast_wrapper = operatorAST::times(lop_ast, rop_ast);
                         break;
                 case BPatch_mod:
-                        /* XXX Not yet implemented. */
-                        assert(0);
+                        ast_wrapper = operatorAST::mod(lop_ast, rop_ast);
                         break;
                 case BPatch_ref:
          ast_wrapper = generateArrayRef(lOperand, rOperand);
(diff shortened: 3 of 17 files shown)


hainest:
I was digging around in BPatch recently and  noticed this. I was really curious why it was never implemented (it was added in 1997...). I'll approve the workflows so they'll run when you're ready to submit for review.

---

This will need to be added to #2086.

@bbiiggppiigg Does AMDGPU have a modulus instruction or a way to get a remainder?

kcossett-amd:
> I was digging around in BPatch recently and noticed this. I was really curious why it was never implemented (it was added in 1997...). I'll approve the workflows so they'll run when you're ready to submit for review.

I suppose I'll put the PR for review. I wanted to check whether the workflows passed, so thank you for letting me know.

bbiiggppiigg:
> @bbiiggppiigg Does AMDGPU have a modulus instruction or a way to get a remainder?

Not that I'm aware of. I can check how they do it in their math API, but I would bet the normal way for general remainder is to do the A - floor(A/B) * B .

kcossett-amd:
Seems `instructionAPI_dwarf5_inline_func_test` failed due to missing binary (see https://github.com/dyninst/dyninst/actions/runs/24228870833/job/70735945337#step:4:2258)

I'm guessing the `git clone` failed https://github.com/dyninst/dyninst/blob/master/tests/cmake/DyninstBinaryTests.cmake#L13 but there is no error checking for me to verify this.

hainest:
> Seems `instructionAPI_dwarf5_inline_func_test` failed due to missing binary (see https://github.com/dyninst/dyninst/actions/runs/24228870833/job/70735945337#step:4:2258)
> 
> I'm guessing the `git clone` failed https://github.com/dyninst/dyninst/blob/master/tests/cmake/DyninstBinaryTests.cmake#L13 but there is no error checking for me to verify this.

The clone happens during CMake config time:

```

-- Installing 'test_binaries' into /tmp/v41Vik/tests/test_binaries
Cloning into '/tmp/v41Vik/tests/test_binaries'...
fatal: unable to access 'https://github.com/dyninst/test_binaries.git/': The requested URL returned error: 403
```

The builds for AMDGPU codegen are failing because the new methods aren't implemented on its [emitter](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/emit-amdgpu.h#L49). I'd recommend waiting until @bbiiggppiigg gets back to us about how modulus is implemented there before proceeding.

hainest:
@bbiiggppiigg Did you have a chance to take a look? If not, we can add a stub implementation for now.

bbiiggppiigg:
> @bbiiggppiigg Did you have a chance to take a look? If not, we can add a stub implementation for now.

Sorry for the delay.
Yeah I recommend you implement a stub.
The mod operator gets lowered into a sequence of instructions that involves multiplying the reciprocal of the divisor to get an estimate of the quotient anf fix it to be more accurate, and then mutiply it by the divisor, which is then used to subtract the divident to get the remainder. 
I doubt we will ever support it for AMDGPU.

PR #2214, opened 2026-04-19
Refactor check for encoding immediates in operationAST

hainest:
I am very dubious that these calculations are fully correct- particularly for x86. For now, I've just collected them all into one place to work toward cross-platform codegen.

Note that for AMDGPU, I converted the assert to a failing case in `can_encode_directly` since that will prevent immediate encoding.


hainest:

kupsch (approved):

PR #2215, opened 2026-04-21
[dyninstAPI] Implement isSharedLib for BPatch_object

kcossett-amd:
I've noticed that there exists `isSharedLib` checker for both `BPatch_module` and `BPatch_function`. However, none exists for `BPatch_object`.

It would be nice to not need to go to the module level to check this. And at the `BPatch_module` level, it returns:
`mod->obj()->isSharedLib();` 

**Reason**: I'd like to implement an option within `rocprofiler-systems` whose purpose is to filter out all shared libraries. From what I gather, I'd realistically only be left with the main executable, which I could find by just finding the executable name in the object list, but I figure that having an `isSharedLib` makes things clearer.

### DIFF
--- a/dyninstAPI/h/BPatch_object.h
+++ b/dyninstAPI/h/BPatch_object.h
@@ -103,11 +103,11 @@ class DYNINST_EXPORT BPatch_object {
 
     // BPatch_object::name
     // Returns the file name of the object
-    std::string name();
+    std::string name() const;
 
     // BPatch_object::pathName
     // Returns the full pathname of the object
-    std::string pathName();
+    std::string pathName() const;
 
     // BPatch_object::isSharedLib
     // Returns true if this object represents a shared library
--- a/dyninstAPI/h/BPatch_object.h
+++ b/dyninstAPI/h/BPatch_object.h
@@ -109,6 +109,10 @@ class DYNINST_EXPORT BPatch_object {
     // Returns the full pathname of the object
     std::string pathName();
 
+    // BPatch_object::isSharedLib
+    // Returns true if this object represents a shared library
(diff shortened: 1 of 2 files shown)


dyninstAPI/h/BPatch_object.h
```diff
@@ -109,6 +109,10 @@ class DYNINST_EXPORT BPatch_object {
     // Returns the full pathname of the object
     std::string pathName();
 
+    // BPatch_object::isSharedLib
+    // Returns true if this object represents a shared library
+    bool isSharedLib();
```
kupsch:
```suggestion
    bool isSharedLib() const;
```

dyninstAPI/src/BPatch_object.C
```diff
@@ -81,6 +81,10 @@ std::string BPatch_object::pathName() {
    return obj->fullName();
 }
 
+bool BPatch_object::isSharedLib() {
```
kupsch:
```suggestion
bool BPatch_object::isSharedLib() const {
```

kupsch (commented):
The only comment is to make the method const. Then we'll run it through the CI.

kcossett-amd:
> The only comment is to make the method const. Then we'll run it through the CI.

Wouldn't it then also make sense to have both `BPatch_object::pathName` and `BPatch_object_::name` be `const` as well?

kupsch:
Yes, you can make both of those changes.  It's on our list of improvements to make more methods const, you now we're making them with other cleanup.

Thanks.

kupsch:
Thanks. I started the CI tests.

hainest (approved):

PR #2216, opened 2026-04-21
Remove 'noCost' param from codegen interfaces

hainest:
It's not exactly clear where it came from or when it was removed, but all of the current interfaces have always ignored it.

### DIFF
--- a/dyninstAPI/src/ASTs/sequenceAST.h
+++ b/dyninstAPI/src/ASTs/sequenceAST.h
@@ -66,7 +66,7 @@ public:
   }
 
 private:
-  bool generateCode_phase2(codeGen &gen, bool noCost, Dyninst::Address &,
+  bool generateCode_phase2(codeGen &genw, Dyninst::Address &,
                            Dyninst::Register &retReg) override;
 };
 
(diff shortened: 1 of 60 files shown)


hainest:
No regressions (I only ran it on one machine of each platform).

dyninstAPI/src/ASTs/sequenceAST.h
```diff
@@ -66,7 +66,7 @@ class sequenceAST : public codeGenAST {
   }
 
 private:
-  bool generateCode_phase2(codeGen &gen, bool noCost, Dyninst::Address &,
+  bool generateCode_phase2(codeGen &genw, Dyninst::Address &,
```
kupsch: I don't think you meant the extra `w`:
```suggestion
  bool generateCode_phase2(codeGen &gen, Dyninst::Address &,
```

kupsch (approved):
one typo, otherwise good.

dyninstAPI/src/ASTs/sequenceAST.h
hainest: No, definitely a weird typo.

PR #2217, opened 2026-04-21
Remove codegen callOptions type

hainest:
It was added by a65441575 in 1994, but never used.


hainest:
No need to run full battery of tests.

kupsch (approved):

PR #2218, opened 2026-04-22
Extract, refactor instMapping class

hainest:


hainest:

dyninstAPI/src/instMapping.h
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+
+    // Fork
+    instMapping(const instMapping *parIM, AddressSpace *child)
+        : func(parIM->func), inst(parIM->inst), where(parIM->where),
+          useTrampGuard(parIM->useTrampGuard), allow_trap(parIM->allow_trap) {
+      for(auto ast : parIM->args) {
+        args.push_back(ast);
+      }
```
kupsch: The for loop is equivalent to assignment and can be moved to the initializer list:
```suggestion
          args(parIM->args),
          useTrampGuard(parIM->useTrampGuard), allow_trap(parIM->allow_trap) {
```

kupsch (approved):
one small improvement you could make, otherwise looks good

dyninstAPI/src/instMapping.h
hainest: Good catch. Updated.

PR #2219, opened 2026-04-22
Remove unused variables from codegen emitter interfaces

hainest:
This wasn't part of #2216 because that one removes one variable with hundreds of references. This PR removes many variables with only a few references each.

Requires #2216 #2218


hainest:
No test regressions (I only tested on one machine of each platform since this is just code removal).

kupsch (approved):

PR #2220, opened 2026-04-22
Remove getFunction(instPoint*) from inst.h

hainest:
This was added by a65441575 in 1994, but I don't think it was every implemented.

Requires #2218


hainest:
No need to run the full battery of tests.

kupsch (approved):

PR #2226, opened 2026-04-24
Extract x86-only emitJmpMC

hainest:
Once platform-independent codegen is in place, this can be put into some x86 namespace and directly called.


hainest:

kupsch (approved):

PR #2227, opened 2026-04-24
Extract writeFunctionPtr from codegen into PCProcess

hainest:
That's where it's used. The AMDGPU version is explicitly removed because dynamic instrumentation isn't supported there, so that code will never get called.


hainest:

bbiiggppiigg (approved):

kupsch (approved):

PR #2229, opened 2026-04-28
Clean up Emitter class

hainest:
Below is a table of which members of Emitter are implemented on the supported platforms (except RISCV). There are two big takeaways:

1. Most of Emitter is x86-specific
2. ppc64le is deeply lacking functionality. For example, `emitOp` drives a lot of functionality on the other platforms.

Overall, I would say this class sits in an odd position between trying to be platform-agnostic and providing useful features. My guess would be that deconstructing the `emitA/V/etc.` functions will likely replace most of the members here. I'm not sure what will happen when compared against `insnCodeGen`. The ultimate goal is to distill a unified interface that every platform supports (or can be made to support) that is driven by the needs of the AST classes.

X means "not implemented"

| function                      | i386 | AMD64 | AMDGPU | aarch64 | ppc64le |
|-------------------------------|------|-------|--------|---------|---------|
| clobberAllFuncCall            |      |       |   X    |         |         |
| emitCall                      |      |       |        |         |         |
| emitCallRelative              |   X  |   X   |   X    |   X     |   [1]   |
| emitGetParam                  |      |       |   X    |         |    X    |
| emitGetRetAddr                |      |       |   X    |   X     |    X    |
| emitGetRetVal                 |      |       |   X    |   X     |    X    |
| emitPLTCall                   |      |       |   X    |         |         |
| emitPLTJump                   |      |       |   X    |         |         |
| getInterModuleFuncAddr        |      |       |   X    |         |         |
| getInterModuleVarAddr         |      |       |   X    |         |         |
| emitAddSignedImm              |      |       |   X    |   X     |    X    |
| emitDiv                       |      |       |   X    |   X     |    X    |
| emitDivImm                    |      |       |   X    |   X     |    X    |
| emitTimesImm                  |      |       |        |   X     |    X    |
| emitLoad                      |      |       |   X    |         |    X    |
| emitLoadConst                 |      |       |        |         |    X    |
| emitLoadFrameAddr             |      |       |   X    |    X    |    X    |
| emitLoadIndir                 |      |       |        |         |    X    |
| emitLoadOrigFrameRelative     |      |       |   X    |    X    |    X    |
| emitLoadOrigRegRelative       |      |       |   X    |         |    X    |
| emitLoadOrigRegister          |      |       |   X    |         |    X    |
| emitLoadRelative              |  X   |       |        |         |         |
| emitLoadShared                |      |       |   X    |         |         |
| emitStore                     |      |       |   X    |         |    X    |
[...]

PR #2230, opened 2026-04-28
Move emitImm to Emitter class

hainest:
Requires #2229


hainest:

bbiiggppiigg:
Why is the emitImm implemented sometimes in emitter(x86/aarch64) some times in inst (amdgpu/power).

kupsch (approved):

hainest:
> Why is the emitImm implemented sometimes in emitter(x86/aarch64) some times in inst (amdgpu/power).

I have no idea. There is no regularity to the implementation files for codegen. I plan to clean all of that up like I did with EmitterIA32.

PR #2231, opened 2026-04-29
Extract emitA into Emitter class

hainest:
Requires #2230


hainest:

kupsch (approved):

PR #2232, opened 2026-04-29
Extract emitR into Emitter class

hainest:
Requires #2231


hainest:

kupsch (approved):

PR #2233, opened 2026-04-29
Explicitly look in lib64 when searching for third-party libraries

hainest:
I found that cmake won't search this subdir on some systems that use lib64 (as opposed to lib).

### DIFF
--- a/cmake/Modules/FindLibDW.cmake
+++ b/cmake/Modules/FindLibDW.cmake
@@ -79,7 +79,7 @@ else()
   find_library(
     LibDW_LIBRARIES
     NAMES libdw dw
-    PATH_SUFFIXES elfutils ${_find_path_args})
+    PATH_SUFFIXES elfutils lib64 ${_find_path_args})
 
   if(EXISTS "${LibDW_INCLUDE_DIRS}/version.h")
     file(STRINGS "${LibDW_INCLUDE_DIRS}/version.h" _version_line
--- a/cmake/Modules/FindLibELF.cmake
+++ b/cmake/Modules/FindLibELF.cmake
@@ -79,7 +79,7 @@ else()
   find_library(
     LibELF_LIBRARIES
     NAMES libelf elf
-    PATH_SUFFIXES elfutils ${_find_path_args})
+    PATH_SUFFIXES elfutils lib64 ${_find_path_args})
 
   macro(_check_libelf_version _file)
     file(STRINGS ${_file} _version_line REGEX "^#define _ELFUTILS_VERSION[ \t]+[0-9]+")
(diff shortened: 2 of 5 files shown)


bigtrak:
So, I used to fix that on our 64 bit systems with making a symlink for that  to work correctly.

However, on a mixed 32/64 bit system (which is legal) searching both might pick up the wrong libraries.
it would be good if you test on a 32/64 system to see if the right library is picked up in that case.  It
would be best to test that in a combined 32/64 VM or container, with both set of libraries installed to
see if that conflict might occur.

I believe as of rhel10 that is no longer an issue, but it is on rhel5-rhel9... I don't recall what
the debuntu based systems do for the multilib "aka both 32 and 64 bit libraries installed" config.

Perhaps I missed it in the "checks list", but I don't see any regression tests on rhel8, rhel9 or rhel10..
many of the DoE systems are based on rhel either directly or through the vendor.

kupsch:
Is the cmake variable `FIND_LIBRARY_USE_LIB64_PATHS` set to true?  It is supposed to be set automatically to true on _platforms that require it_, and controls searching in `lib64` variants in addition to `lib`.

hainest:
> So, I used to fix that on our 64 bit systems with making a symlink for that to work correctly.

I used to do that, but it's a massive pain when installing with spack. It really shouldn't be an issue since it works on some installs but not others- even for the same version of OS.

> However, on a mixed 32/64 bit system (which is legal) searching both might pick up the wrong libraries. it would be good if you test on a 32/64 system to see if the right library is picked up in that case. It would be best to test that in a combined 32/64 VM or container, with both set of libraries installed to see if that conflict might occur.

CMake searches through its known list before it tries the PATH_SUFFIXES. I suppose it's technically possible it could confuse an install with lib, lib32, and lib64, but that seems unlikely. I have some systems with 32-bit libraries installed, so I should be able to keep an eye on that.
 
> I believe as of rhel10 that is no longer an issue, but it is on rhel5-rhel9... I don't recall what the debuntu based systems do for the multilib "aka both 32 and 64 bit libraries installed" config.

Debian systems use the `ARCH-linux-LIBC`, so this shouldn't be an issue. I was having this issue on RHEL 8.

hainest:
> Is the cmake variable `FIND_LIBRARY_USE_LIB64_PATHS` set to true? It is supposed to be set automatically to true on _platforms that require it_, and controls searching in `lib64` variants in addition to `lib`.

It is not. Nor is `CMAKE_FIND_LIBRARY_CUSTOM_LIB_SUFFIX`.

hainest:
It seems to be a CMake issue of sorts. I'm not sure it's an actual bug, but I've done other installs on Fedora systems subsequent to this and had no issues.

PR #2235, opened 2026-04-30
[symtabAPI] Implement support for ELF's DT_RELR

### LINKED ISSUE #1290: Support for ELF's new DT_RELR
The just released glibc-2.36 will have an additional relocation that we may eventually need to support.   

    Support for DT_RELR relative relocation format has been added to
    glibc. This is a new ELF dynamic tag that improves the size of
    relative relocations in shared object files and position independent
    executables (PIE). DT_RELR generation requires linker support for
    z pack-relative-relocs option, which is supported for some targets
    in recent binutils versions. Lazy binding doesn't apply to DT_RELR.

glibc-2.36 will be available in the future fedora-37, along with binutils-2.38.

---

Original proposal for ELF: https://groups.google.com/g/generic-abi/c/bX460iggiKg/m/GxjM0L-PBAAJ?pli=1
Implementation in lld: https://reviews.llvm.org/D48247
Blog post from maskray project: https://maskray.me/blog/2021-10-30-relative-relocations-and-relr
Implementation in FreeBSD's rtld: https://reviews.freebsd.org/D32524

kcossett-amd:
Fixes https://github.com/dyninst/dyninst/issues/1290

## Context

Our project is trying to add CI support for `rhel 10` workflows. However, a single binary rewrite test is failing during the run stage with the error:
```
ROCPROFSYS: Executing './ls.inst --coreutils-prog=ls'...
./ls.inst: error while loading shared libraries: ./ls.inst: DT_RELR without GLIBC_ABI_DT_RELR dependency
```
It is blocking our RHEL 10 support :(

Upon investigation, I learned that with `rhel 10`, the `coreutils` binary is distributed so that it uses `DT_RELR` relative relocation.
```
// RHEL 8.10
readelf -dW /usr/bin/coreutils | grep RELA
 0x0000000000000014 (PLTREL)             RELA
readelf -dW /usr/bin/coreutils | grep RELR
// Nothing
```

```
// RHEL 10
readelf -dW /usr/bin/coreutils | grep RELA
 0x0000000000000014 (PLTREL)             RELA
readelf -dW /usr/bin/coreutils | grep RELR
 0x0000000000000024 (RELR)               0x6580
```

After further digging, I discovered https://github.com/dyninst/dyninst/issues/1290
It would be beneficial to have this implemented, so I'm here with some changes to, at the very least, get the ball rolling.

## Reproducer

On a RHEL 10 container, compile the following `reproducer.cpp`
```
#include <iostream>
int main()
{
    std::cout << "hello " << "world" << '\n';
    return 0;
}
```
With `g++ reproducer.cpp -fPIE -pie -Wl,-z,pack-relative-relocs -o reproducer`.

Then compile a `rewriter.cpp` program:
```
#include <BPatch.h>
#include <BPatch_binaryEdit.h>

#include <iostream>

int
main(int argc, char** argv)
{
    if(argc != 3)
    {
        std::cerr << "Usage: " << argv[0] << " <input-binary> <output-binary>\n";
        return 1;
    }

    BPatch bpatch;

    BPatch_binaryEdit* app = bpatch.openBinary(argv[1], false);
    if(!app)
    {
        std::cerr << "Failed to open binary: " << argv[1] << "\n";
        return 2;
    }

    if(!app->writeFile(argv[2]))
    {
        std::cerr << "Failed to write binary: " << argv[2] << "\n";
        return 3;
    }

    return 0;
}
```

With:
```
export LD_LIBRARY_PATH="$DYNINST_ROOT/lib64:$DYNINST_ROOT/lib:${LD_LIBRARY_PATH:-}"
export DYNINSTAPI_RT_LIB="$DYNINST_ROOT/lib64/libdyninstAPI_RT.so"

g++ rewriter.cpp -std=c++17 \
  -I"$DYNINST_ROOT/include" \
  -L"$DYNINST_ROOT/lib" -L"$DYNINST_ROOT/lib64" \
  -ldyninstAPI -lsymtabAPI -lcommon \
  -Wl,-rpath,"$DYNINST_ROOT/lib" \
  -Wl,-rpath,"$DYNINST_ROOT/lib64" \
  -o rewriter

// Then run
./rewriter ./reproducer ./reproducer.inst
```

Upon running `./reproducer.inst`, it fails with:
```
./reproducer.inst: error while loading shared libraries: ./reproducer.inst: DT_RELR without GLIBC_ABI_DT_RELR dependency
```

I did attempt to patch in `GLIBC_ABI_DT_RELR`, but it then failed with a `segmentation fault (core dumped)`, which led me to discover #1290 

## Changes

Generally, I tried to follow how the existing `RT_REL` and `RT_RELA` were implemented. Of course, `RELR` is a different relocation format that is a compressed relative-relocation table encoded with address and bitmap entries, so it needs separate decode logic.

Overall, `elf.h` defines for `RELR` were added and their cases were added in the relevant sections. Also:

In `Object-elf` files:
 - `decodeRelrEntries(...)`: Decodes the entries following https://dram.page/p/relative-relocs-explained/#:~:text=The%20idea%20originates,handled%20needs%20relocating.
 - `get_relocationRelr_entries(...)`:  Finds `.relr.dyn`, determine whether `uintptr_t` is 4 or 8 bytes from the ELF class, validate `DT_RELRENT` + `DT_RELRSZ` against the section data, then call `decodeRelrEntries(...)`.

In `emitElf` files:
 - `.relr.dyn` was not added to `updateLinkInfoSecs` as `sh_info` and `sh_link` are unused.
 - Within `emitElf<ElfTypes>::driver`:
   - For every `RELR` entry, adjust the value stored in the memory slot needing load-base adjustment.
   - In the section that checks for pointer-table sections `(.init_array`, `.fini_array`, etc...), prevent re-applying the offset to an address that `RELR` offset adjustment logic already handled.
 - Handle emitting `RT_RELR` in `createLoadableSections`
 - `createRelrRelocationSection(...)`: Rebuilds the new `.relr.dyn` section from the decoded `RELR` relocation locations shifted by `library_adjust`. 
- Within `createSymbolVersions(...)`, preserve original `.gnu.version_r` version-need entries that were not recreated through symbol version references. This keeps symbolless data such as `GLIBC_ABI_DT_RELR` when it was present in the input binary.

**NOTE**: `createRelrRelocationSection(...)` does not recompress bitmap entries. It emits each decoded location as a direct `RELR` entry (which is still valid `RELR` encoding... see https://gabi.xinuos.com/elf/06-reloc.html#:~:text=unit%20is%20relocatable.-,Note,that%20a%20simple%20list%20of%20(even)%20addresses%20is%20a%20valid%20encoding.,-%E2%86%90%205.). 

Consulted:
 - https://dram.page/p/relative-relocs-explained/
 - https://gabi.xinuos.com/elf/06-reloc.html
 - https://github.com/dyninst/dyninst/issues/1290
 - I had AI help me understand the issues + help review the changes

## Testing

Rebuilding Dyninst and recompiling the `rewriter` our `rewriter`, and running it against the `reproducer`, it now works and prints `hello world`.
A `readelf` comparison shows:
```
== original reproducer ==
 0x0000000000000024 (RELR)               0x7a8
 0x0000000000000023 (RELRSZ)             16 (bytes)
 0x0000000000000025 (RELRENT)            8 (bytes)
  [12] .relr.dyn         RELR            00000000000007a8 0007a8 000010 08   A  0   0  8
== rewritten reproducer.inst ==
 0x0000000000000024 (RELR)               0x1014a4
 0x0000000000000023 (RELRSZ)             24 (bytes)
 0x0000000000000025 (RELRENT)            8 (bytes)
readelf: Warning: [ 6]: Unexpected value (1) in info field.
readelf: Warning: [ 9]: Unexpected value (2) in info field.
  [36] .relr.dyn         RELR            00000000001014a4 0054a4 000018 08   A  0   0  8
```

**Note**: I don't think the warnings are related to the `RELR` changes

Running the `rewriter` with `coreutils` in `RHEL 10` and using it: 
```
[root@1814c85c78ed Experiment]# ./rewriter /usr/bin/coreutils ./coreutils.inst
[root@1814c85c78ed Experiment]# ./coreutils.inst --coreutils-prog=ls
coreutils.inst  relr_stress       reproducer       reproducer.inst.objdump.txt  reproducer.no_relr.inst  rewriter
dyninst         relr_stress.cpp   reproducer.cpp   reproducer.inst.readelf.txt  reproducer.objdump.txt   rewriter.cpp
dyninst-head    relr_stress.inst  reproducer.inst  reproducer.no_relr           reproducer.readelf.txt
```

### Testing With `rocprof-sys-instrument` + `rocprof-sys-run` (binary rewrite mode)

I slightly modified the code in this PR to work with our fork of Dyninst that we use. I then modified all the example executables to be compiled to use `RELR` relocation.

Running our `binary-rewrite` ctest suite with these examples yielded successes (excluding `rccl`, `lulesh` and `jacobi` tests for unrelated reasons). 

The list of examples used in our projects can be found here: https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler-systems/examples


hainest:
I was curious when we'd run into this. I don't have access to any RHEL 10 machines, so we'll want to add at least one regression test in the https://github.com/dyninst/test_binaries repo.

@kupsch This touches the code you're working on. Can I leave it with you?

kcossett-amd:
> I was curious when we'd run into this. I don't have access to any RHEL 10 machines, so we'll want to add at least one regression test in the https://github.com/dyninst/test_binaries repo.

This is reproducible on Ubuntu 24.04 with `reproducer` and `rewriter` compiled with the same flags. I was just so focused on RHEL that I forgot to mention that.

kupsch:
@hainest I can take this, but it will be a bit before I can get to it.

hainest:
> I had AI help me understand the issues + help review the changes

Could you expound on this a bit? We don't have an official policy on machine-generated contributions, but we are very cognizant about attribution of foreign code.

kcossett-amd:
> > I had AI help me understand the issues + help review the changes
> 
> Could you expound on this a bit? We don't have an official policy on machine-generated contributions, but we are very cognizant about attribution of foreign code.

Of course. 

AI was first used to help identify the problem, helping me understand what the `GLIBC_ABI_DT_RELR` error meant. Afterwards, I used it to do a quick scan of the codebase to see whether `RELR` was supported, which it was not. 

I then started doing some surface level digging, getting it to explain to me at a higher level what would be needed to add support for this (to determine if it would be a very large code change). I also used it to present a brief overview of the call chain from binary rewrite to relocation-section emission and how existing `RT_REL` and `RT_RELA` relocation handling worked whilst I was going through the code.

After that, I used AI to help identify likely implementation points so I could make sure I was not missing any required parse or emit paths. I then wrote the code and followed up with AI assistance while debugging compiler errors and reviewing the resulting implementation. It also helped suggest additional sanity checks and improve some comments. 

As a validation check, I also had it verify that my code was doing what I was expecting it to do. For example, with the `decodeRelrEntries` function, I gave it the information from https://dram.page/p/relative-relocs-explained/#:~:text=The%20idea%20originates,handled%20needs%20relocating and had it validate that I was doing just that.

**Side-note**:
I have updated the `changes` section to go more into detail, as I realized it was quite lackluster.
I have also performed extra testing with our `rocprof-sys-instrument` + `rocprof-sys-run` tool (binary-rewrite).

bigtrak:
I'm noticing the magic constant of "libc.so.6" being part of this change.   If that changes,
this change breaks.   If another library uses the new relocations for space efficiency, that
won't work either.

How about detecting the relocation in a library, and making it work if detected?

kcossett-amd:
> I'm noticing the magic constant of "libc.so.6" being part of this change. If that changes, this change breaks. If another library uses the new relocations for space efficiency, that won't work either.
> 
> How about detecting the relocation in a library, and making it work if detected?

That's a fair point. I'll remove the hard-coded `libc.so.6` / `GLIBC_ABI_DT_RELR` handling.

From what I gather, the issue is that dyninst parses the original `.gnu.version_r` table, but the rebuild path only carries forward entries that are referenced by `.gnu.version` entries corresponding to dynamic symbols. `GLIBC_ABI_DT_RELR` is present in `.gnu.version_r`, but no dynamic symbol references its version index, so it never makes it into the reconstructed `verneedEntries`.

I'll update the `.gnu.version_r` reconstruction to merge in original version-need entries that were not recreated through symbol version references, while still skipping provider libraries that were explicitly removed.

symtabAPI/h/Region.h
```diff
@@ -77,6 +77,7 @@ class DYNINST_EXPORT Region : public AnnotatableSparse {
       RT_SYMVERNEEDED,
       RT_REL,
       RT_RELA,
+      RT_RELR,
```
bbiiggppiigg: Maybe its better to move it the end so it doesn't shift the enum values for others. But not that important.

symtabAPI/src/emitElf.C
```diff
@@ -1990,6 +2053,26 @@ void emitElf<ElfTypes>::createRelocationSections(std::vector<relocationEntry> &r
     updateDynamic(dsize_type, dynamic_reloc_size);
 }
 
+template<class ElfTypes>
+void emitElf<ElfTypes>::createRelrRelocationSection(std::vector<Offset> &relr_table) {
+    if (relr_table.empty()) return;
+
+    Elf_Relr *relrs = (Elf_Relr *) malloc(sizeof(Elf_Relr) * relr_table.size());
+    for (unsigned i = 0; i < relr_table.size(); ++i) {
+        relrs[i] = static_cast<Elf_Relr>(relr_table[i] + library_adjust);
```
bbiiggppiigg: Would you consider implementing the relr packing logic as part of this PR?
The following link is an AI assisted prototype of the logic, which seems to work for your reproducer, but we don't have a way of testing it extensively ourselvs.
https://github.com/dyninst/dyninst/commit/d7b47dd64e572d56bc0a0f768a6e701599b7e09e
Without it, the .relr.dyn will bloat from 712 bytes to 12960.

kcossett-amd: Sure

kcossett-amd: Done. I've followed your implementation. 
I've done some testing on our side and our RHEL 10 + Noble workflows pass. 
Locally, I've also observed that the compression works and that the binary can still run.

bbiiggppiigg:
@hainest @kupsch

bbiiggppiigg (approved):
Wait for others.

symtabAPI/src/Object-elf.C
```diff
@@ -107,6 +107,26 @@ static bool pdelf_check_ehdr(Elf_X &elf) {
...
+        } else {
+            entry >>= 1;
+            for (size_t bit = 0; entry != 0; ++bit, entry >>= 1) {
+                if (entry & 1)
+                    relocs.push_back(next + bit * sizeof(RelrT));
+            }
+            next += (sizeof(RelrT) * 8 - 1) * sizeof(RelrT);
+        }
+    }
+}
```
kupsch: Let's make this function return the vector instead of using an out parameter (it's a common bad idiom in Dyninst, but no need to add more). Also eliminating `bit` simplifies things:

```suggestion
static std::vector<Offset> decodeRelrEntries(const RelrT *entries, size_t count) {
    std::vector<Offset> relocs;
    Offset relAddr = 0;  // last address added
    for (size_t i = 0; i < count; ++i) {
        RelrT entry = entries[i];
        if ((entry & 1) == 0) {
            relAddr = static_cast<Offset>(entry);  // is addr, add it
            relocs.push_back(relAddr);
        } else {
            auto lastBitRelAddr = relAddr + sizeof(RelrT) * (8 * sizeof(RelrT) - 1);
            // skips bit 0, adds addr for all other 1 bits
            while ( entry >>= 1 ) {
                relAddr += sizeof(RelrT);
                if (entry & 1)
                    relocs.push_back(relAddr);
            }
            relAddr = lastBitRelAddr;
        }
    }
    return relocs;
}
```

symtabAPI/src/Object-elf.C
```diff
@@ -948,6 +1006,50 @@ void ObjectELF::parseDynamic(Elf_X_Shdr *&dyn_scnp, Elf_X_Shdr *&dynsym_scnp,
...
+    // must describe the byte size of the encoded RELR table
+    if (relr_entry_size_ != entry_size) return false;
+    if (relr_size_ != relrdata.d_size()) return false;
+    if (relr_size_ % entry_size != 0) return false;
+
+    size_t entry_count = relr_size_ / entry_size;
+    if (entry_size == sizeof(uint32_t)) {
+        decodeRelrEntries(reinterpret_cast<uint32_t *>(relrdata.d_buf()),
+                          entry_count,
+                          relr_relocation_table_);
```
kupsch: use new interface:
```suggestion
        relr_relocation_table_ =
            decodeRelrEntries(reinterpret_cast<uint32_t *>(relrdata.d_buf()),
                          entry_count);
```

symtabAPI/src/Object-elf.C
```diff
@@ -948,6 +1006,50 @@ void ObjectELF::parseDynamic(Elf_X_Shdr *&dyn_scnp, Elf_X_Shdr *&dynsym_scnp,
...
+    } else {
+        decodeRelrEntries(reinterpret_cast<uint64_t *>(relrdata.d_buf()),
+                          entry_count,
+                          relr_relocation_table_);
```
kupsch: use new interface:
```suggestion
        relr_relocation_table_ =
            decodeRelrEntries(reinterpret_cast<uint64_t *>(relrdata.d_buf()),
                          entry_count);
```

symtabAPI/src/emitElf.C
```diff
@@ -1990,6 +2053,69 @@ void emitElf<ElfTypes>::createRelocationSections(std::vector<relocationEntry> &r
...
+                    break;                       // beyond this bitmap window
+                bits |= static_cast<Elf_Relr>(1) << (idx + 1);
+            }
+            if (!bits)
+                break;
+            packed.push_back(bits | 1);
+            next += bits_per_entry * word_size;
+            i = j;
+        }
+    }
```
kupsch: I find the three loops not clear. I find this single loop clearer:
```suggestion
    const Offset addrBytesPerRelrBitmap = (8 * word_size - 1) * word_size;

    std::vector<Elf_Relr> packed;
    Offset bitmapBeginAddr = 0;  // first addr covered by bitmap
    Offset bitmapEndAddr = 0;    // first addr not covered by bitmap
    Elf_Relr bitmap = 0;
    for (size_t i = 0; i < addrs.size(); ) {
        auto a = addrs[i];
        if (bitmapBeginAddr <= a && a < bitmapEndAddr) {
            auto offset = a - bitmapBeginAddr;
            if ((offset % word_size) == 0) {                 // check if aligned
                ++i;                                         // consume addr
                bitmap |= 1UL << ((offset / word_size) + 1); // set bit
            } else {
                bitmapBeginAddr = bitmapEndAddr;  // unaligned, invalidate bitmap, try again
            }
            continue;
        }
        if (bitmap) {
            bitmap |= 1;                             // set bit 0, indicating a bitmap
            packed.push_back(bitmap);                // add RELR entry
            bitmap = 0;
            if (bitmapBeginAddr != bitmapEndAddr) {  // set new bitmap range, if valid
                bitmapBeginAddr = bitmapEndAddr;
                bitmapEndAddr += addrBytesPerRelrBitmap;
            }
            continue;  // try again, the value might be in new bitmap range
        }
        ++i;                              // consume addr
        packed.push_back(a);              // add RELR entry
        bitmapBeginAddr = a + word_size;  // set new bitmap range
        bitmapEndAddr = bitmapBeginAddr + addrBytesPerRelrBitmap;
    }
    if (bitmap) {                         // if final bitmap
        bitmap |= 1;                      // set bit 0, indicating a bitmap
        packed.push_back(bitmap);         // add RELR entry
    }
```

symtabAPI/src/emitElf.C
```diff
@@ -631,7 +652,18 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
                 (strcmp(name, ".init_array") == 0 || strcmp(name, ".fini_array") == 0 ||
                  strcmp(name, "__libc_subfreeres") == 0 || strcmp(name, "__libc_atexit") == 0 ||
                  strcmp(name, "__libc_thread_subfreeres") == 0 || strcmp(name, "__libc_IO_vtables") == 0)) {
+            std::vector<Offset> &relr_relocs = object->getRelrDynRelocs();
             for(std::size_t off = 0; off < newdata->d_size; off += sizeof(void*)) {
+                Offset addr = shdr->sh_addr + off;
+                bool is_relr_location = false;
+                for (unsigned ri = 0; ri < relr_relocs.size(); ++ri) {
+                    if (relr_relocs[ri] == addr) {
```
kupsch: Please add a comment describing the purpose of this?

Any idea if this is a performance issue `O(num_relr * size_of_newdata)`?

kcossett-amd: Adding comment.
Regarding performance, it shouldn't be a real issue. The loop would only run for a small set of sections which do not hold many entries. That said, I'll switch to building an `unordered_set` and using `count()` for the check.

kupsch (approved):
This looks good. @hainest, any comments.

kupsch:
resolves #1290

hainest:
@kupsch @bbiiggppiigg I'll take a look. I'm trying to make a reproducer regression test based on a binary from Ubuntu 25.04. I might just add newer Ubuntus to our collection of test containers instead. TBD. I don't want to hold this PR up much longer, so I'll probably just run it through the usual testing battery for now.

symtabAPI/src/emitElf.C
```diff
@@ -559,6 +559,27 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
             newshdr->sh_entsize = 0x0;
         }
 
+        if (library_adjust > 0 && newdata->d_buf && newdata->d_size) {
+            std::vector<Offset> &relr_relocs = object->getRelrDynRelocs();
+            for (unsigned ri = 0; ri < relr_relocs.size(); ++ri) {
```
hainest:
```suggestion
            for (Offset relr_addr : object->getRelrDynRelocs())
```

symtabAPI/src/emitElf.C
```diff
@@ -559,6 +559,27 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
...
+                    relr_addr + sizeof(Elf_Relr) > shdr->sh_addr + shdr->sh_size)
+                    continue;
+
+                Offset relr_off = relr_addr - shdr->sh_addr;
+                if (relr_off + sizeof(Elf_Relr) > newdata->d_size)
+                    continue;
+
+                char *loc = static_cast<char *>(newdata->d_buf) + relr_off;
+                Elf_Relr val = 0;
+                memcpy(&val, loc, sizeof(val));
```
hainest: Please use the memory helpers in `unaligned_memory_access.h`.

```suggestion
                auto val = Dyninst::read_memory_as<Elf_Relr>(loc);
```

symtabAPI/src/emitElf.C
```diff
@@ -559,6 +559,27 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
...
+                if (!val) continue;
+                val += library_adjust;
+                memcpy(loc, &val, sizeof(val));
```
hainest:
```suggestion
                Dyninst::write_memory_as(loc, val);
```

symtabAPI/src/emitElf.C
```diff
@@ -631,7 +652,15 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
...
                  strcmp(name, "__libc_subfreeres") == 0 || strcmp(name, "__libc_atexit") == 0 ||
                  strcmp(name, "__libc_thread_subfreeres") == 0 || strcmp(name, "__libc_IO_vtables") == 0)) {
+            std::vector<Offset> &relr_relocs = object->getRelrDynRelocs();
+            // Every pointer-sized word in this section is shifted by library_adjust.
+            // A word that is also a RELR relocation target was already shifted by the
+            // earlier RELR loop, so skip to avoid offsetting it twice
+            std::unordered_set<Offset> relr_addrs(relr_relocs.begin(), relr_relocs.end());
             for(std::size_t off = 0; off < newdata->d_size; off += sizeof(void*)) {
+                Offset addr = shdr->sh_addr + off;
+                if (relr_addrs.count(addr)) continue;  // already adjusted as a RELR reloc
```
hainest, line 659: @kupsch This check can go away once `library_adjust` is removed, no?

symtabAPI/src/emitElf.C
```diff
@@ -952,6 +981,19 @@ void emitElf<ElfTypes>::fixPhdrs() {
 #if !defined(DT_TLSDESC_GOT)
 #define DT_TLSDESC_GOT 0x6ffffef7
 #endif
+// Older elf.h headers may not define RELR dynamic tag constants
```
hainest, line 981: Side note for future us: the current minimum required elfutils is 0.186, RELR was added 0.190.

symtabAPI/src/emitElf.C
```diff
@@ -1990,6 +2050,75 @@ void emitElf<ElfTypes>::createRelocationSections(std::vector<relocationEntry> &r
...
+    memcpy(relrs, packed.data(), sizeof(Elf_Relr) * packed.size());
+
+    dyn_hash_map<int, Region *> secTagRegionMapping = object->getTagRegionMapping();
+    string name;
+    if (secTagRegionMapping.find(DT_RELR) != secTagRegionMapping.end())
+        name = secTagRegionMapping[DT_RELR]->getRegionName();
+    else
+        name = ".relr.dyn";
+
+    obj->addRegion(0, relrs, packed.size() * sizeof(Elf_Relr), name,
```
hainest:
```suggestion
    obj->addRegion(0, relrs, packed.size() * sizeof(Elf_Relr), std::move(name),
```

hainest (changes requested):
There were no regressions on 20 machines. I also created a tiny rewriter test that successfully rewrote and executed `/usr/bin/iconv` in Ubuntu-25.04 which has RELR. I will add it in a separate PR since it needs an update to dyninst/external-tests.

symtabAPI/src/emitElf.C
kupsch: yes, this will go away

symtabAPI/src/emitElf.C
kcossett-amd: I'll resolve this comment since no code-changes are required.

symtabAPI/src/emitElf.C
kcossett-amd: I'll mark this as resolved

hainest (approved):
No regressions on 19 machines. Confimed can rewrite iconv on Ubuntu-25.04. LGTM. Thanks so much for your patient work!

PR #2236, opened 2026-05-01
Remove BPatch_threadIndexExpr

hainest:
It generates a call to DYNINSTthreadIndex, but that function was removed by 0e791cce8c95f in 2015.


hainest:
No regressions.

kupsch (approved):

PR #2237, opened 2026-05-01
Remove BPatch_shadowExpr

hainest:
It generates a call to RThandleShadow which was disabled by 89919d982 in 2012 and removed by 0b9df83c9 in 2021.


hainest:
No regressions.

kupsch (approved):

PR #2238, opened 2026-05-02
Convert dyninstAPI docs to LaTeX

hainest:
This also removes references to removed classes. I removed Appendix B that talked about how to run the test suite since it is now outside of the Dyninst repo and has its own instructions.

The docs say that `BPatch_Set` was deprecated ages ago. It's still being used in several interfaces. Should we remove now while we can take an ABI break?


kupsch:
I think `BPatch_Set` can be removed.

kupsch (approved):

PR #2242, opened 2026-05-04
Remove disabledItem heap class

hainest:
It only exists in inferiorHeap::disabledList, but its usage was removed by 40607f1e2 in 2002.


hainest:
No test regressions.

kupsch (approved):

PR #2244, opened 2026-05-06
[AMDGPU GFX908] [Emitter] Allow emitMoveRegToReg to move register blocks

rochauha:
So far, we supported moving single registers using the emitter API on AMDGPU. For register blocks, this required getting individual registers in the blocks and moving them separately.

This PR makes the emitter API handle this rather.

### DIFF
--- a/dyninstAPI/src/ASTs/atomicOperationAST.C
+++ b/dyninstAPI/src/ASTs/atomicOperationAST.C
@@ -76,12 +76,7 @@ bool atomicOperationAST::generateCode_phase2(codeGen &gen, Address &retAddr,
   // TODO this needs to pick up the register from placeholderReg
   Dyninst::Register baseRegPair = Dyninst::Register(OperandRegId(94), RegKind::SCALAR, BlockSize(2));
 
-  // TODO: make movs work with blocks.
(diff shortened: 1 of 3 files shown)


hainest (approved):
Looks reasonable to me. Is there an instruction that moves more than one register at a time?

kupsch (approved):

rochauha:
> Looks reasonable to me. Is there an instruction that moves more than one register at a time?

There is a `s_mov_b64` that can be used for aligned pairs only, otherwise the assembler complains. Do you think I should add support for it? If yes, should it be in this or a separate PR?

Then the general approach will then be a combination of paired mov's and regular mov's when the block is big enough.

hainest:
> > Looks reasonable to me. Is there an instruction that moves more than one register at a time?
> 
> There is a `s_mov_b64` that can be used for aligned pairs only, otherwise the assembler complains. Do you think I should add support for it? If yes, should it be in this or a separate PR?

It would just be an optimization, so we don't have to do it unless you feel like it.

> Then the general approach will then be a combination of paired mov's and regular mov's when the block is big enough.

Are any of the ASTs currently using blocks of size more than two?

bbiiggppiigg (approved):

PR #2247, opened 2026-05-09
GitHubCI: update docker login action

hainest:
GitHub CI is moving not Node.js 24 in June which drops support for this version of the action.


hainest:
I can run the dev deploy workflow with this, so it's working.

kupsch (approved):

PR #2248, opened 2026-05-10
Extract emitVstore into Emitter class

hainest:


bigtrak:
Tim; keep emitVstore() as part of operator AST, (or whatever it is -- having auth problems with github
at the moment so I can't review at a larger sacle)  then just have that do gen.emitter()->vstore()
under the covers.   The source code churn is smaller that way, and it can always be mentioned
as inline if code generation is a concern.   Doing it that way also makes it easier to debug
emitter issues, as the local decl can be used to debug emitter operations on any architecture,
instead of needing to go hack individual architectures for generic debugging issues.

hainest:

hainest:
> Tim; keep emitVstore() as part of operator AST, (or whatever it is -- having auth problems with github at the moment so I can't review at a larger sacle) then just have that do gen.emitter()->vstore() under the covers. The source code churn is smaller that way, and it can always be mentioned as inline if code generation is a concern. Doing it that way also makes it easier to debug emitter issues, as the local decl can be used to debug emitter operations on any architecture, instead of needing to go hack individual architectures for generic debugging issues.

The goal is to move all of the free functions into the Emitter class so I can split them up into meaningful functions. The hope here is to build an interface that actually applies to all architectures, rather than having the very sparsely-implemented version that exists.

kupsch (approved):

PR #2250, opened 2026-05-11
Extract, update isPowerOf2

hainest:
This introduces a redundant calculation of `ilog2` (e.g., once in `Emitterx86::can_optimize_as_shift` and again in `Emitterx86::emitDivImm`). However for GNU-ish compilers that use the instrinsics, the number of instructions is a quarter of the number needed for the previous loop implementation.

I don't know why aarch doesn't use this optimization for mul and div or why ppc only uses it for mul.


hainest:

kupsch (approved):

PR #2251, opened 2026-05-11
Move BPatch source files into subdirectory

hainest:
This is purely a cosmetic change to help organize the dyninstAPI toolkit. It will allow `git log --follow` to work correctly since no substantial content has changed.


hainest:
No regressions.

kupsch (approved):

PR #2254, opened 2026-05-13
Fix function_cache::contains implementation

### LINKED ISSUE #2253: Is function_cache::find_if implemented in reverse?
I was using AI to implement some AMDGPU stuff, and it noted me that the logic of find_if might have been implemented in reverse.

Namely

https://github.com/dyninst/dyninst/blob/14f360e7814e0a9606a3499b26317f68403da539/dyninstAPI/src/function_cache.h#L46-L52

contains return true when find_if reaches func.end(), which means it actually didn't find a match.
And this might be relevant to the some slight performance boost someone have observed?

@hainest

bbiiggppiigg:
Previous logic implemented contains in reverse.
Fixes #2253

### DIFF
--- a/dyninstAPI/src/function_cache.h
+++ b/dyninstAPI/src/function_cache.h
@@ -48,7 +48,7 @@ namespace Dyninst { namespace DyninstAPI {
         return c->addr() == f->addr() &&
                c->typedName() == f->typedName();
       };
-      return std::find_if(funcs.begin(), funcs.end(), cmp) == funcs.end();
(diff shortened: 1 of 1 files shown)


kupsch (approved):
I fixed the spelling of `contains` (was `conatains`) in the PR title, but is wrong in the commit message. Change the spelling in the commit before merging.

hainest (approved):
I'll try to re-check the performance regression this afternoon. If it's still an issue, I'll submit a separate update.

Issue #2256, opened 2026-05-13
Incorrect container iteration in IntervalMap


bbiiggppiigg:
**Describe the bug**

In both 
https://github.com/dyninst/dyninst/blob/14f360e7814e0a9606a3499b26317f68403da539/dataflowAPI/rose/util/IntervalMap.h#L224

and

https://github.com/dyninst/dyninst/blob/14f360e7814e0a9606a3499b26317f68403da539/dataflowAPI/rose/util/IntervalMap.h#L236

The iteration through be through otherIter intead of other

PR #2257, opened 2026-05-13
Fix iterator variable in IntervalMap copy ctor and assignment

bbiiggppiigg:
The loops in the templated copy constructor and assignment operator used `other` (the container) instead of `otherIter` (the iterator) in the loop condition and increment, which would fail to compile if these templates were ever instantiated.


bbiiggppiigg:
Fixes #2256

hainest:
This is in old ROSE code, but we can probably fix it anyway.

hainest:

hainest (approved):

PR #2258, opened 2026-05-16
Reorganize dyninstAPI CMake structure

hainest:
This is purely cosmetic, but really needed because the number of files will continue to grow as codegen is made platform-independent.

### DIFF
--- a/dyninstAPI/CMakeLists.txt
+++ b/dyninstAPI/CMakeLists.txt
@@ -140,8 +140,7 @@ if(DYNINST_CODEGEN_ARCH_I386 OR DYNINST_CODEGEN_ARCH_X86_64)
     src/codegen-x86.C
     src/IAPI_to_AST.C)
 elseif(DYNINST_CODEGEN_ARCH_PPC64LE)
-  list(APPEND _sources src/inst-power.C src/codegen-power.C src/RegisterConversion-ppc.C
-       src/trampolines/baseTramp-ppc.C)
+  list(APPEND _sources src/inst-power.C src/codegen-power.C src/RegisterConversion-ppc.C)
 elseif(DYNINST_CODEGEN_ARCH_AARCH64)
   list(APPEND _sources src/inst-aarch64.C src/emit-aarch64.C src/codegen-aarch64.C
        src/RegisterConversion-aarch64.C)
--- a/dyninstAPI/CMakeLists.txt
+++ b/dyninstAPI/CMakeLists.txt
@@ -40,8 +40,10 @@ set(_public_headers
 add_subdirectory(src/ASTs)
 add_subdirectory(src/codegen)
 add_subdirectory(src/registerSpace)
+add_subdirectory(src/Relocation)
 
-set(_subprojects dyninstAPI_ASTs dyninstAPI_codegen dyninstAPI_registerSpace)
+set(_subprojects dyninstAPI_ASTs dyninstAPI_codegen dyninstAPI_registerSpace
+                 dyninstAPI_relocation)
 
 set(_private_headers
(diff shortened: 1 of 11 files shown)


hainest:
No build issues.

hainest:
I'm a dumb-dumb. I could just make them interface targets instead of carrying around a bunch of variables. I'll change them.

dyninstAPI/src/ASTs/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+    ${_base_dir}/scrambleRegistersAST.C
+    ${_base_dir}/sequenceAST.C
+    ${_base_dir}/snippetAST.C
+    ${_base_dir}/stackAST.C
+    ${_base_dir}/stackInsertionAST.C
+    ${_base_dir}/stackRemovalAST.C
+    ${_base_dir}/variableAST.C)
+
+if(DYNINST_CODEGEN_ARCH_AMDGPU_GFX908)
+  list(APPEND _ASTS_headers ${_base_dir}/AmdgpuEpilogue.h ${_base_dir}/AmdgpuPrologue.h)
```
bbiiggppiigg: This should be _ASTs_headers not the upper case _ASTS_headers ?

bbiiggppiigg: And I wonder why this is not caught be CI.
Are these headers not really needed.

bigtrak: Do the CI builds include building for the AMDGPU architectures, as well as installing
all the AMD GPU software install/dependency to do so?

hainest: > And I wonder why this is not caught be CI. Are these headers not really needed.

CI just does a build once. If you were modifying the files, the bug here would just cause them to not be recompiled.

hainest: > This should be _ASTs_headers not the upper case _ASTS_headers ?

Derp, yes. Good catch.

hainest: > Do the CI builds include building for the AMDGPU architectures, as well as installing all the AMD GPU software install/dependency to do so?

The [codegen-cross-build](https://github.com/dyninst/dyninst/actions/workflows/codegen-cross-build.yaml) workflow sets `DYNINST_CODEGEN_ARCH`, but doesn't do anything, otherwise. In essence, we just check that the code builds and that the codegen tests pass.

hainest:
I've updated to use INTERFACE targets which simplifies things greatly. No regressions on 19 machines.

dyninstAPI/CMakeLists.txt
```diff
@@ -192,20 +186,13 @@ if(DYNINST_CODEGEN_ARCH_I386 OR DYNINST_CODEGEN_ARCH_X86_64)
     src/inst-x86.C
     src/emit-x86.C
     src/codegen-x86.C
-    src/IAPI_to_AST.C
-    src/trampolines/baseTramp-x86.C)
+    src/IAPI_to_AST.C)
 elseif(DYNINST_CODEGEN_ARCH_PPC64LE)
   list(APPEND _sources src/inst-power.C src/codegen-power.C src/RegisterConversion-ppc.C
        src/trampolines/baseTramp-ppc.C)
```
bbiiggppiigg: Is there a reason why baseTramp is kept for ppc but no the other architectures?

hainest: No, that was a rebase error. Good catch. Fixed.

bbiiggppiigg (approved):

PR #2263, opened 2026-05-25
[dyninstAPI] Use simple pointer equality in function_cache

hainest:
The current test doesn't seem to be correctly unique. This reverts back to the old pointer-based comparison.

I evidently just completely forgot to run the test battery on #2254 because it caused all sorts of regressions.


hainest:
No regressions on 19 machines (and fixes failures). I actually checked this time.

@bbiiggppiigg Could you take a look at this since Jim is out this week?

bbiiggppiigg (approved):

Issue #356, opened 2017-03-22
assert fails at insnCodeGen::generateBranchViaTrap


stanfordcox:
insnCodeGen::generateBranchViaTrap on ppc linux using dyninst 9.3.1.1 can get an assertion failure.
Seen in e.g. test1_4, test1_10
(gdb) list
	    if (ABS(disp) <= MAX_BRANCH) {
	        // We shouldn't be here, since this is an internal-called-only func.
	        return generateBranch(gen, disp, isCall);
	    }
	    assert (isCall == false); // Can't do this yet

(gdb) bt 10
#0  0x00003fffb7750f50 in __GI_raise (sig=<optimized out>) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00003fffb7752f58 in __GI_abort () at abort.c:90
#2  0x00003fffb7746cac in __assert_fail_base (fmt=0x3fffb789bc00 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x3fffb75f7b08 "isCall == false", 
    file=0x3fffb75f7ab0 "/builddir/build/BUILD/dyninst-9.3.1/dyninst-9.3.1/dyninstAPI/src/codegen-power.C", line=<optimized out>, 
    function=<optimized out>) at assert.c:92
#3  0x00003fffb7746d9c in __GI___assert_fail (assertion=0x3fffb75f7b08 "isCall == false", 
    file=0x3fffb75f7ab0 "/builddir/build/BUILD/dyninst-9.3.1/dyninst-9.3.1/dyninstAPI/src/codegen-power.C", line=<optimized out>, 
    function=0x3fffb75f74d8 <insnCodeGen::generateBranchViaTrap(codeGen&, unsigned long, unsigned long, bool)::__PRETTY_FUNCTION__> "static void insnCodeGen::generateBranchViaTrap(codeGen&, Dyninst::Address, Dyninst::Address, bool)") at assert.c:101
#4  0x00003fffb75b56d8 in insnCodeGen::generateBranchViaTrap (gen=..., from=234946896, to=to@entry=268507212, 
    isCall=<optimized out>) at /usr/src/debug/dyninst-9.3.1/dyninst-9.3.1/dyninstAPI/src/codegen-power.C:259
#5  0x00003fffb75b6c40 in insnCodeGen::modifyJump (target=268507212, insn=..., gen=...)
[...]

PR #2265, opened 2026-05-29
[dyninstAPI] Clean up generateBranchViaTrap on PowerPC

hainest:
There was a lot of discussion in
https://github.com/dyninst/dyninst/issues/175 and
https://github.com/dyninst/dyninst/issues/356 about what the correct thing to do here is. It looks like neither suggestion was ever implemented. I'm removing the commented-out code to reduce noise.

The "potential strategy" outline is _NOT_ implemented. I'm leaving that for possible future usage.

### DIFF
--- a/dyninstAPI/src/codegen-power.C
+++ b/dyninstAPI/src/codegen-power.C
@@ -435,49 +435,29 @@ void insnCodeGen::generateBranchViaTrap(codeGen &gen, Dyninst::Address from, Dyn
         // We shouldn't be here, since this is an internal-called-only func.
         return generateBranch(gen, disp, isCall);
     }
-    //assert (isCall == false); // Can't do this yet
-    if (isCall) {
-      // Screw using a trap, just emit a call and save/restore all registers (painful but whatever).
-      //emitCall()
-      assert(isCall == false);       
-      //assert(shouldAssertIfInLongBranch != true);
-      // failedLongBranchLocal = true;
-    } else {    
-      if (gen.addrSpace()) {
-          // Too far to branch.  Use trap-based instrumentation.
+    assert(isCall == false);
+    if (gen.addrSpace()) {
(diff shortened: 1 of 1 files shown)


hainest:

dyninstAPI/src/codegen-power.C
```diff
@@ -435,49 +435,29 @@ void insnCodeGen::generateBranchViaTrap(codeGen &gen, Dyninst::Address from, Dyn
         // We shouldn't be here, since this is an internal-called-only func.
         return generateBranch(gen, disp, isCall);
     }
-    //assert (isCall == false); // Can't do this yet
-    if (isCall) {
-      // Screw using a trap, just emit a call and save/restore all registers (painful but whatever).
-      //emitCall()
-      assert(isCall == false);
```
bbiiggppiigg, line 442: I think there should still be some sort of error reporting to the users who intended to call generateBranchViaTrap with isCall == true.

hainest: There is. Instead of the confusing

```c++
if (isCall) {
  assert(isCall == false);
}
```

I just used the previous `assert(isCall == false);` from line 438 and removed the `if(isCall)`.

bbiiggppiigg: I don't know how I missed it.
But I think we should keep the comment // Can't do this yet.

bbiiggppiigg (approved):

dyninstAPI/src/codegen-power.C
hainest: > But I think we should keep the comment // Can't do this yet.

I was going to add an error message, but ultimately decided it wasn't any worse than what was already there. I went ahead and added some logging for this case.

bbiiggppiigg (approved):

PR #2267, opened 2026-05-31
Fix type width in IndirectAnalyzer::ReadTable

### LINKED ISSUE #2264: CodeObject::parse() runs forever on python shared library
**Describe the bug**

One of our users reported the following problem.

`Parser::parse()` (called from hpcstruct) runs forever on a 9.5 Meg
python shared library, `hf_transfer.abi3.so`.

I used `hpcrun` and `gdb` to investigate a little.
It's not exactly a "hang", it's not stuck on a lock.

Instead, `parse()`, when run with multiple threads, quickly comes down
to one thread while all others sit idle.  It's not stuck in one single
function.  Most of the call stacks look like (omitting most frames):
```
Dyninst::ParseAPI::Parser::parse()
IndirectControlFlowAnalyzer::FindJunkInstruction()
Dyninst::InstructionAPI::InstructionDecoder::decode()
```
But many of the call stacks end in a destructor of some type:
instruction, shared_ptr, pair, etc.

It's curious that memory size does NOT increase over time.
AFAICT, the code thinks it's making progress, but at some point,
it does something that erases or invalidates the progress and then
it starts over and keeps repeating.  But that's just a guess.

**To Reproduce**

The binary is `hf_transfer.abi3.so`, a python library for fast data
transfers across network.
```
-rwxr-xr-x. 1 krentel users 9653960 May 24 10:22 hf_transfer.abi3.so*
md5sum: c84afb99615151688ffb23e9df4b1cee  hf_transfer.abi3.so
```
You can get this from `pip3 install hf_transfer` if your system is new
enough, then fish it out of:
```
~/.local/lib/python3.14/site-packages/hf_transfer/
```
I also put a copy on ufront in:
```
/home/krentel/Files/hf_transfer.abi3.so
```
I ran this from `hpcstruct` as:
```
hpcstruct -j 4 --time hf_transfer.abi3.so
```
But you could also use any unit test that calls `CodeObject::parse()`.

This is on x86_64, RHEL 8.10, glibc-2.28,
spack built dyninst and prereqs with gcc 14.3.0.

Dyninst head of master as of yesterday, May 26,
commit 5d2c62a4ca3638f8bb04d5c596b4d928ea595998.

**Additional context**

Notes:

1. The binary is only 9.5 Meg and I let it run for 16 hours, so I
really think it's stuck not making progress and will never finish.

2. There is one large function, reported by objdump:
```
00000000003e2d80 <ossl_aes_gcm_decrypt_avx512>:
0000000000435480 <ossl_aes_gcm_finalize_avx512>:
```
But that's only 300K, I've seen much worse that finish in less than 30 min.

3. The binary has a lot of vector instructions:
`vpxorq`, `vpslldq`, `vpclmulhqhqdq`, `vpternlogq`, etc.
Maybe dyninst has trouble with some of these.

4. I ran this with hpcstruct and there are a lot of vector instructions,
so probably it heavily uses the unknown instruction interface.
Perhaps that causes trouble.

Any suggestions for turning on debug info inside `parse()`?

I could prepare an hpctoolkit database for you, if you wish.

bbiiggppiigg:
indexBound.stride is 64 bit, v and start is 32 bit.
If v + indexBound.stride results in number with low 32 bit 0,
we get an infinite loop.

This PR changes v and start to be int64_t so the truncation 
no longer happens.

Also fix an logic error with respect to assignment of start, so
the lowerbound is actually used.

Fixes #2264

### DIFF
--- a/parseAPI/src/IndirectAnalyzer.C
+++ b/parseAPI/src/IndirectAnalyzer.C
@@ -339,7 +339,7 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
             if (jtrv.readAddress > maxReadAddress) maxReadAddress = jtrv.readAddress;
             if (prevReadAddress > 0) indexStride = jtrv.readAddress - prevReadAddress;
             prevReadAddress = jtrv.readAddress;
-            parsing_printf("\t index %d, table address %lx, target address %lx\n", v, jtrv.readAddress, jtrv.targetAddress);
+            parsing_printf("\t index %ld, table address %lx, target address %lx\n", v, jtrv.readAddress, jtrv.targetAddress);
             jumpTargets.insert(jtrv.targetAddress);
             entries[jtrv.readAddress] = jtrv.targetAddress;
         } else {
--- a/parseAPI/src/IndirectAnalyzer.C
+++ b/parseAPI/src/IndirectAnalyzer.C
@@ -324,10 +324,10 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
                                             std::map<Address, Address>& entries) {
     CodeSource *cs = block->obj()->cs();
     set<Address> jumpTargets;
-    int start = 0;
+    int64_t start = 0;
     Address prevReadAddress = 0;
     if (indexBound.low > 0) start = indexBound.low = start;
-    for (int v = start; v <= indexBound.high; v += indexBound.stride) {
+    for (int64_t v = start; v <= indexBound.high; v += indexBound.stride) {
         JumpTableReadVisitor jtrv(index, v, cs, block->region(), isZeroExtend, memoryReadSize);
         jumpTargetExpr->accept(&jtrv);
        if (jtrv.valid && cs->isCode(jtrv.targetAddress)) {
--- a/parseAPI/src/IndirectAnalyzer.C
+++ b/parseAPI/src/IndirectAnalyzer.C
@@ -326,7 +326,7 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
     set<Address> jumpTargets;
     int64_t start = 0;
     Address prevReadAddress = 0;
-    if (indexBound.low > 0) start = indexBound.low = start;
+    if (indexBound.low > 0) start = indexBound.low;
     for (int64_t v = start; v <= indexBound.high; v += indexBound.stride) {
         JumpTableReadVisitor jtrv(index, v, cs, block->region(), isZeroExtend, memoryReadSize);
         jumpTargetExpr->accept(&jtrv);
--- a/parseAPI/src/IndirectAnalyzer.C
+++ b/parseAPI/src/IndirectAnalyzer.C
@@ -324,10 +324,9 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
                                             std::map<Address, Address>& entries) {
     CodeSource *cs = block->obj()->cs();
     set<Address> jumpTargets;
-    int64_t start = 0;
     Address prevReadAddress = 0;
-    if (indexBound.low > 0) start = indexBound.low;
-    for (int64_t v = start; v <= indexBound.high; v += indexBound.stride) {
+    auto start = std::max(indexBound.low,0L);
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
@mwkrentel 
I believe this PR fixes your issue.
Worked on my side.
Can you test this PR when you have the time.

hainest:
If I've read this correctly, it seems that `indexBound.high` is more than 4GB ahead of `indexBound.low`. That seems like we're making a wild guess as to where the jump table is. Is this the result of bad parsing due to missing/unparseable blocks stemming from undecodable instructions?

bbiiggppiigg:
I don't think so.
The indirect control flow instruction is a jmp rax, and Dyninst claim to find a bound for rax which is 0x1,0000,0000.
image
As a result (int) (v+indexBound.stride) == 0.

bbiiggppiigg:
For this issue in particular, it is not related to invalid instructions. But we can never know whether the next reported bug is related to it, so on and so forth.

hainest:
> I don't think so. The indirect control flow instruction is a jmp rax, and Dyninst claim to find a bound for rax which is 0x1,0000,0000.  As a result (int) (v+indexBound.stride) == 0.

I understand the bug and the fix. My concern is _why_ Dyninst thinks the bound is 4GB away. That seems like the parser is just making a wild guess based on arbitrary values representable by integers rather than anything to do with the actual data in the file (e.g., the ranges of any .text sections). I think that would be the more fundamental bug.

bbiiggppiigg:
> > I don't think so. The indirect control flow instruction is a jmp rax, and Dyninst claim to find a bound for rax which is 0x1,0000,0000.  As a result (int) (v+indexBound.stride) == 0.
> 
> I understand the bug and the fix. My concern is _why_ Dyninst thinks the bound is 4GB away. That seems like the parser is just making a wild guess based on arbitrary values representable by integers rather than anything to do with the actual data in the file (e.g., the ranges of any .text sections). I think that would be the more fundamental bug.

It seems like the problem is that Dyninst disassemble the movzbl at 0x4a19af as movzx instead of movzbl.
movzbl implies a single byte bound, whereas movzx 0x1f(%rsp), %eax pretty much gives no bound information, and got worsen by the shlq 0x20, %rax at 0x4a198c.

image

bbiiggppiigg:
> > I don't think so. The indirect control flow instruction is a jmp rax, and Dyninst claim to find a bound for rax which is 0x1,0000,0000.  As a result (int) (v+indexBound.stride) == 0.
> 
> I understand the bug and the fix. My concern is _why_ Dyninst thinks the bound is 4GB away. That seems like the parser is just making a wild guess based on arbitrary values representable by integers rather than anything to do with the actual data in the file (e.g., the ranges of any .text sections). I think that would be the more fundamental bug.

You're right. The problem is that Dyninst/rose doesn't deal with shift constants > 32 correctly.
I've posted a separate PR #2269.

parseAPI/src/IndirectAnalyzer.C
```diff
@@ -324,10 +324,10 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
                                             std::map<Address, Address>& entries) {
     CodeSource *cs = block->obj()->cs();
     set<Address> jumpTargets;
-    int start = 0;
+    int64_t start = 0;
     Address prevReadAddress = 0;
-    if (indexBound.low > 0) start = indexBound.low = start;
-    for (int v = start; v <= indexBound.high; v += indexBound.stride) {
+    if (indexBound.low > 0) start = indexBound.low;
```
hainest: The other option is to use `auto start = std::max(indexBound.low, 0L);`. Since `std::max` requires the type of each parameter to be the same, that gives us a compile-time check on this in the future.

parseAPI/src/IndirectAnalyzer.C
```diff
@@ -324,10 +324,10 @@ void IndirectControlFlowAnalyzer::ReadTable(AST::Ptr jumpTargetExpr,
...
+    for (int64_t v = start; v <= indexBound.high; v += indexBound.stride) {
```
hainest:
```suggestion
    for (auto v = start; v <= indexBound.high; v += indexBound.stride) {
```

This will make sure the type of `v` is always correct.

hainest (approved):
A couple of minor suggestions. No regressions on 19 machines.

kupsch (approved):

PR #2268, opened 2026-05-31
Implement Optional Zydis decoding for x86

bbiiggppiigg:
This PR adds a default unknown instruction callback using Zydis,
when the user specify the options DYNINST_ENABLE_ZYDIS and
Zydis_ROOT.


hainest:
Why are we bringing Zydis into Dyninst? The whole purpose of the callback mechanism is to let users handle the decoding.

bbiiggppiigg:
Because whenever a user reports a bug, we will never know whether it is because of an unknown instruction or something else.

The first step of debugging would then be to implement the callback to make sure we are at least looking at the right instruction, then proceed.

bbiiggppiigg:
@kupsch What are your thoughts on this, as a debugging knob until we finally get capstone support integrated.

bigtrak:
This is pretty nice -- this  mechanism can be used as a prototype for _any_ of the outboard 
instruction decoders to discover the size of an unknown instruction; zydis, xed, or even capstone 
as a tool used by dyninst in it's current form.   Or for example when dyninst is lagging behind
capstone (or capstone is lagging behind reality) and other tools are updated faster.

That has always been an issue with capstone, it is not always the most up-to-date set of tools.

I think it's a worthwhile thing to add... and it even lets us offload problems and effort
to a bug  reporter, as they can turn it on and say "this is what is going on with detail" and send in
better / easier trouble reports.  Especially folks like hpcstruct which often deal with newer
machines than we have access to on a regular basis.

cmake/DyninstOptions.cmake
```diff
@@ -57,6 +57,11 @@ option(DYNINST_ENABLE_FILEFORMAT_PE "Enable PE (Windows .exe/.dll) parsing" OFF)
 
 option(DYNINST_ENABLE_CAPSTONE "Enable Capstone" OFF)
 
+option(
```
hainest, line 60: Add a comment here that this should not be used and will be removed in the future without notice.

cmake/DyninstOptions.cmake
```diff
@@ -57,6 +57,11 @@ option(DYNINST_ENABLE_FILEFORMAT_PE "Enable PE (Windows .exe/.dll) parsing" OFF)
...
+  DYNINST_ENABLE_ZYDIS
+  "Use Zydis to recover the length of x86-64 instructions InstructionAPI cannot decode"
+  OFF)
+
```
kupsch, line 65: Let's mark this as advanced also:
```suggestion
  DYNINST_ENABLE_ZYDIS
  "Use Zydis to recover the length of x86-64 instructions InstructionAPI cannot decode"
  OFF)
mark_as_advanced(DYNINST_ENABLE_ZYDIS)
```

instructionAPI/src/InstructionDecoder.C
```diff
@@ -31,14 +31,43 @@
...
 #include <array>
 
+#if defined(DYNINST_ENABLE_ZYDIS)
+#include <Zydis/Zydis.h>
+#endif
+
 namespace {
   namespace ia = Dyninst::InstructionAPI;
   using ui = ia::InstructionDecoder::unknown_instruction;
   ui::callback_t callback{};
```
kupsch: If you're going to have this option, how about initializing `callback{default_x86_callback}`?

instructionAPI/src/InstructionDecoder.C
```diff
@@ -31,14 +31,43 @@
...
+
+#if defined(DYNINST_ENABLE_ZYDIS)
+  // Built-in "unknown instruction" callback for x86-64. InstructionAPI calls
+  // this when it cannot decode a byte sequence; we only need the length of the
+  // instruction so the decoder can skip over it, so we hand back a no-op of the
+  // recovered length (its bytes and control flow are never inspected).
+  ia::Instruction default_x86_callback(ia::InstructionDecoder::buffer seqn) {
```
kupsch: better function name:
```suggestion
  ia::Instruction zydis_unknown_instruction_callback(ia::InstructionDecoder::buffer seqn) {
```

instructionAPI/src/InstructionDecoder.C
```diff
@@ -91,4 +120,13 @@ namespace Dyninst { namespace InstructionAPI {
     ::callback = nullptr;
     return c;
   }
+
+  bool InstructionDecoder::unknown_instruction::register_default_callback() {
+#if defined(DYNINST_ENABLE_ZYDIS)
+    ::callback = &default_x86_callback;
```
kupsch: use `register_callback` instead of assignment

kupsch (commented):
I'm OK with the temporary concept of this, but I have suggestions. The functions to set the default can stay, but users shouldn't have to call them.

instructionAPI/src/InstructionDecoder.C
bbiiggppiigg: I'm not sure if I'll be able to do that if default_x86_callback is only available when DYNINST_ENABLE_ZYDIS is set. Unless you mean 

#if defined(...)
callback{default_x86_callback}
#else
callback{}
#endif

kupsch: yes that is what I meant. If it's set then users don't have get zydis unimplemented instruction decoding without any addition changes to their code. It can still be unset and installed again. If not set then the behavior is as before.

cmake/tpls/DyninstZydis.cmake
```diff
@@ -0,0 +1,37 @@
...
+#
+# Zydis ships a CMake package config that exports the
+# Zydis::Zydis target. If the user specifies Zydis_ROOT_DIR,
+# we restrict the search to that location (or CMAKE_PREFIX_PATH).
+#
+#===========================================================
+
+include_guard(GLOBAL)
+
+if(Zydis_ROOT)
```
hainest, line 17: By convention, the name is the same capitalization as the project's cmake directory. Zydis uses `zydis`, so this should be `zydis_ROOT`. It's fine to have an uppercase version here since it's not a public feature. This is a surprising convention given that `find_package` is case-insensitive.

instructionAPI/CMakeLists.txt
```diff
@@ -153,16 +153,25 @@ else()
...
 
+# Optional decoder used by the built-in "unknown instruction" callback to
+# recover the length of x86-64 instructions InstructionAPI cannot decode.
+set(_unknown_insn_deps)
+set(_unknown_insn_defines)
+if(DYNINST_ENABLE_ZYDIS)
+  list(APPEND _unknown_insn_deps Dyninst::Zydis)
+  list(APPEND _unknown_insn_defines DYNINST_ENABLE_ZYDIS)
+endif()
+
```
hainest, line 162: CMake variables can be used uninitialized, so you can do just

```cmake
if(...)
  set(...)
endif()
```

You don't need to change it here; just for future reference.

hainest (approved):

instructionAPI/CMakeLists.txt
bbiiggppiigg: You mean Line 158 and Line 159 can go away, and then Line 171 and Line 174 will simply take empty strings that have no effect?

hainest: Yes, sorry, probably should have made a more complete example. I would use `set(_unknown_insn_deps Dyninst::Zydis)` instead of `list(APPEND _unknown_insn_deps Dyninst::Zydis)` just to make it clearer that this is the first declaration/definition.

kupsch (approved):

cmake/tpls/DyninstZydis.cmake
bbiiggppiigg: I'm looking at how it is done for capstone, which seems a little bit chaotic.

We exxpect Capstone_ROOT, and use it to set capstone_ROOT.

Then if capstone_ROOT_DIR is set, we also set Capstone_ROOT_DIR, which then sets capstone_ROOT and capstone_DIR.

I just wonder where can I read about this convention.

hainest: > I'm looking at how it is done for capstone, which seems a little bit chaotic.

Yeah, I should probably clean these up a bit. `Package_ROOT_DIR` should match the capitalization of `Package_ROOT` (and `Package_DIR`). Capstone's cmake directory is spelled `capstone`, so I may have added `Capstone_ROOT{_DIR}` as an alias for `capstone_ROOT{_DIR}`.
 
> I just wonder where can I read about this convention.

I can't recall that I've ever seen it documented. It's just one of those little bits of knowledge I've picked up along the way.

bbiiggppiigg: So is there any required change for this or is it fine for now.

Btw zydis uses zydis-config.cmake, but exports Zydis::Zydis, which seems like a mismatch on their end.

hainest: > So is there any required change for this or is it fine for now.

No, this is fine as-is.

> Btw zydis uses zydis-config.cmake, but exports Zydis::Zydis, which seems like a mismatch on their end.

Heh. I didn't even check. That's the nature of CMake- there are no hard rules. Although I now work with the person who is writing the updated CMake user guides, so I'll pass this along to her.

PR #2269, opened 2026-06-01
Fix 64-bit shift count masking in ROSE x86-64 semantics

bbiiggppiigg:
x86 masks shift/rotate counts modulo the operand size: mod 32 for 8/16/32-bit operands, but mod 64 for 64-bit operands. The ROSE semantics extracted the count as Word(5) (extract<0,5>, i.e. mod 32) for all sizes, so a 64-bit shift like `shrq $0x28` (count 40) was truncated to 40 & 0x1f = 8.

In jump-table index analysis this produced a bogus bound: an index computed as `(imm >> 40)` was evaluated as `imm >> 8`, scaling the StridedInterval by 2^32 and yielding strides/ranges like 2^32[0,10*2^32]. Combined with the 32-bit loop counter in ReadTable that became an infinite loop.

Fix the 64-bit (case 8) branches of shl/shr/sar/ror to mask the count mod 64 using Word(6)/extract<0,6>. The <=32-bit branches keep their correct mod-32 masking.

Also harden StridedInterval::ShiftLeft/ShiftRight: `1 << rhs.low` used a 32-bit int literal, which is undefined for counts >= 31. Use (int64_t)1 << rhs.low and bail to top for counts >= 64. Without this, the now-correct large shift counts would overflow.

### DIFF
--- a/dataflowAPI/rose/x86_64InstructionSemantics.h
+++ b/dataflowAPI/rose/x86_64InstructionSemantics.h
@@ -1388,11 +1388,13 @@ struct X86_64InstructionSemantics {
                         break;
                     }
                     case 8: {
+                        // 64-bit shifts mask the count modulo 64, not 32
+                        Word(6) shiftCount64 = extract<0, 6>(read8(operands[1]));
                         Word(64) op = read64(operands[0]);
-                        Word(64) output = policy.shiftLeft(op, shiftCount);
+                        Word(64) output = policy.shiftLeft(op, shiftCount64);
                         Word(1) newCf = policy.ite(shiftCountZero,
                                                    policy.readFlag(x86_flag_cf),
-                                                   extract<63, 64>(policy.shiftLeft(op, policy.add(shiftCount, number<5>(63)))));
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
Fixes #2264 .
Independent of #2267

hainest:
Thanks for chasing these down. They are some very pernicious bugs. I'm curious if we handle the constant shift arguments for [aarch64](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Shifts-applied-to-a-register/Constant-shifts) and riscv64 correctly. You don't need to examine that here, but I have a feeling it's likely broken, as well. Fortunately, those shift values are much smaller so maybe they are ok.

kupsch (approved):
Looks good.  It is strange that `rol`, `shrd` and `shld` handling are commented out, and that there is no 8-byte cases for them.

bbiiggppiigg:
> Looks good. It is strange that `rol`, `shrd` and `shld` handling are commented out, and that there is no 8-byte cases for them.

Those were commented out at the start (when the file is first committed to add "partial support" for x86_64 semantics).

PR #2273, opened 2026-06-01
remove dyninstAPI_RT async functions

bbiiggppiigg:
It was used by BPatch_asyncEventHandler that was removed by https://github.com/dyninst/dyninst/commit/7a737b6a09ee354569db226759309dc968a6a21d in 2010.


kupsch:
Does anyone think that this code is/has been used?  I find no use of it in Dyninst or the testsuite.

The error handling is bad (no error is ever returned), the error message on line 177 is always printed, truncation is still not detected. I pushed a new commit that fixes these.

hainest:
> Does anyone think that this code is/has been used? I find no use of it in Dyninst or the testsuite.

It's not. It was used by `BPatch_asyncEventHandler` that was removed by 7a737b6a09ee354569db226759309dc968a6a21d in 2010. The whole function can go away. This is the second such case in the last couple of weeks. We should probably audit the runtime.

Is there a particularly good reason the runtime has to be in C? I wouldn't think that having it link against lib[std]c++ would be a big deal since that's required for any mutator. Maybe there were portability issues in the past?

kupsch:
I'll delete the connect and disconnect functions, and associate variables.

No idea why it was on C, maybe not bringing in the stdc++ library, maybe linkage.

hainest:
> I'll delete the connect and disconnect functions, and associate variables.

There are also some friend declarations that can go:
```
dyninstAPI/h/BPatch_function.h:80:    friend class BPatch_asyncEventHandler;
dyninstAPI/h/BPatch_process.h:139:  friend class BPatch_asyncEventHandler;
dyninstAPI/h/BPatch_point.h:95:    friend class BPatch_asyncEventHandler;
```

> No idea why it was on C, maybe not bringing in the stdc++ library, maybe linkage.

Has `extern "C"` been around forever?

kupsch:
done

hainest (approved):
No build issues anywhere. I didn't run the full battery of tests since this is just dead code.

PR #2274, opened 2026-06-03
[dyninstAPI] Convert CMake subprojects to internal libraries

hainest:
When creating static libraries (ENABLE_STATIC_LIBS=ON), the install
target needs the object files created by each subproject (e.g.,
dyninstAPI_bpatch). That information is not propagated for INTERFACE
targets, but it is for OBJECT ones.


hainest:
No build or test issues. The CMake Build Options workflow now passes.

bbiiggppiigg (approved):

PR #2275, opened 2026-06-03
Old Fix failing to write instrumented shared library (#2081)

bbiiggppiigg:
Rewriting some shared libraries (e.g. libtorch_python.so) produced a broken or unloadable output. Two assumptions in the ELF emitter break on binaries whose program/section layout is unusual; both must be fixed.

1. Insertion point for new loadable sections. findSegmentEnds() computed dataSegEnd as max(p_vaddr + p_memsz) over all PT_LOAD segments, and driver() only appended the new instrumentation sections when some section satisfied sh_addr + sh_size == dataSegEnd. When the highest-addressed loadable segment holds relocated read-only metadata (.gnu.hash/.dynstr) and is padded past its last section, no section ends on that boundary, so createLoadableSections() was never called and the result came out with a zero-sized .dynamic ("object file has no dynamic section"). Replace findSegmentEnds() with findLastLoadableSec(), which returns the start address of the highest-addressed section contained in the last loadable segment; the trigger becomes shdr->sh_addr == lastLoadableSecStart. This is equivalent to the previous behavior for ordinary binaries (still triggers on .bss).

2. Program-header insertion order. fixPhdrs() located the slot for the new PT_LOAD by scanning for the first LOAD -> non-LOAD transition, assuming PT_LOAD entries are contiguous in the program header table. libtorch_python.so has GNU_STACK as the second program header (right after the first LOAD), so the new highest-vaddr segment was inserted at index 1, leaving the PT_LOAD entries unsorted by p_vaddr. glibc's _dl_map_segments relies on ascending p_vaddr order and crashes. Insert the new segment before the first PT_LOAD whose p_vaddr is greater than newSegmentStart (else append at the end), keeping PT_LOAD entries sorted regardless of interspersed non-loadable entries.

With both fixes, the rewritten library has a valid .dynamic, a correctly ordered new loadable segment, and loads under the dynamic linker the same way the original does; instrumenting and running ordinary binaries is unaffected.

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -1434,12 +1434,31 @@ void EmitterAMD64::emitRestoreFlagsFromStackSlot(codeGen &gen)
    emitSimpleInsn(0x9D, gen);
 }
 
+// SysV AMD64 caller-saved ("volatile") GPRs. A standard-ABI callee may clobber
+// these freely, and a caller normally does not keep them live across a call --
+// EXCEPT for GCC IPA-SRA local clones (.isra/.constprop), which co-allocate
+// scratch registers across the call between a clone and its callers. dyninst
+// cannot see that contract from the callee, so it must conservatively preserve
+// any of these that an inserted snippet clobbers (see #2081 follow-up: %r11
+// trashed by FEntryCoverage->printf inside an instrumented operator=.isra.0).
+static bool isCallerSavedGPR(int enc)
+{
+   switch (enc) {
+      case REGNUM_RAX: case REGNUM_RCX: case REGNUM_RDX:
+      case REGNUM_RSI: case REGNUM_RDI:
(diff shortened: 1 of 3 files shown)


hainest:
This has clear overlapping concerns with #2166 and #2119. Could you compare/contrast this PR against them?

bbiiggppiigg:
> This has clear overlapping concerns with #2166 and #2119. Could you compare/contrast this PR against them?

#2166 and #2119 are two separate approaches trying to fix the issue.
This PR is built on top of the changes of #2166.
I've closed #2119 and I'm planning on closing #2166 in favor of this one as well, but right now I'm first running the testsuite on corriander.

bbiiggppiigg:
@kupsch 



I think the real bug that we were seeing regarding the segfault in _dl_map_segments  is related to the fact that the PT_LOAD segments should be ordered by p_vaddr. 

I created this PR just as a POC to see if there is any other bug regarding this binary.

You can probably just apply this to your fix on inserting the new section at the end, and we will get a working and cleaner solution.

PR #2276, opened 2026-06-05
Add isClone to Symbol

kupsch:
- add isClone method and static method to check if a symbol is a clone (contains a '.')


kupsch:
@hainest @bbiiggppiigg this is ready for review. It only adds a `isClone` method and static method on `Symbol`. `IsColdClone` is unchanged, I thought `.part` clones were not real functions, but they are and that is what broke tests on fedora 43.

bbiiggppiigg:
```
@kupsch Have you figured out what caused the fedora43 tests to fail previously ?
```

Nvm I see it now.

bbiiggppiigg (approved):

PR #2277, opened 2026-06-05
[dyninstAPI] Move patching classes into subdirectory

hainest:
This is purely cosmetic, but will help to make future changes cleaner. The file changes are kept minimal so that `git log --follow` will work.

PR #2278, opened 2026-06-05
[dyninstAPI] Move proccontrol classes into subdirectory

hainest:
This is purely cosmetic, but will help to make future changes cleaner. The file changes are kept minimal so that `git log --follow` will work.

I was going to put the stackwalk classes in a separate directory. It turns out they are really only used by the dynproc classes, so I just put them in that directory.

Requires #2277


hainest:
No build or test issues on all supported platforms.

kupsch (approved):

PR #2279, opened 2026-06-05
Unify definition of linear variable classes

hainest:
I'm not sure why this was duplicated as the two implementations were added around the same time.


hainest:
No build or testing regressions on all platforms.

bbiiggppiigg (approved):

PR #2281, opened 2026-06-06
[dyninstAPI] Move insnCodeGen classes into codegen subdirectory

hainest:
This will get a refactor in the future. This is a pure move to ensure `git log --follow` will work.

Requires #2278

PR #2282, opened 2026-06-06
[dyninstAPI] Extract RegisterConversion classes into registerSpace subdirectory

hainest:
This will get a refactor in the future. This is a pure move to ensure `git log --follow` will work.

Requires #2281


hainest:
No build regressions.

bbiiggppiigg (approved):

PR #2285, opened 2026-06-07
Add more filesystem functions: exists, extension, and suffix updater

hainest:
These are needed for an upcoming PR.

### DIFF
--- a/tests/unit/common/filesystem.cpp
+++ b/tests/unit/common/filesystem.cpp
@@ -8,7 +8,30 @@
 #include <iostream>
 #include <string>
 
+static int test_canonicalize();
+static int test_exists();
+static int test_replace_extension();
+static int test_append_filename_suffix();
+
 int main() {
+  std::array<int(*)(), 4> tests = {{
+      test_canonicalize,
+      test_exists,
+      test_replace_extension,
+      test_append_filename_suffix
+  }};
+
+  bool failed = false;
+  for(auto t : tests) {
+    if(t() == EXIT_FAILURE) {
+      failed = true;
+    }
+  }
+  std::cout << "failed = " << std::boolalpha << failed << "\n";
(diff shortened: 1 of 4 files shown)


hainest:
No build or test issues on supported platforms.

tests/unit/common/filesystem.cpp
```diff
@@ -103,3 +126,71 @@ int main() {
 
   return (failed) ? EXIT_FAILURE : EXIT_SUCCESS;
 }
+
+int test_exists() {
+  auto file = "test.out";
+  std::ofstream fs{file};
```
kupsch, line 132:
```suggestion
  std::ofstream fs{file};
  if (!fs)  {
    std::cerr << "'" << file << "' ofstream failed to create file.\n";
    return EXIT_FAILURE;
  }
```

kupsch (approved):
You might want to check for the file creation failure before checking for existence.

Othewise looks good.

tests/unit/common/filesystem.cpp
hainest: Updated.

tests/unit/common/filesystem.cpp
```diff
@@ -103,3 +126,75 @@ int main() {
 
   return (failed) ? EXIT_FAILURE : EXIT_SUCCESS;
 }
+
+int test_exists() {
```
bbiiggppiigg, line 130: Should text.out be removed after the test?

bbiiggppiigg (approved):

tests/unit/common/filesystem.cpp
hainest: Ideally, yes, we should fully cleanup after the test. There's just not a portable way of removing a file outside of `boost::filesystem::remove`, but I didn't feel like adding that to the `dyninst_filesystem` interface just for that one use.

PR #2287, opened 2026-06-08
Remove AddressSpace::usesDataLoadAddress

hainest:
It has always only returned false, so it has no use.


hainest:
No build issues on all platforms.

bbiiggppiigg (approved):

kupsch (approved):

PR #2288, opened 2026-06-09
Preserve caller-saved GPRs clobbered by an inserted instrumentation call

bbiiggppiigg:
The base trampoline's register guard decided which registers to save using intra-procedural liveness (shouldSaveReg in emit-x86.C). A caller-saved GPR marked "dead" at the instrumentation point was skipped -- correct for a standard-ABI function, since a caller never keeps a caller-saved scratch register live across a call.

That assumption is wrong for GCC local clones (.isra/.constprop/...), which co-allocate scratch registers across the call between a clone and its callers: the caller legitimately keeps e.g. %r11 live across the call because the clone promises not to touch it. That contract is invisible from the callee, so dyninst skipped saving %r11 -- and the inserted snippet's call (here a coverage reporter that calls printf) then clobbered it, corrupting the caller's value.

Observed on an instrumented PyTorch libtorch_python.so: pybind11::detail::string_caster<std::string,false>::load keeps &local in %r11 across a call to std::string::operator=.isra.0; the instrumented operator='s guard saved only {rsi,rdi}, printf clobbered %r11, and the following _M_dispose dereferenced 0xffffffff -> SIGSEGV during `import torch`.

Fix: in shouldSaveReg, when the instrumented function is a clone (SymtabAPI Symbol::isClone -- mangled name carries a GCC clone suffix), a caller-saved GPR that the inserted snippet clobbers is saved even when intra-procedural liveness marks it dead. The check is gated on isClone (and checked first, as it is false for almost all functions) so ordinary functions keep the liveness optimization -- they cannot have a caller holding a caller-saved register live across the call. Callee-saved registers are unaffected (the inserted call preserves them by ABI). No restore-side change is needed: markSavedRegister() marks the reg spilled, the restore loop pops every spilled reg, and the num_to_save counting loop uses the same predicate, so saves/pops stay balanced.

Verified on x86-64: the operator=.isra.0 guard now pushes the full caller-saved set {rax,r10,r11,r8,r9,rcx,rdx,rsi,rdi}, and instrumented `import torch` exits 0 with results identical to the uninstrumented baseline. (Note: writing an instrumented libtorch_python.so also requires the separate insertion-point / program-header emit fix; the run above used a build that included it.) AArch64/PPC base tramps may need the analogous change.

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -1460,12 +1460,21 @@ static bool isCallerSavedGPR(int enc)
 // which a caller keeps a caller-saved register live across a call, so the
 // conservative save below is restricted to them -- ordinary functions keep the
 // normal liveness-based pruning. Uses SymtabAPI's clone predicate so the
-// detection lives in one place (Symbol::isClone: mangled name contains '.').
+// detection lives in one place (Symbol::isClone: name carries a clone suffix).
+//
+// A function can have several symbol-table names (aliases), so check all of
+// them rather than just the primary symTabName(); treat the function as a clone
+// if any name is a clone name.
 static bool isCloneFunc(baseTramp *inst)
 {
    if (!inst || !inst->point() || !inst->point()->func())
       return false;
-   return Dyninst::SymtabAPI::Symbol::isClone(inst->point()->func()->symTabName());
+   func_instance *f = inst->point()->func();
+   for (auto it = f->symtab_names_begin(); it != f->symtab_names_end(); ++it) {
+      if (Dyninst::SymtabAPI::Symbol::isClone(*it))
+         return true;
+   }
+   return false;
 }
 
 bool shouldSaveReg(registerSlot *reg, baseTramp *inst, bool saveFlags)
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -40,6 +40,7 @@
 #include "compiler_annotations.h"
 #include "dyninstAPI/src/codegen.h"
 #include "patching/function.h"
+#include "Symbol.h"
 #include "dyninstAPI/src/emit-x86.h"
 #include "dyninstAPI/src/inst-x86.h"
 #include "dyninstAPI/src/debug.h"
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
We still need the section/segment insertion fix for #2081 to work.

bbiiggppiigg:
Also this is only handled for x86_64. In theory aarch64 and power can have similar flaw.

dyninstAPI/src/emit-x86.C
```diff
@@ -1434,21 +1435,72 @@ void EmitterAMD64::emitRestoreFlagsFromStackSlot(codeGen &gen)
...
+// .part/.cold). Only such clones can carry a non-standard calling convention in
+// which a caller keeps a caller-saved register live across a call, so the
+// conservative save below is restricted to them -- ordinary functions keep the
+// normal liveness-based pruning. Uses SymtabAPI's clone predicate so the
+// detection lives in one place (Symbol::isClone: mangled name contains '.').
+static bool isCloneFunc(baseTramp *inst)
+{
+   if (!inst || !inst->point() || !inst->point()->func())
+      return false;
+   return Dyninst::SymtabAPI::Symbol::isClone(inst->point()->func()->symTabName());
```
hainest: It's technically possible that multiple symbols can be associated with a function, so we should check all of the names returned by [`symtab_names_begin`](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/patching/function.h#L111).

hainest (changes requested):

kupsch (approved):

hainest:
It would be really good to have a test for this against the pytorch binary. We can store the binary in the test-binaries repo.

bbiiggppiigg:
> It would be really good to have a test for this against the pytorch binary. We can store the binary in the test-binaries repo.

The binary doesn’t fail at instrumentation time.
The test would be to instrument it, find the relocated tramp that is associated with isra, then check for spilled register?

bigtrak:
Having the pytorch as something to run would be good...

Random thoughts:

One thing is pytorch  binaries should be able to run without a GPU, so instrumenting
and running that binary is one possibility -- but does the code fail when the GPU
isn't being used?

Creating a test that will guarantee register use and verify the content of
the registers before/after an instrumentation point that is added by the test... to see that
everything is OK.  I will say that given the complexity of optimizers and available registers,
that could be difficult to do.    If the source for that pytorch binary is available... just try pulling
out that one chunk of code (single function and dependencies)  to simplify things and
add a simple harness around it?

I'd try the non-gpu execution as the lesser workload that is known to fail... and does
it fail the same way with or without a GPU?    So that the test could run automated
on the existing harnesses in githhub?

bbiiggppiigg:
> Having the pytorch as something to run would be good...
> 
> Random thoughts:
> 
> One thing is pytorch binaries should be able to run without a GPU, so instrumenting and running that binary is one possibility -- but does the code fail when the GPU isn't being used?
> 
> Creating a test that will guarantee register use and verify the content of the registers before/after an instrumentation point that is added by the test... to see that everything is OK. I will say that given the complexity of optimizers and available registers, that could be difficult to do. If the source for that pytorch binary is available... just try pulling out that one chunk of code (single function and dependencies) to simplify things and add a simple harness around it?
> 
> I'd try the non-gpu execution as the lesser workload that is known to fail... and does it fail the same way with or without a GPU? So that the test could run automated on the existing harnesses in githhub?

User reported issue instrumenting a libtorch_python.so built for cuda. Fortunatley I'm able to reproduce his problem and fix his problem on a openmp version of it, so it is not tied to GPU.

The prolem is indeed related to compiler optimization that optimize out saving caller-saved registers when it has the knowledge that they weren't used by the callee. But grepping the source and find the right optimization option to trigger that behavior might not be easier in any sense.

To test the instrumentation for this binary under real "run", we need to 
1. Install pytorch (and its dependencies)
2. Run a simple test program and save its result for comparison for validation.
3. Backup its libtorch_python.so
4. Instrument hte libtorch_python.so
5. Swap in the instrumented version
6. Execute the simple test program with the instrumented version
7. If no segfault, we can compare the output with that produced in step 2 to verifyh correctness
8. Look at the trace output to see if the result is reasonable.

bbiiggppiigg:
@hainest whats ur thought on what kind of the test you wanted to add for this binary, and does this PR needs to be blocked until that test is added.

hainest:
> @hainest whats ur thought on what kind of the test you wanted to add for this binary, and does this PR needs to be blocked until that test is added.

We shouldn't block this. My concern is that changes like this that account for very specific compiler behaviors will haunt us in the future and testing will make that much easier to deal with. I'm running the testsuite on the Oregon machines. Things look fine thus far, but there are a couple still going. I'll approve when they're done.

hainest (approved):
No testing regressions.

bbiiggppiigg:
> > @hainest whats ur thought on what kind of the test you wanted to add for this binary, and does this PR needs to be blocked until that test is added.
> 
> We shouldn't block this. My concern is that changes like this that account for very specific compiler behaviors will haunt us in the future and testing will make that much easier to deal with. I'm running the testsuite on the Oregon machines. Things look fine thus far, but there are a couple still going. I'll approve when they're done.

I can certainlly add it to the testsuite. But what kind of test make sense for it is a separate question.
What I can think of for now is maybe we hadd a special debug option, that outputs relavent information when this Pass for isra(or other clones) is triggered, then a script that checks the instrumented binary based on the debug output to see  if they match.

Let me know if this make sense or whatever other test you have in mind that you want me to implement.

PR #2289, opened 2026-06-11
[dyninstAPI] Extract, refactor parsing classes

hainest:


hainest:
No build regressions on all platforms.

dyninstAPI/src/parsing/parse_func.C
```diff
@@ -7,7 +7,7 @@ namespace st = Dyninst::SymtabAPI;
 namespace parse = Dyninst::ParseAPI;
```
bbiiggppiigg, line 7: Just curious, when do we want to create alias for name spaces and when don't we?

bbiiggppiigg (dismissed):

dyninstAPI/src/parsing/DynCFGFactory.C
```diff
@@ -0,0 +1,107 @@
...
+      if (stf && stf->getFirstSymbol()) {
+        auto *ret = parse_func::plt_func(stf, pdmod, _img, obj, reg, isrc, src);
+        // PLT stubs are typically are undefined symbols in the binary,
+        // so there is no corresponding SymtabAPI::Function at Symtab level.
+        // PLTFunction is a subclass of SymtabAPI::Function to represent PLT stubs.
+        // However, since there is no easy way to add a PLTFunction back to the
+        // Symtab object, we need to add PLTFunction to a data structure for
+        // future lookup.
+        _img->insertPLTParseFuncMap(stf->getName(), ret);
+        _mtx.unlock();
```
bbiiggppiigg: Double unlock (reported by AI).

dyninstAPI/src/parsing/DynParseCallback.h
```diff
@@ -76,10 +79,6 @@ namespace Dyninst { namespace DyninstAPI {
 
     bool updateCodeBytes(Dyninst::Address target) override;
 
-#if defined(DYNINST_HOST_ARCH_POWER) || defined(DYNINST_HOST_ARCH_AARCH64)
-    void instruction_cb(ParseAPI::Function *, ParseAPI::Block *, Dyninst::Address,
```
bbiiggppiigg: According to the documentation, 
This function is invoked for each instruction decoded during parsing. Implementing this callback may incur significant overhead.
Are you sure we want to do that for x86_64?

dyninstAPI/src/parsing/DynParseCallback.C
```diff
@@ -33,6 +33,20 @@ namespace Dyninst { namespace DyninstAPI {
         ->hasWeirdInsns();
   }
 
+  void DynParseCallback::instruction_cb(ParseAPI::Function *f, ParseAPI::Block *, Address,
+                                        insn_details *det) {
+    parse_func *ifunc = static_cast<parse_func *>(f);
+    /* In case we do callback from a place where the leaf function analysis has not been
```
bbiiggppiigg, line 58: Again, according to AI:

The body runs leaf-function / return-address-save analysis (isReturnAddrSave, _is_leaf_function, saves_return_addr_) — semantics that are meaningful for link-register architectures (POWER/AArch64) but not for x86, where the return address lives on the stack. 

If ParseAPI fires instruction_cb during x86 parsing, this is a behavioral change that could set saves_return_addr_ incorrectly.

dyninstAPI/src/parsing/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ dyninst_library(
 set(_inc_dirs
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_CURRENT_SOURCE_DIR}/..
+  ${CMAKE_CURRENT_SOURCE_DIR}/../Relocation
```
bbiiggppiigg, line 19: Why do we need this change?

dyninstAPI/src/parsing/DynCFGFactory.C
```diff
@@ -0,0 +1,160 @@
+#include "DynCFGFactory.h"
```
bbiiggppiigg, line 1: Do we want the license header here?

dyninstAPI/src/parsing/DynCFGFactory.C
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  ParseAPI::Function *DynCFGFactory::mkfunc(Address addr, FuncSource src,
+                                            std::string name, ParseAPI::CodeObject *obj,
+                                            ParseAPI::CodeRegion *reg,
+                                            InstructionSource *isrc) {
+    _mtx.lock();
+    parse_func *ret;
+    SymtabAPI::Symtab *st;
+    SymtabAPI::Function *stf = NULL;
```
bbiiggppiigg: maybe replace with nullptr?

dyninstAPI/src/image.h
```diff
@@ -54,6 +54,8 @@
 #include "dyninstAPI/src/codeRange.h"
 #include "dynproc/infHeap.h"
 #include "dyninstAPI/h/BPatch_enums.h"
+#include "parsing/DynCFGFactory.h"
+#include "parsing/DynParseCallback.h"
```
bbiiggppiigg, line 58: Move these includes to image.C and replace them with
  namespace Dyninst { namespace DyninstAPI {
    class DynCFGFactory;
    class DynParseCallback;
  }}
since image.h only uses pointer to these types?

bbiiggppiigg (changes requested):

dyninstAPI/src/parsing/DynCFGFactory.C
```diff
@@ -0,0 +1,107 @@
...
+
+  ParseAPI::Function *DynCFGFactory::mkfunc(Address addr, ParseAPI::FuncSource src,
+                                            std::string name, ParseAPI::CodeObject *obj,
+                                            ParseAPI::CodeRegion *reg,
+                                            InstructionSource *isrc) {
+    boost::lock_guard<decltype(_mtx)> _lock{_mtx};
+
+    SymtabAPI::Symtab *st = _img->getObject();
+    SymtabAPI::Function *stf = NULL;
+    pdmodule *pdmod = _img->getOrCreateModule(st->getDefaultModule());
```
kupsch, line 65: initializing `pdmod` here changes the behavior slightly. If not found in `obj->cs()->linkage.find(add)` and `st_findFuncBeEntryOffset(stf, addr)` is true, then `_img->getOrCreateModule(st->getDefaultModule())` is never called, and the default module is not added to `mods_`. Does the default module being added to `mods_` in this case have any consequence?

kupsch (commented):
Other than some of the other reported comments, and my `mkfunc` comment this looks good

dyninstAPI/src/parsing/parse_func.C
hainest: I don't really have any particular rules I use. You can always use them whenever you like.

dyninstAPI/src/parsing/DynCFGFactory.C
hainest: Derp. That was a bad rebase.

dyninstAPI/src/parsing/DynParseCallback.h
hainest: > Implementing this callback may incur significant overhead. Are you sure we want to do that for x86_64?

`ParseAPI::ParseCallback::instruction_cb` is called for every instruction regardless of architecture. I meant to add a runtime check for this override to keep the correctness for non-ppc. I just forgot. Good catch.

dyninstAPI/src/parsing/DynParseCallback.C
hainest: Same as above.

dyninstAPI/src/parsing/CMakeLists.txt
hainest: I don't remember the specific file, but there was a transitive dependency that was broken when I extracted `DynCFGFactory` into its own file.

dyninstAPI/src/parsing/DynCFGFactory.C
hainest: I don't put them in the source files. There doesn't seem to be a particular consistency to where they are added and not.

dyninstAPI/src/parsing/DynCFGFactory.C
hainest: Default initialized.

dyninstAPI/src/image.h
hainest: Those kinds of forward declarations force consumers of `image.h` to include the other headers if they use the types. It's why you have to include `Symtab.h` before `Function.h` if you use `SymtabAPI::FunctionBase`. It's incredibly annoying.

dyninstAPI/src/parsing/DynCFGFactory.C
hainest: [`image::getOrCreateModule`](https://github.com/dyninst/dyninst/blob/v13.0.0/dyninstAPI/src/image.C#L1782) checks if the module already exists, so it's a nop. I didn't notice that edge case, so that's a good catch.

hainest:
1. Fixed double unlock in DynCFGFactory::mkfunc
2. Add runtime arch check in instruction_cb
3. Default-init more variables in DynCFGFactory::mkfunc
4. Confirm no module behavior changes in DynCFGFactory::mkfunc

Testing OK on all platforms.

dyninstAPI/src/parsing/DynParseCallback.C
bbiiggppiigg: @hainest I this callback used to be used for aarch64 as well, so you might need to fix the switch case above.

dyninstAPI/src/parsing/DynParseCallback.h
```diff
@@ -0,0 +1,88 @@
...
+
+class image;
+
+namespace Dyninst { namespace DyninstAPI {
+
+  class DynParseCallback : public ParseAPI::ParseCallback {
+  public:
+    DynParseCallback(image *img) : ParseAPI::ParseCallback(), _img(img) {
+    }
+    ~DynParseCallback() {
```
bbiiggppiigg, line 46: We want to use the default destructor here with =default?

dyninstAPI/src/parsing/DynParseCallback.C
```diff
@@ -0,0 +1,132 @@
...
+    case Arch_riscv64:
+    case Arch_cuda:
+    case Arch_amdgpu_gfx908:
+    case Arch_amdgpu_gfx90a:
+    case Arch_amdgpu_gfx940:
+    case Arch_intelGen9:
+      return;
+    case Arch_ppc32:
+    case Arch_ppc64:
+      break;
```
bbiiggppiigg, line 54: Do we want a default case here?
Or maybe match three cases (ppc32, ppc64, aarch64) at the start, and default to return?

dyninstAPI/src/parsing/DynParseCallback.C
hainest: That's a good catch. However, it turns out not to matter because `isReturnAddrSave` is always false on aarch64 and `_ret_addr` is a write-only variable (it needs to be removed). `DynParseCallback` needs some refactoring. I'll add the behavior back in for aarch64 here, but likely remove it on a future refactor.

dyninstAPI/src/parsing/DynParseCallback.C
hainest: Since we are using `-Wswitch`, we want to avoid using a default unless you absolutely have to since it let's us use the compiler to detect code that has arch-specific runtime behavior.

kupsch (approved):

dyninstAPI/src/parsing/DynParseCallback.C
hainest: @bbiiggppiigg I decided to just put aarch64 back in. After looking at how this callback is used in more detail, I'm pretty sure it can be removed and the relevant pieces moved directly into the ppc emitter. That's for a future refactor.

bbiiggppiigg (approved):

hainest:
No regressions after final update for aarch64.

PR #2291, opened 2026-06-12
Unify, improve implementations of AddressSpace::getDyninstRTLibName

hainest:
This simplifies the 32-bit and static runtime detection and handling. It also removes printing the PID since that requires knowing about the PCProcess derived class.

I'll take a look at the mechanism used in `PCProcess::getDyninstRTLibName` later. It's not at all clear why dynamic instrumentation and static instrumentation have different methods to find the RT library. I'm pretty sure they should both be the same.


hainest:
No regressions on 18 machines (llnl.cs.rice.edu was down).

common/src/dyninst_filesystem.C
```diff
@@ -179,17 +179,30 @@ bool exists(std::string const& path) {
...
-  x += val;
-  x += ext;
-  return x.string();
+  auto loc = path.find('.');
+
+  if(loc == std::string::npos) {
+    return path + val;
+  }
+
+  return path.substr(0, loc) + val + path.substr(loc);
```
kupsch: To be consistent with filesystem treatment of extensions, should this check `has_extension` for the simple append? Then `.foo` become `.foo<suffix>` instead of `<suffix>.foo`.  Also the code doesn't work correctly if there is a `.` in the path of the parent directory. I think something like this should work:

```suggestion
  if(path.has_extension()) {
    return path + val;
  }
  auto parentPath = path.parent_path();
  auto filename = path.filename().c_str();

  auto loc = filename.find('.', 1);   // skip . in filename[0]
  if(loc != std::string::npos) {
    filename = filename.substr(0, loc) + val + filename.substr(loc);
  }

  return parentPath / filename;
```

kupsch (approved):
Looks good except `append_filename_suffix` could be improved to handle some corner cases.

common/src/dyninst_filesystem.C
hainest: > To be consistent with filesystem treatment of extensions, should this check `has_extension` for the simple append?  Also the code doesn't work correctly if there is a `.` in the path of the parent directory. I think something like this should work:

Yes, it should have checked for `has_extension`. I didn't think about directories with a dot in them.

>Then `.foo` become `.foo<suffix>` instead of `<suffix>.foo`.

The documented behavior of `append_filename_suffix("foo.bar/bar.txt.tar", "_sfx")` returns `foo.bar/bar_sfx.txt.tar`. It really only exists to append the `_m32` suffix to find `libdyninstAPI_RT_m32.so.13.0.0`.

dyninstAPI/src/freebsd.C
```diff
@@ -160,86 +160,6 @@ bool BinaryEdit::archSpecificMultithreadCapable() {
...
-                 std::string("DYNINSTAPI_RT_LIB") +
-                 std::string(" has not been defined for process ") +
-                 std::to_string(proc->getPid());
-           }
-           else {
-              msg = std::string("Environment variable ") +
-                 std::string("DYNINSTAPI_RT_LIB") +
-                 std::string(" has not been defined");
-           }           
-           showErrorCallback(101, msg);
```
bbiiggppiigg, line 184: The old error reporting using showErrorCallBack calls to BPatch::reportError, which output the error to stdout unconditionally. startup_cerr is masked behind DYNISNT_DEBUG_STARTUP.

hainest: The two locations ([1](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/dynproc/dynProcess.C#L214), [2](https://github.com/dyninst/dyninst/blob/master/dyninstAPI/src/dynproc/dynProcess.C#L616)) where this is called use a mixture of the two so I chose the common one. For dynamic instrumentation and rewriting statically-linked binaries, this should probably be a fatal error.

hainest: I changed it back to `showErrorCallback` so it will always print.

bbiiggppiigg (approved):

PR #2293, opened 2026-06-14
[dyninstAPI] Extract arch-specific AddressSpace functions

hainest:
To enable cross-arch codegen, these will need to be made dispatchable at runtime.


hainest:
No build or test issues on all platforms.

kupsch (approved):

dyninstAPI/src/inst-x86.C
```diff
@@ -962,34 +945,6 @@ bool emitAddSignedImm(Address addr, long int imm, codeGen &gen) {
...
-   static Dyninst::DyninstAPI::EmitterIA32Dyn emitter32Dyn;
-   static Dyninst::DyninstAPI::EmitterIA32Stat emitter32Stat;
-
-#if defined(DYNINST_CODEGEN_ARCH_X86_64)
-   static EmitterAMD64Dyn emitter64Dyn;
-   static EmitterAMD64Stat emitter64Stat;
-
-   if (getAddressWidth() == 8) {
-       if (proc()) {
-           return &emitter64Dyn;
```
bbiiggppiigg, line 976: When DYNINST_CDOEGEN_ARCH_X86_64 is defined, but getAddressWidth is not 8, the original code would return a 32 bit emitter.
I'm still checking whether this can happen or not

bbiiggppiigg: It seems like getAddressWidth comes from the address of the target binary, so this will be a behavioral change, if truely intended,

hainest: It would technically be possible (depending on how `AddressSpace::getAddressWidth` is implemented). I don't think it makes sense since the _codgen_ architecture is x86_64. I think this was just a sloppy implementation.

dyninstAPI/src/IAPI_to_AST.C
```diff
@@ -1,96 +0,0 @@
...
-#else
-    MachRegister reg = r->getID();
-    reg = reg.getBaseRegister();
-    Register astreg = reg.val() & ~reg.getArchitecture();
-    m_stack.push_back(operandAST::origRegister((void*)(astreg)));
-#endif
-}
-void ASTFactory::visit(MultiRegisterAST* )
-{
-    fprintf(stderr,"ASTFactory doesn't support MultiRegisterAST!");
```
bbiiggppiigg, line 93: Are you sure if we want to remove the warning assert with slient default overrdie?

dyninstAPI/src/inst-x86.C
bbiiggppiigg: I mean if your target binary is 32 bit, shouldn't you emit 32 bit instructions instead of 64 bit instructions?

hainest: > I mean if your target binary is 32 bit, shouldn't you emit 32 bit instructions instead of 64 bit instructions?

If `DYNINST_CODEGEN_ARCH=x86_64`, I wouldn't expect 32-bit codegen to work for the same reason that I wouldn't expect that 64-bit codegen to work when `DYNINST_CODEGEN_ARCH=i386`. That it does is an accident. I guess I can preserve the old behavior. `getEmitter` will eventually get dispatched at runtime based on the binary's architecture to give actual cross-architecture codegen.

bbiiggppiigg: I think it is because that these were all HOST_ARCH before the separation of host and codegen arch. So the original behavior is that, on a x86_64 host, we pick emitter based on the address width of the target binary.

dyninstAPI/src/IAPI_to_AST.C
hainest: A vector register can't be the source/target of a call.

dyninstAPI/src/inst-x86.C
hainest: Derp. You are quite right. I'll change this back.

dyninstAPI/src/IAPI_to_AST.C
bbiiggppiigg: If you mean something like call rax, I think it will be the case for AMDGPU, when we do 
s_swappc_b64 s[4:5], s[30:31], one of them would be the call target and one of them  will be the return address

hainest: `ASTFactory` is only used on x86.

bbiiggppiigg (approved):

PR #2294, opened 2026-06-17
Fix failing to write instrumented shared library (#2081)

bbiiggppiigg:
Rewriting some shared libraries (e.g. libtorch_python.so) produced a broken or unloadable output. Two assumptions in the ELF emitter break on binaries whose program/section layout is unusual; both must be fixed.

1. Insertion point for new loadable sections. findSegmentEnds() computed dataSegEnd as max(p_vaddr + p_memsz) over all PT_LOAD segments, and driver() only appended the new instrumentation sections when some section satisfied sh_addr + sh_size == dataSegEnd. When the highest-addressed loadable segment holds relocated read-only metadata (.gnu.hash/.dynstr) and is padded past its last section, no section ends on that boundary, so createLoadableSections() was never called and the result came out with a zero-sized .dynamic ("object file has no dynamic section"). Replace findSegmentEnds() with findLastLoadableSec(), which returns the start address of the highest-addressed section contained in the last loadable segment; the trigger becomes shdr->sh_addr == lastLoadableSecStart. This is equivalent to the previous behavior for ordinary binaries (still triggers on .bss).

2. Program-header insertion order. fixPhdrs() located the slot for the new PT_LOAD by scanning for the first LOAD -> non-LOAD transition, assuming PT_LOAD entries are contiguous in the program header table. libtorch_python.so has GNU_STACK as the second program header (right after the first LOAD), so the new highest-vaddr segment was inserted at index 1, leaving the PT_LOAD entries unsorted by p_vaddr. glibc's _dl_map_segments relies on ascending p_vaddr order and crashes. Insert the new segment before the first PT_LOAD whose p_vaddr is greater than newSegmentStart (else append at the end), keeping PT_LOAD entries sorted regardless of interspersed non-loadable entries.

With both fixes, the rewritten library has a valid .dynamic, a correctly ordered new loadable segment, and loads under the dynamic linker the same way the original does; instrumenting and running ordinary binaries is unaffected.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -79,7 +79,7 @@ emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, ObjectELF *ob
         symTabData(NULL), dynsymData(NULL), dynData(NULL),
         phdrs_scn(NULL), verneednum(0), verdefnum(0),
         newSegmentStart(0), firstNewLoadSec(NULL),
-        dataSegEnd(0), dynSegOff(0), dynSegAddr(0),
+        dynSegOff(0), dynSegAddr(0),
         phdrSegOff(0), phdrSegAddr(0), dynSegSize(0),
         secNameIndex(0), currEndOffset(0), currEndAddress(0),
         linkedStaticData(NULL), loadSecTotalSize(0),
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
This should be the last piece of #2275 to address the issue #2081, so it produce a correct ELF.
I've testes this on coriander, where I don't see testsuite regression.
I'll test on zeroah and llnl to see if anything regresesion comes up.

bbiiggppiigg:
Actually I was not able to run on zeroah since something seems to be broken for the default loaded tbb modules.

hainest:
> Actually I was not able to run on zeroah since something seems to be broken for the default loaded tbb modules.

Yeah, I use

```
export PATH=/opt/cmake-3.20.6/bin:$PATH

BOOST="/opt/boost-1.78.0"
TBB="/opt/intel-tbb-2020.3"
ELFUTILS="/opt/elfutils-0.186"

export LD_LIBRARY_PATH=$BOOST/lib:$TBB/lib:$ELFUTILS/lib:$LD_LIBRARY_PATH
```

bbiiggppiigg:
@hainest 

I tried running on llnl, but for some reason it only runs the following tests, and i'm not sure what happened on zeroah.

```
test_pt_ls                        none 64  create   NA      dynamic nonPIC  FAILED (Running Test)
test2_1                           none 64  create   NA      dynamic nonPIC  PASSED
test2_2                           none 64  create   NA      dynamic nonPIC  PASSED
test_pt_ls                        none 64  rewriter NA      dynamic nonPIC  PASSED
aarch64_decode                    none 64  disk     NA      dynamic nonPIC  PASSED
power_decode                      none 64  disk     NA      dynamic nonPIC  PASSED
aarch64_simd                      none 64  disk     NA      dynamic nonPIC  PASSED
power_cft                         none 64  disk     NA      dynamic nonPIC  PASSED
aarch64_decode_ldst               none 64  disk     NA      dynamic nonPIC  PASSED
aarch64_cft                       none 64  disk     NA      dynamic nonPIC  PASSED
```

hainest (approved):

PR #2297, opened 2026-06-18
[dyninstAPI] Clean up emitLoadPreviousStackFrameRegister

hainest:
There doesn't seem to be any particular reason to having multiple implementations that are identical. Except for ppc which is completely different- but again without any explanation.


hainest:
No regressions.

bbiiggppiigg (approved):

kupsch (approved):

PR #2298, opened 2026-06-19
Regression Tests For AMDGPU Indirect Control Flow

bbiiggppiigg:
Requires #2299 and #2301 to create a correct reference answer


bbiiggppiigg:
The test_binaries structure for this PR right now looks like:
parseAPI/IndirectControlFlowAnalysis/
{
1-908.gpubin
1-908.ref.out
2-90a.gpubin
2-90a.ref.out 
3-940.gpubin
3-940.ref.out
}

Let met know what you guys think.

bbiiggppiigg:
By the way right the reference answers does not match the dynisnt output right now, and it seems like there are mistakes on both sides (incorrect handling of s_trap instruction v.s. duplicate edges), so I guess I need to fix these bugs first so I can come up with a better reference file, and then I can upload the test_binaries with the reference answers.

kupsch:
It seems like your test of each binary is split into two tests

1. run test program, generating output file
2. compare file against known results

Since the tests can run in parallel step 2 could be run before or simultaneously with 1 resulting in incorrect results. If you can combine these steps into one program (or script) that would fix the problem. If not there are ways of specifying dependencies between tests.

kupsch:
I think it would be better to structure the files with the binaries in on directory and the test results in another as a binary may be the input to multiple tests, and a test may test multiple binaries.

**expected_result**/_\<component>_**/**_\<test_name>_**/**_\<input_binary_name>_**.out**
_\<arch>_**/**_\<binary_descriptive_name>_**-**_\<arch>\<optional_suffix>_

Use the standard dyninst arch names for _arch_ (the top level directory can use the arch family name (aarch, ppc, x86, amdgpu, riscv, ...). The _binary_descriptive_name_ can be anything. If the binaries are the same except for architecture use the same _binary_descriptive_name_.

So if the 1-, 2-, 3- are the same binary, different architecture, then I would use something like this:

amdgpu
{
controlflow-gfx908.gpubin
controlflow-gfx90a.gpubin
controlflow-gfx940.gpubin
}

expected_results/parseAPI/IndirectControlFlowAnalysis/
{
controlflow-gfx908.gpubin.out
controlflow-gfx90a.gpubin.out
controlflow-gfx940.gpubin.out
}

kupsch:
Perhaps the expected output files should be named after the ctest name which would be:

**expected_result**/_\<ctest_name>_**.out**

or

**expected_result**/_\<component>_**\_**_\<test_name>_**\_**_\<arch>_**.out**

hainest:
A few items:

1. The DWARF test has a bug that I fixed in #2300. The test needs to not include `DyninstTestBinaries.cmake` and be guarded by `DYNINST_ENABLE_EXTERNAL_BINARY_TESTS`.
2. ctest doesn't guarantee tests will run in any particular order, so you have to add the `DEPENDS` property on the test to force an ordering.
3. I would recommend changing the names of the files so you can just loop over the architecture names. The change I made here isn't necessarily what needs to happen- it's just an example.

```cmake
if(DYNINST_ENABLE_EXTERNAL_BINARY_TESTS)
  set(_targ IndirectControlFlowAnalysis)
  add_executable(${_targ} branchInfo.cpp)
  target_compile_options(${_targ} PRIVATE ${SUPPORTED_CXX_WARNING_FLAGS})
  target_link_libraries(${_targ} PRIVATE parseAPI)
  set(_fpath "${DYNINST_TEST_BINARIES_DIR}/parseAPI/IndirectControlFlowAnalysis")

  foreach(_arch 908 90a 940)
    add_test(NAME parseAPI_${_targ}_${_arch} COMMAND ${_targ} "${_fpath}/${_arch}.gpubin"
                                                     ${_arch}.out)
    add_test(NAME compare_${_targ}_${_arch}
             COMMAND ${CMAKE_COMMAND} -E compare_files ${_arch}.out
                     "${_fpath}/${_arch}.ref.out")

    set_tests_properties(parseAPI_${_targ}_${_arch} PROPERTIES LABELS "regression")
    set_tests_properties(compare_${_targ}_${_arch} PROPERTIES LABELS "regression" DEPENDS
                                                              parseAPI_${_targ}_${_arch})
  endforeach()
endif()
```

kupsch:
`DEPENDS` is only a runs after dependency and does not require success. I think you want to use [FIXTURES](https://crascit.com/2016/10/18/test-fixtures-with-cmake-ctest/).

hainest:
> `DEPENDS` is only a runs after dependency and does not require success. I think you want to use [FIXTURES](https://crascit.com/2016/10/18/test-fixtures-with-cmake-ctest/).

Bah. I didn't notice that. Yes, fixtures is the way to ensure a logical grouping of tests. I've not been using a real unit testing framework because the tests have been so simple. I think it might be time to consider using catch2, GTest, or Boost.Test.

bbiiggppiigg:
@hainest I've updated the cmake to have fixtures / include guard and foreach.
@kupsch added expected_result to the path.
I'm not sure I fully understand what your suggestion.

The corresponding file layout I have for this PR rn is reflected here
https://github.com/dyninst/test_binaries/tree/bbiiggppiigg/add_amdgpu_testt/parseAPI/IndirectControlFlowAnalysis

kupsch:
Let's discuss this when we meet

bbiiggppiigg:
@hainest  @kupsch 

I've removed freopen, using namespace std and updated the naming of the files.

bbiiggppiigg:
@hainest Is there supposed to be a delay between when the test_binaries repo is updated vs when the dyninst CI picked up the latest change in that repo.

hainest:
> @hainest Is there supposed to be a delay between when the test_binaries repo is updated vs when the dyninst CI picked up the latest change in that repo.

There shouldn't be. The binary-tests workflow just does a `git clone`.

bbiiggppiigg:
> > @hainest Is there supposed to be a delay between when the test_binaries repo is updated vs when the dyninst CI picked up the latest change in that repo.
> 
> There shouldn't be. The binary-tests workflow just does a `git clone`.

It seems like it is only failing the regression test when building dyninst with clang?
I guess I can try to build with clang and mahir and see if I can reproduce the failure.

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+        printf("Usage: %s <binary path> <output path>\n", argv[0]);
+        return -1;
+    }
+    char *binaryPath = argv[1];
+    std::ofstream of(argv[2]);
+    if(!of){
+      std::cerr << "Cannot open output file !" << std::endl;
+    }
+    SymtabCodeSource *sts;
+    CodeObject *co;
```
kupsch: define where used and can be initialized

```suggestion
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+    SymtabAPI::Symtab *symTab;
+    std::string binaryPathStr(binaryPath);
+    bool isParsable = SymtabAPI::Symtab::openFile(symTab, binaryPathStr);
+    if(isParsable == false){
+        const char *error = "error: file can not be parsed";
+        of << error;
+        return - 1;
+    }
+    sts = new SymtabCodeSource(binaryPath);
+    co = new CodeObject(sts);
```
kupsch: declare with auto

```suggestion
    auto sts = new SymtabCodeSource(binaryPath);
    auto co = new CodeObject(sts);
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+    //parse the binary given as a command line arg
+    co->parse();
+
+    //get list of all functions in the binary
+    const CodeObject::funclist &all = co->funcs();
+    if(all.size() == 0){
```
kupsch: use the more descriptive name `func` instead of `all`

```suggestion
    const CodeObject::funclist &func = co->funcs();
    if(func.size() == 0){
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+    for(;fit != all.end(); ++fit){
+        f = *fit;
+        //get address of entry point for current function
+        //
+        auto bit = f->blocks().begin();
+        for(; bit != f->blocks().end(); bit++){
+            Block * b = * bit;
+            bb_addrs.insert(b->start());
+        }
+    }
```
kupsch: 1. `bb_addr` is never used delete it and the for loop.
2. move the `decoder` and `instr` variables to where they are used

```suggestion
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+    fit = all.begin();
+    for(;fit != all.end(); ++fit){
+        f = *fit;
```
kupsch: 1. use range-based for
2. declare the `decoder here

```suggestion
    for(auto f : funcs){
        InstructionDecoder decoder(f->isrc()->getPtrToInstruction(f->addr()),
                InstructionDecoder::maxInstructionLength,
                f->region()->getArch());
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+        //get address of entry point for current function
+        of << "Parsing function " << f->name() << " at addreess 0x" << std::hex << f->addr() << std::endl;
+        auto bit = f->blocks().begin();
+        for(; bit != f->blocks().end(); bit++){
+            Block * b = * bit;
```
kupsch: use range-based for loop:

```suggestion
            for (auto & edges : b->targets() ){
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,102 @@
...
+            of << "Parsing block ( " << std::hex << b->start() << "," << b->end() << ")" << std::endl;
+            for (auto & edges : b->targets() ){
+                Address branch_addr = edges->src()->lastInsnAddr();
+
+                void * insn_ptr = f->isrc()->getPtrToInstruction(branch_addr);
+                instr = decoder.decode((unsigned char * ) insn_ptr);
```
kupsch: declare `instr` here:
```suggestion
                auto instr = decoder.decode((unsigned char * ) insn_ptr);
```

kupsch (approved):
There's a bunch of code cleanup to do, but otherwise is OK. You'll need to test the changes I suggested.

bbiiggppiigg:
Will mark as ready when #2310 is in.

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,81 @@
...
+int main(int argc, char **argv){
+    if(argc != 3){
+        printf("Usage: %s <binary path> <output path>\n", argv[0]);
+        return -1;
+    }
+    char *binaryPath = argv[1];
+    std::ofstream of(argv[2]);
+    if(!of){
+      std::cerr << "Cannot open output file !" << std::endl;
+      return -1;
```
kupsch: Do not return -1. The return value from a process is an uint_8 (0 to 255), so even though main returns an int it is cast to uint_8.

The convention is that 0 mean success, and in the shells if process exits by a signal `N`, shells return `128 + N` (in your own code you can distinguish between a normal exit and exit by signal). Given this it's best to restrict exit values to 0 to 127.

I would suggest using 1. The C standard defines `EXIT_SUCCESS` and `EXIT_FAILURE`, but 1 is fine.


```suggestion
      return 1;
```

tests/regression/parseAPI/IndirectControlFlowAnalysis/branchInfo.cpp
```diff
@@ -0,0 +1,81 @@
...
+    auto co = new CodeObject(sts);
+    //parse the binary given as a command line arg
+    co->parse();
+
+    //get list of all functions in the binary
+    const CodeObject::funclist &funcs = co->funcs();
+    if(funcs.size() == 0){
+        const char *error = "error: no functions in file";
+        of << error;
+        return - 1;
```
kupsch: same as above:


```suggestion
        return 1;
```

kupsch (approved):
fix return value

PR #2301, opened 2026-06-20
[parseAPI] Treat AMDGPU s_trap 0x02 (llvm.trap) as a block terminator

bbiiggppiigg:
s_trap was decoded as an ordinary scalar instruction with no control-flow category, so parseAPI gave it an implicit fall-through and kept parsing the bytes after it. For non-returning traps (s_trap 0x02, the lowering of llvm.trap) the following bytes are padding/garbage, leading to mis-parses.

Override IA_amdgpu::isSoftwareException() to recognize s_trap and inspect its trap ID (low 8 bits of the SIMM16 immediate). Only trap ID 0x02 terminates the wavefront, so only it is reported as a software exception, which makes the parser end the block with no fall-through edge. Trap IDs 0x01 (debugger breakpoint) and 0x03 (llvm.debugtrap) resume execution, so they keep their fall-through; all other (reserved) trap IDs are left unchanged.

The trap-ID semantics follow the LLVM AMDGPU trap handler ABI: https://llvm.org/docs/AMDGPUUsage.html#trap-handler-abi

The distinction depends on the immediate operand, so it cannot be expressed in the opcode-only entryToCategory()/allowsFallThrough() paths and is handled in the AMDGPU parse adapter instead.

### DIFF
--- a/parseAPI/src/IA_amdgpu.C
+++ b/parseAPI/src/IA_amdgpu.C
@@ -140,3 +140,37 @@ bool IA_amdgpu::isNopJump() const
 {
     return false;
 }
+
+bool IA_amdgpu::isSoftwareException() const
+{
+    Instruction ci = curInsn();
+    auto id = ci.getOperation().getID();
+    if(id != amdgpu_gfx908_op_S_TRAP &&
+       id != amdgpu_gfx90a_op_S_TRAP &&
+       id != amdgpu_gfx940_op_S_TRAP)
+        return IA_IAPI::isSoftwareException();
+
+    // s_trap takes a 16-bit immediate whose low 8 bits are the trap ID.
+    //
+    // The per-trap-ID behavior is defined by the AMDGPU trap handler ABI in the
+    // LLVM AMDGPU backend documentation:
+    //   https://llvm.org/docs/AMDGPUUsage.html#trap-handler-abi
+    // (see the "AMDGPU Trap Handler for AMDHSA OS Code Object V3 / V4 and Above"
+    // tables). Summarizing the relevant trap IDs:
+    //   - 0x01 (debugger breakpoint): the wave is halted and the debugger
+    //     resumes it, re-executing the overwritten instruction. Execution
+    //     continues, so the fall-through must be preserved.
+    //   - 0x02 (llvm.trap): the wave is halted and its queue is put into the
+    //     error state, terminating the dispatch's waves. It does not return,
+    //     so this is the only s_trap that ends the block with no fall-through.
+    //   - 0x03 (llvm.debugtrap): a no-op when no debugger is attached, otherwise
+    //     reported to the debugger which resumes the wave. Either way execution
+    //     continues, so the fall-through must be preserved.
+    // All other trap IDs are reserved and are left to fall through.
+    Immediate::Ptr imm =
+        boost::dynamic_pointer_cast<Immediate>(ci.getOperand(0).getValue());
+    if(!imm)
+        return false;
+    Result trapId = imm->eval();
+    return trapId.defined && ((trapId.val.s16val & 0xff) == 0x02);
+}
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
One crash on coriander.
image

Honestly I don't know how to tell if there is a regression from the testsuite result.

hainest:
> Honestly I don't know how to tell if there is a regression from the testsuite result.

Just run it again. If it shows up again, it's probably a regression. I also recommend running on multiple machines to see if it shows up on any of them. The test suite is just unreliable. Running on 19 machines gives a statistical measure of breakages, not any guarantees (see the previous PRs fixing old regressions).

bbiiggppiigg:
@hainest The same test CRASHED again when I rerun the testsuite.

To rerun that specific test I assume it should be the following command?

LD_LIBRARY_PATH=$TESTSUITE_LD_LIBRARY_PATH ./test_driver -v -log output.log -debugPrint -test test_thread_1 -dynamic -PIC

hainest:
> @hainest The same test CRASHED again when I rerun the testsuite.
> 
> To rerun that specific test I assume it should be the following command?
> 
> LD_LIBRARY_PATH=$TESTSUITE_LD_LIBRARY_PATH ./test_driver -v -log output.log -debugPrint -test test_thread_1 -dynamic -PIC

The `dynamic` option is spelled `-dynamiclink` (because why not?). Also add `-g++ -create` to get just the version that failed.

bbiiggppiigg:
@hainest 
After running it multiple times, I finally get 
1. a run where master gets the same crash
2. a run where this branch does not crash

Can you test for regression on the other machines?

bigtrak:
1) The trap value is 16 bits wide, but the upper bits are  masked off
instead of including them in the check of a value of 02... which
means other trap values that are unrelated could cause a difference.
Please inspect the full 16 bit word.   For example a trap value of 0xff02 
would be treated the same as 0x02

2) I'm concerned that if the value can't be created, that this just returns; it should
be a fatal error, as that immediate value must exist, OTW, handling can
become non-deterministic.   For example a malformed binary comes to mind,
or internal errors somewhere else.

bbiiggppiigg:
> 1. The trap value is 16 bits wide, but the upper bits are  masked off
>    instead of including them in the check of a value of 02... which
>    means other trap values that are unrelated could cause a difference.
>    Please inspect the full 16 bit word.   For example a trap value of 0xff02
>    would be treated the same as 0x02

Right now the value is only defined for up to 0xff, with 13 out 16 being marked as reserved.
But I agree that we should not mask off the upper bits.

> 2. I'm concerned that if the value can't be created, that this just returns; it should
>    be a fatal error, as that immediate value must exist, OTW, handling can
>    become non-deterministic.   For example a malformed binary comes to mind,
>    or internal errors somewhere else.

I don't think a fatal error on this is a good idea.
If someone comes up with a ill-formed binary with undefined trap value, they will need to face whatever undefined behavior the hardware is going to slap them with. 
For our purpose alone, we are just trying to differentiate whether the parsing should have a fallthrough edge (where the trap is known to return or not)

Here the default is to continue parsing unless it is provable that the trap wouldn't return (case 0x02), and I don't see a problem with that.

bbiiggppiigg:
> > 1. The trap value is 16 bits wide, but the upper bits are  masked off
> >    instead of including them in the check of a value of 02... which
> >    means other trap values that are unrelated could cause a difference.
> >    Please inspect the full 16 bit word.   For example a trap value of 0xff02
> >    would be treated the same as 0x02
> 
> Right now the value is only defined for up to 0xff, with 13 out 16 being marked as reserved. But I agree that we should not mask off the upper bits.
> 

Actually, based on the documentaton, it only uses 8 bits for trap reason, so there is no need to change at all.

hainest (approved):
Assuming we can add tests for this in #2298, this looks fine.

PR #2302, opened 2026-06-20
Add DYNINST_REPORT_PROGRESS rewrite-progress reporting

bbiiggppiigg:
Rewriting a very large binary (e.g. a multi-hundred-MB shared library) spends long stretches in silent loops -- building reloc blocks, applying transforms, code generation, the emit address-fixpoint, speculative gap parsing, and assembling/emitting the output file -- so the tool can look hung for many minutes with no output.

Add a coarse, opt-in progress channel gated on the DYNINST_REPORT_PROGRESS environment variable (and the dyn_debug_progress flag), following the existing debug-channel convention in debug.{h,C}. Each progress line is prefixed with a wall-clock timestamp so phases can be timed directly from the log.

Instrumented phases:
  - CodeMover: per-function reloc-block build loop and per-RelocBlock codegen loop (periodic counts + "codegen done").
  - addressSpace::relocateInt/generateCode: phase markers (transforms, code generation, emit address-fixpoint attempts, patching).
  - CodeBuffer::generate: per-pass / periodic progress over the buffer elements in the do/while regeneration fixpoint.
  - image::analyzeImage: brackets around the (single-threaded) speculative gap-parse phase per text region.
  - BinaryEdit::writeFile / BPatch_binaryEdit::writeFile: committing instrumentation, in-memory relocation, copying tracked regions into the output buffer (periodic), building the symbol table, the final symtab emit (the on-disk write), and per-dependent-library writes -- the span from "patching jumps/springboards" through the end of writeFile.
  - Parser-speculative probabilistic_gap_parsing: self-contained, DYNINST_REPORT_PROGRESS-gated timestamped progress (parseAPI sits below dyninstAPI and cannot use its progress channel directly). The env-var check matches dyninstAPI's check_env_value() truthiness (set and non-zero) so the single knob behaves identically across both layers.

No behavior change when the env var is unset.

### DIFF
--- a/common/src/debug_common.C
+++ b/common/src/debug_common.C
@@ -31,6 +31,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
+#include <time.h>     // progress timestamps (progress_printf_int)
 
 #include "common/src/debug_common.h"
 
@@ -38,6 +39,7 @@ int common_debug_dwarf = 0;
 int common_debug_addrtranslate = 0;
 int common_debug_lineinfo = 0;
 int common_debug_parsing = 0;
+int common_debug_progress = 0;
 int common_debug_initialized = 0;
 
 bool init_debug_common() {
@@ -59,6 +61,15 @@ bool init_debug_common() {
     
     if(getenv("COMMON_DEBUG_PARSING")){
         common_debug_parsing = 1;  
+    }
+
+    // Match the dyninstAPI check_env_value() truthiness (set and non-zero) so the
+    // single DYNINST_DEBUG_PROGRESS knob behaves identically across both layers.
+    {
+        const char *v = getenv("DYNINST_DEBUG_PROGRESS");
+        if (v && atoi(v) != 0) {
+            common_debug_progress = 1;
+        }
     }
 
     return true;
@@ -112,6 +123,32 @@ int common_parsing_printf_int(const char *format, ...)
    init_debug_common();
   if (!common_debug_parsing) return 0;
   if (NULL == format) return -1;
+
+  va_list va;
+  va_start(va, format);
+  int ret = vfprintf(stderr, format, va);
+  va_end(va);
+
+  return ret;
+}
+
+int progress_printf_int(const char *format, ...)
+{
+   init_debug_common();
+  if (!common_debug_progress) return 0;
+  if (NULL == format) return -1;
+
+  // Unlike the other debug channels, prefix each progress line with a wall-clock
+  // timestamp so phase timing is visible during long rewrites/parses.
+  time_t now = time(NULL);
+  struct tm tmv;
+  char ts[40];
+#if defined(os_windows)
+  if (localtime_s(&tmv, &now) == 0 && strftime(ts, sizeof(ts), "[%Y-%m-%d %H:%M:%S] ", &tmv))
+#else
+  if (localtime_r(&now, &tmv) && strftime(ts, sizeof(ts), "[%Y-%m-%d %H:%M:%S] ", &tmv))
+#endif
+    fputs(ts, stderr);
 
   va_list va;
   va_start(va, format);
--- a/common/src/debug_common.C
+++ b/common/src/debug_common.C
@@ -143,16 +143,16 @@ int progress_printf_int(const char *format, ...)
   time_t now = time(NULL);
   struct tm tmv;
   char ts[40];
-#if defined(os_windows)
-  if (localtime_s(&tmv, &now) == 0 && strftime(ts, sizeof(ts), "[%Y-%m-%d %H:%M:%S] ", &tmv))
-#else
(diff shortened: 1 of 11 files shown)


bbiiggppiigg:
One test1_12 crash on coirander.

hainest:
I have lots of questions/complaints for this. It might be easier to have a quick chat about it. I have some time after the Rice meeting tomorrow, if that works.

bbiiggppiigg:
> I have lots of questions/complaints for this. It might be easier to have a quick chat about it. I have some time after the Rice meeting tomorrow, if that works.

That works.

bbiiggppiigg:
@hainest @kupsch 

Ready for review.

Rename the controlling environment variable DYNINST_REPORT_PROGRESS -> DYNINST_DEBUG_PROGRESS 
Drop the counter-throttled in-loop progress prints
Drop unnecessary call to fprintf

Example output
```
[2026-06-29 13:38:32] write: committing pending instrumentation to out.so
[2026-06-29 13:38:32] write: instrumentation committed
[2026-06-29 13:38:32] write: writing primary output file out.so
[2026-06-29 13:38:32] write: relocating instrumented code (in memory)
[2026-06-29 13:38:32] write: relocation done; assembling output image
[2026-06-29 13:38:32] write: copying 20 tracked regions into output buffer
[2026-06-29 13:38:32] write: copied tracked regions into output buffer
[2026-06-29 13:38:32] write: adding 0 instrumentation symbols to symbol table
[2026-06-29 13:38:32] write: emitting output file out.so (final on-disk write; can take a while for large binaries)
[2026-06-29 13:38:32] write: wrote output file out.so
[2026-06-29 13:38:32] write: writeFile complete
```

common/src/debug_common.C
```diff
@@ -120,3 +131,29 @@ int common_parsing_printf_int(const char *format, ...)
...
+  // Unlike the other debug channels, prefix each progress line with a wall-clock
+  // timestamp so phase timing is visible during long rewrites/parses.
+  time_t now = time(NULL);
+  struct tm tmv;
+  char ts[40];
+#if defined(os_windows)
+  if (localtime_s(&tmv, &now) == 0 && strftime(ts, sizeof(ts), "[%Y-%m-%d %H:%M:%S] ", &tmv))
+#else
+  if (localtime_r(&now, &tmv) && strftime(ts, sizeof(ts), "[%Y-%m-%d %H:%M:%S] ", &tmv))
+#endif
```
kupsch: Do we need windows support?  Does Dyninst compile on Windows? If not, I'd suggest removing the ifdef since we can't test it.

kupsch (approved):

common/src/debug_common.C
bbiiggppiigg: @kupsch 

I'm not sure if we support windows or not.
But as long as we don't offically remove the check of if define(os_windows), 
I would argue it doesn't hurt to have it here, just in case someone is trying to use it.
Other usage of this checks:

./CodeSource.C:#if defined(os_windows)
./IA_x86.C:#if defined(os_windows)
./IA_x86.C:#if defined(os_windows)
./IdiomModelDesc.C:  #if defined(os_windows)
./IndirectASTVisitor.C:#if defined(os_windows)
./IndirectASTVisitor.C:#if defined(os_windows)
./IndirectASTVisitor.C:#if defined(os_windows)
./ProbabilisticParser.C:#if defined (os_windows)
./SymbolicExpression.C:#if defined(os_windows)
./Parser-speculative.C:  #if defined(os_windows)

kupsch: The harm is that it adds noise to the source code. It needs to be maintained.It isn't tested. There are likely lots of things that won't work/compile on windows, so it can be added if there is serious windows work.

@hainest do you have any opinion?

hainest: > I'm not sure if we support windows or not.

There is a lot of code for Windows that hasn't been compilable for at least a decade. Forget it exists. Pretty much the same for FreeBSD.

> @hainest do you have any opinion?

Delete it. My suggestion is to never worry about any code that is guarded by a check for Windows.

hainest: Once we move to C++20, we can do

```c++
std::cerr << "[" << std::chrono::current_zone()->to_local(std::chrono::steady_clock::now()) << "]\n";
```

hainest (approved):
As I noted, I recommend dropping the Windows check but I don't have a hard opinion about it.

bbiiggppiigg:
@hainest @kupsch 

I've removed the windows line
i've also added flockfile and funlockfile on stderr to prevent the prefix output from being separated.
Let met know if you guys have any objection/concern. on doing so

hainest:
> @hainest @kupsch
> 
> I've removed the windows line i've also added flockfile and funlockfile on stderr to prevent the prefix output from being separated. Let met know if you guys have any objection/concern. on doing so

That's a good idea. The one in dyninstAPI uses a mutex, but I think flock is better. The one in dataflowAPI uses `omp critical`, but none of the others even try to synchronize.

Issue #916, opened 2020-11-09
Find new small-object allocator for InstructionAPI


hainest:
#835 fixed a large memory leak at the expense of using global new/delete. We should find a suitable small-object allocator to enhance the performance of InstructionAPI. Notes from Xiaozhu are

---

Another opinion is that TBB's malloc proxy has such sub-allocators functionality and malloc in newer versions of glibc also has this functionality built-in. So, we may not benefit from the singleton_object_pool layer as much as we hope. If you can find other existing allocators, the benchmarking should consider (1) different versions of glibc and (2) single threads vs multiple threads. If a malloc level allocator is good enough, then we should just get rid of this singleton_object_tool layer.

At rice, we have been using a microbenchmark written by Mark that mimics what hpcstruct does. It will print out the time spent and memory usage for symtab, parse, and hpcstruct. You can get the microbenchmark at:

https://github.com/HPCToolkit/hpctoolkit-tests/blob/master/dyninst/openmp-loops.cpp

---

Just an example. I ran a quick experiment on gpu.cs.rice.edu, which has glibc-2.17.

**glibc:**

    $ ./openmp-loops -jp 1 -js 1 -v ~/test-binaries/_pywrap_tensorflow_internal.so > trash.txt
    file: /home/xm13/test-binaries/_pywrap_tensorflow_internal.so
    symtab threads: 1  parse threads: 1

    init:         0.0 sec          0 meg          4 meg
    symtab:      14.5 sec       1235 meg       1240 meg
    parse:      150.3 sec       7758 meg       8998 meg
    struct:     121.3 sec        352 meg       9350 meg
    total:      286.1 sec       9345 meg       9350 meg

**tbbmalloc_proxy:**

    $ ./openmp-loops -jp 1 -js 1 -v ~/test-binaries/_pywrap_tensorflow_internal.so > trash.txt
    file: /home/xm13/test-binaries/_pywrap_tensorflow_internal.so
    symtab threads: 1  parse threads: 1

    init:         0.0 sec          0 meg          5 meg
    symtab:      14.8 sec       1309 meg       1314 meg
    parse:      133.3 sec       7757 meg       9072 meg
    struct:     106.5 sec        357 meg       9429 meg
    total:      254.6 sec       9424 meg       9429 meg

You can see that by linking against tbbmalloc_proxy from TBB, the whole program has about 1.12X speedup. I expect the difference between glibc version and tbbmalloc_proxy version will be small if we run this experiment on a Linux with newer glibc.

PR #2304, opened 2026-06-22
Do not link tbbmalloc_proxy

blue42u:
`tbbmalloc_proxy` is a `malloc` replacement library that calls TBB's `scalable_malloc` equivalents. It was at one time recommended for Dyninst users since it improved performance, although [it's not clear whether these benefits apply to modern Glibc](https://github.com/dyninst/dyninst/issues/916). Unfortunately replacing `malloc` is tricky business and `tbbmalloc` is known to do things that are [explicitly documented as unsafe like `dlopen`/`dlsym`](https://sourceware.org/glibc/manual/latest/html_node/Replacing-malloc.html). As of TBB `2023.0.0` we have finally observed it crashing when combined with ParseAPI (for details see https://gitlab.com/hpctoolkit/hpctoolkit/-/work_items/998#note_3475868658). 

This PR drops Dyninst's dependency on `tbbmalloc_proxy`, making it the users' prerogative to link it if they so desire. This is more correct anyway, `malloc` replacements need to be linked to the _main executable_ (or `LD_PRELOAD`'d) to ensure proper symbol resolution, but Dyninst is a library and not in (enough) control of the user's executable link line to make this happen.

### DIFF
--- a/cmake/tpls/DyninstTBB.cmake
+++ b/cmake/tpls/DyninstTBB.cmake
@@ -28,7 +28,7 @@ endif()
 
 find_package(
   TBB ${_min_version}
-  COMPONENTS tbb tbbmalloc tbbmalloc_proxy
+  COMPONENTS tbb tbbmalloc
   REQUIRED ${_find_path_args})
 
 # Don't let TBB variables seep through
@@ -36,13 +36,11 @@ mark_as_advanced(TBB_DIR)
 
 if(NOT TARGET Dyninst::TBB)
   add_library(Dyninst::TBB INTERFACE IMPORTED)
-  target_link_libraries(Dyninst::TBB INTERFACE TBB::tbb TBB::tbbmalloc
-                                               TBB::tbbmalloc_proxy)
+  target_link_libraries(Dyninst::TBB INTERFACE TBB::tbb TBB::tbbmalloc)
   target_include_directories(
     Dyninst::TBB SYSTEM
     INTERFACE $<TARGET_PROPERTY:TBB::tbb,INTERFACE_INCLUDE_DIRECTORIES>
-              $<TARGET_PROPERTY:TBB::tbbmalloc,INTERFACE_INCLUDE_DIRECTORIES>
-              $<TARGET_PROPERTY:TBB::tbbmalloc_proxy,INTERFACE_INCLUDE_DIRECTORIES>)
+              $<TARGET_PROPERTY:TBB::tbbmalloc,INTERFACE_INCLUDE_DIRECTORIES>)
 endif()
 
 message(STATUS "Found TBB ${TBB_VERSION}")


hainest:
I'd have to go back to the PR that defaulted to using TBB's allocator, but I would imagine the rationale was to give "better" performance by default. We don't have any performance analysis tests, so it's hard to say how much users on old glibcs would be affected by this. I'd like to dig into the HPCToolkit bug report a bit more.

mwkrentel:
@hainest   See:  https://gitlab.com/hpctoolkit/hpctoolkit/-/work_items/998

I can reproduce the segfault with a dyninst unit test outside of
hpctoolkit if the unit test links or preloads `libtbbmalloc_proxy`.

I tried a smaller example with `boost::assign::map_list_of`
inside an init ctor which seems to be where the segv originates,
but so far I can't get it to fail.

I'd like to find an example without Parseapi.  That would give me a
clear bug report to send to TBB.  But so far I can't get it to fall over
without Parseapi.

Anyway, I agree with @blue42u that Dyninst should not be linking with malloc
proxy.  It should be the dyninst app that decides to insert it or not.

blue42u:
To add: if TBB's allocator _does_ provide better performance in general for Dyninst, the right approach is (and always was) to use the `tbb::scalable_allocator` and related interfaces inside Dyninst. That would avoid tainting the symbol space with a potentially incompatible `malloc` (e.g. if the application wanted to instead use TCMalloc), and avoids forcing an allocator for application code that may not see benefits (or may see detriments) when using TBB's allocator.

That's a much larger and API-breaking change, and not one I have the time to tackle at this time. This patch to work around a visible bug is about the maximum time I can spare at the moment.

hainest:
> To add: if TBB's allocator _does_ provide better performance in general for Dyninst, the right approach is (and always was) to use the `tbb::scalable_allocator` and related interfaces inside Dyninst. That would avoid tainting the symbol space with a potentially incompatible `malloc` (e.g. if the application wanted to instead use TCMalloc), and avoids forcing an allocator for application code that may not see benefits (or may see detriments) when using TBB's allocator.

Agreed. It looks like all of the concurrent containers in Dyninst are explicitly using `std::allocator`. That should be fixed instead of using the proxy.

> That's a much larger and API-breaking change, and not one I have the time to tackle at this time. This patch to work around a visible bug is about the maximum time I can spare at the moment.

No worries. It shouldn't be too much to make the containers better, so I'll work on that. I think there are only a couple of public interfaces that expose the TBB containers (Ideally, there would be none).

hainest:
> @hainest See: https://gitlab.com/hpctoolkit/hpctoolkit/-/work_items/998
> 
> I can reproduce the segfault with a dyninst unit test outside of hpctoolkit if the unit test links or preloads `libtbbmalloc_proxy`.

That is just weird.
 
> I tried a smaller example with `boost::assign::map_list_of` inside an init ctor which seems to be where the segv originates, but so far I can't get it to fail.
> 
> I'd like to find an example without Parseapi. That would give me a clear bug report to send to TBB. But so far I can't get it to fall over without Parseapi.

Definitely let me know if you find one. There is a weekly CI job that runs effectively the same code as your example to parse all system libraries in all the Dyninst containers (Ubuntu 20-24, Fedora 39-41), but have never seen any failure like this.

> Anyway, I agree with @blue42u that Dyninst should not be linking with malloc proxy. It should be the dyninst app that decides to insert it or not.

Agreed. I'll get this merged after the testing finishes.

mwkrentel:
After we add Jonathon's patch, I also want to add a conflict to the
Dynist spack recipe to conflict `intel-tbb@2023:` with `dyninst@:13`,
that is, new TBB with old Dyninst.

That would block the bad pairing while allowing dyninst master with
the latest intel-tbb (after the MR).

Everyone Ok with that?

hainest:
> After we add Jonathon's patch, I also want to add a conflict to the Dynist spack recipe to conflict `intel-tbb@2023:` with `dyninst@:13`, that is, new TBB with old Dyninst.
> 
> That would block the bad pairing while allowing dyninst master with the latest intel-tbb (after the MR).
> 
> Everyone Ok with that?

That should be fine. We do a weekly spack build of Dyninst against the latest spack dependencies, but we don't guarantee old Dyninst will build with new versions of dependencies.

mwkrentel:
@hainest  If you build Dyninst master with the latest spack intel-tbb @2023.0.0,
you should be seeing a segfault with any app that runs ParseAPI.

Do you run anything with ParseAPI or just build it?

hainest:
> @hainest If you build Dyninst master with the latest spack intel-tbb @2023.0.0, you should be seeing a segfault with any app that runs ParseAPI.
> 
> Do you run anything with ParseAPI or just build it?

Ah, nevermind. That workflow uses the distro-provided TBB. I'd have to check, but I'm pretty sure even Ubuntu 24 only has tbb@2023. We should probably have a workflow to run tests against the latest spack develop to keep up with the bleeding edge. I'd also like to have containers for rolling releases of Ubuntu and Rawhide, but wishes and fishes...

mwkrentel:
@hainest Note: despite the name, intel-tbb `2023.0.0` (aka oneTBB) was released Apr 30 of
this year and added to spack packages on May 17.

Unliess you're spack-building the latest TBB, you're not using the one
triggers the segfault.

mwkrentel:
I tried some timing experiments with/without `tbbmalloc_proxy`.  This
is one datapoint, John's `qmcpack.m` directory on aurora.  It's a real
world example: 400+ separate files, some large files with 24 threads,
lots of small files with 1 thread.

It's not a targeted example with lots of opportunities for threads.
I'm guessing a single, large file would show more of a difference,
assuming it's not the kind of file where parseapi serializes itself
(like libmpi.so).

no tbb malloc proxy:    4:50--4:55
with malloc proxy:      3:30--3:35
with jemalloc preload:  3:10--3:15

Just one data point, but the takeaways are:

(1) tbb malloc proxy does help, it's noticeable, not really a deal breaker,
(2) jemalloc works just as well, even a tad better.

hainest:
> no tbb malloc proxy: 4:50--4:55 with malloc proxy: 3:30--3:35 with jemalloc preload: 3:10--3:15

What is OMP_NUM_THREADS? Are you using the same example code from the GitLab issue? That's a non-trivial speedup for tbbmalloc_proxy. I'd like to see how things change if we update the default allocator in the Dyninst concurrent data structures.

mwkrentel:
@hainest 

> some large files with 24 threads,

24 threads.  hpcstruct on a measurements directory (with many files)
looks at the size of the text segment in the file and allocates the full amount
for large files run one at a time and 1 or 2 for small files run in parallel.
It's not a targeted unit test, but it is an example of how we really use hpcstruct.
Sorry if that wasn't clear.

I tried a few other files like libmkl, libmpi, but those are mostly "bad" binaries
that have one large function and thus quickly come down to one thread.
jemalloc helped a small amount and didn't segfault, which is what I was really testing.

I'm still looking for a large binary that offers lots of parallelism.
Most binaries (like libparseAPI) run too fast to be a meaningful test.

hainest:
> @hainest
> 
> > some large files with 24 threads,

I clearly need to work on my reading comprehension...

> It's not a targeted unit test, but it is an example of how we really use hpcstruct. Sorry if that wasn't clear.

Real-world uses are better than any synthetic unit test, so this is great.

> I'm still looking for a large binary that offers lots of parallelism. Most binaries (like libparseAPI) run too fast to be a meaningful test.

If you find one, let me know. I've been searching for something like that for a while. Ideally, a .text >1GB would be a great stress test.

hainest:
No regressions on 19 machines. I'm going to merge since it's causing hard failures in HPCToolkit. However, we should find a way to try to evaluate and recuperate performance losses. Maybe we can _briefly_ talk about this at today's UW/Rice meeting?

hainest (approved):

PR #2146, opened 2026-02-06
Update RISC-V64 decoder to latest Capstone

hainest:
Capstone recently revamped their RISCV decoder and changed several representations.

**TODO:**

- [ ] Determine a category for `RISCV_FEATURE_ISRVE`
- [ ] Implement decoding for `RISCV_OP_FP`, `RISCV_OP_CSR`
- [ ] Fix instruction decoder test failures
- [ ] Update base containers with latest Capstone build

PR #2305, opened 2026-06-23
Fix capstone 6 riscv mode constant name

kripticni:
Capstone 6 renamed CS_MODE_RISCVC -> CS_MODE_RISCV_C. Since Dyninst's CMakeLists
already requires capstone >= 6.0.0-Alpha5, this is a straightforward rename of
the affected constant in the riscv decoder.

### DIFF
--- a/instructionAPI/src/decoder/riscv/decoder.C
+++ b/instructionAPI/src/decoder/riscv/decoder.C
@@ -31,6 +31,11 @@
 #include "arch-riscv64.h"
 #include "capstone/capstone.h"
 #include "capstone/riscv.h"
+
+#ifndef CS_MODE_RISCV_C
+#define CS_MODE_RISCV_C CS_MODE_RISCVC
+#endif
+
 #include "categories.h"
 #include "debug.h"
 #include "decoder/riscv/decoder.h"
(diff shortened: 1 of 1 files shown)


hainest:
What changes are needed for RISCV depends sensitively on which alpha tag of Capstone you are using. Does this change work with alpha-5? I can't remember exactly, but I believe the big set of changes in https://github.com/dyninst/dyninst/pull/2146 were introduced in alpha-6 which includes the update proposed here.

I'd really like to finish that PR, but Capstone added support for a bunch of system registers that requires some big updates to the Capstone importer scripts.

hainest:
@bbiiggppiigg I don't believe Angus is working on Dyninst anymore. I've taken up work on most of the RISCV PRs.

kripticni:
you're right to question this. i checked the v6.0.0 C headers and CS_MODE_RISCVC is still the canonical name - CS_MODE_RISCV_C doesn't exist in this release. i've updated the PR: instead of a straight rename, it now adds a #ifndef guard so the code uses CS_MODE_RISCV_C if it exists (future capstone versions), with a fallback to CS_MODE_RISCVC otherwise. functionally no change for v6.0.0, but it won't break when/if capstone eventually renames it.

hainest:
@kripticni This isn't an ideal solution, but it will work until I get the other PR finished. Thanks!

kripticni:
@hainest , anything else needed on my end to get this merge? And is help needed for something elsewhere?

hainest:
> @hainest , anything else needed on my end to get this merge? And is help needed for something elsewhere?

No. I'm just waiting on some external tests to finish.

hainest:
Everything looks good. I will try to get the other updates needed for the latest Capstone. No timeline on that work, though.

hainest (approved):

PR #2306, opened 2026-06-23
Use CMake CTest module

hainest:
This is needed to upload ctest results to CDash.

This is part of my work on the DoE PESO project.


hainest:
No issues on all supported platforms.

kupsch (approved):

PR #2309, opened 2026-07-01
Add spack CI automation to run on Frank enclave

hainest:
This adds the necessary GitLab CI files to build and test Dyninst, dyninst/examples, dyninst/external-tests, and dyninst/testsuite on the [mirror](https://gitlab.spack.io/tools-sdk/dyninst) I made (it's not permanent, but it could be). The testsuite isn't run yet because it will require additional processing of its output to convert into a CDash submission. [Here](https://open.cdash.org/index.php?project=TOOLS-SDK) is an example of the output from the x86_64 gary machine at Frank.

The *only* action needed on the Dyninst side is the creation of the GITHUB_TOKEN. Everything else is for me to complete. 

Once complete, pull requests in this repo that have been approved to run on CI will automatically be picked up and run on the spack CI at the Frank enclave and the results posted to CDash with a link to the results and a pass/fail status.

**TODO**
- [x] Set up SSH keys
- [x] Create fine-grain token for updating statuses
- [x] Create GitLab pipeline to poll PRs
- [ ] Convert testsuite results to CDash format
  - This will be done later. Currently, the testsuite is just built.
- [ ] Group CDash results under a single SHA so the status can point to the CDash entry
  - TBD. Requires more analysis of CDash.

This work is supported by the DOE PESO project.

### DIFF
--- a/.gitlab-ci.yaml
+++ b/.gitlab-ci.yaml
@@ -1,0 +1,18 @@
+variables:
+  CDASH_SERVER: "https://open.cdash.org/submit.php?project=TOOLS-SDK"
+
+stages:
+  - sync
+  - test
+
+# Synchronize GitHub with GitLab
+# Only runs when scheduled on gitlab.spack.io
+include:
+  - local: '/.gitlab/sync-spack.yaml'
+    stage: sync
+    rules:
+    - if: $CI_SERVER_HOST == "gitlab.spack.io" && $CI_PIPELINE_SOURCE == "schedule"
+  - local: '/.gitlab/.uofrank.gary.yaml'
+    stage: test
+    rules:
+      - if: $CI_SERVER_HOST == "gitlab.spack.io" && $CI_PIPELINE_SOURCE != "schedule"
--- a/.gitlab/.uofrank.gary.yaml
+++ b/.gitlab/.uofrank.gary.yaml
@@ -1,0 +1,39 @@
+uo-gary:
+  interruptible: true
+  stage: test
+  tags: [ 'gary' ]
+  script:
+    - module load gcc/14.0.1
+    - module load e4s/25.06/mvapich4.0-rocm6.4.0
+    - source /storage/packages/e4s/25.06/mvapich4.0-rocm6.4.0/spack/share/spack/setup-env.sh
+    - spack load --first boost+atomic elfutils libiberty+pic intel-tbb
+    - git clone --depth=1 https://gitlab.spack.io/tools-sdk/dyninst
+    - cd dyninst
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst -DSUBMIT_URL='$CDASH_SERVER' "
+    - export DYNINST_INSTALL_DIR=$PWD/install
+    - cmake .. -DCMAKE_INSTALL_PREFIX=$DYNINST_INSTALL_DIR -DDYNINST_WARNINGS_AS_ERRORS=ON -DDYNINST_ENABLE_TESTS=ALL -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Configure -T Build -T Test -T Submit
+    - cmake --install .
+    - git clone --depth=1 https://github.com/dyninst/external-tests
+    - cd external-tests
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst-external-tests -DSUBMIT_URL='$CDASH_SERVER' "
+    - cmake .. -DDYNINST_DISABLE_STATIC_TESTS=ON -DDyninst_ROOT=$DYNINST_INSTALL_DIR -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Build -T Test -T Submit
+    - git clone --depth=1 https://github.com/dyninst/examples
+    - cd examples
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst-examples -DSUBMIT_URL='$CDASH_SERVER' "
+    - cmake .. -DDyninst_ROOT=$DYNINST_INSTALL_DIR -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Build -T Test -T Submit
+    - git clone --depth=1 https://github.com/dyninst/testsuite
+    - cd testsuite
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst-testsuite -DSUBMIT_URL='$CDASH_SERVER' "
+    - cmake .. -DDyninst_ROOT=$DYNINST_INSTALL_DIR -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Build -T Submit
--- a/.gitlab/scripts/SpackCIBridge.py
+++ b/.gitlab/scripts/SpackCIBridge.py
@@ -1,0 +1,825 @@
+#!/usr/bin/env python3
+
+import argparse
+import atexit
+import base64
+from datetime import datetime, timedelta, timezone
+import dateutil.parser
+import github
+import json
+import os
+import re
+from requests import Session
+from requests.adapters import HTTPAdapter, Retry
+import subprocess
+import sys
+import tempfile
+import time
+import urllib.parse
+import urllib.request
+
+import sentry_sdk
+
+sentry_sdk.init(traces_sample_rate=0.1)
+
+
+def _durable_subprocess_run(*args, **kwargs):
+    """
+    Calls subprocess.run with retries/exponential backoff on failure.
+    """
+
+    max_attempts = 5
+    for attempt_num in range(max_attempts):
+        try:
+            return subprocess.run(*args, **kwargs, check=True)
+        except subprocess.CalledProcessError as e:
+            if attempt_num == max_attempts - 1:
+                raise e
+            print(
+                f"Subprocess failed ({e}), retrying ({attempt_num+1}/{max_attempts})",
+                file=sys.stderr,
+            )
+            time.sleep(2 ** (1 + attempt_num))
+
+
+def _write_secret(file: str, content: str):
+    # Create the file with the oatuh token
+    with open(file, "wb") as fd:
+        fd.write(b"")
+    os.chmod(file, 0o600)
+    with open(file, "wb") as fd:
+        fd.write(content.encode())
+
+
+class SpackCIBridge(object):
+
+    def __init__(self, gitlab_repo="", gitlab_host="", gitlab_project="", github_project="",
+                 disable_status_post=True, disable_protected_sync=True, disable_tag_sync=True, sync_draft_prs=False,
+                 main_branch=None, prereq_checks=[], required_label=""):
+        self.gitlab_repo = gitlab_repo
+        self.github_project = github_project
+        github_token = os.environ.get('GITHUB_TOKEN')
+        self.github_repo = "https://{0}@github.com/{1}.git".format(github_token, self.github_project)
+        self.py_github = github.Github(auth=github.Auth.Token(github_token))
+        self.py_gh_repo = self.py_github.get_repo(self.github_project)
+
+        self.session = Session()
+        self.session.mount(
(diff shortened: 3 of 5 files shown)


sameershende:
@hainest Omnia is an Icelake system. If you want Sapphire Rapids, you may want to run on headroom or picard.

hainest:
@sameershende Is there a list of tag names for the systems available to use as runners?

.gitlab/.uofrank.gary.yaml
```diff
@@ -0,0 +1,39 @@
+uo-gary:
+  interruptible: true
+  stage: test
+  tags: [ 'gary' ]
+  script:
+    - module load gcc/14.0.1
+    - module load e4s/25.06/mvapich4.0-rocm6.4.0
+    - source /storage/packages/e4s/25.06/mvapich4.0-rocm6.4.0/spack/share/spack/setup-env.sh
+    - spack load --first boost+atomic elfutils libiberty+pic intel-tbb
+    - git clone --depth=1 https://gitlab.spack.io/tools-sdk/dyninst
```
bbiiggppiigg, line 10: Is this testing against master?
Do we need something like --branch  "$CI_COMMIT_REF_NAME"?
Also well this fail when run a second time, because the rpo already exist? or will it wipe?

.gitlab/.uofrank.gary.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - cd dyninst
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst -DSUBMIT_URL='$CDASH_SERVER' "
```
bbiiggppiigg, line 14: This might end up like, so maybe drop the single quotes?
-DSUBMIT_URL='https://open.cdash.org/submit.php?project=TOOLS-SDK'

bbiiggppiigg: line 22,29,36 also

.gitlab/.uofrank.gary.yaml
hainest: > Is this testing against master?

No, it uses the [sync script](https://github.com/spack/spack-infrastructure/blob/main/images/gh-gl-sync/SpackCIBridge.py) to discover all branches that exist in GitHub but not in GitLab _and_ that haven't been tested. I plan on updating the sync action to allow filtering based on tags so we can limit the number of branches that get tested on Frank. I'll fix that before merging here.

> Also well this fail when run a second time, because the rpo already exist? or will it wipe?

The test runners on Frank use temporary directories, so you get a fresh clone each time.

.gitlab/.uofrank.gary.yaml
hainest: I put them there because I wasn't sure if the shell would try to interpret the characters in the URL (e.g., ':' or '/').

.gitlab-ci.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+stages:
+  - sync
+  - test
+
+# Synchronize GitHub with GitLab
+# Only runs when scheduled on gitlab.spack.io
+include:
+  - local: '/.gitlab/sync-spack.yaml'
+    stage: sync
```
bbiiggppiigg, line 12: According to AI: The stage and rules keys belong inside the jobs of your included file, not directly under the include: keyword.

The example fix I was given was something like:
variables:
  CDASH_SERVER: "https://open.cdash.org/submit.php?project=TOOLS-SDK"

stages:
  - sync
  - test

# Synchronize GitHub with GitLab
# Only runs when scheduled on gitlab.spack.io
include:
  - local: '/.gitlab/sync-spack.yaml'
    rules:
      - if: $CI_SERVER_HOST == "gitlab.spack.io" && $CI_PIPELINE_SOURCE == "schedule"

  - local: '/.gitlab/.uofrank.gary.yaml'
    rules:
      - if: $CI_SERVER_HOST == "gitlab.spack.io" && $CI_PIPELINE_SOURCE != "schedule"

And then  under .gitlab/.uofrank.gary.yaml


your-sync-job-name:
  stage: sync  # <--- Define the stage here, inside the job!
  script:
    - echo "Syncing Spack..."

and 

your-test-job-name:
  stage: test  # <--- Define the stage here, inside the job!
  script:
    - echo "Running tests..."

.gitlab-ci.yaml
```diff
@@ -0,0 +1,18 @@
+variables:
```
bbiiggppiigg, line 1: GitLab's default CI config path is .gitlab-ci.yml.
I don't know if this is changed for this project or not

.gitlab/scripts/SpackCIBridge.py
```diff
@@ -0,0 +1,825 @@
...
+                           gitlab_project=args.gitlab_project,
+                           github_project=args.github_project,
+                           disable_status_post=args.disable_status_post,
+                           disable_protected_sync=args.disable_protected_sync,
+                           disable_tag_sync=args.disable_tag_sync,
+                           sync_draft_prs=args.sync_draft_prs,
+                           main_branch=args.main_branch,
+                           prereq_checks=args.prereq_check,
+                           required_label=args.required_label)
+    bridge.setup_git_auth(gitlab_ssh_key_base64, gitlab_token)
```
bbiiggppiigg, line 824: setup_git_auth() runs before sync(), but sync() does os.chdir(tmpdirname) and only then git init. So _write_secret(".git-credentials", …) and git config credential.helper "store --file .git-credentials" both act on the startup CWD: the git config either fails outright (no repo there yet) or writes into an unrelated repo, and the credentials file is invisible to the tmpdir repo that actually pushes. The SSH path survives the chdir only because its state lives in env vars — that's the asymmetry. Fix: do the auth setup inside sync() after setup_git_repo(), and use an absolute path for the credentials file.

.gitlab/scripts/SpackCIBridge.py
```diff
@@ -0,0 +1,825 @@
...
+            self.setup_ssh(ssh_key_base64)
+        elif token:
+            self.setup_oauth(token)
+        else:
+            raise Exception("Failed to setup auth for pushing to git remote.")
+
+    def setup_oauth(self, token: str):
+        """Configure the origin"""
+
+        url = urllib.parse.urlparse(self.gitlab_repo)
```
bbiiggppiigg, line 131: urlparse will return '' for git@ssh.gitlab.spack.io:tools-sdk/dyninst

so neither scheme == "ssh" nor scheme == "https" fires. A missing GITLAB_SSH_KEY_BASE64 now fails late with the generic
  Failed to setup auth for pushing to git remote. instead of upstream's specific message — a regression in exactly the
  configuration this PR ships. Fix: scheme = gl_url.scheme or "ssh", or write the URL as
  ssh://git@ssh.gitlab.spack.io/tools-sdk/dyninst.

.gitlab/sync-spack.yaml
```diff
@@ -0,0 +1,13 @@
+include:
+  - component: $CI_SERVER_FQDN/dav-sdk/gh-gl-sync/github-sync@~latest
+    inputs:
+      github_project: hainest/dyninst
```
bbiiggppiigg, line 4: should this be dyninst/dyninst

.gitlab/.uofrank.gary.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - export DYNINST_INSTALL_DIR=$PWD/install
+    - cmake .. -DCMAKE_INSTALL_PREFIX=$DYNINST_INSTALL_DIR -DDYNINST_WARNINGS_AS_ERRORS=ON -DDYNINST_ENABLE_TESTS=ALL -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Configure -T Build -T Test -T Submit
```
bbiiggppiigg, line 17: replace CMAKE_BUILD_PARALLEL_LEVEL=16 with some environmental variable, so we can remove the hard coded constant on line 25,32,39.

bbiiggppiigg, line 17: Is there a reason to reconfigure ctest when it should be done by the cmake ..?

.gitlab/scripts/SpackCIBridge.py
```diff
@@ -0,0 +1,825 @@
...
+                raise e
+            print(
+                f"Subprocess failed ({e}), retrying ({attempt_num+1}/{max_attempts})",
+                file=sys.stderr,
+            )
+            time.sleep(2 ** (1 + attempt_num))
+
+
+def _write_secret(file: str, content: str):
+    # Create the file with the oatuh token
```
bbiiggppiigg, line 46: oauth

.gitlab/.uofrank.gary.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - cmake --install .
+    - git clone --depth=1 https://github.com/dyninst/external-tests
+    - cd external-tests
+    - mkdir build
+    - cd build
+    - cdash_opts="-DBUILDNAME=dyninst-external-tests -DSUBMIT_URL='$CDASH_SERVER' "
+    - cmake .. -DDYNINST_DISABLE_STATIC_TESTS=ON -DDyninst_ROOT=$DYNINST_INSTALL_DIR -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc ${cdash_opts}
+    - CMAKE_BUILD_PARALLEL_LEVEL=16 ctest -VV -T Build -T Test -T Submit
```
bbiiggppiigg, line 25: Do we need --no-tests=error for here (and other places?

bbiiggppiigg (changes requested):

.gitlab/.uofrank.gary.yaml
hainest: This only runs on Gary, so 16 is really all we should use. Once I add other runners, I'll refactor these build/test steps into a separate component that will take the number of jobs as an input.

hainest: The first `cmake` run is to prime the cache with the `${cdash_opts}`. This is a known corner case for CDash. This was the fix recommended by the main CDash dev.

.gitlab/.uofrank.gary.yaml
hainest: We could certainly add it. That would make this closer to the pr-tests workflow on GitHub.

.gitlab/sync-spack.yaml
hainest: Yes. I missed that one.

.gitlab/sync-spack.yaml
```diff
@@ -0,0 +1,13 @@
...
+      gitlab_repo: git@ssh.gitlab.spack.io:tools-sdk/dyninst
+      gitlab_host: https://gitlab.spack.io
+      gitlab_project: tools-sdk/dyninst
+      sync_draft_prs: false
+      environment: ghgl-bridge
+      script_path: ".gitlab/scripts/SpackCIBridge.py"
```
hainest, line 10: Derp. This and the whole bridge script should go away, as well. I thought I rebased those commits out of the branch...

.gitlab-ci.yaml
hainest: I based this on another repo. It works as-is, but I'll check if the default filename was changed in the GitLab config.

.gitlab-ci.yaml
hainest: The unstructured nature of declaring which steps execute in a given stage in GitLab annoys me. I put it here to be explicit, but I also added it to the `sync-spack` component. I'll remove it from `sync-spack` and leave it here.

The `rules` check is carried over from the `dav-sdk/gh-gl-sync/github-sync` component because that is too far away from the logic here and isn't under our control. I'd have to go back and check if the `include` directive can accept a "top level" `if`. I don't think it can which is why it's structured like this now.

.gitlab/scripts/SpackCIBridge.py
hainest: This file shouldn't be here. I thought I rebased it out.

kupsch (approved):
Look good to me.

A couple of things to consider now or in the future:

1. Did you want the tests to run in parallel in addition to the builds?  If so you need to also set `CTEST_PARALLEL_LEVEL` (or pass `-j`) and probably OMP_NUM_THREADS (so the parallel jobs don't overcommit the CPUs.
2. You could dynamically set to the number of CPUs using nproc `PARALLELISM=$(( 1  + ( $(nproc) - 1) / frac ))` to get the `1/frac` rounded up number of CPUs.

hainest:
Oh, sorry, I didn't realize this PR was from a branch on my fork. This isn't ready yet. I am testing a new implementation. I'll put this in draft mode.

>     1. Did you want the tests to run in parallel in addition to the builds?  If so you need to also set `CTEST_PARALLEL_LEVEL` (or pass `-j`) and probably OMP_NUM_THREADS (so the parallel jobs don't overcommit the CPUs.
> 
>     2. You could dynamically set to the number of CPUs using nproc `PARALLELISM=$(( 1  + ( $(nproc) - 1) / frac ))` to get the `1/frac` rounded up number of CPUs.

The tests usually take about 1-2 seconds to run serially everywhere I've tested, so I didn't think about it. Have you seen them take substantial time?

PR #2310, opened 2026-07-03
Define _REENTRANT when OpenMP is enabled to fix Clang thread-safety

bbiiggppiigg:
Dyninst's bundled Sawyer/ROSE only compiles its pool allocator and shared-pointer reference counting as thread-safe when _REENTRANT is defined (dataflowAPI/rose/util/Sawyer.h). GCC's -fopenmp defines _REENTRANT implicitly, but Clang's -fopenmp/-fopenmp=libomp does not, so Clang builds ran the OpenMP-parallel parser with no-op allocator locks. This corrupted the pool free list and SValue reference counts, causing assertion failures and segfaults during parallel jump-table symbolic evaluation (only reproducible with >1 OpenMP thread).

Define _REENTRANT globally when USE_OpenMP is set, matching GCC's behavior and keeping the setting consistent across all translation units to avoid ODR mismatches in Sawyer's shared-object layout.

Required for #2298.

### DIFF
--- a/dataflowAPI/rose/util/Sawyer.h
+++ b/dataflowAPI/rose/util/Sawyer.h
@@ -8,6 +8,22 @@
 #ifndef Sawyer_H
 #define Sawyer_H
 
+// Dyninst: define _REENTRANT before any system header so Sawyer selects its
+// thread-safe path (the SAWYER_MULTI_THREADED / SAWYER_THREAD_LOCAL block
+// below keys on _REENTRANT). Dyninst runs its parser in parallel (OpenMP), so
+// Sawyer's pool allocator, shared-pointer reference counting, and per-thread
+// PRNG (fastRandomIndex) must be thread-safe. The compiler only defines
+// _REENTRANT when invoked with -pthread, and Dyninst cannot rely on that:
+// CMake's Threads::Threads adds -pthread as a *compile* flag only when
+// THREADS_HAVE_PTHREAD_ARG is true, which is false whenever pthread lives in
+// libc (glibc >= 2.34), leaving Threads::Threads an empty target. When
+// _REENTRANT is undefined these classes compile with no-op locks and shared
+// "thread-local" state, which crashed the parallel parser under clang.
+// Guarded so it is a no-op where -pthread already defined it.
+#ifndef _REENTRANT
+#define _REENTRANT 1
+#endif
+
 #include <stddef.h>
 #include <stdint.h>
 #include <vector>


bbiiggppiigg:
no testuiste failure on coriander.

kupsch:
Enabling OpenMP does cause internal parallelism, but what if Dyninst was compiled with OpenMP disabled and the application still used parallelism that used ROSE classes? Should this just be unconditionally set? It is unlikely to ever happen, but it could. Thoughts?

hainest:
> Dyninst's bundled Sawyer/ROSE only compiles its pool allocator and shared-pointer reference counting as thread-safe when _REENTRANT is defined (dataflowAPI/rose/util/Sawyer.h).

That is just the wrong way to detect if openmp has been enabled.

> GCC's -fopenmp defines _REENTRANT implicitly, but Clang's -fopenmp/-fopenmp=libomp does not

https://www.man7.org/linux/man-pages/man7/feature_test_macros.7.html

That macro has been deprecated since glibc-2.25 (released in 2017). Since the Rose code is checking for it explicitly, we'll need to use it but I'm not sure we should enable it for all sources. I'd suggest moving this to dataflowAPI/CMakeLists.txt and using

```cmake
if(USE_OpenMP)
  set_source_files_properties(${PROJECT_SOURCE_DIR}/dataflowAPI/rose/util/Sawyer.h
  PROPERTIES COMPILE_DEFINITIONS "_REENTRANT")
endif()
```

hainest:
> Enabling OpenMP does cause internal parallelism, but what if Dyninst was compiled with OpenMP disabled and the application still used parallelism that used ROSE classes? Should this just be unconditionally set? It is unlikely to ever happen, but it could. Thoughts?

The manpage I linked suggests using `_POSIX_C_SOURCE`. Maybe we should do that instead?

bbiiggppiigg:
> ```cmake
> if(USE_OpenMP)
>   set_source_files_properties(${PROJECT_SOURCE_DIR}/dataflowAPI/rose/util/Sawyer.h
>   PROPERTIES COMPILE_DEFINITIONS "_REENTRANT")
> endif()
> ```

Setting source file properties on header does not propogate to the source files that include it.
So if we want to do it this way we need to do it for all the files that include it?

Maybe we can just set it unconditionally, as I don't see what could go wrong if we make something thread safe and run it with a single thread.

bbiiggppiigg:
> Enabling OpenMP does cause internal parallelism, but what if Dyninst was compiled with OpenMP disabled and the application still used parallelism that used ROSE classes? Should this just be unconditionally set? It is unlikely to ever happen, but it could. Thoughts?
> 
> The manpage I linked suggests using `_POSIX_C_SOURCE`. Maybe we should do that instead?

The man page says This macro is automatically defined if one compiles with cc -pthread.
I've tested it and it seems like clang does define _REENTRANT when passed with -pthread.
So maybe the fix is to compile parseAPI with -pthread, (maybe when openmp is used?)

kupsch:
A simpler solution would be to add `#define _REENTRANT` (and a comment) in `Sawyer.h` before the `#ifdef _REENTRANT` and its preceding comment.

hainest:
The latest version of ROSE uses this check in many more places than the version we have. I think the best thing to do is link parseAPI against `Thread::Thread`.

bbiiggppiigg:
@hainest 

Can you check if #2311 is more inline with what you think, and whether we want to link with pthread unconditionally.

bbiiggppiigg:
closed in favor of #2311

bbiiggppiigg:
@kupsch @hainest 

This is reopend since the #2311 fix couldn't work. 
See explanation in #2322, which is going to revert the merged attempted fix.

Fortunately these two PRs are independent in terms of edits.

kupsch (approved):

PR #2311, opened 2026-07-03
Compile parseAPI with -pthread to fix Clang thread-safety

bbiiggppiigg:
Dyninst's bundled Sawyer/ROSE only compiles its pool allocator and shared-pointer reference counting as thread-safe when _REENTRANT is defined (dataflowAPI/rose/util/Sawyer.h). GCC's -fopenmp defines _REENTRANT implicitly, but Clang's -fopenmp=libomp does not, so Clang builds ran the OpenMP-parallel parser with no-op allocator locks. This corrupted the pool free list and SValue reference counts, causing assertion failures and segfaults during parallel jump-table symbolic evaluation (only reproducible with >1 OpenMP thread).

Rather than defining the deprecated _REENTRANT macro directly, set THREADS_PREFER_PTHREAD_FLAG and link parseAPI against Threads::Threads when USE_OpenMP is enabled. On GCC/Clang, -pthread defines _REENTRANT as a documented side effect (see feature_test_macros(7)), so this is the standard, compiler-independent way to compile threaded code. All Sawyer/ROSE sources compile into the single parseAPI target and none of the affected types are exposed in public headers, so scoping the flag to parseAPI keeps the shared-object layout consistent (no ODR mismatch) without touching the vendored Sawyer header.

This is an alternative to the global add_compile_definitions(_REENTRANT) approach in PR #2310, offered for reviewer comparison.

### DIFF
--- a/parseAPI/CMakeLists.txt
+++ b/parseAPI/CMakeLists.txt
@@ -94,6 +94,13 @@ dyninst_library(
   DYNINST_DEPS common instructionAPI ${SYMREADER}
   PUBLIC_DEPS Dyninst::Boost_headers
   PRIVATE_DEPS OpenMP::OpenMP_CXX
+               # -pthread (via Threads::Threads) defines _REENTRANT so the
+               # bundled Sawyer/ROSE sources compiled into parseAPI use their
+               # thread-safe pool allocator and refcounting. Linked
+               # unconditionally so these classes are thread-safe even in
+               # non-OpenMP builds where the application drives them in
+               # parallel.
+               Threads::Threads
 )
 # cmake-format: on
 
--- a/parseAPI/CMakeLists.txt
+++ b/parseAPI/CMakeLists.txt
@@ -95,11 +95,7 @@ dyninst_library(
   PUBLIC_DEPS Dyninst::Boost_headers
   PRIVATE_DEPS OpenMP::OpenMP_CXX
                # -pthread (via Threads::Threads) defines _REENTRANT so the
-               # bundled Sawyer/ROSE sources compiled into parseAPI use their
-               # thread-safe pool allocator and refcounting. Linked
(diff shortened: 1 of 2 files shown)


hainest:
@bbiiggppiigg Yes, this would be my preferred solution.

bbiiggppiigg:
@kupsch @hainest 

What are you guys opinion on whether this linking should be conditional or unconditional.
Once you guys reach a decision, I'll do a test on this again.

hainest:
> @kupsch @hainest
> 
> What are you guys opinion on whether this linking should be conditional or unconditional. Once you guys reach a decision, I'll do a test on this again.

The conditional linking to `Threads::Threads` is fine. Ultimately, I don't think it would be detrimental to always have it be used. I also rather doubt that anyone is using Dyninst without OpenMP enabled. I don't have a strong opinion.

parseAPI/CMakeLists.txt
```diff
@@ -94,6 +94,13 @@ dyninst_library(
   DYNINST_DEPS common instructionAPI ${SYMREADER}
   PUBLIC_DEPS Dyninst::Boost_headers
   PRIVATE_DEPS OpenMP::OpenMP_CXX
+               # -pthread (via Threads::Threads) defines _REENTRANT so the
+               # bundled Sawyer/ROSE sources compiled into parseAPI use their
+               # thread-safe pool allocator and refcounting. Linked
+               # unconditionally so these classes are thread-safe even in
+               # non-OpenMP builds where the application drives them in
+               # parallel.
```
kupsch: This comment could less verbose

```suggestion
               # -pthread (via Threads::Threads) defines _REENTRANT so the
               # Sawyer/ROSE pool allocator is thread safe
```

kupsch (approved):

Issue #2155, opened 2026-02-18
DynInst examples fail on some SpecCPU 2017 benchmarks


Capo80:
**Intention**
We were trying to run the ```codeCoverage``` example on the SpecCPU benchmarks and some fail either with a crash or invalid output.
We tried to write our own code to instrument only some functions but the problem remains the same.

**Describe the bug**
The following SpecCPU benchmarks fail with the following behavior.

|  Program | Source Language | Result |
|---|---|---|
|   gcc | C | Segmentation fault |  
|   omnetpp | C++ | Segmentation fault | 
|   leela |  C++ | Aborted at the very end, when its trying to exit | 
|   deepsjeng | C++ |Changes output |

We tried the same program with the native gcc on the machine and it works, so the problem is related to the benchmark. 

**To Reproduce**

Using the codeCoverage example from [here](https://github.com/dyninst/examples/tree/master/codeCoverage)

First we compile all the binaries then we access them directly in the folder, it is needed to copy all files from the run folder into the folder where we are running this commands, i use the test size input to speed up the process:
```
GCC
> ./code_coverage /paht/to/spec_cpu/benchspec/CPU/602.gcc_s/exe/sgcc_peak.mytest-m64 test_gcc
... lots of output, no errors ...
> ./test_gcc /paht/to/spec_cpu/benchspec/CPU/602.gcc_s/run/test_run_name/gcc-pp.c -O5 -fipa-pta -o gcc-pp.opts-O5-fipa-pta.s

/home/user/spec_cpu/benchspec/CPU/602.gcc_s/run/run_peak_refspeed_mytest-m64.0009/gcc-pp.c: In function 'fibheap_delete_node':
/home/user/spec_cpu/benchspec/CPU/602.gcc_s/run/run_peak_refspeed_mytest-m64.0009/gcc-pp.c:19958:49: warning: overflow in implicit constant conversion
Segmentation fault
```
```
OMNETPP

> ./code_coverage /home/user/spec_cpu/benchspec/CPU/620.omnetpp_s/exe/omnetpp_s_peak.mytest-m64 test_omne
... lots of output, no errors ...
> ./test_omne -c General -r 0
OMNeT++ Discrete Event Simulation  (C) 1992-2008 Andras Varga, OpenSim Ltd.
Version: 4.0, build: 090310-10709, edition: Academic Public License -- NOT FOR COMMERCIAL USE
See the license for distribution terms and warranty disclaimer
Setting up Cmdenv...
[...]

PR #2312, opened 2026-07-05
Relocation: classify springboards on a complete graph (issue #2155)

bbiiggppiigg:
Test case / binary (Related to #2155 )
------------------
Reproduced with dyninst's code_coverage example instrumenting the SPEC CPU2017 602.gcc_s benchmark binary ("sgcc"), built -O2 + LTO. code_coverage inserts a counter at every function entry, so every function is relocated and every jump-table target block needs a springboard. gcc's source is full of large switches, and at -O2/LTO these compile to dense PIC jump tables whose case blocks are tiny landing pads -- sometimes only a few bytes long and packed immediately adjacent. In function build_unary_op the crash site has two such case-target blocks 3 bytes apart:

    4eb514: 48 89 eb     mov %rbp,%rbx          ; jump-table target (3 bytes)
    4eb517: 8b 43 38     mov 0x38(%rbx),%eax    ; the next jump-table target

Both are reached by indirect jumps (0x4eb353 -> 0x4eb514, and 0x4eb57c/0x4eb748 -> 0x4eb517). All addresses below are original vaddrs in that sgcc binary. A dyninst springboard is a 5-byte jmp rel32, so a springboard at 0x4eb514 spans 0x4eb514..0x4eb519 and overwrites 0x4eb517.

Bug
---
CodeMover::finalizeRelocBlocks walked the RelocGraph once, calling linkRelocBlocks() and determineSpringboards() back-to-back per block. But an in-edge whose source is another RelocBlock is deferred by getPredecessors() and only materialized when that source block is later linked (via its getSuccessors -> makeEdge). So a block was classified against a half-built graph: determineSpringboards()'s IndirBlockEntry upgrade requires the block's INDIRECT in-edge to already exist, but for a jump-table target whose dispatch block appears later in the list (a backward indirect jump: dispatch addr > target addr) that edge is not present yet, so the target is mis-classified as Suggested.

Here 0x4eb514's dispatch (0x4eb353) precedes it -> correctly IndirBlockEntry; 0x4eb517's dispatches (0x4eb57c, 0x4eb748) follow it -> mis-classified Suggested. Springboards install per priority pass (FuncEntry > IndirBlockEntry > Suggested), high addr to low within a pass. So 0x4eb514 (IndirBlockEntry) is installed first and lays its 5-byte jump; 0x4eb517 (Suggested, later pass) then cannot fit any springboard in the bytes 0x4eb514 already claimed and gets none. At runtime the switch dispatches to 0x4eb517, which is now 3 bytes into 0x4eb514's jump, executing its displacement bytes as an instruction -> SIGSEGV.

Fix
---
Split finalizeRelocBlocks into two passes -- link all RelocBlocks first, then classify all of them -- so determineSpringboards always reads a complete graph. 0x4eb517 is then correctly IndirBlockEntry, installs its springboard first (higher addr, same pass), and 0x4eb514 falls back to a trap instead of clobbering it. Edge-set preserving; only the read timing changes. Verified: instrumented sgcc runs to completion (exit 0); deepsjeng unaffected.

### DIFF
--- a/dyninstAPI/src/Relocation/CodeMover.C
+++ b/dyninstAPI/src/Relocation/CodeMover.C
@@ -118,8 +118,26 @@ void CodeMover::finalizeRelocBlocks() {
    if (finalized_) return;
 
    finalized_ = true;
+   // Two phases, deliberately NOT fused. linkRelocBlocks() builds a block's
+   // edges, but an in-edge whose source is another RelocBlock is deferred and
+   // only materialized when that *source* block is later linked (via its
+   // getSuccessors). determineSpringboards() classifies a block from its
+   // inEdges_, and the IndirBlockEntry upgrade requires the block's INDIRECT
+   // in-edge to already exist. If the two are fused into one loop, a block is
+   // classified against a half-built graph: a jump-table target whose dispatch
+   // block appears later in the list (e.g. a backward indirect jump) has no
+   // INDIRECT in-edge yet, so it is mis-classified as Suggested instead of
+   // IndirBlockEntry. In a later, lower-priority pass it can then lose the
+   // springboard slot to an adjacent block whose 5-byte jump springboard
+   // overwrites its entry, so an indirect dispatch to it lands mid-jump ->
+   // SIGSEGV. This bites -O2/LTO code with dense jump tables and tiny (<5-byte)
+   // case blocks packed together -- e.g. two case targets 3 bytes apart in one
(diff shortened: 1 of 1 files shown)


kupsch (approved):
Think if there is a better title for the PR and commit, otherwise looks good

bbiiggppiigg:
> Think if there is a better title for the PR and commit, otherwise looks good

Somthing like Fix springboard conflict resolution logic?

Issue #449, opened 2018-04-04
processCreate crashed on aarch64


wangzhijun2005:
Platform: aarch64:Linux:4.1.12:#46 SMP PREEMPT, CentOS

yum install elfutils-devel
# rpm -qa|grep elf
elfutils-libelf-devel-0.163-3.h1.aarch64
elfutils-0.163-3.h1.aarch64
elfutils-devel-0.163-3.h1.aarch64
elfutils-libelf-0.163-3.h1.aarch64
elfutils-libs-0.163-3.h1.aarch64

Pull the code from arm64/feature/relocation branch, turn on all the debug options of dyninst in debug.h,  then build dyninstAPI.

Test case 1:  Use the test program in https://github.com/dyninst/dyninst/issues/436, then run : 
export LD_LIBRARY_PATH=/usr/local/lib
export DYNINSTAPI_RT_LIB=/usr/local/lib/libdyninstAPI_RT.so
# ./test /usr/bin/echo 1
DYNINSTAPI_RT_LIB=/usr/local/lib/libdyninstAPI_RT.so
Creating process /usr/bin/echo in directory 
Arguments: (2)
   0: /usr/bin/echo
   1: 1
Environment: (0)
[281473411620864]dynProcess.C[85]: stdin: 0, stdout: 1, stderr: 2
[281473411620864]unix.C[353]: Candidate RTLib is /usr/local/lib/libdyninstAPI_RT.so
[281473411620864]unix.C[353]: Candidate RTLib is /usr/local/lib/libdyninstAPI_RT.so
[281473411620864]unix.C[353]: Candidate RTLib is /usr/local/lib/libdyninstAPI_RT.so
Created process 17343
[281473411620864]dynProcess.C[348]: attempting to bootstrap process 17343
[281473411620864]Processing initial shared objects
[281473411620864]----
[281473411620864]mapped_object.C[151]:  about to parseImage
[281473411620864]mapped_object.C[153]: name /usr/bin/echo, codeBase 0x0, dataBase 0x0
[281473411620864]image.C[1220]:  about to create image
[281473411620864]image.C[1415]:  opening file /usr/bin/echo
[281473411620864]image.C[1506]:  before findMain
[281473411620864]image.C[1513]: findMain analysis succeeded.
[281473411620864]image.C[1544]:  before determineImageType
New variable, mangled program_name, module echo...
New variable, mangled quote_quoting_options, module echo...
New variable, mangled exit_failure, module echo...
[...]

PR #2314, opened 2026-07-06
Make C++ exceptions catchable through relocated / instrumented code (#1437)

### LINKED ISSUE #1437: Dyninst instrumentation breaks C++ exception handling on Linux
This is an old issue with Dyninst going back decades, but after talking with the Dyninst team it may need to be redocumented.

In short, Dyninst instrumentation will break any any C++ exception handling that's implemented according to the Linux Standards Base (so on all Linux). If a function that is dyninst-relocated is on the stack when an exception is thrown, then the program will prematurely terminate when the the __throw() implementation encounters the function on the stack while unwinding.

This is because __throw() is extracting PC-return values from the stack, and looking up those return values in the ehframe table. If __throw() does not find an entry in ehframe for a PC then it terminates the program with an uncaught exception. Relocation modifies the PC-return values in the stack, and causes the lookup to always fail.

I've attached a reproducer:

- The mutatee throws and exception and tries to catch it in a function called 'relocateme'. If the exception is caught, it prints a message about working as expected and terminates.
- The mutator instruments every basic block in 'relocateme' and runs the mutatee.

When run alone the mutatee prints:

```
$ ./eh_mutatee
About to throw an exception...
Caught an exception. Program running as expected.
```

When run under the mutator, the mutatee prints:
```
$ ./eh_mutator
About to throw an exception...
terminate called after throwing an instance of 'int'
```

The easiest way to fix this might be with DWARF editting. The ehframe table is a slightly-modified version of DWARF3's FDE and CIE tables. If Dyninst could update the PC values in the ehframe table as it relocates code, then the exception system should work.

[ehtest.tar.gz](https://github.com/dyninst/dyninst/files/11638669/ehtest.tar.gz)

bbiiggppiigg:
## Summary

When Dyninst relocates a function — whether by the static rewriter or dynamic
instrumentation — it copies the instructions to new addresses but does not
reconstruct the DWARF unwind info (`.eh_frame`) or the C++ LSDA
(`.gcc_except_table`) that describe them. A C++ exception thrown from, or
propagating through, relocated code then finds no CFI and no call-site table at
the new PC, the unwinder fails, and the process calls `std::terminate` instead
of running the matching `catch`.

This PR synthesizes both tables for relocated code from the original function's
unwind info, mapped through the relocation's orig→reloc tracker, and delivers
them so libgcc can unwind through instrumented frames.

Fixes #1437.

## Approach

- **Generator** (`eh_frame_gen.C`): three passes over the `CodeTracker`s build an
  orig→reloc address map and per-function runs, decode each LSDA-bearing
  function's call sites, and emit one CIE/FDE (+LSDA) per relocated run. Handles
  the awkward cases that otherwise silently `terminate`: instrumentation snippets
  sharing an original address, hot/cold-split functions (`.cold` fragments),
  stack-realigning frames, mixed C/C++ personalities, interleaved neighbors, and
  foreign code pulled in by the runtime. (The per-decision rationale lives in the
  commit messages.)
- **Delivery** — static: the blobs are added as loadable regions and a
  `DT_DYNINST_EH_FRAME` dynamic tag points the runtime at them; the RT registers
  them at load via `dl_iterate_phdr` + `__register_frame`. Dynamic: an inferior
  region is allocated, the tables written in, and `__register_frame` invoked over
  an iRPC.
- **PIE**: the generator is load-bias–aware so link-time CFI/LSDA addresses
  resolve against runtime tracker addresses (non-PIE path is byte-for-byte
  unchanged).

## Platform support

| Arch | Static rewrite | Dynamic instrumentation |
|------|:--:|:--:|
| x86-64 | ✅ | ✅ |
| aarch64 | ✅ | — (dynamic instrumentation unimplemented, #449) |
| ppc64le | ✅ | — (blocked on a separate TOC/r2 relocation gap, not EH) |

## Testing

New in-tree CTest integration tests under
`tests/integration/dyninstAPI/exceptions/`: a driver + mutatees covering
single/nested/rethrow/cleanup/multi/base-catch/deep and hot-cold-split throws,
run through both the static rewriter and dynamic instrumentation, each built
position-independent and not, with per-test artifact isolation. Tests match only
the success sentinel, so any wrong-catch/crash/terminate/timeout fails.

- x86-64: full matrix green; the #1437 reproducer now catches.
- aarch64 / ppc64le: static + hot-cold matrices green.
- SPEC CPU 2017 spot checks (via `code_coverage`): leela byte-identical,
  deepsjeng clean, omnetpp instruments to a valid PIE ELF (0 section overlaps,
  4003 FDEs) and runs to completion.

## Merge dependencies

For ppc64le, this depends on the instruction-decoder fixes in #2335 / #2337
(`bc`/`bclr`) and #2338 (`isel`) — those are kept out of this branch and should
merge first.

## Notes

- Exception support is intentionally excluded for static (non-dynamic) binaries —
  the runtime is linked at emit time and would mis-relocate its reference to the
  region.
- Debug output is on the `DYNINST_DEBUG_EH` channel.

### DIFF
--- a/dwarf/h/dwarfFrameParser.h
+++ b/dwarf/h/dwarfFrameParser.h
@@ -98,6 +98,30 @@ public:
             std::vector<VariableLocation> &locs,
             FrameErrors_t &err_result);
 
+    // the classified CFI rule for one register over
+    // [range.first, range.second), one entry per CFI row. Says where the
+    // caller's value of `reg` lives at each PC sub-range; used to re-emit
+    // unwind info for relocated code. Rules that cannot be expressed as one
+    // of the simple kinds (DWARF expressions, val_offset) report Undefined.
+    struct FrameRegRule {
+        Address lowPC, hiPC;
+        enum Kind { SameValue = 0, AtCFAOffset = 1, InRegister = 2,
+                    Undefined = 3 } kind;
+        long offset;      // AtCFAOffset: value is stored at CFA+offset
+        unsigned regnum;  // InRegister: DWARF number of the holding register
+    };
+    bool getRegRulesForFunction(
+            std::pair<Address, Address> range,
+            MachRegister reg,
+            std::vector<FrameRegRule> &rules,
+            FrameErrors_t &err_result);
+    // As above but keyed on a raw DWARF register number, so a CFI column with
+    // no Dyninst MachRegister (e.g. ppc64 return address, DWARF 65) is reachable.
+    bool getRegRulesForDwarf(
+            std::pair<Address, Address> range,
+            int dwarf_reg,
+            std::vector<FrameRegRule> &rules,
+            FrameErrors_t &err_result);
 
 private:
 
--- a/dwarf/src/dwarfFrameParser.C
+++ b/dwarf/src/dwarfFrameParser.C
@@ -218,28 +218,136 @@ bool DwarfFrameParser::getRegsForFunction(
         {
             Dwarf_Frame * frame = NULL;
             int result = dwarf_cfi_addrframe(cfi_data[i], next_pc, &frame);
-            if(result==-1) break;
+            // No FDE covers next_pc: a padding gap between functions. When the
+            // caller's range legitimately spans more than one function (e.g.
+            // sampling a hot/cold-split function whose partitions bracket other
+            // functions), such gaps must be skipped, not treated as the end of
+            // the range -- otherwise rows past the first gap are lost. Advancing
+            // one byte is fine: inter-function padding is tiny.
+            if(result==-1) { next_pc++; continue; }
 
             Dwarf_Addr start_pc, end_pc;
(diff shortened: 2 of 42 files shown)


bbiiggppiigg:
Tested on 

x86_64: mahir coirander 
cyclops: ppc64le
zeroah: aarch64

hainest:
I think we should also remove the [note](https://github.com/dyninst/dyninst/blob/master/README.md?plain=1#L41) in the README about this.

bbiiggppiigg:
> I think we should also remove the [note](https://github.com/dyninst/dyninst/blob/master/README.md?plain=1#L41) in the README about this.

I don't know. Felt like  maybe we should put the support matrix there.
For example, your table in #1922 
And my table in the PR description.

bbiiggppiigg:
depends on #2379

PR #2316, opened 2026-07-07
AMDGPU: fix SOPP short-branch relocation off-by-one (S_BRANCH PC+4)

bbiiggppiigg:
gfx908 S_BRANCH / s_cbranch_* (SOPP) compute their destination relative to the instruction AFTER the branch: target = (PC_of_branch + 4) + SIMM16*4. The short-branch emitters in codegen-amdgpu.C encoded SIMM16 = disp/4, omitting the -1 that accounts for that +4, so every displacement-computed short branch landed one instruction (4 bytes) past its intended target.

Fix the three displacement-based emitters to compute SIMM16 = disp/4 - 1:
  - insnCodeGen::generateBranch (short-jump path)
  - insnCodeGen::modifyJump
  - insnCodeGen::modifyJcc (the jump-to-target of the conditional-branch expansion)

The -1 is direction-agnostic (it is +4 expressed in dword units) and holds for both forward and backward jumps. emitLongJump already accounts for this (fromAddress + 4); the literal emitShortJump(1) skip-jumps in modifyJcc are intentionally SIMM16=1 and are unchanged.

This affects short branches only, which are emitted when relocating a function that has internal control flow (loops, if/else); straight-line functions relocate using only long jumps, so the bug was previously latent.

### DIFF
--- a/dyninstAPI/src/codegen/codegen-amdgpu.C
+++ b/dyninstAPI/src/codegen/codegen-amdgpu.C
@@ -50,7 +50,9 @@ void insnCodeGen::generateIllegal(codeGen & /* gen */) {
 void insnCodeGen::generateBranch(codeGen &gen, Dyninst::Address from, Dyninst::Address to,
                                  bool /* link */) {
   long disp = (to - from);
-  long wordOffset = disp / 4;
+  // S_BRANCH target = (PC_of_branch + 4) + SIMM16*4, so SIMM16 = disp/4 - 1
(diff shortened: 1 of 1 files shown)


bigtrak:
I'd suggest moving that computation to a common function that does that computation,
instead of having that  particular computation buried in a few random source file
locations.     That way all places doing that will be using the same code.

If there is a different displacement computation for something else, then
that could also be put there, and used for the other displacement methods.

kupsch (approved):

PR #2317, opened 2026-07-07
AMDGPU: represent vcc/exec/flat_scratch as their 32-bit halves

bbiiggppiigg:
Summary

On AMDGPU, MachRegister::getBaseRegister() aliased the 32-bit halves to a combined 64-bit register (vcc_lo/vcc_hi → vcc, exec_lo/hi → exec, flat_scratch_lo/hi → flat_scratch_all). Those combined registers are not the entries used by any consumer — the liveness index map (machRegIndex_amdgpu_*), the register-conversion maps (RegisterConversion-amdgpu.C), and the register space are all keyed on the _lo/_hi halves. This made two things silently misbehave:

- Liveness: LivenessAnalyzer::calcRWSets computes getIndex(getBaseRegister(reg)); for the halves this resolved to the combined register, which isn't in the map, so getIndex returned -1 and the if (index >= 0) guard dropped the read and the write. VCC/EXEC/FLAT_SCRATCH were never tracked — e.g. VCC was reported dead across a live carry chain (v_add_co_u32 defines VCC, v_addc_co_u32 uses it).
- Register conversion: convertRegID() looked up the combined register, missed, and returned ignored.

Investigating the fix also surfaced inconsistencies in the generated decoder/register data across architectures (decodeOPR_VCC returned vcc_lo on gfx908 but the combined vcc on gfx90a/gfx940; decodeOPR_FLAT_SCRATCH returned the combined register everywhere; flat_scratch_lo was declared BITS_64 on gfx908/gfx940 but BITS_32 on gfx90a).

Changes

Commit 1 — treat the halves as base registers. getBaseRegister returns *this for vcc_lo/vcc_hi, exec_lo/exec_hi, and flat_scratch_lo/flat_scratch_hi on all three AMDGPU arches. They are independently addressable 32-bit registers and exactly what the maps above are keyed on. A 64-bit VCC/EXEC operand is already decoded as a MultiRegister of its two halves, so the combined form is not needed.

Commit 2 — remove the now-redundant combined registers. With nothing aliasing to them, vcc, exec, and flat_scratch_all are deleted from the register definitions. Supporting changes for uniformity/correctness:
- Decoders emit the _lo half (expanded to [lo, hi]) for VCC and FLAT_SCRATCH on all three arches.
- flat_scratch_lo is BITS_32 everywhere.
- Dead vcc/exec cases removed from MachRegister.C's type-query switches (the halves already carry the control-status classification).
- Unit tests (base_registers, type_queries) updated: each half is now its own base.

Removing the combined registers means any future code — or inconsistent generator data — that references a whole special register now fails to compile rather than silently corrupting liveness.

Testing

- Builds clean; all six AMDGPU MachRegister unit tests pass.
- Verified (via an instruction-level liveness dump) that VCC is now live across v_add_co/v_addc_co carry chains and each half reports itself as its base register.
- A downstream liveness-driven register-spill pass that previously faulted (clobbering an untracked live VCC) now runs correctly.

### DIFF
--- a/common/src/registers/MachRegister.C
+++ b/common/src/registers/MachRegister.C
@@ -172,56 +172,23 @@ namespace Dyninst {
         return *this;
       }
 
-      case Arch_amdgpu_gfx908: {
-        if(category == amdgpu_gfx908::MISC) {
-          switch(val()) {
-            case amdgpu_gfx908::ivcc_lo:
-            case amdgpu_gfx908::ivcc_hi:
-              return amdgpu_gfx908::vcc;
-            case amdgpu_gfx908::iexec_lo:
-            case amdgpu_gfx908::iexec_hi:
-              return amdgpu_gfx908::exec;
-            case amdgpu_gfx908::iflat_scratch_lo:
-            case amdgpu_gfx908::iflat_scratch_hi:
-              return amdgpu_gfx908::flat_scratch_all;
-          }
-        }
+      case Arch_amdgpu_gfx908:
+      case Arch_amdgpu_gfx90a:
+      case Arch_amdgpu_gfx940:
+        // On AMDGPU each of vcc_lo/vcc_hi, exec_lo/exec_hi and
+        // flat_scratch_lo/flat_scratch_hi IS a base register: an independently
+        // addressable 32-bit SGPR, and exactly the entries used by the liveness
+        // index map (machRegIndex_amdgpu_*) and the register-conversion maps.
+        // These must NOT alias to the combined 64-bit vcc/exec/flat_scratch_all:
+        // those combined registers are absent from those maps, so aliasing made
+        //   - LivenessAnalyzer::calcRWSets getIndex() return -1 -> VCC/EXEC/
+        //     FLAT_SCRATCH reads AND writes were silently dropped, so e.g. VCC was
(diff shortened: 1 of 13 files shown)


common/src/registers/MachRegister.C
```diff
@@ -172,56 +172,23 @@ namespace Dyninst {
...
+        // index map (machRegIndex_amdgpu_*) and the register-conversion maps.
+        // These must NOT alias to the combined 64-bit vcc/exec/flat_scratch_all:
+        // those combined registers are absent from those maps, so aliasing made
+        //   - LivenessAnalyzer::calcRWSets getIndex() return -1 -> VCC/EXEC/
+        //     FLAT_SCRATCH reads AND writes were silently dropped, so e.g. VCC was
+        //     reported dead across a live carry chain (v_add_co/v_addc_co) and a
+        //     liveness-driven register spill would clobber it, and
+        //   - convertRegID() miss the lookup and return "ignored".
+        // A 64-bit VCC/EXEC operand is already emitted by the decoder as a
+        // MultiRegister of its two halves, so nothing needs the combined form here.
```
kupsch: This comment makes sense in the context of this PR, but once the change is made I don't think it is useful. It can probably be deleted

kupsch (approved):

PR #2318, opened 2026-07-09
proccontrol: fix UAF in ProcessSet::terminate

bbiiggppiigg:
ProcessSet::terminate snapshots the group's processes as raw int_process* pointers, then calls waitAndHandleEvents(false) to drain preTerminate side effects before looping over the snapshot to terminate() and finally delete each process.

If one of the snapshotted processes finishes exiting during that intervening event handling, its int_process is deleted: ~int_process nulls the owning Process wrapper's llproc_, but the raw int_process* captured in the snapshot is left dangling.  The later terminate() then makes a virtual call (plat_terminate) on freed memory -- observed as "pure virtual method called" / SIGABRT once the freed block still carries the base-class vtable -- and the final loop double-deletes it.

The window is narrow, so this is normally latent; it becomes a reliable, roaming crash under anything that shifts teardown event timing (e.g. a full testsuite -all run, where whichever managed process exits inside the window is the victim, while each test passes in isolation).

Snapshot the reference-counted Process::ptr wrappers instead of raw int_process*.  The wrapper always outlives the int_process, so llproc() is always safe to call and returns NULL exactly when the int_process has been freed.  Re-derive llproc() after each waitAndHandleEvents and skip any process that was freed, in both the terminate loop and the final delete loop, so a process reclaimed mid-teardown is never terminated or double-deleted.

### DIFF
--- a/proccontrol/src/procset.C
+++ b/proccontrol/src/procset.C
@@ -1624,12 +1624,19 @@ bool ProcessSet::terminate() const
    int_process::waitAndHandleEvents(false);
 #endif
 
-   // Clean out the event queue before we terminate; otherwise we can race
-   set<int_process *> procs;
+   // Clean out the event queue before we terminate; otherwise we can race.
+   // Hold the reference-counted Process wrappers rather than raw int_process*:
+   // the waitAndHandleEvents() calls below can finish an in-flight exit and
+   // delete the underlying int_process, which nulls the wrapper's llproc_.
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
tested on coriander.

bbiiggppiigg:
Required for #2315

kupsch (approved):
Assuming all the locking is correct, which would take a lot of time, this looks good.

PR #2319, opened 2026-07-09
dyninstAPI: fix -fipa-ra caller register clobber

### LINKED ISSUE #724: Buggy register liveness detection, rdi gets clobbered
**Intention**
The main intention is to insert instrumentation code at both the entry point of a function and the exit point of a function.  It only seems to be a problem in the instrumentation code that is inserted at the end of the function. Seems to be similar to issue #434.

**Describe the bug**
The instrumentation code involves some syscalls which make use of the rdi register, this register is unfortunately not saved properly and gets filled with an a syscall parameter. Later in the code the rdi contents of the rdi register are used as a memory address, but the contents are garbage so the program segfaults. If I disable the liveness analysis the newly mutated program works properly. Listed below is the relevant assembly.
```
...
  50ad3b:       50                      push   %rax
  50ad3c:       51                      push   %rcx
  50ad3d:       52                      push   %rdx
  50ad3e:       48 8d 64 24 f8          lea    -0x8(%rsp),%rsp
  50ad43:       48 b8 00 00 00 00 00    movabs $0x0,%rax
  50ad4a:       00 00 00
  50ad4d:       ff 15 45 b8 ff ff       callq  *-0x47bb(%rip)        # 506598 <fg_perf_stop@Base>
  50ad53:       48 bf 24 00 00 00 00    movabs $0x24,%rdi
  50ad5a:       00 00 00
  50ad5d:       48 b8 00 00 00 00 00    movabs $0x0,%rax
  50ad64:       00 00 00
  50ad67:       ff 15 33 b8 ff ff       callq  *-0x47cd(%rip)        # 5065a0 <fg_perf_read@Base>
  50ad6d:       48 b8 00 00 00 00 00    movabs $0x0,%rax
  50ad74:       00 00 00
  50ad77:       ff 15 2b b8 ff ff       callq  *-0x47d5(%rip)        # 5065a8 <fg_perf_start@Base>
  50ad7d:       48 8d 64 24 08          lea    0x8(%rsp),%rsp
  50ad82:       5a                      pop    %rdx
  50ad83:       59                      pop    %rcx
  50ad84:       58                      pop    %rax
  50ad85:       48 8b 24 24             mov    (%rsp),%rsp
...
```

The calls in fg_perf_* use ioctls syscalls which ends up clobbering the rdi register. The following code gets called later, which rdi is properly saved but it is too late as it contains garbage contents, which results in a segfault because the garbage contents are then dereferenced.
```
...
  50ac6e:       5f                      pop    %rdi
  50ac6f:       5e                      pop    %rsi
  50ac70:       59                      pop    %rcx
  50ac71:       41 58                   pop    %r8
  50ac73:       48 8b 24 24             mov    (%rsp),%rsp
  50ac77:       48 8b 07                mov    (%rdi),%rax
...
```

**To Reproduce**
It is directly reproducible on most of the programs in my test bench. I can include any other requested information.

**Expected behavior**
The contents of the rdi register are saved and restored properly and the mutated program doesn't segfault.

x86_64
DynInst version 10.1
glibc 2.30, gcc 9.2

### LINKED ISSUE #2313: Caller-saved register clobber under GCC `-fipa-ra`
Sub issue of #2155 

Binaries compiled with `-fipa-ra` (on at `-O2`/LTO) lets a caller keep a live value in a *caller-saved*
register across a call to a **local** callee that provably never writes it. 

In the sgcc binary provided,
`fold_builtin_inf` keeps `&local` in `%rsi` across `call real_inf` because `real_inf` only writes `{rax,rcx,rdx,rdi}`.
Dyninst instruments `real_inf`'s entry; its intraprocedural liveness sees `%rsi` dead inside`real_inf`, so the base trampoline doesn't preserve it — but `incFuncCoverage` (standard ABI)
clobbers `%rsi` → the caller's live pointer is destroyed → SIGSEGV on the following `rep movsl`.

**Why the existing #2288 fix missed it.** #2288 gated the "save it anyway" logic in
`shouldSaveReg` (emit-x86.C) on the callee being a *clone* (`Symbol::isClone`, mangled name has a
dot: `.isra`/`.constprop`/…). `real_inf` is a plain **local** function, no dot → not covered.
`-fipa-ra` applies to *any* local callee, not just clones.

I'm not sure what the right fix should be. 

1.  AI suggested extend the gate to also fire when the instrumented function has `SL_LOCAL` linkage, which worked for the binary, but I'm not sure if that fix is too limited or too broad.
2. We could save all caller-saved registers, at the cost of high runtime cost
3. We could compute the clobber set of the callee. Then a caller-saved register used in the instrumentation can only not be saved if is dead at entry, and the callee clobbers it. (As otherwise we cannot tell whether the caller is using some weird calling convention that tries to not save it).

bbiiggppiigg:
## Problem

GCC -fipa-ra (interprocedural register allocation, enabled at -O2 and
aggressive under LTO) narrows a callee's clobber set from a caller's
point of view whenever the callee's body is fully visible and the call
cannot be interposed. A direct caller may then keep a value in a
caller-saved register live across the call, because the callee provably
never writes that register.

Instrumenting such a callee breaks this contract: dyninst's
intra-procedural liveness sees the register dead inside the callee, the
base trampoline does not preserve it, and the inserted snippet (a
standard-ABI call) clobbers it -- silently corrupting the caller.
Observed on an LTO-built SPEC CPU 2017 gcc instrumented at every
function entry with the code_coverage example: fold_builtin_inf keeps
&local in %rsi across a call to the plain local function real_inf
(which only writes {rax,rcx,rdx,rdi}); the instrumentation destroys
%rsi and the caller SIGSEGVs on a following rep movsl.

PR https://github.com/dyninst/dyninst/pull/2288 handled the special case of GCC local clones
(.isra/.constprop/..., detectable by name suffix), but -fipa-ra applies
to any callee whose definition binds directly, clone or not.

## Fix

Extend the conservative-save gate in shouldSaveReg (emit-x86.C): when
the instrumented callee's definition binds directly -- so a caller may
hold a narrowed -fipa-ra contract with it -- preserve every
caller-saved GPR the inserted snippet clobbers, even those liveness
reports dead. A definition binds directly when:

- any of its symbols has SL_LOCAL linkage; or
- any of its symbols is GLOBAL/WEAK with non-DEFAULT visibility
  (hidden / protected / internal): not preemptible by another object,
  so same-object callers bind directly and -fipa-ra may narrow it
  exactly as for a local. This is decisive under LTO, gcc's most
  aggressive -fipa-ra setting: LTO internalizes symbols as
  GLOBAL/HIDDEN in .symtab rather than demoting them to LOCAL. A survey
  of the LTO-built SPEC CPU 2017 binaries found ~31% of all functions
  carry a GLOBAL/HIDDEN symbol (5892 of 18814), which a linkage-only
  gate would leave on the unprotected fast path. (On distro-shipped
  binaries the difference is ~0.05%: the static linker demotes hidden
  globals to LOCAL in shared libraries, so GLOBAL/HIDDEN survives
  mainly in executables.); or
- it has no symbol in the binary at all: functions discovered by
  parsing (ParseAPI src() != HINT -- recursive traversal, gap
  heuristics, on-demand) get a synthesized "targXXX" symbol with
  SL_GLOBAL linkage, so a linkage check alone would wrongly take the
  fast path. In stripped binaries the vulnerable direct-bound callees
  are precisely the functions whose symbols are gone.

Ordinary default-visibility globals are interposable and keep the
existing liveness-pruned fast path; the check short-circuits before any
per-register work.

## Narrowing the save set

The saved set is then reduced by gcc's own promise condition: -fipa-ra
can only let a caller keep register R live across a call if the callee
(transitively) never writes R, so a register the callee's own code
writes needs no conservative save. The set becomes:

  snippet-clobbered  intersect  caller-saved GPRs  intersect
  never-written-by-the-callee

calleeWrittenCallerSavedGPRs scans the instrumented function's ParseAPI
blocks once (cached per parse_func, computed only for functions that
reach the gate). The direction of conservatism is inverted relative to
a clobber summary -- an omission from the mask only adds a harmless
save, while a wrong addition can suppress a required save and corrupt a
caller -- so call/syscall/interrupt instructions and undecodable bytes
add NOTHING (gcc's interprocedural summary may know an unresolved
transitive callee preserves R and let a caller rely on it). The scan is
exact for leaf functions, -fipa-ra's main beneficiaries; for callees
that use the usual scratch registers the extra save set is empty, so
the gate costs nothing there. Non-x86-64 code keeps the unnarrowed
conservative save.

## Verification

- SPEC CPU 2017 gcc (LTO, PIE), every function entry instrumented:
  previously SIGSEGV in fold_builtin_inf; now compiles its workload to
  completion with correct output. The narrowing preserves exactly the
  needed save (%rsi at real_inf) while dropping the unnecessary ones.
- SPEC deepsjeng: no regression.
- Testsuite -rewriter -allcomp (gcc+g++): 113/113 PASSED (glibc 2.36).

## Scope and known limitations

- Does not fix https://github.com/dyninst/dyninst/issues/724 (same bug family): there the victim is a plain
  default-visibility global in an executable, where interposition never
  applies and gcc narrows globals too -- no symbol-level signal exists.
  Covering it needs an object-type (executable vs shared library) term
  in the gate, a policy trade-off left for follow-up; with the
  per-register narrowing here that term would be near-zero-cost.
- A function reachable through both a default-visibility global and a
  local alias (gcc's foo/foo.localalias, glibc's __GI_*) takes the
  conservative path by design -- the local alias is the direct-bound,
  narrowable route.
- Shared libraries built with -fno-semantic-interposition / -Bsymbolic
  are indistinguishable from the binary; not covered.

## Related

- https://github.com/dyninst/dyninst/pull/2288 -- clone-only precursor of the gate; retained and extended.
- https://github.com/dyninst/dyninst/pull/1712 -- orthogonal: reworks the ABI liveness tables. The ABI model
  at a return is correct for standard-ABI callers; the issue here is
  that -fipa-ra callers aren't standard.
- https://github.com/dyninst/dyninst/pull/2325 -- complementary: computes the snippet-side clobber set
  precisely (including opaque PatchAPI snippets). It composes with this
  change through the unchanged definedRegs interface, no code
  dependency: once both land, the saves narrow to what the snippet
  really writes intersect what the callee never writes.

### DIFF
--- a/dyninstAPI/src/emit-x86.C
+++ b/dyninstAPI/src/emit-x86.C
@@ -41,6 +41,11 @@
 #include "codegen/codegen.h"
 #include "patching/function.h"
 #include "Symbol.h"
+#include "Function.h"
+#include "Instruction.h"
+#include "registers/x86_64_regs.h"
+#include <unordered_map>
+#include <set>
 #include "dyninstAPI/src/emit-x86.h"
 #include "dyninstAPI/src/inst-x86.h"
 #include "dyninstAPI/src/debug.h"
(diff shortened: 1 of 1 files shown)


dyninstAPI/src/emit-x86.C
```diff
@@ -1477,6 +1478,50 @@ static bool isCloneFunc(baseTramp *inst)
...
+      return true;   // no parse-level info: be conservative and save
+   if (pf->src() != Dyninst::ParseAPI::HINT)
+      return true;   // discovered by parsing: no symbol in the binary
+   Dyninst::SymtabAPI::Function *sf = pf->getSymtabFunction();
+   if (!sf)
+      return true;   // no symtab function: no symbol
+   std::vector<Dyninst::SymtabAPI::Symbol *> syms;
+   if (!sf->getSymbols(syms) || syms.empty())
+      return true;   // symbol-less
+   for (auto *s : syms) {
```
bbiiggppiigg, line 1538: Flip the loop

hainest:
@bbiiggppiigg @kupsch Are we still planning to replace this with a save-what-you-use strategy?

kupsch:
I think that was the conclusion, with further optimizations in the future

PR #2321, opened 2026-07-09
Harden proccontrol against concurrent process exit during event handling

bbiiggppiigg:
# Harden proccontrol against concurrent process exit during event handling

## Summary

Follow-up to #2318. That PR fixed a use-after-free in `ProcessSet::terminate`
where a raw `int_process*` was cached across a `waitAndHandleEvents()` call that
could delete the process. This PR fixes the **same bug class** at every other
site I could find where it occurs.

The pattern, in all cases: a function snapshots a raw `int_process*` (in a
local, a `set<int_process*>`, or a `pair<int_process*, ...>`), then calls a bare
event-processing routine (`waitAndHandleEvents` / `waitForAsyncEvent` /
`waitfor_startup`), then dereferences the raw pointer. If a process finishes
exiting during that call, its `int_process` is deleted — `~int_process()` nulls
the owning `Process` wrapper's `llproc_`, but the raw pointer that was snapshot
is left dangling. The later dereference is a UAF (typically a virtual call on
freed memory → `pure virtual method called` / `SIGABRT`, or a double-free).

The fix idiom (matching #2318): hold the reference-counted `Process::ptr`
wrapper instead of the raw pointer, and after each event-handling call re-derive
`int_process *p = wrapper->llproc()` and skip the process if it is now `NULL`
(NULL means its `int_process` was freed).

## The sites fixed

- **`infMalloc`** (`process.C`). Cached raw `int_process*` in `active_mallocs`,
  ran the inferior-malloc RPCs via `waitAndHandleEvents()`, then dereferenced
  the raw pointers. The RPC runs the target, so it can exit during the wait.
  Its sibling `infFree` was already written the safe way (`Process::ptr` +
  re-derive + NULL-check) — this makes `infMalloc` match.

- **process create / attach bootstrap** (`process.C`). `int_process::create`
  and `int_process::attach` re-dereferenced a raw `set<int_process*>` across the
  `waitForAsyncEvent` / `waitAndHandleEvents` / `waitfor_startup` calls that
  drive `post_create` / `post_attach` and startup; a process that exits during
  bootstrap is deleted inside those calls. (`attach` even read `proc->pid` after
  `waitfor_startup` had already deleted `proc`.) Also fixes a latent infinite
  loop in `attach`'s `plat_attachThreadsSync` loop, where an `errorstate`
  `continue` never advanced the iterator.

- **`LWPTrackingSet::refreshLWPs`** (`procset.C`). Held raw `int_process*` in
  `all_procs` / `change_procs` across `waitForAsyncEvent` and
  `waitAndHandleEvents`, then re-dereferenced them.

- **breakpoint install** (`procset.C`, `process.C`). `addBreakpointWorker`
  cached raw `int_process*` in `bp_installs` and called
  `addBreakpoint_phase2/phase3` after two `waitForAsyncEvent`s;
  `sw_breakpoint::create` held its `int_process*` argument across the
  memory/result waits. Both can lose the process if it exits mid-install.
  `addBreakpointWorker` now carries `Process::ptr` (worker set + its three
  callers); `sw_breakpoint::create` checks the wrapper's liveness after each
  wait.

## Not changed (investigated)

- **`ProcessSet::rmBreakpoint`**: looks like the pattern, but its `resp_to_proc`
  map is populated from an always-empty local, so the raw pointer is never
  actually dereferenced. There is a separate latent bug there (an error-path
  `resp_to_proc[resp]` yields NULL and trips `assert(proc)`), but it is not a
  UAF; left alone to keep this PR focused.

- **`freebsd.C`** has an analogous `set<int_thread*>` in its thread-cleanup
  path. It does not build on the Linux CI here, so it is untested and left for a
  platform owner to confirm/port.

## Verification

- Full testsuite `-all`: **1534/1534 pass, 0 crashes**, run on this branch
  rebased onto current `master` (which contains #2318). The create/attach and
  breakpoint paths are exercised by every test, so a clean run is strong
  evidence the refactor does not regress normal operation.
- The `terminate` UAF that motivated #2318 originally surfaced as a roaming
  ~1-per-`-all`-run flake (each affected test passing in isolation); the same
  presentation is what these siblings would produce under the right timing.

## Notes for reviewers

- Each commit is one site/family for reviewability; all use the identical
  `Process::ptr` + re-derive-`llproc()` + skip-NULL idiom introduced in #2318.
- The `attach` change is the largest (that function has ~9 loops over the
  process set); the infinite-loop fix in it is incidental but real.

### DIFF
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -5068,6 +5068,12 @@ sw_breakpoint *sw_breakpoint::create(int_process *proc, int_breakpoint *bp, Dyni
    is.addr = addr;
    is.bp = bp;
 
+   // The waitForAsyncEvent calls below can finish an in-flight exit of this
+   // process and delete it; hold the wrapper and check liveness after each
+   // wait.  (While the wrapper's llproc() is non-NULL it is still this same
+   // int_process; ~int_process nulls it on deletion.)
+   Process::ptr wrapper = proc->proc();
+
    result = proc->addBreakpoint_phase1(&is);
    if (!result)
       return NULL;
@@ -5075,6 +5081,10 @@ sw_breakpoint *sw_breakpoint::create(int_process *proc, int_breakpoint *bp, Dyni
       result = int_process::waitForAsyncEvent(is.mem_resp);
       if (!result) {
          perr_printf("Error waiting for result of memory response\n");
+         return NULL;
+      }
+      if (!wrapper->llproc()) {
+         perr_printf("Process exited during breakpoint install\n");
          return NULL;
       }
    }
@@ -5086,6 +5096,10 @@ sw_breakpoint *sw_breakpoint::create(int_process *proc, int_breakpoint *bp, Dyni
       result = int_process::waitForAsyncEvent(is.res_resp);
       if (!result) {
          perr_printf("Error waiting for result of result response\n");
+         return NULL;
+      }
+      if (!wrapper->llproc()) {
+         perr_printf("Process exited during breakpoint install\n");
          return NULL;
       }
    }
--- a/proccontrol/src/process.C
+++ b/proccontrol/src/process.C
@@ -87,13 +87,18 @@ void Process::version(int& major, int& minor, int& maintenance)
 
 bool int_process::create(int_processSet *ps) {
    bool had_error = false;
-   set<int_process *> procs;
-   transform(ps->begin(), ps->end(), inserter(procs, procs.end()), ProcToIntProc());
+   // Hold reference-counted Process wrappers, not raw int_process*: a process
+   // can exit/crash during bootstrap and be deleted inside the
+   // waitForAsyncEvent() below (post_create), which nulls the wrapper's
+   // llproc_.  Re-derive llproc() after each event-handling point and skip
+   // any process that was freed.
+   set<Process::ptr> procs(ps->begin(), ps->end());
 
    //Should be called with procpool lock held
    pthrd_printf("Calling plat_create for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); ) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); ) {
+      int_process *proc = (*i)->llproc();
+      if (!proc) { i = procs.erase(i); had_error = true; continue; }
       bool result = proc->plat_create();
       if (!result) {
          pthrd_printf("Could not create debuggee, %s\n", proc->executable.c_str());
@@ -106,8 +111,9 @@ bool int_process::create(int_processSet *ps) {
    }
 
    pthrd_printf("Creating initial threads for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); i++) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); i++) {
+      int_process *proc = (*i)->llproc();
+      if (!proc) continue;
 	  // Because yo, windows processes have threads when they're created...
 	  if (proc->threadPool()->empty()) {
         int_thread::createThread(proc, NULL_THR_ID, NULL_LWP, true, int_thread::as_created_attached);
@@ -122,8 +128,9 @@ bool int_process::create(int_processSet *ps) {
 
 
    pthrd_printf("Waiting for startup for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); ) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); ) {
+      int_process *proc = (*i)->llproc();
+      if (!proc) { i = procs.erase(i); continue; }
       string executable = proc->executable;
       Dyninst::PID pid = proc->pid;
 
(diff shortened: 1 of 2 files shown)


bbiiggppiigg:
tested on coirander.

proccontrol/src/process.C
```diff
@@ -87,13 +87,18 @@ void Process::version(int& major, int& minor, int& maintenance)
 
 bool int_process::create(int_processSet *ps) {
    bool had_error = false;
-   set<int_process *> procs;
-   transform(ps->begin(), ps->end(), inserter(procs, procs.end()), ProcToIntProc());
```
hainest, line 91: Since `ProcToIntProc` isn't used anymore, we can remove it.

proccontrol/src/process.C
```diff
@@ -87,13 +87,18 @@ void Process::version(int& major, int& minor, int& maintenance)
...
+   // Hold reference-counted Process wrappers, not raw int_process*: a process
+   // can exit/crash during bootstrap and be deleted inside the
+   // waitForAsyncEvent() below (post_create), which nulls the wrapper's
+   // llproc_.  Re-derive llproc() after each event-handling point and skip
+   // any process that was freed.
+   set<Process::ptr> procs(ps->begin(), ps->end());
```
hainest, line 95: This copy isn't needed anymore. We can directly modify `ps` because `int_processSet` is just a typedef for `set<Process::ptr>`.

proccontrol/src/process.C
```diff
@@ -87,13 +87,18 @@ void Process::version(int& major, int& minor, int& maintenance)
...
 
    //Should be called with procpool lock held
    pthrd_printf("Calling plat_create for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); ) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); ) {
```
hainest, line 99: While we're here, might as well tidy this up so we don't have to know that `int_processSet` is a typedef.

```suggestion
   for (auto i = procs.begin(); i != procs.end(); ) {
```

proccontrol/src/process.C
```diff
@@ -87,13 +87,18 @@ void Process::version(int& major, int& minor, int& maintenance)
 
 bool int_process::create(int_processSet *ps) {
    bool had_error = false;
-   set<int_process *> procs;
-   transform(ps->begin(), ps->end(), inserter(procs, procs.end()), ProcToIntProc());
+   // Hold reference-counted Process wrappers, not raw int_process*: a process
+   // can exit/crash during bootstrap and be deleted inside the
+   // waitForAsyncEvent() below (post_create), which nulls the wrapper's
+   // llproc_.  Re-derive llproc() after each event-handling point and skip
+   // any process that was freed.
```
hainest, line 94: > a process can exit/crash during bootstrap

Yes, but only after `int_process::plat_create()` has been called.

> and be deleted inside the `waitForAsyncEvent()` below (post_create), which nulls the wrapper's llproc_.

This was a bit confusing to me the first time I read it. `Process::llproc_` is not nulled in `~Process`, but is _indirectly_ nulled in `~int_process`: neither of which is directly called in `waitForAsyncEvent()`. I would reword this to something like

"which invalidates the contained low-level process (llproc_)"

>Re-derive llproc() after each event-handling point

What does this mean? `int_process::llproc()` just returns the member.

>and skip any process that was freed.

A minor nit, but I would reword this as "and ignore any invalid processes."

proccontrol/src/process.C
```diff
@@ -106,8 +111,9 @@ bool int_process::create(int_processSet *ps) {
    }
 
    pthrd_printf("Creating initial threads for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); i++) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); i++) {
+      int_process *proc = (*i)->llproc();
+      if (!proc) continue;
```
hainest, line 116: I'm not sure why this stuff is being done in a separate pass. If it's safe to use the value returned by `llproc()` five lines after then it should be safe to use 20 lines after. If that isn't true, then there is a *much* bigger problem around synchronizing the value returned by `llproc()`. Checking it here is either redundant or disguising asynchronous updates to the low-level process.

proccontrol/src/process.C
```diff
@@ -122,8 +128,9 @@ bool int_process::create(int_processSet *ps) {
 
 
    pthrd_printf("Waiting for startup for %d processes\n", (int) procs.size());
-   for (set<int_process *>::iterator i = procs.begin(); i != procs.end(); ) {
-      int_process *proc = *i;
+   for (set<Process::ptr>::iterator i = procs.begin(); i != procs.end(); ) {
+      int_process *proc = (*i)->llproc();
+      if (!proc) { i = procs.erase(i); continue; }
```
hainest, line 133: Here, the `ProcPool()` is no longer under a mutex, so it's critical we understand if the value of `llproc()` can change asynchronously. I think we should have a chat about this.

hainest (changes requested):
I've only reviewed the changes for `int_process::create` because I think we need to have a better understanding of how `llproc` can be asynchronously updated before we can reason about the efficacy of these changes.

We also need to add a check for the return value of `int_process::create` at the callsite in [`ProcessSet::createProcessSet`](https://github.com/dyninst/dyninst/blob/v13.0.0/proccontrol/src/procset.C#L795).

This whole business of creating processes is incredibly convoluted. It is entirely unclear to me why the `int_process` factories can't do this for us instead of forcing users to call down into `int_process` and then _still_ have to call `int_process::create`. Why does an instance of `int_process` even know that it is wrapped in a `Process` object? So many of these checks could go away if we were doing this while constructing a `Process` instead of doing this dance between `Process` and `int_process`. Just sillybuggers.

PR #2323, opened 2026-07-10
instructionAPI: decode ptwrite

### LINKED ISSUE #1523: Dyninst generates unnecessary spill code (pnnl/memgaze)
Reported by @nrtallent and @yasodha-sk.

From [memgaze](https://github.com/pnnl/memgaze/tree/master/xlib#dyninst):

>MemGaze needs the mapping of each new ptwrite (instrumentation) to the original memory access instruction. Not only does this provide source line information, but it is also needed for the memory-access class (e.g., load classes) associated with a given ptwrite instrumentation.
>
>Further, MemGaze's instrumentation has been designed to never require that registers be saved or restored, i.e., spill code. Consequently, DynInst should never insert spill code!
>
>There are two DynInst issues.
>
>First, DynInst has no interface for the desired mapping. Consequently, we currently apply a hack/patch to print it for MemGaze's consumption.
>
>Second, for an unknown reason, DynInst sometimes inserts unnecessary spill code. Further, the spill code is added after the mapping of new instrumentation to original code is generated!
>
>As an aside, it is important to note that the resolution of the mapping is the basic-block level rather than per-instruction. The basic-block mapping is typically sufficient to obtain an instruction level one as long as the order of instrumented instructions does not change. MemGaze does not change ordering.
>
>The two problems above have two results. First, it adds time overhead to instrumented code. Second, it means that DynInst's mapping of new instrumentation to original code is wrong.

[mg-spmm-mat-dyinst.tar.gz](https://github.com/dyninst/dyninst/files/12588100/mg-spmm-mat-dyinst.tar.gz)

bbiiggppiigg:
ptwrite (F3 0F AE /4, SDM Vol 2B) has never decoded: the register form was rejected outright and the memory form misdecoded as REP xsave with a bogus write set (RCX and memory written). The group 15 tables had no way to see the F3 prefix: sse_prefix[0xAE] was 0, so the prefix parser filed F3 as a plain REP prefix and getOpcodePrefix() returned 0.

Mark 0F AE as F3-discriminated and route group 15 /4 through t_sse with two new sseMap rows: the memory form selects xsave without a prefix and ptwrite with F3; the register form (mod == 11) is only defined with F3. ptwrite only reads its operand -- its sole output is a PTW packet in the Intel PT trace stream -- so it writes no register and no flag, which get_written_flags already reports correctly for entries without a flag table entry.

This matters beyond disassembly: MemGaze-style instrumenters insert raw ptwrite bytes as PatchAPI snippets, and any later analysis of the instrumented binary (including Dyninst re-instrumenting it) misparsed them. It is also the prerequisite for deciding instrumentation clobber sets by decoding generated code.

Adds a unit test covering ptwrite register/memory/32-bit forms (empty write set included), the group 15 neighbors (xsave, fxsave, stmxcsr, xrstor, fences, clflush), and rejection of the undefined encodings in the same slot.

### DIFF
--- a/common/src/arch-x86.C
+++ b/common/src/arch-x86.C
@@ -208,7 +208,8 @@ enum {
   SSEE0, SSEE1, SSEE2, SSEE3, SSEE4, SSEE5, SSEE6, SSEE7,
   SSEE8, SSEE9, SSEEA, SSEEB, SSEEC, SSEED, SSEEE, SSEEF,
   SSEF0, SSEF1, SSEF2, SSEF3, SSEF4, SSEF5, SSEF6, SSEF7,
-  SSEF8, SSEF9, SSEFA, SSEFB, SSEFC, SSEFD, SSEFE
+  SSEF8, SSEF9, SSEFA, SSEFB, SSEFC, SSEFD, SSEFE,
+  SSEAE04M, SSEAE04R
 };
 /** END_DYNINST_TABLE_DEF */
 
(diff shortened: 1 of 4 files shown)


tests/unit/instructionAPI/ptwrite-x86.cpp
```diff
@@ -0,0 +1,195 @@
...
+  bool writes_must_be_empty;       // no register and no flag may be written
+  bool reads_memory;
+  bool writes_memory;
+};
+
+bool run_one(const decode_test &t) {
+  di::InstructionDecoder dec(t.bytes.data(), t.bytes.size(), t.arch);
+  di::Instruction insn = dec.decode();
+
+  if (t.opcode == e_No_Entry) {
```
hainest: `Instruction::isValid` checks this, so it's redundant here.

hainest (changes requested):
Please move this into the integration tests with the x86 call instruction tests.

The test checks included here have been implemented in a standard form (e.g, `tests/opcode_tests.h`). Please use those, if possible.

hainest (approved):
LGTM. Thanks for refactoring the tests.

bbiiggppiigg:
@hainest Github still automatically closed issue-1523 even though I didn't use the keyword fix, and that issue is linked to PRs.
Or maybe I need to do sub issues?

PR #2324, opened 2026-07-10
Fix cmake format issue for AMDGPU regression tests

bbiiggppiigg:
This is required otherwise all future PRs' CI will die on the cmake-format check.


bbiiggppiigg:
@hainest @kupsch this one is needed since it will bug CI for all subsequent PRs.

kupsch (approved):

PR #2325, opened 2026-07-10
dyninstAPI: skip flags spill when instrumentation cannot clobber flags

### LINKED ISSUE #1523: Dyninst generates unnecessary spill code (pnnl/memgaze)

bbiiggppiigg:
Issue #1523 (reported by the MemGaze project, pnnl/memgaze): Dyninst wraps instrumentation in "spill code" even when the instrumentation is engineered to clobber nothing. The observed sequence around a bare `ptwrite` snippet is the baseTramp condition-flags save:

```
lea    -0x80(%rsp),%rsp     ; red-zone skip
push   %rax
lahf                        ; save flags into %rax
seto   %al
push   %rax
ptwrite %rdi                ; the actual instrumentation
pop    %rax
add    $0x7f,%al
sahf
pop    %rax
lea    0x80(%rsp),%rsp
```

~26 bytes of overhead per insertion point, paid at every point where the flags are live (e.g. any load scheduled between a `cmp` and its `jcc`). It also corrupts MemGaze's instrumentation-to-original mapping: the relocation tracker points at the first byte of the tramp — the spill `lea` — instead of the `ptwrite`.

Root cause: `EmitterAMD64::emitBTSaves` decides the flags save from liveness alone (`checkVolatileRegisters(live)`). GPR saves are pruned by the `definedRegs` regeneration pass in `baseTramp::generateCode`, but flags are outside that machinery entirely: no emitter ever calls `markRegDefined` for a flag, and an opaque PatchAPI snippet (raw bytes via `Buffer::copy`) marks nothing at all — so Dyninst must assume every snippet clobbers the flags.

## Fix

After the tramp body is generated, decode exactly the body's bytes with InstructionAPI and record every register it writes — **including the individual condition flags** — via the existing `markRegDefined` mechanism. Then:

- `shouldRegenBaseTramp` requests one more generation pass when flags were saved but the body never wrote a flag (the exact parallel of the existing unneeded-GPR-save check);
- `emitBTSaves` skips the flags save once optimization info shows the body leaves the flags alone.

The unbounded cases degrade conservatively, with the bound matched to what is actually known:

- a **call, syscall, or interrupt** in the body transfers control to code we do not scan, but the System V ABI bounds its effect: callee-saved registers come back intact, so the call-clobbered GPRs and all flags (which no ABI preserves) are marked;
- a **decode failure** means nothing is known about that instruction or the rest of the body, so every GPR and flag is marked.

Branch targets are not analyzed — AST-generated control flow stays inside the body.

### Latent soundness fix

Decoding the body also closes a real corruption hole: a **live GPR written only by an opaque snippet** had its save dropped by the regeneration pass — `definedRegs` never saw the write, so `shouldSaveReg` pruned the save and the snippet's write escaped into the caller. With the body decode, such registers stay saved. (Repro: a `mov $5,%rbx` snippet at a point where `%rbx` is live.)

## Scope and interactions

- **x86_64 only.** 32-bit mutatees (served by this codegen arch via `cap_32_64`) keep the conservative saves: their flag state is a single `IA32_FLAG_VIRTUAL_REGISTER` slot, so the per-flag numbering here does not apply (documented in the code).

- Composes with the conservative-save work for `-fipa-ra` (#2288/#2319): those gate on `definedRegs` too, so a provably clobber-free body saves nothing extra, and a clobbering body still triggers the conservative paths.

## Testing

- **Verified against the test case attached to #1523**: instrumenting the provided `spmm_mat` binary at the exact reported sites (`0x7800`, `0x780a` in `ptiOmpSparseMatrixMulMatrix_Reduce._omp_fn.2`, flags live from `cmp` to `jne`) with MemGaze's ptwrite encodings now produces bare ptwrites directly between the `cmp` and its `jne` — zero `lahf`/`seto` in the entire rewritten binary, instruction order preserved, and the instrumentation's first byte is the `ptwrite` itself (the mapping lands on the instrumentation, not on spill code).
- A flags-clobbering snippet (`cmp %rax,%rax`) at a flags-live point keeps the full save/restore.
- A live-GPR-clobbering snippet keeps the register saved (previously corrupted, see above); the rewritten binary runs correctly.
- External testsuite: 1534/1534 pass, 0 failures (create/attach/rewriter modes — broad coverage of call-containing snippets exercising the conservative paths). Full `ctest` ALL: 121/121. Both on x86_64.

Closes #1523


dyninstAPI/src/trampolines/baseTramp.C
```diff
@@ -47,6 +47,15 @@
 #include "patching/instPoint.h"
 #include "Point.h"
 
+#if defined(DYNINST_CODEGEN_ARCH_X86_64)
```
hainest: Please don't do this. Move this code into the arch-specific files.

hainest (changes requested):

bbiiggppiigg:
No failure on coriander.

bbiiggppiigg:

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,149 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+namespace {
+
+// The individual flag slots the x86-64 register space tracks, plus
+// REGNUM_EFLAGS, the whole-flags-register aggregate that liveness overrides
+// and decoded aggregate writes use (see registerSpace::checkVolatileRegisters
+// and saveVolatileRegisters, which handle the same aggregate-vs-individual
+// split). Listed explicitly: no assumption about enum ordering.
+const Dyninst::Register x86_64FlagSlots[] = {
+    REGNUM_OF, REGNUM_SF, REGNUM_ZF, REGNUM_AF, REGNUM_PF, REGNUM_CF,
+    REGNUM_TF, REGNUM_IF, REGNUM_DF, REGNUM_NT, REGNUM_RF, REGNUM_EFLAGS};
```
kupsch: The current code to save and restore flags only saves the flags in the lower 8-bits of the flags register `(cf, pf, af, zf, sf)` and the `of` flag. The other flags are unlikely to be used, but it is misleading to list them here and not handle them. If you want to support the other flags, then a different instruction sequence needs to be emitted using the `pushf` and `popf` instructions (shorter in bytes, longer in execution cost).

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,149 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+      if (r != REGNUM_IGNORED) gen.markRegDefined(r);
+    }
+  }
+}
+
+// False only when the generated body provably never writes a flag: every
+// flag slot (and the EFLAGS aggregate) is clear in definedRegs, which
+// baseTramp_x86::accumulateBodyClobbers filled from the generated code.
+// Conservative (true) until the body has been generated once.
+bool baseTramp_x86::mayClobberVolatileRegs() {
```
kupsch: name is misleading, what is `VolatileRegs`? This is only flags, perhaps `mayClobberFlags()`?

hainest: I swear I had seen a set of ABI registers referred to as "volatile", but I can't find it in any of the SystemV or ARM docs. I _think_ they refer to the [global](https://github.com/dyninst/dyninst/pull/1712/changes#diff-7d35ec48b80318dae1bdb33ca535c5857fb8da48e7456e4b972582586cc66de0R117) set. Clearly, that's not what is actually being done in this function.

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,149 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+  return false;
+}
+
+// The flags parallel of shouldRegenBaseTramp's unneeded-GPR
+// check. emitBTSaves decides the flags save before the body exists, so on the
+// first pass live flags are always saved (lahf/seto + rax spill). If the
+// generated body never wrote a flag (see accumulateBodyClobbers above),
+// report the save as unneeded so the tramp regenerates; with
+// validOptimizationInfo set, emitBTSaves then skips the save.
+bool baseTramp_x86::savedUnneededVolatileRegs(registerSpace *rs) {
```
kupsch: I don't understand what `Volatile` means in the function name. Perhaps it should just go away. (I see the terminology `VolatileRegisters` is used in the x86 registerSpace and elsewhere, but no need to carry this over here)

dyninstAPI/src/trampolines/baseTramp.C
```diff
@@ -204,6 +204,17 @@ bool baseTramp::shouldRegenBaseTramp(registerSpace *rs)
                          "App liveness?\n",  __FILE__, __LINE__, reg->number.getId());         
       }
    }
+   // Parallel of the unneeded-GPR check above for the volatile
+   // registers (condition flags). The save decision is made before the body
+   // exists, so on the first pass live flags are always saved. If the
+   // generated body never wrote a flag (see accumulateBodyClobbers),
+   // regenerate: with validOptimizationInfo set, the emitter skips the save.
+   if (savedUnneededVolatileRegs(rs)) {
+      regalloc_printf("[%s:%d] - baseTramp saved volatile registers the body "
```
kupsch: `volatile` should go away. What is different between `volatile register` and `register`.

bbiiggppiigg:
According to google search, it seems like volatile registers might mean caller saved registers, but I agree they should just go away.
https://www.sra.uni-hannover.de/Lehre/SS25/V_BSB/doc/x86-abi.html

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,149 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+      default:
+        break;
+    }
+    std::set<di::RegisterAST::Ptr> writes;
+    insn.getWriteSet(writes);
+    for (const auto &w : writes) {
+      Dyninst::Register r = convertRegID(w->getID());
+      if (r != REGNUM_IGNORED) gen.markRegDefined(r);
+    }
+  }
```
kupsch: The use of `continue` makes for non-obvious control flow (skip what comes after the switch`.  It would be clearer to rewrite as:
```suggestion
      case di::c_InterruptInsn:
        markCallClobbers(gen);
        break;
      default:
        {
          std::set<di::RegisterAST::Ptr> writes;
          insn.getWriteSet(writes);
          for (const auto &w : writes) {
            Dyninst::Register r = convertRegID(w->getID());
            if (r != REGNUM_IGNORED) gen.markRegDefined(r);
          }
        }
        break;
    }
  }
```

kupsch (commented):
Some cleanup, and an issue about some flags are checked, but not saved. Should the code be enhanced to handle these flags. It would be unlikely that they would be modified and require handling, but it could happen.

kupsch (approved):
I'm OK with this if @hainest agrees that we only need to be concerned the following flags: `dr::of, dr::sf, dr::zf, dr::af, dr::pf, dr::cf, dr::flags` (seems likely for generated code). It doesn't appear that Dyninst saves the other flags, so this is the same as current behavior.

bbiiggppiigg:
It seemse like df register is the only other register that can be modified.
Under the System V AMD64 ABI, the Direction Flag (DF) is a callee-saved register with a strict rule: it must be clear (0) upon both function entry and function return.
So unless we instrument between a live std cld window, our instrumentation not saving df should be fine.

If we do want to save df, we need to use pushfq popfq, which will boost the latency from 1 cycle to 3-20 cycles.

kupsch:
I agree.  The only normal privilege flag that is not save is df.  I also agree that instrumentation likely doesn't modify it, so it should be OK as is.  Since it is no worse than before I think you can merge this.

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,152 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+    Dyninst::Register r = convertRegID(mr);
+    if (r != REGNUM_IGNORED) gen.markRegDefined(r);
+  }
+}
+
+// The condition flags a call clobbers -- the same set the spill saves (see
+// x86_64FlagSlots): no ABI preserves them across a call.
+void markFlagClobbers(codeGen &gen) {
+  namespace dr = Dyninst::x86_64;
+  static const Dyninst::MachRegister flags[] = {
```
hainest: A `MachRegister` is really an `int32_t` (although it should probably be an `int64_t` to make the fields more manageable), so it's not terribly expensive to store them on the stack. It likely has less overhead than the thread-safe init check on each call when it's a function-local static.

```suggestion
  const Dyninst::MachRegister flags[] = {
```

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,152 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+      dr::of, dr::sf, dr::zf, dr::af, dr::pf, dr::cf, dr::flags};
+  markClobbers(gen, flags);
+}
+
+// The System V AMD64 call-clobbered GPRs: what a standard-ABI callee may
+// leave modified.
+void markCallClobbers(codeGen &gen) {
+  namespace dr = Dyninst::x86_64;
+  static const Dyninst::MachRegister volatiles[] = {
```
hainest: Same as above.

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,152 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+      dr::r8,  dr::r9,  dr::r10, dr::r11};
+  markFlagClobbers(gen);
+  markClobbers(gen, volatiles);
+}
+
+// Undecodable code: nothing is bounded, mark every GPR and flag. (RSP is
+// included for completeness; the save logic never saves it.)
+void markAllClobbers(codeGen &gen) {
+  namespace dr = Dyninst::x86_64;
+  static const Dyninst::MachRegister gprs[] = {
```
hainest: Same as above.

dyninstAPI/src/trampolines/baseTramp-x86.C
```diff
@@ -9,3 +19,152 @@ bool baseTramp_x86::generateSaves(codeGen &gen, registerSpace *) {
...
+  size_t consumed = 0;
+  while (consumed < len) {
+    di::Instruction insn = dec.decode();
+    if (!insn.isValid() || insn.size() == 0) {
+      markAllClobbers(gen);
+      return;
+    }
+    consumed += insn.size();
+    switch (insn.getCategory()) {
+      case di::c_CallInsn:
```
hainest: I'd prefer to use `Instruction::isCall()` and friends, but this is probably fine for now.

dyninstAPI/src/trampolines/baseTramp.h
```diff
@@ -127,7 +128,24 @@ class baseTramp {
...
     virtual bool generateSaves(codeGen &, registerSpace *) { return false; }
     virtual bool generateRestores(codeGen &, registerSpace *) { return false; }
-    
+
+    // Record every register (including condition flags) that the
+    // just-generated instrumentation body [bodyStart, gen.getIndex())
+    // writes, through codeGen::markRegDefined, so that unneeded saves can
+    // be dropped when the tramp is regenerated.
+    // Default: record nothing; conservative saves are kept.
+    virtual void accumulateBodyClobbers(codeGen &, codeBufIndex_t) {}
```
hainest: It looks like these could be `const` member functions.

hainest (changes requested):

hainest (approved):
No regresions on 17 machines.

@kupsch Do you want to go ahead and merge since Hsuan-Heng is on vacation?

PR #2326, opened 2026-07-12
Improve AMDGPU Indirect Control Flow Parsing

### LINKED ISSUE #1470: Dyninst doesn't continue parsing AMDGPU code after a procedure call

bbiiggppiigg:
## Summary

On AMDGPU (gfx908/gfx90a/gfx940), ParseAPI abandons a function at the first
procedure call whose target it cannot resolve statically. Everything after such
a call — blocks, edges, instrumentation points, liveness/footprint — is missing
from the CFG. This is the concrete form of #1470 ("Dyninst doesn't continue
parsing AMDGPU code after a procedure call"); the precise trigger is not "a
call" but an **unresolved-target (indirect) call**.

## Background: AMDGPU call/return encoding

AMDGPU has no call instruction with a wide immediate target. The compiler lowers
**every** call — direct or indirect — through a register pair:

- **Call**: `s_swappc_b64 s[30:31], <reg-pair>` — saves the return PC into the
  destination pair and jumps. It returns, so the next instruction is a valid
  return point (a real fall-through).
- **Direct call**: the target is materialized first with the PC-relative idiom
  `s_getpc_b64` / `s_add_u32` / `s_addc_u32`, then `s_swappc_b64` through that
  pair. Statically knowable, but only via data-flow, not from the call
  instruction alone.
- **Indirect call**: the target is loaded (e.g. from a function-pointer table)
  into a pair, then `s_swappc_b64`. Not statically knowable.
- **Return / tail call / jump table**: `s_setpc_b64` — a terminator with no
  fall-through.

## Root cause

Three independent defects combine:

1. **`s_swappc_b64` was classified as a branch.** An indirect `s_swappc` has a
   register target that can't be resolved, so it fell into the indirect-jump
   path — and that path **couples the fall-through edge to target resolution**:
   when the target can't be resolved, the block is abandoned with no
   fall-through, so all code after the call is left unparsed. A returning call
   must always get a call fall-through regardless of whether its target is
   known.

2. **Direct call targets weren't recovered on the call path.** Because there's
   no wide-immediate call, direct calls are also register-targeted, so
   `getCFT()` fails for them too; their target is only recoverable by
   backward-slicing the `getpc`/`add`/`addc` chain. That resolution existed only
   inside the jump-table machinery reachable from the (wrong) branch path.

3. **A register-modeling bug blocked `vcc`-materialized targets.** When the
   target pair is `vcc` (rather than an SGPR pair), the ROSE->Dyninst register
   reverse-conversion reconstructed MISC-class registers from `src_scc` (a
   `BITS_1` register), producing the wrong size encoding and an invalid register
   (`INVALID_REG`). The slice couldn't fold the value, so `vcc`-based calls
   never resolved.

## Impact

- **Correctness-level, not cosmetic.** Any code dominated by an unresolved call
  disappears from ParseAPI's view of the function.
- It bites function-pointer / virtual dispatch, `-O0`/`-fno-inline`, and any code where a callee isn't
  statically resolvable — notably the device-function-pointer pattern GPU binary
  instrumentation often targets.
- Reproduced on gfx908 with a function-pointer test kernel: ParseAPI stopped
  dead at the indirect `s_swappc`, dropping the trailing `global_store` +
  `s_endpgm`; and on the gfx940 regression binary, whole function tails after
  indirect calls were absent.

## Fix

Split across two changes:

**1. Reorganize AMDGPU indirect control flow analysis.** Reclassify
`s_swappc_b64`/`s_call_b64` as calls so an unresolved target yields a
`CALL`-to-sink edge **plus an unconditional call fall-through**, letting parsing
continue; `s_setpc_b64` stays a branch (return / tail call / jump table) and
never gets a fall-through. Recover direct/register-materialized targets with a
dedicated `CallTargetPred` backward slice over the `getpc`/`add`/`addc`/`mov`
affine chain, invoked from the call path when `getCFT()` fails and cached into
the CFT so `isDynamicCall()` and callee parsing stay consistent. Give
`s_call_b64` real decoder successors so its immediate target resolves directly.

**2. Fix MISC-class register reverse-conversion.** Reconstruct MISC registers
(`vcc`, `exec`, `flat_scratch`, `m0`) from a size-appropriate template
(`src_scc` for 1-bit flags, `vcc_lo` for 32-bit) preserving `arch|category|size`,
so `vcc`-materialized call targets resolve.

## Known limitation

Resolving a target across intervening calls currently treats `s_swappc` as
non-clobbering (the value is assumed to survive the call). That's correct when
the holding register is callee-saved, which is the case here, but it's an
unmodeled assumption — Dyninst has no AMDGPU calling-convention register model
(`ABI::initialize64` is a stub for these arches). Making it fully sound
(correctly refusing to resolve a target held in a caller-saved register)
requires populating the AMDGPU callee-saved/caller-saved sets and teaching the
call-clobber check that `s_swappc`/`s_call` are calls — a follow-up. Genuinely
indirect calls (target from memory / a live-in argument) correctly remain
unresolved (sink call edge + fall-through).

### DIFF
--- a/dataflowAPI/rose/semantics/SymEvalSemantics.C
+++ b/dataflowAPI/rose/semantics/SymEvalSemantics.C
@@ -124,8 +124,9 @@ Dyninst::Absloc SymEvalSemantics::RegisterStateAST_amdgpu_gfx908::convert(const 
         break;
     }
     case amdgpu_regclass_misc : {
-        Dyninst::MachRegister base = Dyninst::amdgpu_gfx908::src_scc;
-        mreg  = Dyninst::MachRegister(base.val() + minor) ;
+        Dyninst::MachRegister base = (size == 1) ? Dyninst::amdgpu_gfx908::src_scc
+                                                 : Dyninst::amdgpu_gfx908::vcc_lo;
+        mreg = Dyninst::MachRegister((base.val() & 0xffffff00) | minor);
         found = true;
         break;
     }
@@ -164,8 +165,9 @@ Dyninst::Absloc SymEvalSemantics::RegisterStateAST_amdgpu_gfx90a::convert(const 
         break;
     }
     case amdgpu_regclass_misc : {
-        Dyninst::MachRegister base = Dyninst::amdgpu_gfx90a::src_scc;
-        mreg  = Dyninst::MachRegister(base.val() + minor) ;
+        Dyninst::MachRegister base = (size == 1) ? Dyninst::amdgpu_gfx90a::src_scc
+                                                 : Dyninst::amdgpu_gfx90a::vcc_lo;
+        mreg = Dyninst::MachRegister((base.val() & 0xffffff00) | minor);
         found = true;
         break;
     }
@@ -197,8 +199,9 @@ Dyninst::Absloc SymEvalSemantics::RegisterStateAST_amdgpu_gfx940::convert(const 
         break;
     }
     case amdgpu_regclass_misc : {
-        Dyninst::MachRegister base = Dyninst::amdgpu_gfx940::src_scc;
-        mreg  = Dyninst::MachRegister(base.val() + minor) ;
+        Dyninst::MachRegister base = (size == 1) ? Dyninst::amdgpu_gfx940::src_scc
+                                                 : Dyninst::amdgpu_gfx940::vcc_lo;
+        mreg = Dyninst::MachRegister((base.val() & 0xffffff00) | minor);
         found = true;
         break;
     }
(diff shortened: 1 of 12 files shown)


bbiiggppiigg:
Cannot merge/Turn to PR before https://github.com/dyninst/test_binaries/pull/4 as CI will break.

bbiiggppiigg:
@kupsch @hainest 

This is correctness and may affect John's student so is probably of higher priority.

bbiiggppiigg:
Tested on coriander.

kupsch (approved):

PR #2331, opened 2026-07-18
instructionAPI: fix x86 decode-table errors against the Intel SDM and AMD APM

bbiiggppiigg:
## Summary

The hand-written x86 decode tables in `common/src/arch-x86.C` were
audited against the *Intel 64 and IA-32 Architectures SDM* (June 2025)
and the *AMD64 Architecture Programmer's Manual*, first by walking the
opcode-extension groups of SDM Vol 2D, Table A-6, and then by a
differential sweep: ~45,000 generated encodings (every one- and
two-byte opcode × none/66/F3/F2 prefixes × ModRM.reg values ×
register/memory forms, the `0F 38`/`0F 3A` maps, and the VEX space
maps 1–3 × pp × L × W) decoded by both Dyninst and GNU binutils
objdump, with every disagreement verified against the manuals before
any table was changed.

This series fixes the confirmed decoding, operand, read/write-set, and
flag errors that turned up. Several affect instructions that appear in
essentially every modern binary (`cmovg`, `tzcnt`/`lzcnt`, FMA3
contractions, `rdtscp`, the `blendv*` write sets); two are
length-desync bugs that corrupt every subsequent instruction in a
linear sweep.

## Commits

**1. Decode every group 15 (0F AE) slot by its mandatory prefix**

Group 15 selects among sixteen instructions by ModRM.reg, ModRM.mod,
and the mandatory prefix (SDM Vol 2D, Table A-6); the tables ignored
the prefix outside the `/4` slot fixed for ptwrite in #2323.
- `xsaveopt` (NP /6 mem) and `clrssbsy` (F3 /6 mem) decoded as `clwb`
  (which actually requires the 66 prefix); `clflushopt` (66 /7)
  decoded as `clflush`.
- `rdfsbase`/`rdgsbase`/`wrfsbase`/`wrgsbase` (F3 /0../3 reg) did not
  decode at all; they are now decoded and rejected outside 64-bit mode
  (#UD per their SDM pages).
- `incsspd`/`incsspq` decoded as `lfence`; `umonitor`, `tpause`, and
  `umwait` decoded as `mfence` — erasing their register reads and flag
  writes from every def-use chain.
- Reserved prefixed encodings now fail to decode instead of aliasing
  the unprefixed instruction.
- The xsave family gains its implicit EDX:EAX feature-bitmap read
  (SDM Vol 1, §13.7).

**2. Fix group 9 (0F C7) entries; decode 0F FB as psubq**

- `xsavec` carried its operand semantics (`s1W`) in the legacy-type
  field, where that bit value means `IS_CALL`: it was typed as a call
  and reported no memory write.
- `xrstors` was described with an XMM operand and empty semantics
  (empty read set); `xsaves` was missing.
- `0F FB` carried the `psubd` identity (the row was marked
  `FIXME: Same????`); it is PSUBQ per SDM Vol 2B.
- `rdrand` now reports its flag writes (CF result, others cleared).

**3. Identity and length fixes from the differential sweep**

- `0F 4F` was a copy of the `0F 4D` entry: every **cmovg** decoded as
  cmovge.
- F3 `0F BC`/`0F BD` were not prefix-discriminated: **tzcnt/lzcnt**
  decoded as REP bsf/bsr (the lzcnt row existed but was unreachable).
- `C6 F8` (**xabort**) decoded as xbegin; `C7 F8` (**xbegin**)
  inherited mov's operands and lost its branch target. Group 11 is now
  split per opcode; xbegin produces a RIP-relative control-flow target.
- `0F 18 /4../7` memory forms are reserved NOPs per Table A-6 but were
  rejected, desyncing linear-sweep parsing.
- SSE4A (AMD APM Vol 4): `extrq`'s immediate form never decoded (wrong
  mod subrow and register field), and the `0F 79` register-pair forms
  consumed **two phantom immediate bytes** — a 4-byte instruction
  decoded as 6 bytes.
- VEX space: `vmovd`/`vmovq` ignored VEX.W in both directions; VEX3
  `vpand` decoded as the EVEX-only `vpandd`; the FMA3 **vfnmsub**
  rows for `0F 38 9E/AE/BE` pointed one row off (packed 213/231 forms
  decoded as scalar 132/213 forms), with two EVEX columns similarly
  shifted; `vaesdeclast` carried the legacy ID and **no destination
  write**; `VEX.66.0F3A 42` decoded as the EVEX-only `vdbpsadbw`
  instead of `vmpsadbw`.
- FMA4 (AMD APM Vol 6) was half present, with `69` mislabeled and
  legacy/wrongly-prefixed encodings decoding via an out-of-bounds
  `fma4Map[..][-1]` index; it now requires VEX with pp = 66.
- `writesOperand()` had no case for the four-operand `s1W2R3R4R`
  semantic, so **blendvps/blendvpd/pblendvb, all FMA4, and XOP forms
  reported an empty write set**.

**4. Decode group 7 (0F 01) mod=11 encodings**

Group 7's mod=11 encodings are selected by reg and r/m together; they
fell through to the memory forms, so **rdtscp** decoded as `invlpg`
with an empty write set (it writes EDX, EAX, *and* ECX — added a
three-write operand semantic), `swapgs` as `invlpg`, and
`monitor`/`mwait`/`xgetbv`/`xsetbv`/`clac`/`stac`/`rdpkru`/`wrpkru`/
VMX forms as lgdt/sgdt. Reserved slots and forms without entry IDs
(SVM, SGX, CET, serialize, pconfig) now decode as invalid rather than
as a wrong instruction.

**5. Discriminate group 9 (0F C7) on mod and mandatory prefix**

Group 9 keyed only on ModRM.reg, so `vmptrld` (NP /6 mem) decoded as
**rdrand** with a bogus memory-written operand, `vmxon` (F3 /6) as REP
rdrand with string-op implicit registers, and `vmclear`/`vmptrst`
misdecoded or failed; `rdseed` and `rdpid` (reg /7) did not decode at
all. Group 9 now dispatches like groups 15 and 7; the VMX pointer
instructions report success through the status flags (SDM Vol 3C) and
rdseed writes CF like rdrand. The REX.W `cmpxchg16b` variant still
reports as cmpxchg8b (documented table limitation).

**6. Decode the remaining prefixed and grouped system instructions**

- **wbnoinvd** (F3 0F 09; decoded as REP wbinvd), plus mnemonic mappings
  for the already-decoding SSE4A movntss/movntsd, which formatted as
  `[INVALID]`.
- **cldemote** (NP 0F 1C /0 mem) and **rdsspd/rdsspq** (F3 0F 1E /1),
  which sat in the reserved-NOP run and decoded as plain NOPs — hiding
  rdssp's register write; the rest of the 0F 1C/1E space keeps decoding
  as the SDM's reserved NOPs, and endbr32/64 recognition follows the
  prefix reclassification.
- **Group 7 (0F 01) completed**: serialize (the only new `entryID` in
  the whole series), CET setssbsy/saveprevssp/rstorssp (F3-selected,
  rstorssp reading and rewriting its m64 token), TSX xend/xtest (ZF via
  flags), vmfunc (EAX-selected), AMD SVM vmrun/vmmcall/vmload/vmsave/
  stgi/clgi/skinit/invlpga with their rAX/ECX implicit reads (APM
  Vol 3), and SGX enclv/encls/enclu plus pconfig as leaf-style entries
  (SDM Vol 3D).
- **GFNI** (SDM Vol 2A): gf2p8mulb / gf2p8affineqb / gf2p8affineinvqb in
  legacy, VEX, and EVEX forms, W-gated per the opcode columns (mulb=W0,
  affine=W1); wrong-W and unprefixed forms decode as invalid. Verified
  byte-for-byte against objdump.

**7. Select REX.W-distinguished instructions**

REX.W denotes a different instruction or mnemonic for some legacy
encodings, but the tables held one entry per slot, so cmpxchg16b
reported as cmpxchg8b and the 64-bit xsave layouts (xsave64 … xrstors64)
and incsspq/rdsspq reported their 32-bit base IDs. A `t_rexw`
continuation table selects on REX.W (mirroring `t_vexw` on VEX.W);
cmpxchg16b joins the LOCK whitelist and reports its ZF write. No new
entry IDs — the enum already had these.

**8. Enforce memory-only and register-only ModRM forms**

Instructions that constrain ModRM.rm to one kind (SDM Vol 2 operand
tables) were described as either-register-or-memory, so undefined
encodings decoded as the wrong operand kind and a few reported the
wrong register bank. Memory-only stores (movnti/movntq/movntdq/
movntps/movntpd) now reject a register ModRM; register-only sources
(pmovmskb, movmskps/pd, maskmovq/dqu, movq2dq/dq2q, extrq/insertq)
reject a memory ModRM, with corrected register banks (e.g. movntq's
source is MMX, not XMM). MOV to/from CR/DR (0F 20-23), which ignores
ModRM.mod, is deliberately left lenient.

**9. Build: raise the frame-size limit for ia32_entry.C**

The entryNames_IAPI mnemonic map is a single map_list_of static
initializer; the names added while decoding more instructions pushed
its one-time init frame just past -Wframe-larger-than=20480, failing
warnings-as-errors builds. Since the frame belongs to a global
initializer that runs once before main, the heuristic is a false
positive here (the same rationale already excludes the power decoder
table, the AMDGPU operand table, and MachSyscall.C), so the file gets a
per-file override with headroom, following the existing pattern.

**10. Tests**

One consolidated `decoding_tests` covering every fix: each case pins
the mnemonic, instruction length, exact register read/write sets
(including implicit operands and flags), and memory access direction;
over forty reserved or mode-restricted encodings must fail to decode.
Every expectation cites the SDM/APM section it was verified against.
The existing ptwrite test's xsave expectations gain the implicit
EDX:EAX reads. 

## Verification

- Every changed encoding cross-checked against objdump; the remaining
  sweep disagreements are objdump naming conventions (AT&T suffixes,
  imm-specialized pseudo-ops) or documented gaps, not misdecodes.
- Full build clean; all integration and unit tests pass at every
  commit in the series.

## Known gaps (documented, out of scope here)

MPX is not decoded (matching its removal from the SDM); AVX-512 mask-op
(kmov/kand/kshift/…) width selection is wrong — the VEX3 column is
hardcoded to the W1 mnemonic, and those rows also carry operand-template
errors (spurious third operands, wrong kmov-store direction) that want
fixing together. A follow-up PR will address the mask family.


bbiiggppiigg:
Tested on coriander.

hainest:
Are there known bugs that this solves? I really don't want to spend time on the x86 decoder unless we need to.

bbiiggppiigg:
I would rather that we fix the semantics once instead of fixing one semantic per bug report, if we still cannot have a fixed date on when the capstone integration could happen.

The test cases under 
[decoding.cpp](https://github.com/dyninst/dyninst/compare/master...bbbiiggppiigg/fix-dyninst-x86-semantics#diff-51929fcc55b8466ec17826d023fa6efe915def2d1150b4d6a1c5aac31abce34b)
could serve as a report of all the latent bugs that gets fixed in this PR.

At least the instructions that decodes to incorrect length should be fixed.

These results are validated through scripts that compare results of dyninst decode vs object as mentioned in the description. 
I can go through the manuals and reverify each of these changes if that is what you wanted, unless you guys decide to reject this PR as a whole.

hainest:
> I would rather that we fix the semantics once instead of fixing one semantic per bug report, if we still cannot have a fixed date on when the capstone integration could happen.

Instruction semantics are a huge update included in #1646 because they are created procedurally based on the information provided by Capstone rather than piecemeal based on which instructions we happen to test in the current decoder. If you look at the unit tests in 1646, there are many that don't pass with the Dyninst decoder because they have incorrect semantics (and some fail due to bad decoding by the Dyninst decoder).

> The test cases under [decoding.cpp](https://github.com/dyninst/dyninst/compare/master...bbbiiggppiigg/fix-dyninst-x86-semantics#diff-51929fcc55b8466ec17826d023fa6efe915def2d1150b4d6a1c5aac31abce34b) could serve as a report of all the latent bugs that gets fixed in this PR.

I very much want to keep these tests no matter what decoder we're using.

> At least the instructions that decodes to incorrect length should be fixed.

Agreed.

> These results are validated through scripts that compare results of dyninst decode vs object as mentioned in the description.

Could you share those? I'm curious how you're generating the bytes for each instruction (e.g., the different addressing modes).

> I can go through the manuals and reverify each of these changes if that is what you wanted, unless you guys decide to reject this PR as a whole.

I'm not principally opposed to these changes. I would rather spend what precious little time we have to figure out how to throw away the current decoder rather than patching it. I'm kind of wearing thin on Capstone for x86. I just don't see that they will get a release anytime soon. I've talked with Jim about using Zydis since it 1) is what Rice is using, 2) is what Capstone will use in the future for their x86 backend, 3) has more complete instruction coverage than Capstone, and 4) has a rewritter interface that we could use in place of the current spaghetti in the Emitter classes (for x86 flavors only). The cost is another dependency (and then another for Capstone for the other architectures).

bbiiggppiigg:
> Instruction semantics are a huge update included in #1646 because they are created procedurally based on the information provided by Capstone rather than piecemeal based on which instructions we happen to test in the current decoder. If you look at the unit tests in 1646, there are many that don't pass with the Dyninst decoder because they have incorrect semantics (and some fail due to bad decoding by the Dyninst decoder).

I can maybe check this PR againsts the tests in #1646 as well.

> Could you share those? I'm curious how you're generating the bytes for each instruction (e.g., the different addressing modes).
> 

https://github.com/dyninst/dyninst/tree/bbiiggppiigg/x86-tool/tests/integration/InstructionAPI/decoder/x86/tools

hainest:
> > Instruction semantics are a huge update included in #1646 because they are created procedurally based on the information provided by Capstone rather than piecemeal based on which instructions we happen to test in the current decoder. If you look at the unit tests in 1646, there are many that don't pass with the Dyninst decoder because they have incorrect semantics (and some fail due to bad decoding by the Dyninst decoder).
> 
> I can maybe check this PR againsts the tests in #1646 as well.

I just updated the branch, so it should all be there now. You will need the fix from this PR for the correct registers read by ptwrite.

bbiiggppiigg:
> I just updated the branch, so it should all be there now. You will need the fix from this PR for the correct registers read by ptwrite.

It is a bit unclear to me what should I test against. I imported the unit tests in #1646 (before you force-pushed) and find that there are two wrong regiser read/write set (missing ecx write for loop, missing df flag writes for some instructions, not introduced by this PR).

There is another failing related to isConditional but that is probably not in the scope of this PR.

I'm thinking that I can try to compare the read/write set of the generated set of instructions against Zydis, if you think that is a good idea.

bbiiggppiigg:
will be fixed by zydis

PR #2333, opened 2026-07-20
symtabAPI: preserve .gnu.version_d entries not recreated from symbol versions

bbiiggppiigg:
emitElf regenerates .gnu.version_d for every rewritten object that has
dynamic symbols, and createSymbolVersions rebuilds the defined-version
table purely from per-symbol version strings. Any verdef referenced by
no symbol is therefore silently dropped. The guaranteed casualty is the
BASE definition -- the verdef at vd_ndx VER_NDX_GLOBAL (1), named after
the object itself -- because fully-versioned libraries reference it from
no symbol at all: a rewritten glibc libc.so.6 goes from 39 to 38 verdef
entries and eu-elflint reports 'section .gnu.version_d: no BASE
definition'. In addition, vd_flags has been hardcoded to 0 since the
reconstruction was written, so even a BASE entry that survives (when
some symbol does parse with the BASE version) is emitted without
VER_FLG_BASE and with a wrong, sequentially allocated vd_ndx.

The glibc runtime loader happens to tolerate the missing entry (it skips
VER_FLG_BASE entries when building l_versions and special-cases version
indices 0 and 1), but the emitted library violates the ELF gABI's
versioning conventions, fails eu-elflint, and misleads link-time
consumers that read the defined-version table.

Fix, mirroring the .gnu.version_r preservation introduced with the RELR
work (https://github.com/dyninst/dyninst/pull/2235) and using the same parse-time maps: after rebuilding
verneed/verdef entries from symbols, walk the object's original version
mapping and re-add every verdef-side entry (those with no provider-file
mapping) that symbols did not recreate, keeping VER_NDX_GLOBAL for the
BASE definition and preserving aux name chains verbatim. A
symbol-supplied BASE version (versym index 1 parses to the BASE name) is
mapped to VER_NDX_GLOBAL instead of a sequential index, and the emission
loop now sets VER_FLG_BASE on the VER_NDX_GLOBAL entry.

Verified in a glibc-2.36 container via the testsuite's test_reloc (the
one test that rewrites libc.so.6 itself): without this patch the
rewritten libc has 38 verdef entries and elflint's BASE complaint; with
it, 39 entries with 'Flags: BASE Index: 1 Name: libc.so.6', elflint
clean, and the rewriter suite (-rewriter -allcomp, gcc+g++) passes.

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -195,6 +195,19 @@ std::string phdrTypeStr(Elf64_Word phdr_type) {
 
 }
 
+// The name of the object's BASE version definition: the verdef at
+// VER_NDX_GLOBAL, conventionally the object's SONAME. Falls back to
+// DT_SONAME when the object carried no verdefs (e.g. versions are being
+// introduced into a previously unversioned object).
+static const char *baseVersionName(ObjectELF *object) {
+    if (!object) return NULL;
+    const auto &vm = object->getVersionMapping();
+    auto it = vm.find(VER_NDX_GLOBAL);
+    if (it != vm.end() && !it->second.empty())
+        return it->second[0].c_str();
+    return object->getSoname();
+}
+
 template<class ElfTypes>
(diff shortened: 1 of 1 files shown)


bbiiggppiigg:
should have been fixed by #2381

PR #2334, opened 2026-07-21
Fix capstone (follow #1646)

bbiiggppiigg:
# Fix and finish the Capstone x86 decoder integration

Branch: `bbiiggppiigg/dyninst@fix-capstone`

## Summary

This branch continues the "Use Capstone for x86 instruction decoding" work
(dyninst/dyninst#1646, by @hainest), which replaces Dyninst's hand-written x86
`InstructionDecoder` with Capstone-backed decoding. On top of that work it fixes
the correctness regressions and consistency gaps that the decoder swap
introduced, makes the branch buildable/testable against a current Capstone, and
clears several of the PR's remaining checklist items.

It is rebased on `dyninst/master` and builds cleanly. All x86 InstructionAPI
decoder tests pass.

## Relationship to #1646

The 43 commits of #1646 are included unchanged. This branch adds **6 commits**
on top of them:

| Commit | Change |
| --- | --- |
| `Bump Capstone requirement to 6.0.0-Alpha9` | The riscv decoder now uses Capstone enum names introduced in Alpha7 (`RISCV_INS_*_AQRL`, renamed from `*_AQ_RL`), so the pinned Alpha5 no longer compiles. Bumps both `docker/dependencies.versions` and the `DyninstCapstone` `_min_version` (the `dependency-version` CI check requires the two to match). |
| `instructionAPI: fix getReadSet documentation example` | The `getReadSet` example used `jc [eax]`, not a valid instruction form. Corrected to `jc 0x1234` reading `{ip, flags, cf}`, with `jmp [eax]` illustrating address-calculation reads. |
| `instructionAPI: categorize prefetches with no Capstone group` | The prefetch check was nested inside the loop over Capstone instruction groups, so prefetches Capstone assigns no group (`prefetchw`, `prefetchwt1`) were left uncategorized: `isPrefetch()` returned false and `readsMemory()` wrongly returned true. Moves the mnemonic check out of the loop and adds the symmetric `isPrefetch()` guard to `writesMemory()`. Includes a new x86 prefetch regression test. |
| `Fix entryID consistency for the Capstone x86 decoder` | The decoder emits `e_ret`/`e_retf` and (on 64-bit) `e_popfq`, but consumers still keyed on the old pseudo-mnemonics `e_ret_near`/`e_ret_far` and omitted `e_popfq`, so x86 return detection and 64-bit `popf` handling were silently broken in dataflowAPI, parseAPI, and dyninstAPI. Renames `e_ret_near`→`e_ret`, `e_ret_far`→`e_retf` in all live consumers and adds the missing `e_popfq` cases. |
| `Remove obsolete x86 test registers (tr0-tr7)` | The x86 test registers TR0–TR7 left the ISA after the 486 and were only produced by the old decoder. Removes their definitions and the now-dead references in the DWARF register maps and `RegisterMap`. The task register `tr` is retained. |
| `parseAPI: bound all SETcc results to [0,1] in BoundFactCalculator` | The transfer function assumed the `SETxx` entry ids were contiguous and matched only `setb`/`setbe`/`sete`, missing the other 13 condition codes. Enumerates all 16 `SETcc` ids explicitly and bounds them to `[0,1]`. |

## Bugs fixed (introduced by the decoder swap)

- **Prefetch mis-categorization.** `prefetchw`/`prefetchwt1` were not recognized
  as prefetches and were reported as reading (and, for `prefetchwt1`, writing)
  memory. Fixed and covered by a new regression test.
- **x86 return detection broken.** The Capstone decoder produces `e_ret`/`e_retf`
  while every downstream consumer (`AbslocInterface`, `stackanalysis`,
  `RoseInsnFactory`, `convertOpcodes`, `StackAccess`, `frameChecker`,
  `IdiomModelDesc`) still matched `e_ret_near`/`e_ret_far`. Returns were silently
  unrecognized on x86.
- **64-bit `popfq` unhandled.** `convertOpcodes`, `stackanalysis`, and `IA_x86`
  handled `e_popf`/`e_popfd` but not `e_popfq`, which is what 64-bit `popf`
  decodes to.
- **Incomplete `SETcc` bounds** in jump-table bound analysis (see above).

## Build

Requires **Capstone ≥ 6.0.0-Alpha9**, built as a shared library with a CMake
config package (the `capstone::capstone_shared` target). Alpha5 (the previous
pin) does not compile this branch.

```sh
# Capstone
git clone --branch 6.0.0-Alpha9 https://github.com/capstone-engine/capstone
cmake -S capstone -B capstone/build \
  -DCAPSTONE_BUILD_SHARED_LIBS=ON -DCAPSTONE_BUILD_STATIC_LIBS=OFF \
  -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX=<cap-prefix>
cmake --build capstone/build --parallel && cmake --install capstone/build

# Dyninst
cmake -S dyninst -B dyninst/build -G Ninja \
  -DDYNINST_ENABLE_TESTS=INTEGRATION \
  -DCMAKE_PREFIX_PATH=<cap-prefix>
cmake --build dyninst/build --parallel
```

## Test status

Full build succeeds with no errors. The x86 InstructionAPI decoder suite passes
100% (`instructionAPI_x86_*`, `sysctl`, `x87`, `syscall`, the new `prefetch`
test, plus the x86 register/machregister tests).

Known pre-existing failures, unrelated to these changes: the `ptwrite` and
`riscv64` InstructionAPI tests fail because their expected disassembly strings
and register sets were pinned to an earlier Capstone and differ under Alpha9
(e.g. riscv `beq t1, 0xc` vs `beq t1, zero, 0xc`). These need their expectations
regenerated for the required Capstone version; they are not affected by this
branch's changes.

### Verification note

InstructionAPI-level behavior is covered by the in-tree tests. The
dataflowAPI/parseAPI/dyninstAPI *functional* paths touched by the entryID and
register changes (stack-analysis return handling, idiom matching, ROSE
conversion, frame detection) are verified only by compilation and by
no-regression in the available tests; they are not exercised by functional tests
in this configuration and should be validated against the external Dyninst test
suite before merge.

## Deferred / follow-up work

These items from the #1646 checklist remain and are best done as separate,
independently-mergeable PRs:

- **Remove the old decode tables** (`common/src/arch-x86.C`, `ia32_entry.C`,
  `arch-x86.h`), still compiled and used by `convertOpcodes`,
  `BoundFactCalculator`, and instruction-length code.
- **Remove the now-dead enum pseudo-mnemonics** `e_ret_near`, `e_ret_far`,
  `e_int80`, `e_fp_generic` (coupled to the table removal; needs an
  entryID-serialization audit).
- **Replace direct `Operation::mnemonic` writes with `updateMnemonic`** (the
  remaining uses are in the PowerPC decoder), then make `updateMnemonic`
  private and remove the non-const `Instruction::getOperation()` (117 callers).
- **Generalize the ad-hoc register checks** into `Instruction` helper methods.
- **Optional modeling:** the EFLAGS IOPL field is not modeled (no `iopl`
  subregister), so POPF/IRET/PUSHF/CLI/STI/IN/OUT do not report IOPL access;
  and implicit x87 control/state registers are not modeled.


bbiiggppiigg:
@hainest 

Do you want the Deferred / follow-up work to be in this PR or a separate PR. 
I'll start testing once that decision is made.

PR #2335, opened 2026-07-22
instructionAPI/PowerPC: fix BO-field decoding for the B-form bc branch

bbiiggppiigg:
The B-form conditional-branch decoder (BO()) extracted individual BO bits with field<N,N>(bo_field). field<> uses 32-bit instruction-word bit numbering (raw >> (31 - end)), so applying it to the already-extracted 5-bit bo_field value reads high-order bits that are always zero, rather than the intended BO[0]/BO[2]/BO[4] bits.

The practical consequence: `branch_always` (BO = 1z1zz) was never detected for B-form branches, so an always-taken branch-and-link was mis-classified as a *conditional call* (a control-flow target with both isCall and isConditional set). The ubiquitous PowerPC PIC prologue idiom `bcl 20,31,$+4` (used to materialize the PC into LR) is exactly this form. When the relocation engine builds a CFWidget for a block ending in such an instruction, CFWidget::generate hits `assert(!isConditional_)` in the call path and aborts -- crashing e.g. the test_reloc test on ppc64le.

Extract the BO bits directly from the 5-bit field (BO[0] is the MSB), the same way the XL-form (bcctr/bclr) path already does with a direct mask. After the fix: `bcl 20,31,$+4` decodes as an unconditional call (isCall=1, isConditional=0); genuine conditional branches (bc, bdnz, bdz) remain conditional.

### DIFF
--- a/tests/integration/InstructionAPI/decoder/ppc64le/branches.cpp
+++ b/tests/integration/InstructionAPI/decoder/ppc64le/branches.cpp
@@ -262,37 +262,31 @@ std::vector<test_insn> make_tests() {
     /*
      *  --- Branch Conditional B-form ---
      */
-    { //  bc +32
+    { //  bc +32  (BO=branch-always)
       0x42f00020, is_branch, !is_return,
-      test_cft{pc_value + 32, {!is_call, is_conditional, !is_indirect, !is_fallthrough}},
-      {},
-      {},
-      {},
-      test_cft{pc_value + 4, {!is_call, !is_conditional, !is_indirect, is_fallthrough}}
+      test_cft{pc_value + 32, {!is_call, !is_conditional, !is_indirect, !is_fallthrough}},
+      {}, {}, {}, {}
     },
-    { //  bca 32
+    { //  bca 32  (BO=branch-always)
       0x42f00022, is_branch, !is_return,
       {},
       {},
       {},
-      test_cft{32, {!is_call, is_conditional, !is_indirect, !is_fallthrough}},
-      test_cft{pc_value + 4, {!is_call, !is_conditional, !is_indirect, is_fallthrough}},
+      test_cft{32, {!is_call, !is_conditional, !is_indirect, !is_fallthrough}},
+      {}
     },
-    { //  bcl +32  (subroutine call to relative address)
+    { //  bcl +32  (subroutine call to relative address; BO=branch-always)
       0x42f00021, !is_branch, !is_return,
-      test_cft{pc_value + 32, {is_call, is_conditional, !is_indirect, !is_fallthrough}},
-      {},
-      {},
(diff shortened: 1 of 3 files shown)


hainest:
Derp. I can't believe I did that. I mean, I _can_, but still. You know, we could technically use Capstone 5.0 for ppc and aarch64 today...

bbiiggppiigg:
> Derp. I can't believe I did that. I mean, I _can_, but still. You know, we could technically use Capstone 5.0 for ppc and aarch64 today...

Well we cannot. There is some RISC-V constant that gets merged in that requires Capstone alpha 6.7.0 And then because of something in the cmake that requires the minimum version = dependency version, I actually need to use Capsteon 6.9.0 (in that PR)

hainest:
> > Derp. I can't believe I did that. I mean, I _can_, but still. You know, we could technically use Capstone 5.0 for ppc and aarch64 today...
> 
> Well we cannot. There is some RISC-V constant that gets merged in that requires Capstone alpha 6.7.0 And then because of something in the cmake that requires the minimum version = dependency version, I actually need to use Capsteon 6.9.0 (in that PR)

When `DYNINST_CODEGEN_ARCH=riscv64` but _without_ #2146, you can use alpha-6.2 up to alpha-6.4(?). For alpha-6.5+, you need 2146. For ppc and aarch64, you could use 5.0. I can't remember which alpha version contains my fixes for x86, but I think it's alpha-3 or so. I started conversion with x86 since it's the once I'm most familiar with and is the most popular for Dyninst users. That doesn't mean we couldn't convert the other architectures before x86. I just don't think Bart would find that a good use of your or Jim's time.

bbiiggppiigg:
@hainest @kupsch updated and ready for review (unless CI breaks).

tests/integration/InstructionAPI/decoder/ppc64le/branches.cpp
```diff
@@ -308,13 +302,16 @@ std::vector<test_insn> make_tests() {
       {},
       {}
     },
-    { // bclrl  (LK=1)
-      0x4e800021, is_branch, !is_return,
+    { // bclrl  (LK=1; BO=branch-always)
+      // The LR target is still marked conditional even though BO says
+      // branch-always (there is no fallthrough successor, so the flag is
+      // inconsistent); this documents current decoder behavior.
+      0x4e800021, !is_branch, is_return,
```
hainest, line 309: This is fixed in #2337?

bbiiggppiigg: > This is fixed in #2337?

Yes

hainest (approved):
No regressions on 5 ppc machines.

PR #2336, opened 2026-07-22
PowerPC: exclude ifunc resolvers from relocation-based instrumentation

bbiiggppiigg:
## Summary

When rewriting a shared library, Dyninst relocated/instrumented `STT_GNU_IFUNC` **resolver** functions along with everything else. On PowerPC64 (ELFv2) this makes the rewritten library crash at process startup, before `main`. This change excludes ifunc resolvers from relocation so they remain at their original addresses.

## Problem

An ifunc resolver is invoked by the dynamic linker during `_dl_relocate_object` (the `IRELATIVE`/BIND_NOW pass), i.e. very early in startup — before the PLT/GOT slots are populated and before any Dyninst runtime or base-tramp exists. If Dyninst has relocated the resolver, the relocated copy makes an indirect call through a PLT stub whose `.plt` slot is still zero:

```
std   r2,24(r1)
ld    r12,-<off>(r2)   ; .plt slot not yet filled by ld.so -> 0
mtctr r12              ; ctr = 0
bctr                   ; -> PC = 0  -> SIGSEGV, before main
```

This reproduces on ppc64le when rewriting `libc.so.6` (e.g. the `test_reloc` test): the relocated `strlen` ifunc resolver crashes the process at load time.

## Fix

In `CodeMover::addFunctions`, build a per-object set of the entry offsets of `STT_GNU_IFUNC` (`Symbol::ST_INDIRECT`) symbols and skip any function whose entry address is in that set. Excluded resolvers are never relocated or springboarded; their `IRELATIVE` relocation keeps pointing at the original resolver.

## Why this is safe

Callers of an ifunc symbol never call the resolver directly — they reach the chosen implementation through the PLT/GOT, which `ld.so` fills by running the (now unmodified, correct) resolver once at load time. So the resolver's original code runs, the right implementation is selected, and every caller works. The implementations and the callers remain fully instrumentable; only the resolver itself — which cannot be safely instrumented, since it runs inside the dynamic linker before the instrumentation runtime exists — is left untouched.

## Testing (ppc64le / POWER9)

- The rewritten `libc`'s `strlen` ifunc resolver is left as original code (no inserted branch to a relocated copy); the ld.so startup SIGSEGV is gone.
- Register-preservation / relocation tests (`test1_8`, `test1_9`, `test1_20`, `test1_22`) pass in both create and rewriter modes, i.e. normal instrumentation is unaffected.

### DIFF
--- a/dyninstAPI/src/Relocation/CodeMover.C
+++ b/dyninstAPI/src/Relocation/CodeMover.C
@@ -37,12 +37,16 @@
 
 #include "dyninstAPI/src/addressSpace.h" // Also for debug
 #include "patching/function.h"
+#include "dyninstAPI/src/mapped_object.h"  // func_instance::obj() -> parse_img()
+#include "dyninstAPI/src/image.h"          // image::getObject() -> SymtabAPI::Symtab
 
 #include "dyninstAPI/src/debug.h"
 #include "CodeTracker.h"
 #include "CFG/RelocGraph.h"
 
 #include <iterator>  // std::distance (rewrite-progress reporting)
+#include <map>       // ifunc-resolver exclusion cache
+#include <set>
 
 using namespace std;
 using namespace Dyninst;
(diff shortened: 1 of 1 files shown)


hainest:
@bbiiggppiigg Does this need any of the other ppc fixes?

bbiiggppiigg:
> @bbiiggppiigg Does this need any of the other ppc fixes?

I don't think so.

hainest (approved):
No regressions on all 4 ppc machines (8, 9, and 10).

PR #2342, opened 2026-07-24
dyninstAPI: fix i386 division snippet codegen (spurious SIGFPE)

bbiiggppiigg:
emitDiv on 32-bit x86 raised #DE (SIGFPE) for correct, nonzero divisors in two independent ways:

  - It called makeRegisterAvail(EAX) after loading the dividend (src1) into EAX. makeRegisterAvail spills the current occupant, so the dividend was evicted from EAX and the following idiv/div operated on a garbage dividend.

  - It emitted cdq unconditionally before both the signed idiv and the unsigned div. cdq sign-extends EAX into EDX, which is correct only for idiv. For an unsigned div with the dividend's top bit set (e.g. 0xffffffff), cdq set EDX = 0xffffffff, so div computed (0xffffffff:EAX) / src2, a quotient that does not fit in 32 bits and faults with #DE.

Fix: mark EAX defined (it is written by the divide) instead of spilling it; emit cdq only on the signed path and zero EDX on the unsigned path; and release src1's claim on EAX before mapping dest onto EAX (mirroring the src1-reused-as-dest pattern elsewhere in this emitter) so the destination assignment does not trip the register-occupied assertion.

Verified on i386 (coriander, 32-bit runtime): division snippets no longer crash with SIGFPE.

### DIFF
--- a/dyninstAPI/src/codegen/emitters/x86/IA32/EmitterIA32.C
+++ b/dyninstAPI/src/codegen/emitters/x86/IA32/EmitterIA32.C
@@ -580,13 +580,30 @@ namespace Dyninst { namespace DyninstAPI {
     gen.rs()->makeRegisterAvail(RealRegister(REGNUM_EDX), gen);
     gen.rs()->noteVirtualInReal(scratch, RealRegister(REGNUM_EDX));
     RealRegister src2_r = gen.rs()->loadVirtual(src2, gen);
-    gen.rs()->makeRegisterAvail(RealRegister(REGNUM_EAX), gen);
-    emitSimpleInsn(0x99, gen); // cdq (src1 -> eax:edx)
+    // EAX holds the dividend (src1, loaded above): cdq sign-extends it into
+    // edx:eax and idiv divides that pair. Do NOT makeRegisterAvail(EAX) here --
+    // that spills src1 out of EAX (makeRegisterAvail -> spillReal), leaving a
+    // garbage dividend and, with a valid nonzero divisor, a #DE / SIGFPE.
+    // Just mark EAX defined (it is written by the divide) without evicting it.
+    gen.markRegDefined(REGNUM_EAX);
     if(s) {
-      emitOpExtReg(0xF7, 0x7, src2_r, gen); // idiv eax:edx,src2 -> eax
+      emitSimpleInsn(0x99, gen); // cdq: sign-extend eax into edx for signed idiv
+      emitOpExtReg(0xF7, 0x7, src2_r, gen); // idiv edx:eax, src2 -> eax
     } else {
-      emitOpExtReg(0xF7, 0x6, src2_r, gen); // div eax:edx,src2 -> eax
+      // Unsigned divide: the dividend is 0:eax, so edx must be ZEROED, not
+      // sign-extended. The old code emitted cdq unconditionally before both
+      // idiv AND div; for an unsigned div with eax's top bit set (e.g. 0xffffffff)
+      // cdq set edx=0xffffffff, making div compute 0xffffffff`eax / src2, whose
+      // quotient overflows 32 bits -> #DE / SIGFPE.
+      emitMovImmToReg(RealRegister(REGNUM_EDX), 0, gen); // edx = 0
+      emitOpExtReg(0xF7, 0x6, src2_r, gen); // div edx:eax, src2 -> eax
     }
+    // The quotient is now in EAX; src1 has been consumed by the divide. Release
+    // src1's claim on EAX (freeRegister -> markVirtualDead clears the mapping)
+    // BEFORE assigning EAX to dest, otherwise noteVirtualInReal(dest, EAX) trips
+    // its "reg already occupied" assert. Mirrors the src1-reused-as-dest pattern
+    // elsewhere in this emitter (freeRegister(src1); noteVirtualInReal(dest,...)).
+    gen.rs()->freeRegister(src1);
     gen.rs()->noteVirtualInReal(dest, RealRegister(REGNUM_EAX));
     gen.rs()->freeRegister(scratch);
   }


hainest:
Let's add a test for this in https://github.com/dyninst/dyninst/blob/master/tests/unit/dyninstAPI/emitter/x86_64.cpp.

tests/unit/dyninstAPI/emitter/x86_64.cpp
```diff
@@ -265,6 +273,76 @@ int main() {
     0x50  // push rax
   }});
 
+  /*
+   * IA-32 emitter
+   *
+   * The 32-bit emitter is compiled into the x86_64 build, so its codegen can be
+   * checked here. Unlike the 64-bit emitter, it operates on *virtual* registers
```
hainest: Let's split this into its own file. Sorry, I should have mentioned that when I commented earlier.

hainest:
@bbiiggppiigg On coriander, I built Dyninst with `BUILD_RTLIB_32=ON ENABLE_STATIC_LIBS=OFF` and the testsuite with `DYNINST_ENABLE_32BIT_TESTS=ON`. With that, I'm seeing 697 of 1100 32-bit tests passing. Is that consistent with your findings? The full report is in `/home/tim/paradyn/2342.results`, if you're interested.

bbiiggppiigg:
@hainest 

Yes, consistent. I ran the same configuration on coriander (64-bit Dyninst with `BUILD_RTLIB_32=ON`, testsuite with `DYNINST_ENABLE_32BIT_TESTS=ON`, gcc-10 for the 32-bit pieces) on current master (03e7e074d) as `runTests -32 -all`, and diffed your 32-bit rows against mine by verdict class (PASSED/FAILED/CRASHED, ignoring the parenthesised reason):

- 1063 of the 1075 rows we both have are identical.
- The only real difference is test1_6 create/attach (8 rows): `CRASHED (Group Teardown)` on master, `FAILED (Running Test)` in your run. That is this PR: the SIGFPE from the unconditional `cdq` before an unsigned `div` is gone, and what remains is the pre-existing FAILED from test1_6's `long long` mul/div subtests (`3*INT_MAX`, `LLONG_MAX/INT_MAX`, `ULLONG_MAX/100`, ...), which i386 snippets can't do with 32-bit registers. Same story for test1_5 (`LLONG_MIN/MAX` comparisons). Both fail identically on master.
- test_reloc create (4 rows) is PASSED on my master run and CRASHED in yours; it's unstable at 32-bit on master too (in isolation it SIGABRTs `test_driver` on master), so probably noise.

The ~400 non-passing 32-bit rows are pre-existing and unrelated to the division fix. Most of them are cascades: runTests batches ~10 tests per `test_driver`, and one crash stamps the whole group `CRASHED (Group Setup/Teardown)`.

1. Every 32-bit `rewriter` group crashes in Group Setup/Teardown (322 of the 329 rewriter rows, in your run and on master alike).
2. `pc_tls` create/attach fail on every variant; `pc_hw_breakpoint` fails and then hangs.
3. test1_2/test1_5/test1_6 (64-bit integer snippet ops), test1_20, test_callback_2 create, test_thread_5 create, snip_ref_shlib_var.

Running each test in isolation (`test_driver -32 -test <T> -all` under a timeout) removes the cascades and also reaches the tail that the grouped run never gets to: `runTests` kills the run when one group exceeds its hard-coded 600 s (`runTests-utils.C`), and the 32-bit `pc_hw_breakpoint` group hangs, so both your run and mine stop right after it (pc_launch, pc_library, pc_singlestep, pc_stat, pc_temp_detach, pc_terminate*, pc_thread*, pc_tls and the symtab tests never ran at 32-bit in either).

| run (32-bit rows only)              | rows | pass | fail | crash | driver died / hung | non-passing |
| ----------------------------------- | ---- | ---- | ---- | ----- | ------------------ | ----------- |
| grouped `runTests -32 -all`, master | 1075 | 701  | 43   | 330   | 0                  | 374         |
| grouped, your run (PR 2342)         | 1100 | 697  | 76   | 326   | 0                  | 403         |
| per-test isolation, master          | 1277 | 1058 | 84   | 90    | 44                 | 219         |

Separately, while looking at this I found that `-DDYNINST_CODEGEN_ARCH=i386` on x86_64 can't run the testsuite at all on master (a `set(${_codegen_arch} ...)` typo in `DyninstPlatform.cmake` that makes the installed config report `ppc64_linux`, a `cap_32_64`-guarded 4-byte write in `trampTrapMappings::writeToBuffer` that corrupts the mutator heap, and the rewriter's library resolver parsing 64-bit candidates the i386 build has no ABI tables for). With those three fixed, the i386 code generator matches the x86_64 one on every 32-bit test except test_pt_ls (it parses the 64-bit system `ls`). I'll send that as its own PR.

For reference, my results are on coriander:

- Grouped `runTests -32 -all` on master, 1075 rows, 701 passed:
  `/home/wuxx1279/dyn/master/results/runs-m32/20260917-171046/stdout.log`
  (results table; `output.log` beside it is the per-subtest detail, `env.txt` records the SHAs and environment)
- Per-test isolated run on master, 1277 rows, 1058 passed:
  `/home/wuxx1279/dyn/master/results/pertest-m32/20260917-173407/RESULTS.txt`
  (`logs/<test>.stdout` and `logs/<test>.log` per test beside it)
- The `-DDYNINST_CODEGEN_ARCH=i386` runs, on the branch with the three fixes (will come with that PR):
  `/home/wuxx1279/dyn/pr-codegen-arch-platform/results/runs-i386/20260917-164316/stdout.log` (grouped, 1100 rows)
  `/home/wuxx1279/dyn/pr-codegen-arch-platform/results/pertest-i386/20260917-175612/RESULTS.txt` (per-test, 1274 rows)

hainest (approved):

PR #2344, opened 2026-07-24
Port PowerPC decoder tests from the testsuite to tests/integration

bbiiggppiigg:
Ports the two PowerPC InstructionAPI tests from the dyninst/testsuite repository (src/instruction/power_decode.C and
src/instruction/power_cft.C) into the modern integration test tree as tests/integration/InstructionAPI/decoder/ppc64le/, following the layout of the existing aarch64/riscv/x86 decoder tests. The testsuite framework boilerplate (TestMutator, test factories, test_results_t) was stripped; the checks themselves carry over unchanged. Both tests decode fixed in-memory byte buffers, so they run on any host architecture.

- decode.cpp (instructionAPI_ppc64le_decode_test): port of power_decode. Verifies register read/write sets for 27 fixed-point, floating-point, load/store, and branch instructions, plus a readsMemory() check for lhzux. Uses the shared register_tests harness (exact set equality, same semantics as the testsuite's verify_read_write_sets). As in the original, decoding is done with Dyninst::Arch_ppc32 and the expected registers come from the ppc32 namespace, because the decoder currently produces ppc32-namespace registers even in 64-bit mode. Once the decoder is switched to the ppc64 register namespace, this test needs a mechanical update (Arch_ppc64 + ppc64:: registers).

- branches.cpp (instructionAPI_ppc64le_branches_test): port of power_cft. Verifies control-flow targets for the branch instruction families (I-form, B-form, XL-form to LR/CTR): each target expression is evaluated by binding PC/CTR/LR to known values and checking the resulting address as well as the per-target call/conditional/ indirect/fallthrough properties, plus isBranch()/isReturn(). This test does not use the shared cft_tests harness because that harness assumes all CFTs of an instruction share the same properties and does not evaluate target addresses; PowerPC conditional branches have a taken target and a fallthrough target with different properties, and the value checks would be lost.

Notes on the original tests, preserved faithfully:
- The buffer comment "lhzux r9, r7, r5" in power_decode was misleading; the encoding 0x7ca74a6e is lhzux r5, r7, r9 (fixed the comment only, the expected sets are unchanged).
- power_decode passed a 1-byte buffer length when re-decoding lhzux; the port passes the correct 4-byte length.
- The bctar/bctarl cases remain commented out (the decoder cannot decode them), same as the testsuite.

Verified with gcc 11.1 / Boost 1.78 (RelWithDebInfo, DYNINST_ENABLE_TESTS=INTEGRATION): both tests pass under ctest -R ppc64le.

### DIFF
--- a/tests/integration/CMakeLists.txt
+++ b/tests/integration/CMakeLists.txt
@@ -3,6 +3,7 @@ include_guard(GLOBAL)
 message(STATUS "Enabling integration tests")
 
 add_subdirectory(InstructionAPI/decoder/aarch64)
+add_subdirectory(InstructionAPI/decoder/ppc64le)
 add_subdirectory(InstructionAPI/decoder/x86)
 
 if(DYNINST_ENABLE_CAPSTONE)
--- a/tests/integration/InstructionAPI/decoder/ppc64le/CMakeLists.txt
+++ b/tests/integration/InstructionAPI/decoder/ppc64le/CMakeLists.txt
@@ -1,0 +1,13 @@
+include_guard(GLOBAL)
+
+add_executable(ppc64le_decode_test decode.cpp)
+target_compile_options(ppc64le_decode_test PRIVATE ${SUPPORTED_CXX_WARNING_FLAGS})
+target_link_libraries(ppc64le_decode_test PRIVATE instructionAPI register_tests)
+add_test(NAME instructionAPI_ppc64le_decode_test COMMAND ppc64le_decode_test)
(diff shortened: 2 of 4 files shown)


hainest (approved):

bbiiggppiigg:
@hainest 
I just realized while this result is correct for master, it will fail when #2335  and #2337 gets merged. Do you want me to merge this (with substituting ppc32 to ppc64 registers) first, and then in 2335 and 2337 update the test accordingly? Or should 2335 and 2337 be merged first and this PR's tests get updated to use the correct information ?

Issue #712, opened 2020-02-25
No registers live after x86_64 retq instructions according to liveness query


wcohen:
I have been experimenting with Dyninst's liveness analysis described in the dataflowAPI manual and wrote a little program that provides a count of the registers live before and after each instruction:

https://github.com/wcohen/quality_info/tree/master/dyninsttools

One thing I noticed was that each retq instruction had zero registers live after the instruction executed. This seems odd with the x86_64 ABI returning  values in rax for some functions.  Is no register live after retq instruction correct behavior?

wcohen:
So the analysis is limited to within a function.  How does the liveness analysis handle calls that leave a function and then return?

mxz297:
> So the analysis is limited to within a function. How does the liveness analysis handle calls that leave a function and then return?

Still rely on the ABI. For example, for a call instruction, registers that are used for passing the parameters are treated as read.

wcohen:
What about the register(s) used to return values for function calls?  Would the liveness analysis possibly get things wrong for a function that calls another function with no arguments but returns a value in register?

mxz297:
> What about the register(s) used to return values for function calls? Would the liveness analysis possibly get things wrong for a function that calls another function with no arguments but returns a value in register?

Similarly, Dyninst treats call instruction reading %rax for this reason:

https://github.com/dyninst/dyninst/blob/master/dataflowAPI/src/ABI.C#L308

Without an inter-procedural analysis, Dyninst will over-approximate the live registers to be safe. An inter-procedural liveness analysis is on my to-do list, but I have not got there yet...

wcohen:
I have been looking at code generated by GCC.  There are cases where the compiler generates cold sections of code that are in a region of text marked as <function_name>.cold.  A function may have branches out to snippet of code in <function_name>.cold and then jump back to the some place in <function_name>, so it is certainly possible for leave the region of code associated with <function_name>.  Can see example of this in the calibrate_delay function in Fedora 34 /usr/lib/debug/lib/modules/5.12.7-300.fc34.x86_64/vmlinux .  There are some values in registers being passed to code snippets in the calibrate_delay.cold

I have some code in https://github.com/wcohen/quality_info/blob/master/dyninsttools/readonly.C that uses the liveness analysis to determine if a variable'st value is no longer used.  Not sure it is entirely correct, but readonly does generate information for calibrate_delay.cold particularly tracking the transitions between calibrate_delay and calibrate_delay.cold.

[...]

Issue #609, opened 2019-06-12
AMD64 ABI registers corrupted by BPatch_basicBlock instrumentation


mothran:
I recently opened a pull request (#603) that attempted to resolve the issue I was seeing.  That pull request was incomplete and below is a full reproduction of the issue I was running into.

It appears that the BPatch emitBT() code for AMD64 misses the liveness of some ABI registers during a callback, so registers used in the callback can leak into the instrumented context causing crashes / undefined behavior.  Below are reproduction steps:

Dyninst version:
fdc4d6a2a82726f1c2d71cac47ae66c204f9c1ea

cmake command:
cmake -DBUILD_RTLIB_32=ON -DCMAKE_INSTALL_PREFIX=/usr/dyninst ../..

System:
ubuntu 18.04
gcc:
Ubuntu 7.4.0-1ubuntu1~18.04

Reference repo:
https://github.com/mothran/dyninst-bb-patch

Dependencies:
```bash
sudo apt install -y libyaml-dev
```

build sets (in reference repo):
```bash
export DYNINST_ROOT=/usr/dyninst/
export LD_LIBRARY_PATH=${DYNINST_ROOT}/lib
export DYNINSTAPI_RT_LIB=${LD_LIBRARY_PATH}/libdyninstAPI_RT.so
make
make yaml #see note below if fails
```

Note on 'make yaml':  The makefile has a hardcoded address to libyaml.a for static linking, if this build stage fails please modify the makefile with the correct location of libyaml.a.

Running the instrumentation
```bash
./patcher ./tests/yaml ./yaml-patch
```

Running the target:
```bash
echo -ne "'AA\n\t " > /tmp/blah.txt
LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:. ./yaml-patch /tmp/blah.txt
Segmentation fault (core dumped)
```

Running without instrumentation:
```bash
./tests/yaml /tmp/blah.txt # no crash
```

Crash location:
```
Block: 0x2176

Original block assembly:
00002176  mov     dword [rdi+0x154], 0x0
00002180  mov     edi, 0x10
00002185  mov     qword [rsp+0x30 {var_118}], 0x0
0000218e  mov     qword [rsp+0x38 {var_110}], 0x0
00002197  mov     qword [rsp+0x40 {var_108}], 0x0
000021a0  mov     qword [rsp+0x50 {var_f8}], 0x0
000021a9  mov     qword [rsp+0x58 {var_f0}], 0x0
[...]

PR #2349, opened 2026-07-27
[DRAFT] ABI (1135)

### LINKED ISSUE #1135: Split dataflowAPI/src/ABI.C into arch-specific files
This will drastically simplify the logic in this file and make it reflect the structure of other arch-specific functionality.

bbiiggppiigg:
# Architecture-independent ABI model; PowerPC register-namespace unification

Supersedes #1712 (Rewrite and fix ABI representation and usage).
Fixes #1135.

This PR does **not** fix the register-liveness bugs #724/#712/#609:
- **#724** (rdi clobbered by exit instrumentation): the issue's own reproducer
  (`files_rdi.tar.gz`) still segfaults with this branch — the crash dereferences
  `rdi` holding a garbage ioctl argument (`rdi=0x5`), exactly as reported — and
  it reproduces identically on current master. Still open.
- **#712** (no registers live after x86_64 `retq`): both master and this branch
  report the return + callee-saved set live *entering* the ret (the reported
  "zero live" was an after-the-ret query); not a live defect on either.
- **#609**: not reproduced.

The rewrite does make the derived clobber/read sets more ABI-accurate (e.g. at a
`retq` the callee-saved `rbp` is live and the caller-saved `rcx` is not, where
the previous tables had it inverted), but that is a correctness cleanup, not a
fix for the above.

## Summary

This PR replaces Dyninst's per-architecture, hand-written ABI/liveness tables
with a single **architecture-independent ABI model** (`Dyninst::ABI`), and
collapses the duplicated 32-bit PowerPC (ppc32) support so that PowerPC uses one
register namespace (ppc64) throughout while 32-bit PowerPC binaries remain
analyzable.

Net change is a removal: **47 files, +1634 / −2486**.

There are three threads of work:

1. **A common ABI model.** A new `Dyninst::ABI` (`common/src/ABI`) describes each
   architecture's calling convention as `registerSet`s — parameters, returns,
   callee-saved, globals, and the syscall convention — from which the
   caller/return/syscall read-and-clobber sets used by liveness are *derived*
   (`callWritten = computational − preserved`, etc.) rather than hand-listed.
2. **Bridging liveness onto it.** `dataflowAPI` no longer keeps per-arch
   `bitArray` tables; `ABIBridge` converts the common ABI's base-register sets
   into the liveness `bitArray` representation (expanding to every width-view
   alias, sizing by max register index). x86/x86_64, ppc64, and aarch64 liveness
   now consume the common ABI.
3. **PowerPC register-namespace unification (ppc32 → analyze-only).** The
   redundant ppc32 register namespace, ABI, emitters, and proccontrol paths are
   removed; PowerPC decodes, parses, and instruments through the ppc64 namespace.
   `Arch_ppc32` is retained as an **analyze-only alias** of ppc64 so 32-bit
   PowerPC binaries still decode and open in SymtabAPI/ParseAPI.

### Common ABI (`common/src/ABI`, `common/h`)
- New `Dyninst::ABI` / `registerSet` / `architecture` types and the
  `registerSet`-based accessors.
- Per-architecture convention tables: `make_x86`, `make_x86_64`, `make_ppc64`,
  `make_aarch64` (including syscall conventions and PowerPC nonvolatile GPRs).
- Clobber sets are derived from the computational-register set, so scratch/flag
  registers are captured without enumerating control/status/segment registers.

### dataflowAPI liveness (`dataflowAPI/src/ABI.C`, `ABIBridge`)
- `ABIBridge::buildABIBitArrays` converts the common ABI sets to liveness
  `bitArray`s; expands base registers to all width-view aliases (e.g. xmm/ymm/zmm)
  so narrow-view liveness queries don't miss a live register.
- Liveness `bitArray`s are sized by **max register index + 1**, not map entry
  count (the index maps are sparse; the old sizing tripped dynamic_bitset's
  bounds check on 32-bit x86 targets).
- `initialize64()` for x86_64, **ppc64, and aarch64** now route through the
  bridge; the legacy hand-written tables are deleted. (The ppc64/aarch64 wiring
  also removes a latent aarch64 bug where `callWritten` aliased the `callRead`
  pointer.)

### PowerPC: one register namespace, ppc32 analyze-only
- Instruction decoding, symbolic evaluation (ROSE), liveness/registerSpace, and
  DWARF/ROSE register conversion all move to the ppc64 register namespace.
- Removed: ppc32 register map, ppc32 ABI, ppc32 syscall tables, ppc32 code
  emitters, ppc32 proccontrol and loadLibrary codegen, ppc32 register-conversion
  paths.
- `Arch_ppc32` stays in the `Architecture` enum as a 4-byte analyze-only alias of
  ppc64: ELF recognition, decode/parse dispatch, and register conversion all
  route through ppc64, while instrumentation of ppc32 targets is dropped.

### Build / compatibility
- Drop an NSDMI on the pimpl pointer for gcc-8 compatibility.
- Handle `Arch_riscv64` / `Arch_amdgpu_gfx950` in the ABI dispatch switch
  (`-Werror=switch`) and re-run cmake-format.

## Architecture coverage

| Arch | Liveness ABI source |
|---|---|
| x86 / x86_64 | common `Dyninst::ABI` (bridge) |
| ppc64 | common `Dyninst::ABI` (bridge) |
| aarch64 | common `Dyninst::ABI` (bridge) |
| riscv64 / amdgpu | unchanged — legacy per-arch tables (no common ABI table yet) |
| ppc32 | analyze-only alias of ppc64 (decode/parse only; no instrumentation) |

## Validation

Build gate (CI `-DDYNINST_WARNINGS_AS_ERRORS=ON`): x86_64 gcc **and** clang
clean; ppc64 (cyclops) and aarch64 (zeroah) native gcc `-Werror` clean.

For each architecture whose liveness moved to the common ABI, the derived ABI
sets and the resulting instrumentation were validated against the previous
implementation:

- **Set-level A/B** (bridge output vs the legacy tables): read sets
  (`returnRead`/`callRead`) are supersets of the legacy sets (conservative); the
  clobber sets (`callWritten`/`syscallWritten`) contain no callee-saved /
  kernel-preserved register, so liveness never marks a live register dead;
  syscall sets match each architecture's kernel ABI.
- **In-tree ctest** (`DYNINST_ENABLE_TESTS=ALL`): x86_64 (coriander) full pass;
  ppc64 (cyclops) 108/108; aarch64 (zeroah) 125/132 (the 7 are pre-existing
  amdgpu/DWARF5 test-fixture gaps, unrelated to this change).
- **Isolated testsuite A/B** — the same `dyninst/testsuite` binaries run against
  the new vs the previous dyninst library (only the `.so` differs): **result
  tables are identical** on ppc64 (793P/13F) and aarch64 (1762P/1F); the
  failures are pre-existing (`test_pt_ls`, `test_thread_3/8`) and reproduce on
  the previous library. x86_64 `runTests -all`: 1512 pass / 0 fail.
- **ppc32 analyze-only**: verified that a ppc32 binary decodes and opens in
  SymtabAPI (arch reported as ppc32, symbols and `.text` decode) with the
  instrumentation paths removed.

## Notes for reviewers

- **riscv64 and amdgpu are intentionally not ported** to the common ABI in this
  PR; their liveness continues to use the existing per-arch tables. The
  `make_*` scaffolding is x86/ppc64/aarch64 only.
- **ppc32 is now analyze-only.** Code that instruments 32-bit PowerPC targets is
  removed; decoding/parsing of ppc32 binaries is preserved. Out-of-tree code that
  referenced the ppc32 register namespace (`ppc32::`) or instrumented ppc32
  targets will need updating.
- The PowerPC decoder-semantics fixes surfaced during this work were extracted
  and merged separately (BO-field, bclr conditional-return, isel dispatch), so
  this branch assumes those are already in master.


bbiiggppiigg:
@hainest @kupsch 

I think this is working now on x86_64 (coriander), aarch64(zeroah) and ppc64le(cyclops).
I'm thinking that maybe I can split this into three PRs, one for base + x86_64, one for aarch64 and one for ppc64le.
Let me know what you guys think.

hainest:
> I'm thinking that maybe I can split this into three PRs, one for base + x86_64, one for aarch64 and one for ppc64le. 

I'm not sure if splitting at the PR level would be helpful, but definitely having clear commits in this PR is necessary (I've not looked at the invididual commits yet, so this may already be the case).

We should chat about ppc32. It is true that #1145 removed the ability to _build_ on a ppc32 platform, but there are still distros that have multilib support so it's possible they could be running 32-bit binaries on ppc64 platforms. I would be surprised if anyone using Dyninst is doing that, but it is possible. I don't have strong opinions on removing it. In any case, the changes here to fully disambiguate ppc32 from ppc64 are absolutely needed.

hainest:
Notes on ppc32 support:

[fedora](https://developer.fedoraproject.org/deployment/secondary_architectures/powerpc.html): not supported after Fedora 22 (EOL since 2016)

[debian](https://wouter.pages.debian.net/webwml/ports/powerpc/index): not supported since Debian 8 ([EOL](https://www.debian.org/releases/) since 2025).

[gentoo](https://wiki.gentoo.org/wiki/Project:PowerPC): still supported, but most mentions are for PowerMacs. I'm not sure if they support Power8 or newer (that's the minimum Dyninst supports).

[openSUSE](https://en.opensuse.org/Portal:PowerPC): dropped after Evergreen 11.4. The last Evergreen release was [13.1](https://en.opensuse.org/openSUSE:Evergreen) which had its support dropped in 2016.

[RHEL](https://access.redhat.com/solutions/509373): dropped in RHEL 7 (last [updated](https://access.redhat.com/articles/red-hat-enterprise-linux-release-dates#RHEL7) in 2020).

https://powerpc.dev/linux/distros.html has a few others, but no citations for their sources.

PR #2350, opened 2026-07-27
Suppress stack frame size warning for aarch64 instruction tests

hainest:


hainest:
After applying the nuclear approach, the Compiler Multibuild workflow is passing.

bbiiggppiigg:
@hainest 

Can we just increase this size limit for the entire project?
Or you think it is better to specify on a per source code basis?

bbiiggppiigg (approved):

kupsch (approved):

kupsch:
It's useful to know that we are not creating large stack objects in general and especially that it does not occur in functions that might be recursive.

PR #2352, opened 2026-08-01
parseAPI: don't treat unrelocated CF displacements as concrete targets

bbiiggppiigg:
Rewriting a statically linked binary crashes intermittently (~60% of runs)
with a NULL dereference in `Parser::parse_frame_one_iteration`
(`Parser.C:1498`). The crash is ASLR- and allocation-order-dependent and
never reproduces under a default gdb session (gdb disables ASLR), which made
it look like a heisenbug.

Root cause: in unlinked object files (`ET_REL`, e.g. the `libc.a` members
loaded by `BinaryEdit::doStaticBinarySpecialCases`), the displacement bytes
of direct calls/branches that name another section or an external symbol are
placeholders awaiting link-time relocation. ParseAPI ignored the relocation
entries and used the placeholder-derived constant as a real control-flow
target. Two failure shapes:

- `call ext_func` decodes as a call to the next instruction, which
  `isRealCall()` misclassifies as a PC-getter, silently folding real call
  sites into straight-line code;
- a tail call at the end of a section decodes as a jump to one past the end
  of the section. Concretely: `libc.a(loadmsgcat.o)` ends section
  `__libc_freeres_fn [0,0xe5)` with an unrelocated `jmp free`
  (`R_X86_64_PLT32`), giving target `0xe5`. Since every section in a `.o`
  is based at 0, `0xe5` lies in the *overlapping* `.text` region, so the
  jump is classified a cross-region tail call. The resulting call work
  element fails both `createAndRecordFunc` and `findFunc` for the frame's
  region, and the parser dereferences the NULL callee.

Whether the crash fires depends on `SymtabCodeSource::lookup_region`
resolving the ambiguous overlapping-region lookup via
`*std::set<CodeRegion*>::begin()` — pointer order, hence the ASLR/allocation
sensitivity.

## Minimal reproducer

No mutator or rewrite is needed — parsing the one archive member is enough.
Extract it and look at the offending instruction:

```console
$ ar x /usr/lib64/libc.a loadmsgcat.o
$ objdump -dr --section=__libc_freeres_fn loadmsgcat.o | tail -3
  e0:   e9 00 00 00 00    jmpq   e5 <.annobin__nl_unload_domain.end>
                          e1: R_X86_64_PLT32     free-0x4
```

The `rel32` field is a placeholder (zeros), so the decoded target is `0xe5` —
one byte past the end of the section `[0, 0xe5)`. Parse the file with a
plain ParseAPI harness:

```cpp
#include "Symtab.h"
#include "CodeSource.h"
#include "CodeObject.h"
using namespace Dyninst;
using namespace Dyninst::ParseAPI;

int main(int argc, char **argv) {
  SymtabAPI::Symtab *st = nullptr;
  if (!SymtabAPI::Symtab::openFile(st, argv[1])) return 2;
  SymtabCodeSource scs(st);
  CodeObject co(&scs);
  co.parse();
  // mirror image::analyzeImage(): gap-parse each text region
  for (auto *r : scs.regions()) {
    auto *scr = static_cast<SymtabCodeRegion *>(r);
    if (scr->symRegion()->isText()) co.parseGaps(scr);
  }
  printf("parsed %zu functions\n", co.funcs().size());
}
```

```console
$ OMP_NUM_THREADS=1 ./harness loadmsgcat.o     # before this PR: SIGSEGV
```

Single-threaded, this crashes **deterministically** (A/B-verified against
unpatched master: 6/6 segfault before, 0/6 after): in a fresh process the
two overlapping regions are allocated in a fixed order, so the
pointer-ordered ambiguous lookup always answers `isCode(0xe5)` from `.text`
and the bogus cross-region tail-call work element is always created. With
default (multi-threaded) parsing the same run crashes only ~2/3 of the time
(4/6 measured; thread scheduling perturbs allocation order), inside a larger
process such as a real rewriter ~60% — and never under a default gdb
session, since gdb disables ASLR.

**Caveat for reproducers**: whether an unfixed build crashes *at all* is
itself decided by that build's allocation order — we have one unfixed master
build that segfaults 6/6 and another (same bug present) that survives every
run and every libc.a member because the ambiguous `is_code` lookup happens
to always pick the non-code region, demoting the bogus edge to a harmless
sink edge. The reliable build-independent observable is the parsing log:

```console
$ OMP_NUM_THREADS=1 DYNINST_DEBUG_PARSING=1 ./harness loadmsgcat.o 2>&1 | \
    grep -E "in other regions, TAIL CALL|unapplied relocation"
```

Unfixed builds print `jump to 0xe5 in other regions, TAIL CALL` (the bogus
edge); fixed builds print `REJECTED: CF target ... carries an unapplied
relocation` instead.

After this PR the same run parses cleanly; with `DYNINST_DEBUG_PARSING=1`
the jump is reported instead as

```
REJECTED: CF target of 'jmp 0x5(%rip)' at 0xe0 carries an unapplied relocation (placeholder target 0xe5)
```

(67 rejection lines at 58 distinct call/branch sites in this one member; the
others are the `call <next insn>` placeholder shape that previously parsed as
PC-getters.)

## Fix

The NULL-callee crash guard in `Parser::parse_frame_one_iteration` was split
out at review request and merged separately as #2386; this PR is rebased on
it and now contains the root-cause fix plus a regression test.

**Root cause:**
- **`CodeRegion::hasUnresolvedRelocs(lo, hi)`** (new, optional): true when
  bytes in `[lo,hi)` are covered by an unapplied static relocation.
  `SymtabCodeRegion` implements it from the region's relocation entries,
  populated **only** when `Symtab::isUnlinkedObjectFile()`. Linked binaries —
  including `--emit-relocs` builds, whose relocation entries are already
  applied — take the empty-vector fast path and are behaviorally unchanged.
  Offsets are rebased by `low()` (ET_REL `r_offset` is section-relative).
- **`IA_IAPI::getCFT()`** rejects a computed target when the instruction's
  bytes carry an unapplied relocation. Downstream this makes such calls
  calls-to-unknown (sink + fallthrough) and such branches
  tail-calls-to-unknown (sink), matching how unresolved indirect control
  flow is already modeled. `isIndirectJump()` correspondingly declines to
  run jump-table analysis on placeholder bytes, and `isRelocatable()` no
  longer asserts on an undefined call CFT.
- **`isDynamicCall()` returns false** for such calls: the callee is
  statically unknown, but the call is not through a register or memory
  operand — so call sites in unlinked objects are *not* classified as
  dynamic call sites (they take the normal unknown-callee sink-call path).
- **Conditional branches** with an undefined taken-target now edge to the
  sink (`ADDRESS_INVALID`) instead of the placeholder constant.

**Known gap**: `Object-elf.C` attaches a relocation to the code region it
applies to only when the relocation's symbol resolves (`if (sym)`,
`Object-elf.C:4106`); unattached relocations are invisible to this check and
fall through to the guard merged in #2386.

## Regression test

`tests/regression/parseAPI/UnappliedRelocations/` adds `unrelocated_cft`, the
harness above, run by ctest as `parseAPI_unrelocated_cft_test` on
`libc.a(loadmsgcat.o)` from dyninst/test_binaries (dyninst/test_binaries#5,
merged). It runs with `DYNINST_DEBUG_PARSING=1` and passes only if the
parsing log reports the `jmp` at `0xe0` as rejected (placeholder target
`0xe5`) and does not report `jump to 0xe5 in other regions, TAIL CALL`. On
an unfixed tree the file parses to the bogus tail call (and segfaults
without #2386), so the test fails there deterministically.

## What this deliberately does NOT do (follow-up work)

The relocation on the displacement names the real callee (e.g. `free`);
this PR does not resolve it:

- **Intra-object cross-section targets** (e.g. hot/cold split `.text` →
  `.text.unlikely`) become unknown-target edges rather than resolving the
  relocation symbol to its real (region, offset). Doing that correctly needs
  region-qualified targets threaded through the call work-element plumbing,
  which is address-only today. Such targets normally carry their own
  symbols, so they are still parsed as functions via hints. Note the old
  code never created the correct cross-section edge either — the placeholder
  bytes cannot encode it — so no capability is lost.
- Callee-name-based analyses (e.g. non-returning detection for `call abort`
  in a `.o`) could use the relocation's symbol name; they don't yet.

## Behavior changes / residual risk

- **Linked binaries: no change, by construction** (the
  `isUnlinkedObjectFile()` gate). Verified by re-parsing a rewritten static
  binary containing `.dyninstInst`: zero rejected targets, zero skipped work
  elements, and cross-region edges from original code into instrumentation
  (`.text` → `.dyninstInst`) still created and followed.
- **Unlinked `.o` CFG shapes change.** Every changed case previously
  produced a wrong CFG, but consumers that parse `.o` files will see
  differences: external call sites now end their basic block with
  CALL(sink) + CALL_FT edges where they previously folded into
  straight-line code. In particular, instrumenting functions *inside*
  archive members during static rewriting now takes the
  call-with-unknown-callee path at such call sites; this is the standard
  unknown-callee path, but it is the least-exercised combination in the
  validation below.
- **32-bit PIC thunk idiom in `.o`**: `call __x86.get_pc_thunk.*` was
  previously (coincidentally) suppressed as a PC-getter; it now yields a
  sink call edge. Linked 32-bit binaries are unaffected.
- **Indirect conditional branches** (e.g. POWER `bcctr`-style): COND_TAKEN
  now targets the sink rather than address 0. Identical outcome unless
  address 0 is code, which only occurs in `.o` files.
- **ABI**: new virtual on `CodeRegion` and a new data member in
  `SymtabCodeRegion` (vtable layout / object size change); out-of-tree
  `CodeRegion` subclasses must be recompiled.
- **Perf**: one gather+sort of relocation offsets per region, only for
  unlinked object files; not measurable in the 1671-member sweep below.

## Validation

- Deterministic mini-repro (ParseAPI-only harness on `libc.a(loadmsgcat.o)`,
  single-threaded): crash/bogus-edge condition fired 10/10 before, 0/10
  after; the parsing log shows all 58 relocated CF sites in that member
  rejected and reclassified. Now the ctest regression test above.
- Sweep-parsed **all 1671 members** of glibc's `libc.a`: 0 crashes,
  0 failures.
- Full static-binary rewrite (insert a `puts` call into `main` of a
  statically linked executable): rewritten binary runs and prints the
  expected output. This was the original failing scenario.
- Re-parsed the rewritten binary (contains `.dyninstInst`): 1090 functions,
  0 rejected CFTs, 0 skipped work elements; direct edges from original code
  into `.dyninstInst` addresses present.
- In-tree ctest (`DYNINST_ENABLE_TESTS=ALL`), branch vs master at the same tip
  (4547a0bf2): x86_64 (mahir, coriander) 134/134 vs 133/133, ppc64le (cyclops)
  109/109 vs 108/108, aarch64 (zeroah) 133/133 vs 132/132 — the +1 is the new
  test, passing on every host.
- dyninst/testsuite, branch vs master at the same tip, on four machines:
  **no regressions** — every run complete, result tables identical to master
  row for row. x86_64 coriander (grouped `-all`, 1172 rows, 0 non-PASSED);
  aarch64 zeroah (grouped `-all`, 1233 rows, only the known `test_pt_ls`
  flake, on both sides); ppc64le cyclops (per-test isolated, 832P/14F/19H on
  both — the pre-existing ppc64le proccontrol/thread set); x86_64 mahir
  (per-test isolated, 1272P/16F/39H/6D on both — pre-existing platform noise).
  Details in the comment below.

### DIFF
--- a/parseAPI/h/CodeSource.h
+++ b/parseAPI/h/CodeSource.h
@@ -86,6 +86,16 @@ class DYNINST_EXPORT CodeRegion : public Dyninst::InstructionSource, public Dyni
 
     virtual bool wasUserAdded() const { return false; }
 
+    /* True if any byte in [lo,hi) is covered by an unapplied static
+       relocation, meaning the raw bytes are placeholder values rather
+       than final values. Only possible in relocatable (unlinked) files,
+       where control-flow displacements naming other sections or external
+       symbols have not been resolved by a linker yet.
+
+       Optional */
+    virtual bool hasUnresolvedRelocs(Address /* lo */, Address /* hi */) const
+    { return false; }
+
 };
 
 template <typename OS>
@@ -243,6 +253,11 @@ class DYNINST_EXPORT SymtabCodeRegion : public CodeRegion {
     SymtabAPI::Symtab * _symtab;
     SymtabAPI::Region * _region;
     std::map<Address, Address> knownData;
+    // sorted offsets of unapplied relocations in this region; only
+    // populated for unlinked object files
+    std::vector<Address> _reloc_offsets;
+
+    void initRelocOffsets();
  public:
     SymtabCodeRegion(SymtabAPI::Symtab *, SymtabAPI::Region *);
     SymtabCodeRegion(SymtabAPI::Symtab *, SymtabAPI::Region *,
@@ -267,6 +282,8 @@ class DYNINST_EXPORT SymtabCodeRegion : public CodeRegion {
     /** interval **/
     Address low() const { return offset(); }
     Address high() const { return offset() + length(); }
+
+    bool hasUnresolvedRelocs(Address lo, Address hi) const;
 
     SymtabAPI::Region * symRegion() const { return _region; }
 };
--- a/parseAPI/src/IA_IAPI.C
+++ b/parseAPI/src/IA_IAPI.C
@@ -434,6 +434,14 @@ bool IA_IAPI::isDynamicCall() const
         bool success;
         boost::tie(success, addr) = getCFT();
         if (!success) {
+            if (cftHasUnresolvedReloc()) {
+                // A direct call whose displacement awaits link-time
+                // relocation (unlinked object file): the callee is
+                // statically unknown, but the call is not through a
+                // register or memory operand.
+                parsing_printf("... Call 0x%lx has an unresolved relocation, not dynamic\n", current);
+                return false;
+            }
             parsing_printf("... Call 0x%lx is indirect\n", current);
             return true;
         }
@@ -486,6 +494,14 @@ bool IA_IAPI::isIndirectJump() const {
     Address target;
     boost::tie(valid, target) = getCFT(); 
     if (valid) return false;
+    if (cftHasUnresolvedReloc()) {
+        // Not a jump table: the target carries an unapplied relocation
+        // (unlinked object file), so any bytes an indirect-target
+        // analysis would read are placeholders. Handled as a tail call
+        // to an unknown target in getNewEdges instead.
+        parsing_printf("... jump with unresolved relocation at 0x%lx, not treating as indirect\n", current);
+        return false;
+    }
     parsing_printf("... indirect jump at 0x%lx, delay parsing it\n", current);
     return true;
 }
@@ -612,7 +628,14 @@ void IA_IAPI::getNewEdges(std::vector<std::pair< Address, EdgeTypeEnum> >& outEd
     {
         if(ci.allowsFallThrough())
         {
-            outEdges.push_back(std::make_pair(getCFT().second,
+            bool takenValid;
+            Address takenTarget;
+            boost::tie(takenValid, takenTarget) = getCFT();
+            // An undefined taken-target (e.g. an unapplied relocation on
+            // the displacement in an unlinked object file) must go to the
+            // sink, not to whatever placeholder constant was encoded.
+            outEdges.push_back(std::make_pair(
+                        takenValid ? takenTarget : Dyninst::ADDRESS_INVALID,
                         COND_TAKEN));
             outEdges.push_back(std::make_pair(getNextAddr(), COND_NOT_TAKEN));
             return;
@@ -646,6 +669,19 @@ void IA_IAPI::getNewEdges(std::vector<std::pair< Address, EdgeTypeEnum> >& outEd
                         FILE__, __LINE__, target);
                 outEdges.push_back(std::make_pair(target, DIRECT));
             }
+            return;
+        }
+        else if(cftHasUnresolvedReloc())
+        {
+            // Branch whose target bytes carry an unapplied relocation
+            // (unlinked object file): the real target is another section
+            // or an external symbol and is unknowable until link time.
+            // Treat it as a tail call to an unknown target.
+            parsing_printf("%s[%d]: tail call to unresolved relocation target at 0x%lx\n",
+                    FILE__, __LINE__, current);
+            outEdges.push_back(std::make_pair(Dyninst::ADDRESS_INVALID, INDIRECT));
+            tailCalls[INDIRECT] = true;
+            tailCalls[DIRECT] = true;
             return;
         }
         else if(isReturn(context, currBlk))
@@ -856,6 +892,12 @@ std::pair<bool, Address> IA_IAPI::getFallthrough() const
     return make_pair(true, curInsnIter->first + curInsnIter->second.size());
 }
 
+bool IA_IAPI::cftHasUnresolvedReloc() const
+{
+    if(!curInsn().getControlFlowTarget()) return false;
+    return _cr->hasUnresolvedRelocs(current, current + curInsn().size());
+}
+
 std::pair<bool, Address> IA_IAPI::getCFT() const
 {
     if(validCFT) return cachedCFT;
@@ -868,7 +910,20 @@ std::pair<bool, Address> IA_IAPI::getCFT() const
 
     Result actualTarget = callTarget->eval();
 
-    if(actualTarget.defined)
+    if(actualTarget.defined
+       && _cr->hasUnresolvedRelocs(current, current + curInsn().size()))
+    {
+        // The target bytes are placeholders awaiting link-time relocation
+        // (e.g. a call or tail call in a .o naming another section or an
+        // external symbol); the placeholder-derived constant is
+        // meaningless — typically the next instruction or one past the
+        // end of the section.
+        cachedCFT = std::make_pair(false, 0);
+        parsing_printf("REJECTED: CF target of '%s' at 0x%lx carries an unapplied relocation (placeholder target 0x%lx)\n",
+                curInsn().format().c_str(), current,
+                actualTarget.convert<Address>());
+    }
+    else if(actualTarget.defined)
     {
         cachedCFT = std::make_pair(true, actualTarget.convert<Address>());
         parsing_printf("SUCCESS (CFT=0x%lx)\n", cachedCFT.second);
@@ -898,8 +953,10 @@ bool IA_IAPI::isRelocatable(InstrumentableLevel lvl) const
         {
             bool valid; Address addr;
             boost::tie(valid, addr) = getCFT();
-            assert(valid);
-            if(!_isrc->isValidAddress(addr))
+            // An undefined CFT here means the target is unknown (e.g. an
+            // unapplied relocation in an unlinked object file), not that
+            // it is invalid; only validate targets we actually know.
+            if(valid && !_isrc->isValidAddress(addr))
             {
                 parsing_printf("... Call to 0x%lx is invalid (outside code or data)\n",
                         addr);
--- a/parseAPI/src/IA_IAPI.h
+++ b/parseAPI/src/IA_IAPI.h
@@ -110,6 +110,11 @@ class IA_IAPI : public InstructionAdapter {
                                 unsigned int,
                                 const std::set<Address> &) const = 0;
         virtual std::pair<bool, Address> getCFT() const;
+        // True if the current instruction has a control-flow target whose
+        // bytes are covered by an unapplied relocation (unlinked object
+        // files only): the encoded target is a placeholder, the real one
+        // is unknowable until link time.
+        bool cftHasUnresolvedReloc() const;
         virtual std::pair<bool, Address> resolveDynamicCallTarget(Dyninst::ParseAPI::Function *,
                                                                   Dyninst::ParseAPI::Block *) const
         {
--- a/parseAPI/src/Parser.C
+++ b/parseAPI/src/Parser.C
@@ -1495,6 +1495,15 @@ Parser::parse_frame_one_iteration(ParseFrame &frame, bool recursive) {
             } else {
                 ct = _parse_data->findFunc(frame.codereg, work->target());
             }
+            if (!ct) {
+                // The target is not code in this frame's region and no
+                // function exists there: e.g. a bogus cross-region target
+                // in a binary with overlapping regions. Skip the work
+                // element rather than dereference a NULL function.
+                parsing_printf("[%s] call target %lx has no function in frame's region; skipping work element\n",
+                        FILE__, work->target());
+                continue;
+            }
             bool frame_not_created = (frame_status(ct->region(),ct->addr())==ParseFrame::BAD_LOOKUP);
             parsing_printf("\tframe %lx, not created %d\n", ct->addr(), frame_not_created);
 
--- a/parseAPI/src/SymtabCodeSource.C
+++ b/parseAPI/src/SymtabCodeSource.C
@@ -90,6 +90,7 @@ SymtabCodeRegion::SymtabCodeRegion(
             knownData[(*sit)->getOffset()] = (*sit)->getOffset() + (*sit)->getSize();
             parsing_printf("Add known data range [%lx, %lx) from symbol %s\n", (*sit)->getOffset(), (*sit)->getOffset() + (*sit)->getSize(), (*sit)->getMangledName().c_str());
         }
+    initRelocOffsets();
 }
 
 
@@ -106,6 +107,31 @@ SymtabCodeRegion::SymtabCodeRegion(
             knownData[(*sit)->getOffset()] = (*sit)->getOffset() + (*sit)->getSize();
             parsing_printf("Add known data range [%lx, %lx) from symbol %s\n", (*sit)->getOffset(), (*sit)->getOffset() + (*sit)->getSize(), (*sit)->getMangledName().c_str());
         }
+    initRelocOffsets();
+}
+
+void
+SymtabCodeRegion::initRelocOffsets()
+{
+    // Only unlinked object files (e.g. archive members) contain
+    // relocations that have not yet been applied to the section bytes;
+    // in linked binaries (including those produced with --emit-relocs)
+    // the bytes already hold their final values.
+    if(!_symtab->isUnlinkedObjectFile()) return;
+
+    // In ET_REL files r_offset is relative to the start of the section
+    // the relocation applies to; rebase it into this region's address
+    // space (in practice unlinked sections live at 0, so low() == 0).
+    for(auto const& rel : _region->getRelocations())
+        _reloc_offsets.push_back(low() + rel.rel_addr());
+    std::sort(_reloc_offsets.begin(), _reloc_offsets.end());
+}
+
+bool
+SymtabCodeRegion::hasUnresolvedRelocs(Address lo, Address hi) const
+{
+    auto it = std::lower_bound(_reloc_offsets.begin(), _reloc_offsets.end(), lo);
+    return it != _reloc_offsets.end() && *it < hi;
 }
 
 void


hainest:
I think there are two separate PRs here. The first is the check in `Parser::parse_frame_one_iteration` that a valid function was found. This could occur for many reasons- the handling of unrelocated targets is just one. The second would then be the unrelocated target fixes. For the second one, it would be nice to have a test to go along with it. I'm fine with putting the loadmsgcat.o in the test_binaries repo and then run the mutator shown here on it.

bbiiggppiigg:
> I think there are two separate PRs here. The first is the check in `Parser::parse_frame_one_iteration` that a valid function was found. This could occur for many reasons- the handling of unrelocated targets is just one. The second would then be the unrelocated target fixes. For the second one, it would be nice to have a test to go along with it. I'm fine with putting the loadmsgcat.o in the test_binaries repo and then run the mutator shown here on it.

So basically extracting the first commit out as a PR that should be merged before this one?
For the test binary thing, when you say run the mutator you mean adding a integration test that does instrumentation?

hainest:
> > I think there are two separate PRs here. The first is the check in `Parser::parse_frame_one_iteration` that a valid function was found. This could occur for many reasons- the handling of unrelocated targets is just one. The second would then be the unrelocated target fixes. For the second one, it would be nice to have a test to go along with it. I'm fine with putting the loadmsgcat.o in the test_binaries repo and then run the mutator shown here on it.
> 
> So basically extracting the first commit out as a PR that should be merged before this one?

Yeah, that would probably be the best way.

> For the test binary thing, when you say run the mutator you mean adding a integration test that does instrumentation?

I was thinking of just adding the example code in the PR description as the "mutator" (even though it just parses) and then grep the output for the `REJECTED: CF target ` as in the description.

bbiiggppiigg:
## Validation summary (head c9ac3b921, rebased on master 4547a0bf2 = #2386)

Every comparison below is branch vs master **at the same tip** (master pinned at the branch's merge-base), built and run with the same recipe on each machine.

### In-tree ctest (`-DDYNINST_ENABLE_TESTS=ALL`)

| host | arch | master | branch |
|---|---|---|---|
| mahir | x86_64 | 133/133 | 134/134 |
| coriander | x86_64 | 133/133 | 134/134 |
| cyclops | ppc64le | 108/108 | 109/109 |
| zeroah | aarch64 | 132/132 | 133/133 |

The +1 is the new `parseAPI_unrelocated_cft_test`, passing on all four hosts (including parsing the x86_64 `.o` on the ppc64le and aarch64 hosts). All builds warning-free.

### External dyninst/testsuite (full `-all` matrix)

| host | arch | method | rows (master / branch) | non-PASSED | verdict |
|---|---|---|---|---|---|
| coriander | x86_64 | grouped `runTests -all` | 1172 / 1172 | none on either side | identical |
| zeroah | aarch64 | grouped `runTests -all` | 1233 / 1233 | `test_pt_ls` 64-bit create FAILED on **both** (known flake) | identical |
| cyclops | ppc64le | per-test isolated (`test_driver -test T -all`, 300 s timeout) | 1045 / 1045 | 832P / 14F / 19 HUNG on **both** — the pre-existing ppc64le `pc_*`/`test_thread_*` cluster, `pc_tls`, `test_thread_3/8`, `test_pt_ls` | identical |
| mahir | x86_64 | per-test isolated (same) | 1335 / 1335 | 1272P / 16F / 39 HUNG / 6 driver-died on **both** — pre-existing platform noise | identical |

Per host: every run completed (per-run-mode row counts match), and a verdict-class diff of the two result tables shows **0 regressions (master PASSED → branch not PASSED), 0 fixes, 0 class changes, 0 missing rows**. Every non-passing row on the branch has the same verdict on master.

Method note: on mahir and cyclops a grouped `-all` cannot complete (one hanging group trips runTests' fixed 600 s limit and the whole run stops at a moving point), so those two were run per-test isolated with a timeout, which bounds each hanger as HUNG and gives complete, diffable coverage. The hang sets are identical on master and branch.

Testsuite @ 8d868d76 everywhere.

bbiiggppiigg:
@hainest @kupsch validated and ready for review.

kupsch (commented):
Here's what I got out of claude:

Open, worth fixing before merge:

1. parseAPI/src/IA_IAPI.C:899 — cftHasUnresolvedReloc() can't distinguish direct from memory-indirect.  Confirmed against the installed library: .text relocations at 0x12, 0x22, 0xc2 fall inside jmpq *0x0(,%rdi,8), callq *0x0(%rip), and jmpq *0x0(%rip) respectively, so the helper returns true for all three. Clearest harm is isDynamicCall() returning false for a genuine memory-indirect call. Tracing it into ParserDetails.C:492-516, the CFG shape survives — has_unres ends up true either way and resolvable_edge is false either way — but ProcessCallInsn is handed dynamic_call=false, so every ParseCallback consumer is told a GOT-indirect call is a static call, and IA_x86::isIATcall() (early-returns on !isDynamicCall()) can never fire in an object file. The Dereference gate I sketched earlier fixes it.

Open, minor:

2. parseAPI/h/CodeSource.h:89 — the header promises "any byte in [lo,hi)", the implementation stores only r_offset start points and ignores field width. Not reachable from the diff's own call sites; it's a public virtual selling a guarantee it doesn't provide. 
3. parseAPI/src/IA_IAPI.C:632 — conditional branches with an unapplied relocation take the allowsFallThrough() early return, so tailCalls[] and parsedJumpTable are never set, leaving getInstLevel()'s assert(0) (InstructionAdapter.C:169) reachable. Verified latent: grep across the tree finds no caller of getInstLevel other than IA_IAPI::getInstLevel delegating to the base. It's an exported virtual, so external consumers could hit it. getReturnStatus() on the same path returns UNSET benignly.
4. tests/regression/parseAPI/UnappliedRelocations/CMakeLists.txt:4 — add_executable outside the DYNINST_ENABLE_EXTERNAL_BINARY_TESTS guard, unlike the sibling IndirectControlFlowAnalysis test.


● The bug

  parseAPI/src/IA_IAPI.C:899:

```
bool IA_IAPI::cftHasUnresolvedReloc() const
{
    if(!curInsn().getControlFlowTarget()) return false;
    return _cr->hasUnresolvedRelocs(current, current + curInsn().size());
} 
```

  The predicate this is meant to express is "this is a direct branch or call
  whose displacement is a placeholder awaiting link-time relocation" — the jmp
  free with R_X86_64_PLT32 in libc.a(loadmsgcat.o) that motivated the commit.
  What it actually tests is "this instruction has a control-flow target, and
  somewhere in its bytes there is a relocation."

  Those come apart on memory-indirect branches and calls.
  getControlFlowTarget() is non-null for them too — it returns the Dereference
  — and their relocation lives on the memory operand's displacement, inside the
  same [current, current+size) range.

  Why every call site reaches it

  All three sites are guarded by isCall() or isBranch(), and each is entered
  only after getCFT() failed. But getCFT() fails for two unrelated reasons:

  1. the new rejection in getCFT() — a direct branch whose defined target was thrown out because of a relocation (the intended case);
  2. a Dereference CFT, which has never evaluated, patch or no patch.

  cftHasUnresolvedReloc() is being used to distinguish case 1 from case 2, and
  it can't: in case 2 the relocation is present too, so the test passes for the
  wrong reason.

  Evidence

  gcc -O2 -fno-pic -fno-plt -c produces both shapes. Confirmed at three levels
  against the patched install/lib64/libparseAPI.so.13.0.0:

  The region really does report these offsets — querying
  SymtabCodeRegion::hasUnresolvedRelocs directly gives 0x12, 0x22, 0xc2, which
  fall inside jmpq *0x0(,%rdi,8) at [0xf,0x16), callq *0x0(%rip) at
  [0x20,0x26), and jmpq *0x0(%rip) at [0xc0,0xc6).

  The wrong branches are taken — DYNINST_DEBUG_PARSING=1 on the parse:

```
... Call 0x4 has an unresolved relocation, not dynamic                <- IA_IAPI.C:441
... jump with unresolved relocation at 0xf, not treating as indirect  <- IA_IAPI.C:501
IA_IAPI.C[685]: tail call to unresolved relocation target at 0xf      <- the new arm
```

  The damage lands — for the call: edge target ffffffffffffffff, ...,
  dynamic_call 0, so ProcessCallInsn and every ParseCallback consumer are told
  a call through the GOT is a static call, and IA_x86::isIATcall() (which
  early-returns on !isDynamicCall()) can never fire in an object file. For the
  jump: no jump table candidate line anywhere in the log — parseJumpTable()
  never ran — and the block gets one sink edge, leaving 0x16–0xbf of sw (the
  entire switch body) unreachable.

  Scope

  initRelocOffsets() early-returns unless isUnlinkedObjectFile(), so
  executables and shared libraries are untouched. Within ET_REL, PIC is also
  safe: -fpic emits jump tables as lea .L4(%rip),%rdx; movslq
  (%rdx,%rax,4),%rax; add %rdx,%rax; jmp *%rax, putting the relocation on the
  lea and leaving the jmp clean. That's likely why validating on libc.a didn't
  surface it. The exposed configuration is -fno-pic ET_REL — how kernel modules
  are built — and -fno-plt for the call case.

  Fix

  Gate on the CF target being an immediate/PC-relative displacement rather than
  a memory read:

```
bool IA_IAPI::cftHasUnresolvedReloc() const
{
    Expression::Ptr cft = curInsn().getControlFlowTarget();
    if(!cft) return false;
    // A memory-indirect branch/call reads its target at run time; any
    // relocation in the instruction applies to the memory operand, not
    // to a control-flow displacement.
    if(boost::dynamic_pointer_cast<Dereference>(cft)) return false;
    return _cr->hasUnresolvedRelocs(current, current + curInsn().size());
} 
```

  Register-indirect (jmp *%rax) needs no special handling — no displacement, so
  no relocation to match. A tighter variant would check that the relocated
  offset falls within the displacement field itself, but the Dereference test
  is sufficient for the cases above and much simpler.

---
  Test objects

  t.c — nine cases forces GCC to a jump table; f tail-calls free:

```
extern void free(void*);
extern int a(void),b(void),c(void),d(void),e(void);
int sw(int x){ switch(x){case 0:return a();case 1:return b();case 2:return c();
  case 3:return d();case 4:return e();case 5:return a()+1;case 6:return b()+2;
  case 7:return c()+3;case 8:return d()+4;default:return 0;} }
void f(void*p){ free(p); }
```

  u.c — the return 1 prevents a tail call, so the GOT call is a real callq:

```
extern void free(void*);
int h(void*p){ free(p); return 1; }
```

build and objdump:
```
gcc -O2 -fno-pic -fno-plt -c $SP/t.c -o $SP/t_nopic.o
gcc -O2 -fno-pic -fno-plt -c $SP/u.c -o $SP/u.o
objdump -dr $SP/t_nopic.o
objdump -dr $SP/u.o
```

  -fno-pic makes the switch dispatch jmpq *0x0(,%rdi,8) with R_X86_64_32S on
  the displacement; -fno-plt makes the external calls callq *0x0(%rip) with
  R_X86_64_GOTPCRELX.

  Test A — does the region report relocations covering those instructions?

  q.C:

```
#include "CodeSource.h"
#include <cstdio>
using namespace Dyninst; using namespace Dyninst::ParseAPI;
int main(int argc,char**argv){
  fprintf(stderr,"constructing...\n");
  SymtabCodeSource *sts=new SymtabCodeSource(argv[1]);
  fprintf(stderr,"constructed\n");
  for(auto r : sts->regions()){
    printf("region [0x%lx,0x%lx)\n",r->low(),r->high());
    for(Address a=r->low();a<r->high();a++)
      if(r->hasUnresolvedRelocs(a,a+1)) printf("  reloc-start-offset
0x%lx\n",a);
  }
  return 0;
}
```


  Test B — which branch does the parser actually take?

  r.C:

```
#include "CodeObject.h"
#include "CodeSource.h"
#include <cstdio>
using namespace Dyninst; using namespace Dyninst::ParseAPI;
int main(int argc,char**argv){
  SymtabCodeSource *sts=new SymtabCodeSource(argv[1]);
  CodeObject *co=new CodeObject(sts);
  co->parse();
  fprintf(stderr,"PARSE DONE\n");
  _exit(0);   // skip teardown entirely
}
```

The evidence is the parsing_printf strings the patch itself added, so the log
lines map one-to-one onto the new branches.

Build and run

Same compile line for both drivers (-lboost_system is needed for
boost::system::generic_category()):

```
g++ -std=c++11 -o $SP/q $SP/q.C -I install/include -L install/lib64 \
    -lparseAPI -lsymtabAPI -linstructionAPI -lcommon -lboost_system \
    -Wl,-rpath,$PWD/install/lib64

g++ -std=c++11 -o $SP/r $SP/r.C -I install/include -L install/lib64 \
    -lparseAPI -lsymtabAPI -linstructionAPI -lcommon -lboost_system \
    -Wl,-rpath,$PWD/install/lib64

# A: relocation offsets seen by the code region
$SP/q $SP/t_nopic.o

# B: which branches the parser takes
DYNINST_DEBUG_PARSING=1 $SP/r $SP/t_nopic.o 2>&1 \
  | grep -E "unresolved relocation|not treating as indirect|jump table
candidate|Call 0x"

DYNINST_DEBUG_PARSING=1 $SP/r $SP/u.o 2>&1 \
  | grep -E "unresolved relocation|not dynamic|dynamic_call|edge target"
```

What came back

**A, on t_nopic.o —** .text region [0x0,0xc6) reports 0x12 0x22 0x32 0x46 0x56
0x66 0x76 0x86 0x92 0xa2 0xc2, matching objdump -r exactly. 0x12 is inside
jmpq *0x0(,%rdi,8) at [0xf,0x16); 0x22 inside callq *0x0(%rip) at
[0x20,0x26); 0xc2 inside jmpq *0x0(%rip) at [0xc0,0xc6).

**B, on t_nopic.o:**

... jump with unresolved relocation at 0xf, not treating as indirect
IA_IAPI.C[685]: tail call to unresolved relocation target at 0xf
Setting edge (0x9/0xffffffffffffffff) to interproc (tail call)

with no jump table candidate line and no Call 0x lines at all — the switch
body from 0x16 on is unreachable once the table is abandoned, which is also
why this object could not exercise isDynamicCall().

**B, on u.o:**

... Call 0x4 has an unresolved relocation, not dynamic
   edge target ffffffffffffffff, is_code 0, is_plt 0, edge type 0,
dynamic_call 0

dynamic_call 0 is from ParserDetails.C:503, i.e. what ProcessCallInsn and the
ParseCallback consumers receive for a call through the GOT.

One caveat worth repeating: q.C's byte-at-a-time query only finds relocation
start offsets (that's finding 3's limitation, used here on purpose to
enumerate r_offset values), and it aborts with corrupted size vs. prev_size
after printing — I didn't chase that, and it's most likely my program rather
than Dyninst.

bbiiggppiigg:
no regression on coriander zeroah and cyclops

tests/regression/parseAPI/UnappliedRelocations/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  # handling: jump-table analysis attempted, calls classified dynamic.
+  set(_tname parseAPI_${_targ}_indirect_test)
+  add_test(NAME ${_tname} COMMAND ${_targ} "${_fpath}/indirect_nopic.o")
+  set_tests_properties(
+    ${_tname}
+    PROPERTIES LABELS
+               "regression"
+               ENVIRONMENT
+               "DYNINST_DEBUG_PARSING=1;OMP_NUM_THREADS=1"
+               PASS_REGULAR_EXPRESSION
```
kupsch, line 42: Just so you're aware, the `PASS_REGULAR_EXPRESSION` property disables checking the return code to determine failure, and there does not seem to be a PROPERTY to say file is return code is non-0. To check the return status and a regex is complicated, so I'd say this is OK as is.

kupsch (approved):
Looks good now.

PR #2353, opened 2026-08-01
GitHub CI: fix fromJson expressions missing output references

bbiiggppiigg:
In build-fedora, the step name read fromJson(steps.compiler-name.outputs) instead of the 'value' output that the step actually sets, as the cmake flags below already do. 

Also add a missing closing parenthesis in a workflow README example.


bbiiggppiigg:
Fix real CI warning and typo in README

hainest (approved):

hainest:
I'm curious how I never noticed that...

kupsch (approved):

bbiiggppiigg:
> I'm curious how I never noticed that...

It only appears as an annotation / warning.
I wouldn’t have noticed it if it weren’t it happened on a failed CI job

PR #2361, opened 2026-08-05
GitHubCI: reduce repository permissions in workflows

hainest:
The OpenSSF scorecard Token-Permissions check requires minimizing permissions on workflows by

1) setting a global permission for the workflow
2) using job-specific permission overrides

### DIFF
--- a/.github/workflows/binary-tests.yaml
+++ b/.github/workflows/binary-tests.yaml
@@ -25,6 +25,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
 
   # Don't run the check if the PR is a draft
--- a/.github/workflows/build-opts.yaml
+++ b/.github/workflows/build-opts.yaml
@@ -9,6 +9,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
 
   # Don't run the check if the PR is a draft
--- a/.github/workflows/cmake-formatting.yaml
+++ b/.github/workflows/cmake-formatting.yaml
@@ -13,6 +13,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   cmake-formatting:
     runs-on: ubuntu-latest
--- a/.github/workflows/codegen-cross-build.yaml
+++ b/.github/workflows/codegen-cross-build.yaml
@@ -19,6 +19,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
 
   # Don't run the check if the PR is a draft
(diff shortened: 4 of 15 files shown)


bbiiggppiigg:
I really don't know much about these stuff but since AI has a strong opinion I'll leave them here and see if any of these are useful to you. Feel free to ignore them if they are just BS.

=============


## 1. `release-containers.yaml` under `permissions: {}`

It flagged `get-latest-release-version` (`:21`): no job-level override, so it inherits `{}` from `:6` and runs with `contents: none`, while the step at `:27` (`pozetroninc/github-action-get-latest-release`) authenticates with `secrets.github_token` and hits `GET /repos/dyninst/dyninst/releases` — which the REST docs apparently list as requiring Contents: read. It couldn't determine whether that endpoint still answers for a public repo when the token has `contents: none`.

Its reasoning for why this one matters more than the rest: everything else here fires on the next source PR, master push, or a weekly/monthly schedule, so a mistake surfaces fast, whereas `release-containers` is `workflow_dispatch`-only and wouldn't run until we next cut a release.

Does that concern make sense, or does a public-repo read work fine regardless of the token's scopes?

## 2. Why `actions: read` on the other 12?

It was explicit that this isn't a score issue — read-level top-level scopes aren't penalized, so `actions: read` gets the same credit as `{}` or `contents: read`. Its question was just whether anything consumes it. It searched `.github/` and found no `actions/download-artifact`, no `gh run`/`gh api`, no `actions/github-script`, and no Actions REST call, and claimed every job either inherits the top-level block while only doing `actions/checkout` plus a shell or local composite step, or declares its own `permissions: packages: read|write` — which it says replaces the top-level block rather than merging with it, so `actions: read` wouldn't reach those jobs at all.

Its suggestion was `contents: read` as the top-level value instead, on the grounds that `checks.md` names it as the remediation, it matches what the checkout-only jobs (`get-oses`, `get-compilers`, `get-build-types`, `dependency-version`) actually do, and it would make item 1 moot. It also noted `{}` works for checkout on a public repo, citing the `cmake-formatting` dispatch you ran.

Is there a consumer of `actions: read` it's missing, or a reason to prefer it here?

## 3. `dev-containers.yaml:11` and its caller

This one doesn't rest on the AI's reading — `a69673602` looks like direct evidence for it. `build-deploy-dev` needed `actions: read` added by hand even though `refresh-containers.yaml:9` already granted it at the workflow level, which means the top-level block doesn't reach a `uses:` job and the job has to cover whatever the called workflow requests. So the top-level `actions: read` at `dev-containers.yaml:11` is what forces `:108`. Its suggestion was `{}` in `dev-containers.yaml` so the caller line could go back to just `packages: write`, since `get-oses` there is checkout-only and the `dev-containers` job already declares `packages: write` at `:34`.

It also noted that `purge` (`:31`), `build-base` (`:62`), and the callee's own `dev-containers` job all end up with `actions: none` via their overrides while `build-deploy-dev` alone gets `read`, and read that asymmetry as possibly unintentional.

Is that reading of the reusable-workflow rule right, and is the asymmetry deliberate?

## 4. Commit message

It suggested the squashed message record two things: that a job-level `permissions:` block replaces rather than merges with the top-level one, and that a `uses:` job must cover what the called workflow requests — on the theory that the scorecard rules already in the message are easy to look up but those two aren't. Same for a one-line comment at `refresh-containers.yaml:108` if the `actions: read` there stays.

## Nit

`refresh-containers.yaml:107-108` lists `packages` then `actions`; `purge-workflows.yaml` is alphabetical. Ignore if you don't care.

bbiiggppiigg (approved):

kupsch (approved):

hainest:
>I really don't know much about these stuff but since AI has a strong opinion I'll leave them here and see if any of these are useful to you. Feel free to ignore them if they are just BS.

The examples I found online actually suggested setting the top-level permissions to `read-all`. I felt like that was too permissive, so just used allowed reading of actions. I can experiment with a top-level `permissions: {}` and see what happens.

PR #2363, opened 2026-08-06
[dyninstAPI] Have DelayedParsing delay CFG parse until needed

kcossett-amd:
**Problem**: When certain binaries that use shared libraries with many functions are instrumented, `processCreate` takes a substantial amount of time to complete even if we later exclude those large libraries from instrumentation. 

Consider `transpose` (https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler-systems/examples/transpose) that was built to use `libLLVM.so` and `libclang-cpp.so` at runtime. In our tool, we by default exclude these from `runtime-instrumentation` to keep instrumentation overhead manageable (the large amount of functions will make fetching the functions from the object take much longer): 
```
[rocprof-sys][exe] [filter] skipping shared lib 'libclang-cpp.so.23.0git' (78832 functions > --max-library-functions=20000)
[rocprof-sys][exe] [filter] skipping shared lib 'libLLVM.so.23.0git' (95293 functions > --max-library-functions=20000)
[rocprof-sys][exe] Filtered objects: 20 of 22 included (2 excluded) (0.000 sec, 1.280 MB)
```
The function count is a lower bound obtained through the SymtabAPI. 

However, the `processCreate` call (done as the first step) reports:
```
[rocprof-sys][exe] processCreate took 74.853 sec
```

`rocprof-sys-intrument -- ./transpose` takes `84` seconds, of which `74` are spent in `processCreate`. Investigation shows a significant chunk of time is spent processing the two libraries that we later exclude from instrumentation:
```
# Custom print statements used
[dyninst][cfg-parse] code object                                         seconds
[dyninst][cfg-parse] ld-linux-x86-64.so.2                                  0.058
[dyninst][cfg-parse] libLLVM.so.23.0git                                   38.965
[dyninst][cfg-parse] libamdhip64.so.7.1.70101                              0.021
[dyninst][cfg-parse] libc.so.6                                             1.240
[dyninst][cfg-parse] libclang-cpp.so.23.0git                              29.128
[dyninst][cfg-parse] libdl.so.2                                            0.000
# ... Other libraries that take <1 second ...
[dyninst][cfg-parse] TOTAL                                                70.837
```

I will note that I had `OMP_NUM_THREADS=2`, and I am aware this will affect performance, but the point remains that if I am going to exclude these libraries later, there is no point on spending the time on parsing the CFG.

**Solution**: The goal here is to implement the `delayedParsing` flag and have it defer `CFG` until it actually needs it

Changes (48334cdbd9e6bd842cfc51a1af6eaceea97f2c5b)
- `image.C`
	- `image::image` respects `delayedParsing` flag IFF `BPatch_normalMode` and we are not using `ppc64`. It sets `ignoreParse` on the `CodeObject`, so hints are recorded but no CFG is built.
	- Removed `analyzeIfNeeded()` from `getAllVariables` as the list should already be completed once the constructor has run `addSymtabVariables()`, and only `createImageVariables()` adds to it afterwards (same holds for `getCreatedVariables()`)
	- The function to gate analysis is `analyzeIfDeferred`, which wraps `analyzeIfNeeded`. Guarding at the call sites leaves the other 7 pre-existing `analyzeIfNeeded` callers untouched.
	- `image::findFuncVectorByMangled` does use `plt_parse_funcs`, which is filled during CFG construction. To avoid this, parse the stub itself through the new `parsePltStubs`, which inverts `cs_->linkage()` (address -> name, from relocations) once and parses only those addresses.
- `analyzeIfDeferred` checks added to the following:
	- `DynCFGMaker.C`'s `makeFunction`, which requires blocks.
	- `PatchObject::createFuncs`, via a new `virtual ensureParsed()` that `mapped_object` overrides. It has to run before the loop: `co()->funcs()` is ordered, so a mid-walk parse misses whatever it inserts below the cursor.
- Value for `analyzeIfDeferred` read in `mapped_object` and passed to `parseImage`, which keeps `image` free of the `BPatch` singleton.
- `Parser.C`
	- `parse_at` assumed a full parse had already run: it cleared `hint_funcs`/`discover_funcs` and downgraded the state to `COMPLETE`. On a deferred `CodeObject` that strands every unparsed function and makes `Parser::parse()` a permanent no-op. It now parks the seed lists when entering at `UNPARSED` and restores them, state included.
	
With the code changes, I now see:
```
[dyninst][cfg-parse] code object                                         seconds
[dyninst][cfg-parse] libLLVM.so.23.0git                                   39.246
# ... Other libraries that take <1 second.
[dyninst][cfg-parse] TOTAL                                                40.982
```

`libLLVM` is still being fully parsed it seems. Problems comes down to when Dyninst searches for `execve`. It seems Dyninst is after libc's `execve`, but we search by name, so we also match the `execve@plt` stub in every shared lib that imports it, which `libLLVM` has, and `libclang-cpp` does not. Instrumenting a stub is what drags in its library's full CFG.

So we drop every match whose address is a PLT slot in its own object, which leaves the definition. If that leaves nothing, every match was a stub and we keep them all. A static binary has no stubs and pays nothing.

Changes (529234f25a397177004e55f285aded91a4049727)
- `dynProcess.C`
	- `installInstrRequests` keeps only real definitions when a request's name resolves to both a definition and PLT stubs. `findFuncsByAll` matches by name across every object, so one `execve` request fans out to libc's definition plus a stub in each importing library. Each stub is expensive: `AddressSpace::relocate()`'s overlap check calls `Block::getFuncs()` on the stub's blocks, which forces the deferred parse of its entire library, hence `libLLVM.so` still being parsed. 
	- The query is object-wide, so deferral cannot help; the stub has to stay out of the modified set (fallback is to do old behavior).

With this change:
```
[dyninst][cfg-parse] code object                                         seconds
# ... Libraries that take <1 second.
[dyninst][cfg-parse] TOTAL                                                 1.732
```

And I see:
```
[rocprof-sys][exe] processCreate took 3.621 sec
```

Overall, the `ctest` that runs `transpose` with `runtime-instrument` went from `84` to `~13` seconds and passes.
Visualizing the generated traces in perfetto showed no differences that could result from this PR.

### DIFF
--- a/dyninstAPI/src/Relocation/DynCFGMaker.C
+++ b/dyninstAPI/src/Relocation/DynCFGMaker.C
@@ -46,6 +46,10 @@ PatchFunction* DynCFGMaker::makeFunction(Function* f,
                                          PatchObject* obj) {
   Address code_base = obj->codeBase();
   mapped_object* mo = SCAST_MO(obj);
+
+  // func_instance being constructed is unusable without blocks
+  if (!f->entry()) mo->analyzeIfDeferred();
+
   parse_func* img_func = SCAST_PF(f);
   if (!img_func) return NULL;
   assert(img_func->getSymtabFunction());
--- a/dyninstAPI/src/dynproc/dynProcess.C
+++ b/dyninstAPI/src/dynproc/dynProcess.C
@@ -1701,6 +1701,30 @@ void PCProcess::installInstrRequests(const std::vector<dapi::instMapping*> &requ
                         FILE__, __LINE__, matchingFuncs.size(), req->func.c_str());
         }
 
+        // Matches in libraries that import the symbol are PLT stubs that
+        // jump to the definition, so instrumenting the definition covers them.
+        // Dropping them also avoids relocate()'s overlap check, whose
+        // Block::getFuncs() forces a full CFG parse of the stub's library.
+        // If nothing is left, every match was a stub and we keep them all,
+        // since then only the stub can carry the probe. Static binaries have
+        // no stubs and are unaffected.
+        if (!matchingFuncs.empty()) {
+            std::vector<func_instance *> definitions;
+            for (func_instance *func : matchingFuncs) {
+                if (func && func->ifunc() && !func->ifunc()->isPLTFunction())
+                    definitions.push_back(func);
+            }
+
+            if (!definitions.empty() && definitions.size() < matchingFuncs.size()) {
+                inst_printf("%s[%d]: %lu of %lu matches for %s are PLT stubs, "
+                            "instrumenting the definition(s) only\n",
+                            FILE__, __LINE__,
+                            matchingFuncs.size() - definitions.size(),
+                            matchingFuncs.size(), req->func.c_str());
+                matchingFuncs.swap(definitions);
+            }
+        }
+
         for (unsigned funcIter = 0; funcIter < matchingFuncs.size(); funcIter++) {
            func_instance *func = matchingFuncs[funcIter];
            if (!func) {
--- a/dyninstAPI/src/image.C
+++ b/dyninstAPI/src/image.C
@@ -1026,7 +1026,6 @@ image::getAllFunctions()
 
 const std::vector<image_variable*> &image::getAllVariables()
 {
-    analyzeIfNeeded();
     return everyUniqueVariable;
 }
 
@@ -1034,7 +1033,6 @@ const std::vector<image_variable*> &image::getExportedVariables() const { return
 
 const std::vector<image_variable*> &image::getCreatedVariables()
 {
-  analyzeIfNeeded();
   return createdVariables;
 }
 
@@ -1103,7 +1101,8 @@ void image::findModByAddr (const Symbol *lookUp, vector<Symbol *> &mods,
 
 image *image::parseImage(fileDescriptor &desc, 
                          BPatch_hybridMode mode, 
-                         bool parseGaps)
+                         bool parseGaps,
+                         bool delayedParse)
 {
   /*
    * Check to see if we have parsed this image before. We will
@@ -1135,7 +1134,7 @@ image *image::parseImage(fileDescriptor &desc,
 #endif
 
   startup_printf("%s[%d]:  about to create image\n", FILE__, __LINE__);
-  image *ret = new image(desc, err, mode, parseGaps); 
+  image *ret = new image(desc, err, mode, parseGaps, delayedParse); 
   if(err) {
     return nullptr;
   }
@@ -1222,6 +1221,12 @@ void image::analyzeIfNeeded() {
 	  clearNewBlocks();
   }
 }
+
+void image::analyzeIfDeferred() {
+  if (deferredParse_)
+      analyzeIfNeeded();
+}
+  
 
 static bool CheckForPowerPreamble(parse_block* entryBlock, Address &tocBase) {
     ParseAPI::Block::Insns insns;
@@ -1328,7 +1333,8 @@ void image::analyzeImage() {
 image::image(fileDescriptor &desc, 
              bool &err, 
              BPatch_hybridMode mode, 
-             bool parseGaps) :
+             bool parseGaps,
+             bool delayedParse) :
    desc_(desc),
    imageOffset_(0),
    imageLen_(0),
@@ -1350,9 +1356,11 @@ image::image(fileDescriptor &desc,
    trackNewBlocks_(false),
    refCount(1),
    parseState_(unparsed),
+   deferredParse_(false),
    parseGaps_(parseGaps),
    mode_(mode),
-   arch(Dyninst::Arch_none)
+   arch(Dyninst::Arch_none),
+   pltStubAddrsInitialized_(false)
 {
 #if defined(os_linux) || defined(os_freebsd)
    string const& file = desc_.file();
@@ -1458,7 +1466,12 @@ image::image(fileDescriptor &desc,
    // Continue ParseAPI init
    img_fact_ = new Dyninst::DyninstAPI::DynCFGFactory(this);
    parse_cb_ = new Dyninst::DyninstAPI::DynParseCallback(this);
-   obj_ = new CodeObject(cs_,img_fact_,parse_cb_,BPatch_defensiveMode == mode);
+
+   // Only normal mode defers CFG when delayedParsing is enabled
+   // Also skipped when arch is ppc64 as it walks functions are their entry blocks below
+   const bool is_ppc64 = (cs_->getArch() == Arch_ppc64);
+   deferredParse_ = delayedParse && (mode == BPatch_normalMode) && !is_ppc64;
+   obj_ = new CodeObject(cs_,img_fact_,parse_cb_,BPatch_defensiveMode == mode, deferredParse_);
 
      if (obj_->cs()->getArch() == Arch_ppc64) {
         // The PowerPC new ABI typically generate two entries per function.
@@ -1804,6 +1817,27 @@ const std::vector<parse_func *> *image::findFuncVectorByPretty(const std::string
     }
 }
 
+
+// A PLT stub has no symbol of its own, so only parsing creates its parse_func
+bool image::parsePltStubs(const std::string &name)
+{
+    // linkage() is address -> name; we need name -> addresses. Build it once
+    if (!pltStubAddrsInitialized_) {
+        pltStubAddrsInitialized_ = true;
+        for (auto const &entry : cs_->linkage()) {
+            pltStubAddrs_[entry.second].push_back(entry.first);
+        }
+    }
+
+    // The parser names stubs from this same table, so absent means none exists
+    auto iter = pltStubAddrs_.find(name);
+    if (iter == pltStubAddrs_.end())
+        return false;
+
+    codeObject()->parse(iter->second, false);
+    return true;
+}
+
 // Return the vector of functions associated with a mangled name
 // Very well might be more than one! -- multiple static functions in different .o files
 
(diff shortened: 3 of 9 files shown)


bbiiggppiigg:
Below are some opinions of this PR with the help of AI. 

For me peronsally the most important one is probably related to #2336 (or should be merged after 2336), and whether we want to add an API that allows the user to specify which library they want to defer parse, but I think the other opinions are worth reading/discussing.

That being said, I would probably wait for both @hainest and @kupsch' opinions before making any changes so you would not waste time on wrong directions.

====================
1. The PLT-stub filter must not treat ifunc resolvers as definitions (`dynProcess.C`)

For an `STT_GNU_IFUNC` symbol, the non-PLT name match *is the resolver*. Instrumenting the
resolver's entry provides no call coverage: callers reach the chosen implementation through
the PLT/GOT, and the resolver only executes inside `ld.so` during dynamic relocation, before
instrumentation is live. So for an ifunc symbol, the PLT stubs this filter drops are the only
matches that actually see calls. None of the current internal instMapping targets (`execve`,
`fork`, `exit`, ...) are ifuncs in glibc, so nothing breaks today, but this is a silent
coverage hole — and #2336 (which excludes resolvers from relocation entirely, for good
reason) will make the failure deterministic once it lands.

The fix is small because your fallback already does the right thing: exclude matches whose
entry offset carries an `ST_INDIRECT` symbol from the `definitions` bucket. If only stubs and
a resolver match, `definitions` is empty and the existing keep-everything fallback restores
old behavior. #2336 builds exactly this per-object resolver-offset set inside
`CodeMover::addFunctions`; please coordinate with that PR and factor the scan into a shared
helper (e.g. `mapped_object::isIfuncResolver(func_instance*)` with the cache on
`mapped_object`) instead of growing two copies.

2. Scope the PLT-stub filter — it currently changes behavior for all users

The filter in `installInstrRequests` applies unconditionally, flag or no flag. Previously we
instrumented the definition *and* every importing library's stub; now only the definition.
That's semantically equivalent for ordinary symbols, but it's a behavior change in
long-standing code motivated by a feature that is opt-in. Please either gate it on
`delayedParsingOn()` (or on the match residing in a deferred object), or make the case in the
PR description for why it should be unconditional — with item 1 fixed I could be convinced,
but it should be a deliberate decision, not a side effect.

3. Design: per-library control instead of (or in addition to) a global flag

A global "defer everything" flag creates a fragile performance cliff: any stray CFG query
silently pulls a multi-second parse back in — your own second commit exists because the
internal `execve` lookup did exactly that. The tool in your motivating use case already knows
which libraries it will exclude. I'd prefer the policy knob be per-library: either

- `BPatch::addDelayedParsingLibrary(const std::string &pattern)` (documented
  basename/substring matching), or
- a load-time callback, e.g. `BPatch::registerShouldDeferParsingCallback(bool (*)(const char
  *path))`, invoked from `mapped_object::createMappedObject` (which already has
  `desc.file()`).

The mechanism you built is what either shape needs; this is only about how objects are
selected. Keeping `setDelayedParsing(true)` as "defer everything" sugar on top is fine.
Importantly, "skip" should keep meaning what your implementation already does — *defer*, with
the `analyzeIfNeeded()` safety net intact — never "refuse to parse."

A callback/registration API would also resolve the layering wrinkle in this PR:
`mapped_object.C` now includes `BPatch.h` and reads the `BPatch::bpatch` singleton. The PR
description rightly keeps `image` free of that dependency, but this moves it down a layer
rather than removing it. Plumbing the policy through `createMappedObject`'s callers (as
`parseGaps`/`analysisMode` already are) or a registered callback both avoid it.

4. Documentation and tests probably worth doing

- `BPatch.h`'s comments for `setDelayedParsing` / `delayedParsingOn` still describe the old
  (inert) flag; they need to document the new semantics, including the normal-mode-only and
  ppc64 restrictions and the "parse happens lazily on first CFG use" contract.

- Please add a testsuite case: `setDelayedParsing(true)` + `processCreate` + find-by-name
  (including a PLT-only name to exercise `parsePltStubs`) + instrument + run. The `parse_at`
  park/restore in particular has no coverage otherwise.

- Changelog entries for (a) the new delayed-parsing semantics, (b) the `parse_at` behavior
  change (see below), and (c) the stub filter if it stays unconditional.

Discussion / please confirm

- **`parse_at` semantics change is visible to plain ParseAPI users.** Previously,
  `parse_at(...)` followed by `parse()` left the full parse skipped (state ended at
  `COMPLETE`); now the full parse runs. I agree the old behavior was a bug — it strands every
  hint function — but it is a public-API behavior change and should be called out. Also,
  `parse_at` mutates `hint_funcs`/`_parse_state` without holding `parse_mutex` (pre-existing,
  but the swap window is wider now); consider taking the lock or adding a comment stating the
  single-threaded assumption.

- **ABI:** the new `virtual void ensureParsed()` on exported `PatchObject` changes its vtable
  layout. Fine if this targets an ABI-breaking release; flagging so it's a conscious call.
  Also consider declaring it `protected` in the base to match the `mapped_object` override.

- **RISC-V:** `parsePltStubs` inverts only `cs_->linkage()`; `CodeSource` also carries
  `reladyn_linkage()`/`symtab_linkage()` for RISC-V. If stubs there surface through those
  tables, the deferred name-lookup fallback misses them (full parse would have found them).
  Please check or note it as a known limitation.

- **`getAllVariables()`/`getCreatedVariables()`:** removing `analyzeIfNeeded()` looks safe —
  variables are only added in `addSymtabVariables()` and `createImageVariable()` — and it
  applies to all modes, not just deferred. Just confirming that was intentional.

Nits

- `image.C`: comment typo — "as it walks functions **are** their entry blocks below" →
  "and their entry blocks"; trailing whitespace after `analyzeIfDeferred()`'s body.

- `analyzeIfDeferred` reads ambiguously (it *fully analyzes* when parsing was deferred);
  `ensureParsedIfDeferred`/`ensureAnalyzed` would be clearer, especially next to the PatchAPI
  hook you named `ensureParsed`.

- `PatchObject.h`: `ensureParsed()` is added to the public section; `protected` seems more
  appropriate for a subclass hook.

kcossett-amd:
> 3. Design: per-library control instead of (or in addition to) a global flag

Having this would probably be preferable. I was always thinking it would be nice if Dyninst had that. 

Something akin to Dyninst providing a list of objects pre-parse, and then you can exclude some by checking some of their properties (such as the lower-bound number of functions reported by SymtabAPI).

bbiiggppiigg:
@kcossett-amd 

By the way, your PR is failing to build on clang in CI, related to overrides, because we have warning as errors.

The exact flags we compile with  
-Wall;-Wextra;-Wpedantic;-Walloca;-Wcast-align;-Wcast-qual;-Wctor-dtor-privacy;-Wdeprecated-copy-dtor;-Wdouble-promotion;-Wenum-conversion;-Wextra-semi;-Wfloat-equal;-Wformat-signedness;-Wformat=2;-Wframe-larger-than=20480;-Wjump-misses-init;-Wmismatched-tags;-Wmissing-braces;-Wmultichar;-Wnon-virtual-dtor;-Woverloaded-virtual;-Wpointer-arith;-Wrange-loop-construct;-Wshadow;-Wuninitialized;-Wvla;-Wwrite-strings;-Werror
2026-08-07T16:09:30.7835682Z -- Using CXX warning flags: -Wall;-Wextra;-Wpedantic;-Walloca;-Wcast-align;-Wcast-qual;-Wctor-dtor-privacy;-Wdeprecated-copy-dtor;-Wdouble-promotion;-Wenum-conversion;-Wextra-semi;-Wfloat-equal;-Wformat-signedness;-Wformat=2;-Wframe-larger-than=20480;-Wjump-misses-init;-Wmismatched-tags;-Wmis

and the errors reported can be found in the logs
https://productionresultssa10.blob.core.windows.net/actions-results/a057e474-3569-4791-a0bc-3116a7fe93b3/workflow-job-run-d0aa77ec-de43-5370-87a9-a03608918cb8/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-08-07T16%3A34%3A55Z&sig=qvmw6yicRD2Qe68rOtEZll6XAAmhRxvhSA7qPSOVUnA%3D&ske=2026-08-07T19%3A33%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-08-07T15%3A33%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-08-07T16%3A24%3A50Z&sv=2025-11-05

kcossett-amd:
> and the errors reported can be found in the logs https://productionresultssa10.blob.core.windows.net/actions-results/a057e474-3569-4791-a0bc-3116a7fe93b3/workflow-job-run-d0aa77ec-de43-5370-87a9-a03608918cb8/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-08-07T16%3A34%3A55Z&sig=qvmw6yicRD2Qe68rOtEZll6XAAmhRxvhSA7qPSOVUnA%3D&ske=2026-08-07T19%3A33%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-08-07T15%3A33%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-08-07T16%3A24%3A50Z&sv=2025-11-05

FYI I cannot access that link.

bbiiggppiigg:
> > and the errors reported can be found in the logs https://productionresultssa10.blob.core.windows.net/actions-results/a057e474-3569-4791-a0bc-3116a7fe93b3/workflow-job-run-d0aa77ec-de43-5370-87a9-a03608918cb8/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-08-07T16%3A34%3A55Z&sig=qvmw6yicRD2Qe68rOtEZll6XAAmhRxvhSA7qPSOVUnA%3D&ske=2026-08-07T19%3A33%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-08-07T15%3A33%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-08-07T16%3A24%3A50Z&sv=2025-11-05
> 
> FYI I cannot access that link.


Try this.


[2363.fail.log](https://github.com/user-attachments/files/30837207/2363.fail.log)

PR #2365, opened 2026-08-09
Update AMDGPU decoders to the 2025-10-22 ISA spec

bbiiggppiigg:
The AMDGPU decoders (gfx908/gfx90a/gfx940) are regenerated from AMD's ISA specification drop dated 2025-10-22, replacing the previous generation based on the2024-09-24 drop. 

  The main structural change: the spec's optional encodings (*_INST_LITERAL, *_VOP_SDWA, *_VOP_DPP, VOP3_SDST_ENC) are now decoded as full encodings dispatched
  by predicates compiled from the spec's EncodingCondition expression trees (e.g. VOP2's default condition is SRC0 ∉ {249, 250, 255}, SDWA is SRC0 == 249),
  instead of the previous hardcoded literal/SDWA special cases. Instructions that use one encoding+opcode under several conditions with different operand lists
  (SOP2_INST_LITERAL with the literal replacing either source) get a runtime dispatch between operand-list variants.

  What this fixes (all reproducible on master)

  - Decoding any MFMA instruction aborts — the 32-register accumulator operands (128 bytes) exceed the widest Result type, tripping an assert in
  sizeToResult(). Fixed by adding an m1024 Result type.
  - Decoding s_setreg_imm32_b32 on gfx940 aborts — the opcode is missing from the opcode table because the old generator skipped the SOPK_INST_LITERAL encoding
  it lives in.
  - SDWA instructions format with no operands — updateMnemonic() only updated m_InsnOp, so appending the _SDWA extension made isCompressed() (which detects
  RISCV compressed instructions by the two Operations differing) misfire and format() print the stale mnemonic with the empty encoded-operand list.
  - MFMA register banks ignored — the ACC/ACC_CD bits selecting arch-VGPR vs AccVGPR per matrix operand were not decoded, so v_mfma ... a[0:31], v0, v1,
  a[32:63] printed v[0:31]/v[32:63].
  - SDWA source banks ignored — the S0/S1 flags selecting VGPR vs scalar sources were not decoded (v1 decoded as s1).
  - hwreg() operands decode as garbage — the decoder switched on  Jump to bottom (ctrl+End) ↓ he ID field, with off-by-one offset/size extraction.
  - DPP instructions were not decoded at all, and literal-carrying VOP2 instructions (v_madmk_f32/v_madak_f32) were missing on gfx908/90a.
  - Implicit vcc/exec/flat_scratch operands now carry their real 64-bit width as 32-bit-half pairs, completing the model from #2317.

  The condition-gated dispatch also introduces a case the old unconditional dispatch couldn't hit — a word claimed by no encoding (e.g. a literal marker on an
  instruction with no literal variant) — which now yields a 4-byte INVALID instruction instead of dereferencing a stale/null insn_in_progress. The never-called
  mainDecodeOpcode() (whose opcode lookup indexed with a never-set member) and decodeOPR_SDWA() (superseded by SDWA becoming a real encoding) are removed.

  The instruction set itself changes only slightly: the drop removes the 31 DS_*_SRC2_* instructions (referenced nowhere else) and introduces the implicit
  gpumem memory operand, added as a MISC register alongside dsmem.

  Validation

  - Instruction boundaries and operands verified against llvm-objdump (ROCm 7.0.2) on a real HIP gfx908 code object and on per-arch stress objects covering
  every literal/SDWA/DPP/HWREG/MFMA/call encoding path, including all MFMA bank combinations and a constant srcC under ACC_CD=1.
  - DYNINST_ENABLE_TESTS=ALL: 131/131 tests pass, including the IndirectControlFlowAnalysis golden-CFG comparisons for all three arches, which parse real
  binaries end-to-end through the new decoders.
  - Builds clean with -DDYNINST_WARNINGS_AS_ERRORS=ON under gcc 15.2 and clang 20.


bbiiggppiigg:
@hainest @kupsch ready for review

kupsch (approved):
Looks OK.

The the newly added lines to the `decoder_impl` files have white space at the end of the line which shows in dark red when I run `git diff ...`.

There are two non-portable issues that you might want to handle better that you should consider handling better at some point:

1) integer types that are expected to be certain number of bytes should use uintXX_t or intXX_t instead of `unigned long long int` or `unsigned int` as the number of bits used to represent these types is implementation defined.
2) In the C and C++ standards the layout of bit fields is implementation defined. You can add a unit test to validate the compile is matching your assumptions (initialize and extract bits to validate), or use bit manipulations to update and extract; I'm OK with the former.

bbiiggppiigg:
> 2. In the C and C++ standards the layout of bit fields is implementation defined. You can add a unit test to validate the compile is matching your assumptions (initialize and extract bits to validate), or use bit manipulations to update and extract; I'm OK with the former.

The code really does not rely on the bit layout at all. I used the bit width to save space, but since it was a union of structs, the total save is like 8 bytes per decoder. So I think I'm just going to remove the width specification, unless you think the unit tests for bit extraction and initialization hve value no matter what.

kupsch:
Sorry, I made the assumption that these were overlaying an int to get the fields within the instruction, if they are not doing that then bit fields are fine.

PR #2366, opened 2026-08-09
AMDGPU: add gfx950 (MI350) decoder support

bbiiggppiigg:
Stacked on the 2025-10-22 decoder-update PR (this branch's base); only the last commit is new here.

Adds `Arch_amdgpu_gfx950` (MI350), generated from the drop's mi350 specification with the same condition-dispatched decoder as the other CDNA targets, plus the architecture plumbing it needs: register file (modeled on gfx940's), entryIDs, ELF `e_flags` mach `0x4f` mapping, rose register conversion, SymEval register state, ABI register map, formatter, branch-instruction table, and MachRegister unit tests.

gfx950 introduces the 128-bit `ENC_VOP3PX2` encoding (the two `V_MFMA_SCALE` instructions): a `V_MFMA_LD_SCALE` prefix dword pair carrying the scale operands, followed by the main MFMA pair. Bits 127:64 of such instructions live in a new `insn_long2` filled by `setupInsnWord`. One spec discrepancy is handled in the generator and documented in the commit: the drop's encoding identifiers for `ENC_VOP3PX2` claim the prefix pair repeats the main opcode (`0xd3ad`/`0xd3ae`), but hardware and LLVM encode the LD_SCALE prefix opcode 44 (`0xd3ac`) there — the identifier list is overridden and the main opcode decoded from the second pair's `OP` field. The MFMA `ACC`/`ACC_CD` register-bank handling extends to this encoding.

### Validation

- Instruction boundaries and operands verified against `llvm-objdump --mcpu=gfx950` (ROCm 7.0.2) on a stress object covering the literal encodings, both `V_MFMA_SCALE` instructions in vgpr and accvgpr banks, MFMA, calls and branches — including correct decode of the surrounding instructions across the 16-byte encoding.
- `DYNINST_ENABLE_TESTS=ALL`: all tests pass, including the new gfx950 MachRegister unit tests; the pre-existing arches' generated files are byte-identical with this commit applied.
- Builds clean with `-DDYNINST_WARNINGS_AS_ERRORS=ON` under gcc 15.2 and clang 20.


bbiiggppiigg:

common/src/registers/MachRegister.C
```diff
@@ -414,6 +417,35 @@ namespace Dyninst {
         }
       }
       break;
+      case Arch_amdgpu_gfx950: {
+        int reg_class = (reg & 0x00ff0000);
+        if(reg_class == amdgpu_gfx950::SGPR || reg_class == amdgpu_gfx950::VGPR) {
+          return 4;
```
kupsch: Why do you check the reg_class here? Since all the SGPR and VGPR registers are `amdgpu_gfx950::BITS_32`, doesn't the switch produce the same value?

dataflowAPI/src/ABI.C
```diff
@@ -793,6 +796,11 @@ void ABI::initialize64(Architecture arch){
             sz = amdgpu_gfx940_map.size();
             break;
         }
+        case Arch_amdgpu_gfx950:{
+            RegisterMap amdgpu_gfx950_map = machRegIndex_amdgpu_gfx950();
+            sz = amdgpu_gfx950_map.size();
+            break;
+        }
         default:
         assert(0 && "This call is currently implemented for AMDGPU gfx908,gfx90a and gfx940 only");
```
kupsch: add gfx950 to message?

common/src/registers/MachRegister.C
```diff
@@ -177,6 +177,9 @@ namespace Dyninst {
       case Arch_amdgpu_gfx940: {
               return *this;
       }
+      case Arch_amdgpu_gfx950: {
+              return *this;
+      }
```
kupsch, line 180: combine with cases above

common/src/registers/MachRegister.C
```diff
@@ -691,6 +731,9 @@ namespace Dyninst {
       case Arch_amdgpu_gfx940: {
         return (base.val() & 0x0000F000);
       }
+      case Arch_amdgpu_gfx950: {
+        return (base.val() & 0x0000F000);
+      }
```
kupsch, line 712: combine with previous cases

common/h/Architecture.h
```diff
@@ -51,6 +51,7 @@ namespace Dyninst {
     Arch_amdgpu_gfx908 = 0x94000000, // future support for gfx908
     Arch_amdgpu_gfx90a = 0x98000000, // future support for gfx90a
     Arch_amdgpu_gfx940 = 0x9c000000, // future support for gfx940
```
kupsch: comments can be deleted

dwarf/src/registers/convert.C
```diff
@@ -39,6 +39,7 @@ namespace DwarfDyninst {
       case Dyninst::Arch_amdgpu_gfx908:
       case Dyninst::Arch_amdgpu_gfx90a:
       case Dyninst::Arch_amdgpu_gfx940:
+      case Dyninst::Arch_amdgpu_gfx950:
```
kupsch, line 42: In the future if you need to add the dwarf register mapping, they seem to be found here:

https://rocm.docs.amd.com/projects/llvm-project/en/latest/LLVM/llvm/html/AMDGPUUsage.html#register-identifier

dwarf/src/registers/convert.C
```diff
@@ -75,6 +76,7 @@ namespace DwarfDyninst {
...
```
kupsch, line 79: see above

kupsch (approved):
Look OK.

A few minor cleanups.

dwarf/src/registers/convert.C
bbiiggppiigg: noted.

PR #2375, opened 2026-08-18
Fix instructionAPI's use of DYNINST_ENABLE_{CAPSTONE,ZYDIS}

kupsch:
After the recent directory restructuring, neither DYNINST_ENABLE_CAPSTONE nor DYNINST_ENABLE_ZYDIS function correctly.

DYNINST_ENABLE_CAPSTONE does not work since riscv/decoder.h includes capstone/capstone.h which is included by InstructionDecoderImpl.C.  After the directory restructuring DYNINST_ENABLE_CAPSTONE no longer causes a dependency on Dyninst::Capstone for InstructionDecoderImpl.C.  Made instructionAPI_riscv have capstone as a public dependency and make are architecture decode libraries be a dependency of instructionAPI_decoder.

DYNINST_ENABLE_ZYDIS does not work after the directory restructuring as the files that need zydis moved to instructionAPI/src/decoder the cmake logic also needs to move there.

### DIFF
--- a/instructionAPI/src/decoder/riscv/CMakeLists.txt
+++ b/instructionAPI/src/decoder/riscv/CMakeLists.txt
@@ -2,6 +2,7 @@ include_guard(GLOBAL)
 
 if(NOT DYNINST_ENABLE_CAPSTONE)
   add_library(instructionAPI_riscv INTERFACE)
+  install(TARGETS instructionAPI_riscv EXPORT dyninst-targets)
   return()
 endif()
 
--- a/instructionAPI/src/decoder/riscv/CMakeLists.txt
+++ b/instructionAPI/src/decoder/riscv/CMakeLists.txt
@@ -21,7 +21,7 @@ dyninst_library(
   SOURCE_FILES ${_sources}
   PRIVATE_HEADER_FILES ${_headers}
   DYNINST_INTERNAL_DEPS common
-  PRIVATE_DEPS Dyninst::Capstone
+  PUBLIC_DEPS Dyninst::Capstone
 )
 
 set(_inc_dirs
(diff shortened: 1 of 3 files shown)


hainest:
Well, that's neat. I thought I had both of those turned on in one of my local tests builds, but I did not. Sorry about that. For good measure, we should run the `CMake build options` workflow against this branch- although that won't check for the Zydis one.

kupsch:
I started `CMake build options`

hainest:
Another way to do this would be to have `instructionAPI_decoder` directly depend on the `instructionAPI_ARCH` targets. That way, you could move all of the decoder-specific stuff out of the main instructionAPI CMakeLists.txt. That said, it seems CMake is clunky in this area. The [docs](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#linking-object-libraries) say that OBJECT libraries are transitive, but I couldn't get it to work without the `TARGET_OBJECTS` hack.

The version in the diff below is building locally when DYNINST_ENABLE_CAPSTONE and DYNINST_ENABLE_ZYDIS are both on. I didn't try them separately.

[diff.txt](https://github.com/user-attachments/files/31200623/diff.txt)

bbiiggppiigg (approved):

instructionAPI/src/decoder/riscv/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ include_guard(GLOBAL)
 
 if(NOT DYNINST_ENABLE_CAPSTONE)
   add_library(instructionAPI_riscv INTERFACE)
+  install(TARGETS instructionAPI_riscv EXPORT dyninst-targets)
```
hainest, line 5: I'm surprised this is needed. There shouldn't be an attempt to install anything for a non-imported interface target.

PR #2381, opened 2026-08-28
cleanup/refactor/bug fix emitElf

kupsch:
This is a large set of fixes that cleans up the emitElf functionality and it also fixes several bugs in the code.  There are many commits that modernize the code, cleans it up so it is easier to understand, and refactors many aspects to remove redundancy.  It also fixes several bugs the produce incorrect ELF objects.  The number of warnings reported by readelf is reduced to an expected couple.  Some of these are not issues with ELF produced by current tool chains, but are with specially crafted binaries or if the tool chain changes.  The bugs fixed include:

- rewritten binaries respect alignment requirement of the original ELF's sections and segments instead of forcing to various fixed alignment values that could violate original alignment constraints
- the .tbss is no longer chosen as the point to insert after as it appears to contribute to the address space, but it does not
- setting the linkage for the symbol table to the string table based on the actual section number instead of assuming it was the next section; although generally correct, it is not required
- adjusted RELR relocations in all cases
- fixed the RELR generation code to use the correct type for 32-bit ELF files so the RELR section contained valid data
- set the permissions of the resulting ELF to the values of the original file
- support ELF files with extended ELF header values that are stored in the NULL section's data
- clear no longer valid sh_link and sh_info fields of old rewritten sections
- correctly handle ELF files that need a .symtab_shndx section
- in the case of failure, no longer leak a file descriptor, and the memory managed by libelf
- updated the _DYNAMIC point to be correct so other tools that depend on can work
- sort symbol so that SHT_LOCAL symbol types come before others to meet ELF's requirement
- fixed garbage symbol index of -1 for unindexed symbols instead of STN_UNDEF
- fixed st_shndx and section-symbol values for sections after newly inserted sections
- section renames now replace the 2nd character of the name with '#', the previous value of 'o' would fail if a future tool chain produced a section name with a 'o' as the 2nd character
- detect additional errors and return the failure instead of producing invalid ELF files, and in the case of failure the remove the partially written ELF file 

Fixes #1764

### DIFF
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -412,7 +412,7 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
     // ".shstrtab" section: string table for section header names
     const char *shnames = pdelf_get_shnames(oldElfHandle);
     if (shnames == NULL) {
-        log_elferror(err_func_, ".shstrtab section");
+        log_elferror(err_func_, ".shstrtab section not found");
         return false;
     }
 
@@ -1059,7 +1059,7 @@ bool emitElf<ElfTypes>::createLoadableSections(Elf_Shdr *&shdr, unsigned &extraA
         sectionNumber++;
         // Add a new loadable section
         if ((newscn = elf_newscn(newElf)) == NULL) {
-            log_elferror(err_func_, "unable to create new function");
+            log_elferror(err_func_, "unable to create new section");
             return false;
         }
         if ((newdata = elf_newdata(newscn)) == NULL) {
@@ -1313,7 +1313,7 @@ bool emitElf<ElfTypes>::addSectionHeaderTable(Elf_Shdr *shdr) {
     Elf_Shdr *newshdr;
 
     if ((newscn = elf_newscn(newElf)) == NULL) {
-        log_elferror(err_func_, "unable to create new function");
+        log_elferror(err_func_, "unable to create new section");
         return false;
     }
     if ((newdata = elf_newdata(newscn)) == NULL) {
@@ -1365,7 +1365,7 @@ bool emitElf<ElfTypes>::createNonLoadableSections(Elf_Shdr *&shdr) {
         secNames.push_back(nonLoadableSecs[i]->getRegionName());
         // Add a new non-loadable section
         if ((newscn = elf_newscn(newElf)) == NULL) {
-            log_elferror(err_func_, "unable to create new function");
+            log_elferror(err_func_, "unable to create new section");
             return false;
         }
         if ((newdata = elf_newdata(newscn)) == NULL) {
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -591,17 +591,16 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
             // A word that is also a RELR relocation target was already shifted by the
             // earlier RELR loop, so skip to avoid offsetting it twice
             std::unordered_set<Offset> relr_addrs(relr_relocs.begin(), relr_relocs.end());
-            for(std::size_t off = 0; off < newdata->d_size; off += sizeof(void*)) {
+            for(std::size_t off = 0; off < newdata->d_size; off += sizeof(Elf_Relr)) {
                 Offset addr = shdr->sh_addr + off;
                 if (relr_addrs.count(addr)) continue;  // already adjusted as a RELR reloc
 
                 char *loc = static_cast<char*>(newdata->d_buf) + off;
-                size_t val{};
-                // The calls to memcpy are required to not break the aliasing rules.
-                memcpy(&val, loc, sizeof(val));
+                Elf_Relr val{};
+                read_memory_as(val, loc);
                 if(val == 0) continue;
                 val += library_adjust;
-                memcpy(loc, &val, sizeof(val));
+                write_memory_as(loc, val);
             }
         }
         // Change offsets of sections based on the newly added sections
--- a/symtabAPI/src/emitElf.C
+++ b/symtabAPI/src/emitElf.C
@@ -70,52 +70,15 @@ unsigned int elfHash(const char *name) {
 template<class ElfTypes>
 emitElf<ElfTypes>::emitElf(Elf_X *oldElfHandle_, bool isStripped_, ObjectELF *obj_, void (*err_func)(const char *),
                                Symtab *st) :
-        oldElfHandle(oldElfHandle_), newElf(NULL), oldElf(NULL),
-        obj(st),
-        newEhdr(NULL), oldEhdr(NULL),
-        newPhdr(NULL), oldPhdr(NULL), phdr_offset(0),
-        textData(NULL), symStrData(NULL), dynStrData(NULL),
-        olddynStrData(NULL), olddynStrSize(0),
-        symTabData(NULL), dynsymData(NULL), dynData(NULL),
-        phdrs_scn(NULL), verneednum(0), verdefnum(0),
-        newSegmentStart(0), firstNewLoadSec(NULL),
-        dynSegOff(0), dynSegAddr(0),
-        phdrSegOff(0), phdrSegAddr(0), dynSegSize(0),
-        secNameIndex(0), currEndOffset(0), currEndAddress(0),
-        linkedStaticData(NULL), loadSecTotalSize(0),
-        isStripped(isStripped_), library_adjust(0),
-        object(obj_), err_func_(err_func),
-        hasRewrittenTLS(false), TLSExists(false), newTLSData(NULL) {
-    oldElf = oldElfHandle->e_elfp();
-    curVersionNum = 2;
-
-    //Set variable based on the mechanism to add new load segment
-    // 1) createNewPhdr (Total program headers + 1) - default
-    //    library_adjust - create room for a new program header in a position-indepdent library
-    //    by increasing all virtual addresses for the library
-
-    //If we're dealing with a library that can be loaded anywhere,
-    // then load the program headers into the later part of the binary,
-    // this may trigger a kernel bug that was fixed in Summer 2007,
-    // but is the only reliable way to modify these libraries.
-    //If we're dealing with a library/executable that loads at a specific
-    // address we'll put the phdrs into the page before that address.  This
-    // works and will avoid the kernel bug.
-
-    isStaticBinary = obj_->isStaticBinary();
-
-    //If we want to try a mode where we add the program headers to a library
-    // that can be loaded anywhere, and put the program headers in the first
-    // page (avoiding the kernel bug), then set library_adjust to getpagesize().
-    // This will shift all addresses in the library down by a page, accounting
-    // for the extra page for program headers.  This causes some significant
-    // changes to the binary, and isn't well tested.
-
-    library_adjust = 0;
-    if (!(object && object->getLoadAddress())) {
-        library_adjust = getpagesize();
-    }
-
+    oldElfHandle{oldElfHandle_},
+    oldElf{oldElfHandle->e_elfp()},
+    obj(st),
+    isStripped(isStripped_),
+    library_adjust{getpagesize()},
+    object(obj_),
+    err_func_(err_func),
+    isStaticBinary{obj_->isStaticBinary()}
+{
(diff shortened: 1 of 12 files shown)


symtabAPI/src/emitElf.C
```diff
@@ -70,52 +70,15 @@ unsigned int elfHash(const char *name) {
...
-
-    //If we want to try a mode where we add the program headers to a library
-    // that can be loaded anywhere, and put the program headers in the first
-    // page (avoiding the kernel bug), then set library_adjust to getpagesize().
-    // This will shift all addresses in the library down by a page, accounting
-    // for the extra page for program headers.  This causes some significant
-    // changes to the binary, and isn't well tested.
-
-    library_adjust = 0;
-    if (!(object && object->getLoadAddress())) {
```
bbiiggppiigg, line 115: Even when object is always non null, object->getLoadAddress can return  loadAddress or 0 on shared object(according to the comments in Object-elf.h getLoadAddress). Are you sure we can set library_adjust unconditionally here?

symtabAPI/src/emitElf.C
```diff
@@ -1365,7 +1365,7 @@ bool emitElf<ElfTypes>::createNonLoadableSections(Elf_Shdr *&shdr) {
         secNames.push_back(nonLoadableSecs[i]->getRegionName());
         // Add a new non-loadable section
         if ((newscn = elf_newscn(newElf)) == NULL) {
-            log_elferror(err_func_, "unable to create new function");
+            log_elferror(err_func_, "unable to create new section");
```
bbiiggppiigg: It would probably be better to have a way to distinguish between all the error messages
Something like
unable to create new non-loadable section
unable to create new section for section header table ...

Of course if log_elferror reports line number then maybe it is fine.

symtabAPI/src/emitElf.C
```diff
@@ -870,9 +870,9 @@ void emitElf<ElfTypes>::fixPhdrs() {
     newPhdr = ElfTypes::elf_newphdr(newElf, newEhdr->e_phnum);
     void *phdr_data = (void *) newPhdr;
 
-    for (unsigned i = 0; i < segments.size(); i++)
+    for (const auto &segment : segments)
```
bbiiggppiigg, line 955: Should we consider something like std::copy(segments.begin(), segments.end(), newPhdr)?

symtabAPI/src/emitElf.C
```diff
@@ -1663,8 +1664,9 @@ bool emitElf<ElfTypes>::createSymbolTables(set<Symbol *> &allSymbols) {
     if (!obj->isStaticBinary()) {
         //reconstruct .dynsym section
         Elf_Sym *dynsyms = (Elf_Sym *) malloc(dynsymbols.size() * sizeof(Elf_Sym));
-        for (i = 0; i < dynsymbols.size(); i++)
-            dynsyms[i] = *(dynsymbols[i]);
+        i = 0;
+        for (const auto &s : dynsymbols)
```
bbiiggppiigg, line 1740: I think the change for this loop actually makes it less readable.

symtabAPI/src/emitElf.C
```diff
@@ -1320,15 +1310,15 @@ bool emitElf<ElfTypes>::addSectionHeaderTable(Elf_Shdr *shdr) {
     newshdr->sh_addralign = 4;
 
     //Set up the data
-    newdata->d_buf = allocate_buffer(secNameIndex);
+    newdata->d_buf = allocate_buffer(secNameTableTotalBytes);
     char *ptr = (char *) newdata->d_buf;
     for (const auto &secName : secNames) {
-        memcpy(ptr, secName.c_str(), secName.length());
-        memcpy(ptr + secName.length(), "\0", 1);
```
bbiiggppiigg: Do we no longer need to fill 0 for the rest of the string?

symtabAPI/src/emitElf.h
```diff
@@ -45,10 +45,15 @@
 #include <unordered_set>
 #include <vector>
 
-using std::cerr;
-using std::cout;
-using std::endl;
-using std::vector;
+#ifndef STT_GNU_IFUNC
+#define STT_GNU_IFUNC 10
```
hainest, line 49: This was added to elfutils in 2009. I think it's safe to assume it's present.

symtabAPI/src/emitElf.h
```diff
@@ -45,10 +45,15 @@
...
+#endif
+#ifndef STB_GNU_UNIQUE
+#define STB_GNU_UNIQUE 10
```
hainest, line 52: Same as above.

symtabAPI/src/emitElf.h
```diff
@@ -45,10 +45,15 @@
...
+#endif
+#ifndef PT_GNU_PROPERTY
+#define PT_GNU_PROPERTY 0x6474e553
```
hainest, line 55: This was added to elfutils in 2020. We require at least 0.186 which was released in 2021. I think we can assume it's always present.

symtabAPI/src/emitElf.C
```diff
@@ -357,87 +290,199 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
...
 template<class ElfTypes>
-void emitElf<ElfTypes>::renameSection(const std::string &oldStr, const std::string &newStr, bool renameAll) {
-    assert(oldStr.length() == newStr.length());
-    for (unsigned k = 0; k < secNames.size(); k++) {
-        if (secNames[k] == oldStr) {
-            secNames[k].replace(0, oldStr.length(), newStr);
-            if (!renameAll)
-                break;
+void emitElf<ElfTypes>::renameSection(const std::string &oldName) {
+    assert(oldName.length() >= 2 && oldName[1] != 'o');
```
hainest: I can't decide if we should abort here or not. If our renaming scheme can't handle a section name, maybe we should just emit a message and then just not preserve the old copy? I can't imagine most users really care that we preserve the rewritten sections by renaming them.

symtabAPI/src/emitElf.C
```diff
@@ -357,87 +290,199 @@ bool emitElf<ElfTypes>::createElfSymbol(Symbol *symbol, unsigned strIndex, vecto
...
+    Region *foundSec = NULL;
+    Region *shstrtabRegion = NULL;
+
+    string newFName = fName + "XXXXXX";
+    auto buf = std::unique_ptr<char[]>(new char[newFName.length() + 1]);
+    strncpy(buf.get(), newFName.c_str(), newFName.length() + 1);
 
-    newfd = mkstemp(buf.get());
+    auto newfd = mkstemp(buf.get());
+    newFName = buf.get();
```
hainest, line 471: https://www.boost.org/doc/libs/latest/libs/filesystem/doc/reference.html#unique_path?

symtabAPI/src/emitElf.h
kupsch: elfutils has elf.h, but it doesn't install elf.h, so elf.h comes from the distribution, and the efl.h in RHEL 8 does not define these constants.

symtabAPI/src/emitElf.C
kupsch: this means that we would have 2 sections that have the same name, which might confuse some tools. How about renaming but instead of changing the second character to `o`, change it to `#`. That is much less likely to conflict. I think that would be a good change to make. What are your thoughts of this change and leaving the assert?

All the ELF section names I've seen (and verified with all executables and libraries on  RHEL 8 & Ubuntu 24.04) just contain the characters: `a-zA-Z0-9._,$-`.  The only `o` as the 2nd character is in NetworkManager (section `connection_defaults`) and the libtcmalloc related libraries (section google_malloc) neither of which Dyninst would rewrite. There are no section names less than 2 characters.

symtabAPI/src/emitElf.C
kupsch: if the `rewrite_print` wasn't present that would be a good idea,

symtabAPI/src/emitElf.C
kupsch: done

symtabAPI/src/emitElf.C
kupsch: In some respects it is less readable, but better in others. I'm going to leave as is.

symtabAPI/src/emitElf.C
kupsch: it never 0-filled. There were two memcpy's. The first copied the string's value, and the second added a null to the end. The new code copies the string and the null with one memcpy using the magic of adding 1 to the length.

symtabAPI/src/emitElf.C
kupsch: this commit was just cleanup. The old behavior was maintained. This is all fixed in later commits where library_adjust is renamed into two variables

symtabAPI/src/emitElf.C
kupsch: `unique_path` just make a path and returns it this is equivalent to `tempnam` and has the same security issues (see the man page). Since we need a fileno it is best as is. If we needed an iostream then `unique_path` would be the correct choice.

symtabAPI/src/emitElf.h
hainest: I'm really surprised RHEL 8 is that far out of sync from elfutils.

symtabAPI/src/emitElf.C
hainest: > this means that we would have 2 sections that have the same name, which might confuse some tools. 

I was thinking more of `if(renameSection(...)) { /* make a copy */ }`, but I see that's now how this function is used.

> How about renaming but instead of changing the second character to `o`, change it to `#`. That is much less likely to conflict. I think that would be a good change to make. What are your thoughts of this change and leaving the assert?

I think `#` is a better option. I'd like to remove the assert since this isn't really a fatal error, but I think error handling in the rest of this code is likely not any better so we can leave it in.

> All the ELF section names I've seen (and verified with all executables and libraries on RHEL 8 & Ubuntu 24.04) just contain the characters: `a-zA-Z0-9._,$-`. The only `o` as the 2nd character is in NetworkManager (section `connection_defaults`) and the libtcmalloc related libraries (section google_malloc) neither of which Dyninst would rewrite. There are no section names less than 2 characters.

That's good to know. I was very curious about sections that have an `o` as a second character conflicting here.

symtabAPI/src/emitElf.C
hainest: > Since we need a fileno it is best as is

Bah, yes, I was too focused on the ugly string hacking.

symtabAPI/src/emitElf.C
kupsch: not a conflict, but the rename is a nop and if something searches the section by name it could incorrectly find the old section.

bbiiggppiigg (approved):

symtabAPI/src/emitElf.C
bbiiggppiigg: I think the rewrite printf is kind of misleading as the program header update took place somewhere else.

symtabAPI/src/emitElf.C
```diff
@@ -491,55 +521,58 @@ bool emitElf<ElfTypes>::driver(std::string fName) {
...
+            shstrtabRegion = foundSec;
             continue;
+        }
 
         sectionNumber++;
-        changeMapping[sectionNumber] = 0;
-        newNameIndexMapping[string(name)] = sectionNumber;
+        changeMapping[sectionNumber] = false;
+        newNameIndexMapping[name] = sectionNumber;
+        if (foundSec)
```
bbiiggppiigg, line 556: foundSec is set by findRegion only when something is found.
If foundSec on line 544 return false, since it was not checked, foundSec will retain the old value of the previous section.

Either set it to null in the  Symtab::findRegion(Region *&ret, std::string const& secName)

Or reset it to null in the loop?

bbiiggppiigg (changes requested):

hainest:
No regressions on 17 machines. I'll wait to approve until after Hsuan-Heng's comments are resolved.

symtabAPI/src/emitElf.C
kupsch: all findRegion methods set ret to null on failure

bbiiggppiigg (approved):

hainest (approved):

PR #2384, opened 2026-09-07
Import new x86 mnemonics from Zydis

hainest:
This includes an additional 390 instructions over what Capstone previously provided.

Two important items:

1. At the mnemonic level, Zydis doesn't distinguish `retf` from `ret`.
    
    Dyninst takes great strides to differentiate short and far returns via
    e_ret_far and e_ret_near. This will need to be added as a manual
    decoding step for 0xCB and 0xCA.

2. The pusha instructions aren't handled consistently when converting to ROSE.

    The alias of pushad to pushal isn't correct. pushad should alias
    pushaw. Additionally, the check in RoseInsnX86Factory::convertKind
    incorrecty returns `x86_pushad` for `e_pusha`. It should return
    `x86_pusha` and another entry should be added for `e_pushad`. This is a
    long-standing bug that should be fixed.


hainest:

instructionAPI/importers/import_registers.py
```diff
@@ -0,0 +1,145 @@
+import argparse
+import x86.registers
+import aarch64.registers
+import ppc.registers
```
kupsch: The imports fail now that the modules are in a subdirectory

instructionAPI/importers/zydis/mnemonics.py
kupsch, line 1: should there be a `importers/zydis/x86` directory that contains these modules, so it is similar in structure to capstone?

instructionAPI/importers/import_mnemonics.py
```diff
@@ -0,0 +1,126 @@
...
+    if mnemonics.pseudo is not None:
+        for p in mnemonics.pseudo:
+            f.write(
+                "{0:s}_{1:s}, /* pseudo mnemonic */\n".format(
+                    mnemonics.dyninst_prefix, p
+                )
+            )
+
+    all_mnemonics = sorted(
+        set(mnemonics.external) | set(mnemonics.missing) - set(mnemonics.pseudo)
```
kupsch, line 52: The non-x86 arches set `missing` and `pseudo` to `None`. That doesn't play well with the set operations here:

```
Processing mnemonics for aarch64
Traceback (most recent call last):
  File "/nobackup/kupsch/kupsch/cleanup-emitelf/instructionAPI/importers/./import_mnemonics.py", line 53, in <module>
    set(mnemonics.external) | set(mnemonics.missing) - set(mnemonics.pseudo)
TypeError: 'NoneType' object is not iterable
```

Instead of `None` using an empty array for value would fix this as it would have a valid type.

kupsch (dismissed):
A couple of small python issues in the non-x86 architectures, otherwise looks good.

instructionAPI/importers/import_registers.py
hainest: Derp. I didn't test the register importer...

instructionAPI/importers/zydis/mnemonics.py
hainest: I considered that, but decided against it since Zydis only supports x86.

instructionAPI/importers/import_mnemonics.py
hainest: Oh, yes, of course. Requiring them to be non-None also makes generating the enums easier.

bbiiggppiigg (commented):
I'm still going through the PR commit by commit, but I'll post the AI review here first, which seems to be reasonable feebacks.

====
## Worth addressing

### pusha/popa — the disclosed pusha deferral is fine, but the popa half changed here is inconsistent, and the prescribed fix isn't achievable at this layer
The pusha/convertKind bug is properly disclosed (commit `e2bb36bbd` NOTE + PR item 2) and left for later — no objection to deferring. Two points about what *was* changed:

- **popa was restructured but pusha wasn't**, leaving an asymmetry and a dead case. `x86_entryIDs.h:730-731` splits `e_popa`/`e_popad` into distinct values and `convertOpcodes.C:698-701` gets two cases — but the decoder (`arch-x86.C` oneByteMap, `0x61`) only ever emits `e_popad`, so `e_popa → x86_popa` (698-699) is unreachable. pusha stays aliased (`e_pushad = e_pusha`) with `e_pusha → x86_pushad` (772-773). Either defer both (revert the popa split too) or finish both, so the siblings match.

- **Heads-up on the eventual fix:** `convertKind` receives only the `entryID`, which carries no operand size, so it structurally can't pick `x86_pusha` (16-bit) vs `x86_pushad` (32-bit). Applying item 2 literally (`e_pusha → x86_pusha`) would make `0x60` emit the 16-bit kind while `0x61` stays 32-bit — a new asymmetry. The real fix likely needs operand-size-aware selection where size is known, not in `convertKind(entryID)`.

### 3DNow removal — commit rationale ("not used") looks inaccurate; it's a silent decode-result change
Commit `685ff6197` "Remove e_3dnow_generic" says *"It's not used"*, but it was the return value for `t_3dnow` decodes, which still occur (`arch-x86.C:8227` advances past the 3DNow opcode suffix). With the `getID()` case removed:
```diff
-      case t_3dnow:
-         return e_3dnow_generic;
```
and the table entry now `e_No_Entry` (`arch-x86.C:1233`), a real 3DNow instruction now reports `e_No_Entry` — conflating a successful generic decode with "no instruction," and changing `format()` output. Please confirm intended and fix the commit rationale.

### `x86_entryIDs.h:391` — `e_invalid` is Zydis's null sentinel, not a mnemonic
It leaks in because `importers/zydis/mnemonics.py`'s `_ignored` only filters `fsetpm287_nop`. Add `"invalid"` to `_ignored` and regenerate. Inert today (unreferenced), but it pollutes the enum.

### Add to the `import_mnemonics.py:52` thread — operator precedence
Separate from the `None` issue there: `set(external) | set(missing) - set(pseudo)` parses as `external | (missing - pseudo)`, so `pseudo` is never subtracted from `external`. Harmless today (no name collisions), but the intent reads clearer parenthesized: `(set(external) | set(missing)) - set(pseudo)`.

## Nits

- **`dataflowAPI/src/RoseInsnFactory.C:196-206`** — `e_popad` is grouped with `e_pushf` while `e_pusha` sits in its own identical `operands.clear()` block; co-locating the pusha/popa cases reads less accidental.
- **`instructionAPI/importers/README.md`** — title still "Capstone importer" though x86 now uses Zydis; add-an-arch step 3 and the `aliases`-shape description are stale vs the new `capstone`/`zydis` grouping.
- **`x86_entryIDs.h:816`** — `e_retf` (hand-injected via the importer's `missing` list) carries no marker; a `/* missing in Zydis */` note would document why it exists.
- **Commit `b53f12d89` "Fix script output"** — bare subject, empty body; a reader can't tell what was wrong or why.
- **Commit `94eebe2b8`** describes the rename as `pfrcpit1 -> pfrcpit`, but the emitted name is `pfcpit1` (Zydis's spelling); the message's stated target doesn't match the result.

bbiiggppiigg:
Looks good to me other than the concerns listed above.

hainest:
> popa was restructured but pusha wasn't

The commit "Remove/rename ATT syntax aliases" is the correct behavior.
The import from Zydis introduced an alias because the entry in the
importer's alias table somehow got unfixed during a rebase. Fixed.

> Heads-up on the eventual fix: convertKind receives only the entryID,
> which carries no operand size, so it structurally can't pick x86_pusha
> (16-bit) vs x86_pushad (32-bit).

That's true now, but Zydis correctly decodes pusha from pushd. I'm not
worried about this since it's been broken forever.

> 3DNow removal — commit rationale ("not used") looks inaccurate; it's
> a silent decode-result change

I know. I also don't care because 3DNow! instructions are no longer
supported by AMD. Moreover, we'll be moving to Zydis ASAP, so I'd rather
make this change now than wait and possibly forget.
@kupsch @bbiiggppiigg if you think we should support these in the interim,
I'm fine reverting this.

> e_invalid is Zydis's null sentinel, not a mnemonic

The sentinel for ZydisMnemonic is `ZYDIS_MNEMONIC_MAX_VALUE`, but the
entry for it is (currently)
`ZYDIS_MNEMONIC_MAX_VALUE = ZYDIS_MNEMONIC_XTEST,`. Since that doesn't
match the regex in `_parse_zydis`, then it's ignored. `ZYDIS_MNEMONIC_INVALID`
has the same meaning as `e_No_Entry`. Since every other architecture
uses `<arch>_op_INVALID`, we should change `e_No_Entry` to `e_INVALID`.

> thread - operator precedence

Heh. I even ran it through pylint... Fixed.

> e_popad is grouped with e_pushf while e_pusha sits in its own identical
> operands.clear() block; co-locating the pusha/popa cases reads less accidental.

I plan to make an importer for Dyninst <-> ROSE conversion like I did
for DWARF on aarch64 (and need to add one for DWARF everywhere else) that
will also put them in numerical order to possibly let the compiler not
implement the enormous switch as a jump table.

> title still "Capstone importer" though x86 now uses Zydis

Fixed.

> dataflowAPI/src/RoseInsnFactory.C:196-206 - e_popad is grouped with e_pushf
> while e_pusha sits in its own identical operands.clear() block; co-locating
> the pusha/popa cases reads less accidental.

This will get adjusted once we get full semantic support in the Instruction interface.

> x86_entryIDs.h:816 - e_retf (hand-injected via the importer's missing list)
> carries no marker; a /* missing in Zydis */ note would document why it exists.

Not needed. The comments in `<arch>_entryIDs.h` will go away later.

> Commit b53f12d89 "Fix script output" - bare subject, empty body; a
> reader can't tell what was wrong or why.

It's just whitespace changes in the output. :shrug:

> Commit 94eebe2b8 describes the rename as pfrcpit1 -> pfrcpit, but the
> emitted name is pfcpit1 (Zydis's spelling); the message's stated target
> doesn't match the result.

It's just a typo. I don't want to rebase to fix it.

bbiiggppiigg (approved):

kupsch (approved):

kupsch:
removing 3dNow! is fine

PR #2387, opened 2026-09-12
[DRAFT] Revamp proccontrol

bbiiggppiigg:


hainest:
I get that this is a work in progress, but I'm curious why choose to do this now? There is so much going on here that it will be hard to reason about these changes. Personally, I'd rather spend time on the ABI updates.

How does this affect #2315 and #2321?

bbiiggppiigg:
> I get that this is a work in progress, but I'm curious why choose to do this now? There is so much going on here that it will be hard to reason about these changes. Personally, I'd rather spend time on the ABI updates.
> 
> How does this affect #2315 and #2321?

#2315 and #2321 should be closed in favor of this, even though I'm not sure if we are ever going to get this merged.

PR #1002, opened 2021-02-09
DWARF supplemental file and type parsing 

sashanicolas:
Allow parsing of DWARF supplemental file and fix parsing of types.

Fixes #909
Fixes #908
Fixes #907
Fixes #906
Fixes #905
Fixes #904
Fixes #903
Fixes #902
Fixes #646

PR #2388, opened 2026-09-15
fix parseSubrange type_id collision with DWZ supplementary DWARF

### LINKED ISSUE #1647: Dyninst dwarf parsing results in uncaught exception.
Dyninst crashes due to an uncaught throw parsing dwarf.

The dwarf parser itself is not causing the exception,
rather library or support code that is involved in parsing
throws a bad_cast (without bolo/symdebug that is all
you will see in the logs).   Alas the exception lacks
origination information.

The dwarf parsing in Dyninst causes exception throws --
back to main -- when parsing dwarf, particularly on
el/centos/rhel-8.   The exception throw wipes out the
call stack for the failures.   This behavior started in
spring 2023.

To debug this, there is a branch  bolo/symbdebug which adds a
firewall --  exception catch -- to all the dispatches in
the dwarf parser.   Grab the diffs from master in that
branch, and you can apply them to most any version of dyninst
with none or little editing.  Use patch.

With that firewall in place, you can get a stack backtrace
from the dwarf parser itself.  it appears that something
may be goes wrong with dwarf parsing of modules in the
postfix of the code after the big switch.   All the failures
I've noticed so far are from the tail part of the dwarf parser
dispatcher.

I've tried using gdb to catch the throw responsible, to
identify the exact failure location outside of dyninst.
I've had no success catching that throw with the debugger;
it has not been able to intercept that throw.

Several people have mentioned that there may be an issue
parsing module information, or in handling unknown dwarf
returned by elfutils.   This should be investigated; newer
elfutils have been helpful in reducing the occurrences of
these failures... until yet newer binaries are installed by
software updates.   Dyninst may be behind in dwarf parsing
awareness, of unknown dwarf that is parsed, and not known
by elfutils.   There could be problems with elfutils and
unknown dwarf, which affect dyninst.

This causes a hard crash in dyninst, which is not caught
because it isn't dyninst failing -- but something it calls.

To repeat this, run the testsuite on el8 after installing
the bolo/symdebug patches on el8.   The core files
generated will show the stack traces from when the unknown
throws are caught in dwarfWalker.  They demonstrate the call stack inside
dyninst up to that point.

This issue is just a big data dump and how to use the bolo/symdebug
branch to be able to do some debugging and call stack analysis.

This occurs on el8  on x86_64.  It does not occur on ubuntu.  It
is repeatable in a variety of el8 releases using versioned OS containers
on that platform as well.

bbiiggppiigg:
DwarfWalker::parseSubrange hardcoded is_sup=false when deriving a subrange's type_id. On toolchains that emit DWZ multi-file DWARF (e.g. Rocky/RHEL 8 glibc, via .gnu_debugaltlink), a subrange DIE can live in the supplementary file while an unrelated type sits at the SAME section offset in the main file. The get_type_id key is {offset, is_sup, module}; hardcoding is_sup=false collapses the two DWARF spaces, so both get the same type_id. addOrUpdateType then hands the struct back the subrange, and parseMember's curEnclosure()->asFieldListType() (an unguarded dynamic_cast<fieldListType&>) throws std::bad_cast, aborting the entire type parse.

readelf on this system confirms offset 0x2025 is a DW_TAG_subrange_type in the glibc .dwz supplementary file and DW_TAG_structure_type cpuid_registers in the main libm debug file. Six testsuite mutatees (test1_3/1_9/1_17/1_27/1_28/test5_7) deterministically crashed test_driver in Symtab::parseTypes via findType.

Compute is_sup from the subrange DIE's own CU, matching type_id() and the parseTypeReferences path already in this file.

## Validation (full external testsuite, mahir x86_64)

`runTests -all`, master `4547a0bf2` vs this branch (same base + the one-line fix), both built clean:

| | master | this PR | Δ |
|---|---:|---:|---:|
| PASSED  | 864 | 884 | **+20** |
| FAILED  | 16  | 16  | 0 |
| CRASHED | 20  | **0** | **−20** |

All 20 crashes are eliminated, the 20 crashed variants now PASS, and FAILED is unchanged — i.e. no regression. The tests that stop crashing are exactly the six above (`test1_3/1_9/1_17/1_27/1_28/test5_7`), and the in-tree `ctest` suite is unchanged at 133/133.

## Note: this is the deterministic core of the long-standing "mahir flake band"

These six tests have long shown up as intermittent/machine-specific failures on our Rocky 8 x86_64 box and were treated as flaky noise. They are **not** flaky: they crash **deterministically** whenever the toolchain emits **DWZ supplementary DWARF** and a supplementary-file subrange collides on section offset with a main-file type (as `readelf` shows for `0x2025`). That's why they only reproduce on DWZ-using distros (RHEL/Rocky 8 glibc) and looked like unexplained noise elsewhere. This fix removes that whole crash band at the source.

fixes #1647


hainest:
Does this fix #1647?

What I wouldn't give to have a way of deterministically generating DWARF to make tests. Maybe one day...

>curEnclosure()->asFieldListType() (an unguarded dynamic_cast<fieldListType&>) throws std::bad_cast, aborting the entire type parse.

The entire type system in DwarfWalker is just so bad. These types of dynamic casts should _never_ be used.

bbiiggppiigg:
### Does this fix #1647?

Very likely — #1647 ("Dyninst dwarf parsing results in uncaught exception") describes this bug almost line-for-line, and the evidence below maps its claims onto the concrete root cause fixed here.

| #1647 says | This PR's finding |
|---|---|
| "uncaught throw parsing dwarf… throws a **bad_cast**" | `dynamic_cast<fieldListType&>(*this)` at `Type.h:351` → uncaught `std::bad_cast` → `std::terminate`/SIGABRT |
| "**not dyninst failing — but something it calls**" | The throw is inside libstdc++'s `dynamic_cast`, reached from `parseMember` — no Dyninst error, just an abort |
| "particularly on **el/centos/rhel-8**… repeatable across el8 releases" | Fires only where Rocky/RHEL-8 `glibc-debuginfo` ships **DWZ supplementary** DWARF; our Rocky 8 box reproduces, non-el8 machines don't |
| "It **does not occur on ubuntu**" | Ubuntu's debuginfo doesn't produce the same cross-file offset collision in the glibc types we hit |
| "from the **tail part of the dwarf parser dispatcher**, after the big switch" | Crash is in `parseMember`, downstream of the type dispatch, once the enclosure has been mis-set to a wrong-kind object |
| "had **no success catching that throw with gdb**" | Same wall here — RelWithDebInfo optimized out the locals and mutatee vforks killed conditional breakpoints; `readelf` is what finally nailed it |
| "This behavior **started spring 2023**… newer binaries installed by software updates" | Latent code bug, exposed by an el8 debuginfo package update — see timeline below |

**Timeline corroborates it.** The defect — `parseSubrange` hardcoding `is_sup=false` — has been **latent since DWZ supplementary support landed (#1002, 2021)**; the code assumed subrange DIEs never live in a supplementary file. It only *manifests* once the installed el8 `glibc-debuginfo` ships a DWZ layout with a colliding section offset, which is a property of the **debuginfo package, not the Dyninst version**. That matches #1647's "started spring 2023 … triggered by newer binaries from software updates" exactly: a latent parser bug uncovered by an OS debuginfo refresh.

**A/B evidence.** On the el8 box the full external testsuite went from **`bad_cast=40` → `bad_cast=0`**, with all six crashing tests (`test1_3 / test1_9 / test1_17 / test1_27 / test1_28 / test5_7`) fixed and **zero regressions**. Full `runTests -all` A/B, master `4547a0bf2` vs this fix `228478ba6`:

| Machine | Arch | master C / bad_cast | fix C / bad_cast | Net |
|---|---|---|---|---|
| el8 (Rocky 8, DWZ) | x86_64 | 20 / 40 | **0 / 0** | **−20 crashes, +20 passes**, 6 tests un-crashed |
| x86_64 (non-DWZ) | x86_64 | 0 / 0 | 0 / 0 | no-regression (1170P/0F both sides) |
| aarch64 | aarch64 | 0 / 0 | 0 / 0 | no-regression (1231P/1F both sides) |
| ppc64le | ppc64le | 0 / 0 | 0 / 0 | no-regression (794P/12F both sides) |

The fix eliminates the crash cluster exactly where the DWZ trigger is present and is provably inert (identical P/F/C, `bad_cast=0`) on the three architectures that don't ship the colliding supplementary debuginfo.

**Caveat.** #1647 is explicitly a "big data dump" — an umbrella report of a *pattern* of el8 dwarf-parse bad_casts, without a single minimal reproducer attached. This PR resolves the mechanism it describes and drove **every** uncaught bad_cast in our el8 testsuite to zero, but that alone can't prove there is no second, independent bad_cast source behind the same symptom. The clean way to close the loop: re-run the reporter's `bolo/symdebug`-firewalled reproducer on this build — if those el8 backtraces disappear, #1647 is fully closed.

So this reads as **"very likely fixes #1647"** rather than an outright auto-close, pending that confirmation from the reporter.

bbiiggppiigg:
> Does this fix #1647?

I believe so.

kupsch (approved):
Looks good.

bbiiggppiigg:
@hainest Do you think this need any further testing.

hainest:
> @hainest Do you think this need any further testing.

It's likely fine, but I'll run it through the standard battery just to be safe. That should finish in an hour or so.

hainest (approved):

PR #2389, opened 2026-09-15
amdgpu: model s_mov + resolve indirect s_setpc control-flow targets

bbiiggppiigg:
An indirect `s_setpc_b64` whose target is a getpc-relative computed address (a relocation veneer or a tail call) was left unresolved: it fell through to jump-table analysis, which only understands memory-read tables, and sank to an ADDRESS_INVALID edge. The identical getpc-relative computation already resolves for `s_swappc_b64` calls via ResolveCallTargetBySlicing, so route the indirect-branch case through the same backward slice.

That alone is not enough: the slice folds only when every step of the address computation has ROSE semantics. The dispatcher modeled s_add_u32/s_addc_u32 but not s_mov, so an offset staged as `s_mov s_off, imm; s_add s_lo, s_lo, s_off` left s_off a free variable and the target never folded to a constant. Compiler-emitted code folds the immediate straight into s_add and never needed it; code that stages the offset through a mov (e.g. a generated veneer) did.

- convertOpcodes: map S_MOV_B32/S_MOV_B64 -> rose_amdgpu_op_s_mov_b32/b64 for gfx908/90a/940/950.
- DispatcherAMDGPU: add IP_s_mov_b32 / IP_s_mov_b64 semantics (register and immediate sources) and register them.
- RoseInsnFactory::massageOperands: split the s_mov_b64 destination pair into 32-bit halves (mirroring setpc), keeping an immediate source whole.
- IA_IAPI::getNewEdges: for an AMDGPU indirect jump, try the call-target slice first and emit a DIRECT edge on a resolved constant, before giving up to jump-table analysis.

No change to existing resolution: the parseAPI IndirectControlFlowAnalysis regression output (gfx908/90a/940) is byte-identical, and s_swappc call targets still resolve.


hainest:
Is it possible to write a test for this?

bbiiggppiigg:
@hainest 
This is part of an effort to let the branchInfo tool (under dyninst regression test) to parse into .dyninstInst.
The sequence generated here are picked by the AMDGPU emitter implementation (not typical compiler generated sequence).
While I certainly have a test case I think it is of little value to put into the test_binaries repo.
We can maybe discuss tomorrow.

hainest:
> @hainest This is part of an effort to let the branchInfo tool (under dyninst regression test) to parse into .dyninstInst. The sequence generated here are picked by the AMDGPU emitter implementation (not typical compiler generated sequence). While I certainly have a test case I think it is of little value to put into the test_binaries repo. We can maybe discuss tomorrow.

That sounds reasonable. I'll leave it to your discretion. I don't have a frame of reference for the change, so I'll let @kupsch review it.

PR #2390, opened 2026-09-15
Add loongarch64 as a supported platform

huadoudouxue:
Please refer to https://docs.kernel.org/arch/loongarch/introduction.html for an introduction to the loongarch architecture


hainest:
@huadoudouxue Many thanks for taking the time to port Dyninst to a new architecture! However, we can't support a new architecture without a working instruction decoder. We have plans to adopt Capstone in the coming months which does have support for loongarch, so that would be the preferred decoder for this architecture.

We do not have access to any loongarch computers, so we have no way to testing any of this code. That is a serious impedement to accepting support for a new architecture.

@kupsch @bbiiggppiigg Thoughts?

bbiiggppiigg:
I actually just dicussed this topic with Bart.

@huadoudouxue 
We are certainly interested in this port to a new architecture. However, as @hainest suggested, we don't realy have a way to test and evaluate on this architecture. If you are serious about getting porting dyninst to loongarch, we should have some meeting/discussion.

Contact me through email hwu337@wisc.edu.

PR #2391, opened 2026-09-18
dyninstAPI: write 32-bit trap table entries whenever the mutatee is 32-bit

bbiiggppiigg:
Rebased onto master after #2393 (the CMake half) merged; this PR is now only the `dyninstAPI` change.

**`dyninstAPI/src/addressSpace.C`: `trampTrapMappings::writeToBuffer` overran its buffer without `cap_32_64`.**
The 4-byte path for a 64-bit mutator writing a 32-bit mutatee's trap table was wrapped in `#if defined(cap_32_64)`, which only x86_64/ppc64le codegen defines. The two asserts right above it already declare that case supported, so in a `DYNINST_CODEGEN_ARCH=i386` build the function passed the asserts and then wrote 8 bytes per field into a buffer sized for 4 — heap corruption in every 32-bit process the mutator touched (valgrind: invalid write of size 8 in `flush()`, then `free(): invalid next size` in `mapped_object::getVariable`; every 32-bit test `CRASHED (Group Setup)`). The runtime `addr_width` check already selects the case; the preprocessor guard is removed. In a default x86_64 build `cap_32_64` is defined, so the compiled code there is unchanged.

**Validation** (re-run on the rebased branch, master 22af9ffbe)
- x86_64 default build (gcc 15, `-DDYNINST_WARNINGS_AS_ERRORS=ON`): clean; in-tree ctest 136/136; external-tests 7/7. clang-15 `-Werror` build: clean. CI container gate (ubuntu-25.04 and fedora-43 images, CI flags incl. `-DDYNINST_CODEGEN_ARCH=x86_64`, INTEGRATION tests): PASS.
- x86_64 `CODEGEN_ARCH=i386` build: in-tree ctest 135/135 (the x86_64 emitter unit test is correctly not built).
- Testsuite at 32-bit on coriander (`BUILD_RTLIB_32=ON`, `DYNINST_ENABLE_32BIT_TESTS=ON`, `runTests -32 -all`) with `CODEGEN_ARCH=i386`: without this change every 32-bit test crashes in group setup; with it the 1100-row 32-bit matrix runs, 693 PASSED, and row-by-row against the same run on a standard x86_64-codegen build of master (the usual way to test 32-bit), 1067 of 1075 common rows have the same verdict. The differences are test_pt_ls (it runs parseThat on the 64-bit system `ls`, which an i386-only code generator cannot parse) and the rewriter groups, see below.

Not included: 32-bit *rewriter* groups in an i386-codegen build still crash at group setup (328 of 329 rewriter rows) because `BinaryEdit::openResolvedLibraryName` fully parses each 64-bit library candidate it finds on the search path before checking its word size, and an i386-codegen mutator has no 64-bit ABI tables. That is a separate change and will come as its own PR once this one is in.


hainest:
These are somewhat related but different enough that they should be two PRs. Additionally, let's add the following logging to the end of `DyninstPlatform.cmake`:

```cmake
message(STATUS "Host arch = ${_host_arch}")
message(STATUS "Codegen arch =  ${_codegen_arch}")
message(STATUS "DYNINST_PLATFORM = ${DYNINST_PLATFORM}")
```

bbiiggppiigg:
Split as requested: the CMake fix is now its own PR, #2393, together with the three `message(STATUS ...)` lines at the end of `DyninstPlatform.cmake`. Once #2393 is in, I'll rebase this PR so it contains only the `trampTrapMappings::writeToBuffer` change.

bbiiggppiigg:
Basically, in writeToBuffer, it either picks line 1070 anb return or line 1074.
On width mismatch, but cap_32_64 not defined, it pick 1074 and overwrites buffer.

hainest (approved):
No regressions on one machine of each platform.

> 693 PASSED

I'm seeing 697 pass on coriander, but that's probably just a few unstable tests.

PR #2393, opened 2026-09-18
cmake: fix DYNINST_CODEGEN_ARCH override assigning the wrong variable; log the selected arches

bbiiggppiigg:
Split out of #2391 at the reviewer's request: the CMake half only, plus the configure-time logging asked for there.

**`cmake/DyninstPlatform.cmake`: the codegen override assigned the wrong variable.**
`set(${_codegen_arch} ${DYNINST_CODEGEN_ARCH})` expands to `set(x86_64 i386)` on an x86_64 host: it creates a variable *named after the host arch* and leaves `_codegen_arch` untouched. The codegen checks below still select the right arch only because `if()` auto-dereferences the bare token `x86_64`. The deprecated `DYNINST_PLATFORM` block dereferences it the same way and falls through to `ppc64_linux`, so every cross-codegen install exports `set(DYNINST_PLATFORM "ppc64_linux")` in `DyninstConfig.cmake`; a testsuite linked against an x86_64 `CODEGEN_ARCH=i386` install compiles the ppc64 mutatee set and test table (no 32-bit tests at all). Fixed to `set(_codegen_arch ...)`. `DYNINST_PLATFORM` exported by an x86_64 host build, before → after:

| `DYNINST_CODEGEN_ARCH` | master | fixed |
|---|---|---|
| x86_64 | x86_64-unknown-linux2.4 | x86_64-unknown-linux2.4 |
| i386 | ppc64_linux | x86_64-unknown-linux2.4 |
| aarch64 | aarch64-unknown-linux | x86_64-unknown-linux2.4 |
| ppc64le | ppc64_linux | x86_64-unknown-linux2.4 |
| amdgpu_gfx908 | ppc64_linux | x86_64-unknown-linux2.4 |
| riscv64 | ppc64_linux | x86_64-unknown-linux2.4 |

The `DYNINST_CODEGEN_ARCH_*` defines are identical before and after for all six (that is why the cross-codegen CI builds never noticed). The aarch64/ppc64le rows only matched their codegen target by coincidence; `DYNINST_PLATFORM` is the *host* platform by definition (it selects the testsuite's mutatee set, which runs on the host), so the fixed value is the intended one for every row.

**Logging.** As requested, the end of the file now prints what was selected:
```
-- Host arch = x86_64
-- Codegen arch = i386
-- DYNINST_PLATFORM = x86_64-unknown-linux2.4
```

**Validation**
- Configure-only probes on x86_64 for all six `DYNINST_CODEGEN_ARCH` values: platform string as in the table; codegen defines unchanged. `cmake-format --check` clean.
- x86_64 build with `-DDYNINST_WARNINGS_AS_ERRORS=ON`, in-tree ctest, external-tests, clang-15 `-Werror` build and the CI container images (ubuntu-25.04, fedora-43) were all green on #2391 with this same fix; the build+ctest is re-running on this branch.
- With this fix an x86_64 `CODEGEN_ARCH=i386` testsuite build produces the 32-bit mutatee set (286 mutatees) instead of none; running them still needs the second half of #2391.

Once this is in, #2391 will be rebased to contain only the `writeToBuffer` change.


hainest (approved):
Tested on an aarch64 and ppc64le machine with no issues. LGTM.

PR #2394, opened 2026-09-19
cmake: quote expanded operands in if() comparisons

bbiiggppiigg:
Mechanical cleanup, no behavior change: every `if()`/`elseif()` that compared an expanded variable did so unquoted, e.g. `if(${_host_arch} STREQUAL "aarch64")`. This PR quotes the operand: `if("${_host_arch}" STREQUAL "aarch64")`. 58 lines in 12 files.

**Why**
CMake's `if()` treats a bare token as a variable name whenever a variable of that name exists, so an unquoted expansion gets looked up a *second* time. If anything ever defines a variable named `x86_64`, `GNU`, `Clang`, `static`, ... every comparison whose value spells that name silently switches to comparing that variable's value, with no diagnostic. That is precisely how the `DYNINST_CODEGEN_ARCH` override went wrong (fixed in #2393): a stray `set(x86_64 …)` made every `${_host_arch}` comparison in `DyninstPlatform.cmake` read `i386` instead of `x86_64`, and the exported platform became `ppc64_linux`. An unquoted expansion of an *empty* variable also drops an argument (`if( STREQUAL "x")`) and aborts the configure.

Under CMP0054 (NEW since the project's 3.14 minimum) a quoted argument is never re-dereferenced and an empty value compares as `""`, so quoting closes both holes. `FindElfutils.cmake` already used this form in one place.

**What is deliberately unchanged**
- The right-hand operand of `IN_LIST` stays a bare name; that operand must be a variable.
- `if(${_flag_name})` in `tests/cmake/DyninstCheckStaticExeSupported.cmake`: there the expansion *is* a variable name and the lookup is the intent.
- `external/` is untouched.

**Ordering.** This needed #2393 first: on the pre-#2393 tree, `_codegen_arch` still held the host arch after an override, and quoting `"${_codegen_arch}"` would have compared the literal `x86_64` and silently disabled the override for every arch. With #2393 merged the branch sits directly on master and depends on nothing else.

**Validation (behavior-neutral)**
- Configure-only probes on x86_64 for `DYNINST_CODEGEN_ARCH` = x86_64, i386, aarch64, ppc64le, amdgpu_gfx908, riscv64: the exported `DYNINST_PLATFORM` and the full `-D` define set of `common` and `dyninstAPI` are identical before and after for all six. `-DDYNINST_ENABLE_LTO=ON` still configures.
- `cmake-format --check` with the repo's `.cmake-format.yaml`: clean on all changed files.
- x86_64 build with `-DDYNINST_WARNINGS_AS_ERRORS=ON`: clean; in-tree ctest: 136/136.


hainest:
> Under CMP0054 (NEW since the project's 3.14 minimum) 

That's not true. The `OLD` behavior was removed in 4.0, but the [docs](https://cmake.org/cmake/help/v3.14/policy/CMP0054.html#policy:CMP0054) say

```
CMake version 3.14.7 warns when the policy is not set and uses OLD behavior.
```

This makes the PR a no-op. We can either set CMP0054 to `NEW` or use bareword variables where they aren't intended to be indirectly dereferenced. I don't have a preference.

bbiiggppiigg:
> This makes the PR a no-op. 
I'm not sure if this is a no-op. The motivation behind the PR is to make sure we don't trip into something similar to #2393.

> We can either set CMP0054 to `NEW` or use bareword variables where they aren't intended to be indirectly dereferenced. I don't have a preference.

You are the cmake expert here. I have no preference as well.
If you think the PR is of no practical value and I can just close it.

hainest:
> > This makes the PR a no-op.
> > I'm not sure if this is a no-op. The motivation behind the PR is to make sure we don't trip into something similar to #2393.
> 
> > We can either set CMP0054 to `NEW` or use bareword variables where they aren't intended to be indirectly dereferenced. I don't have a preference.
> 
> You are the cmake expert here. I have no preference as well. If you think the PR is of no practical value and I can just close it.

These need to be fixed. I've been meaning to do some CMake cleanup for a while, so now is as good a time as any. I think the easiest solution is to set the policy to `NEW` in the root CMakeLists.txt before calling `project`. Then the rest of the PR fixes any lingering bugs.

```cmake
# Only interpret if() arguments as variables or keywords when unquoted.
if(POLICY CMP0054)
  cmake_policy(SET CMP0054 NEW)
endif()
```

hainest:
@bbiiggppiigg Doing a rebase and adding the policy check I mentioned should make this PR ready to go.

bbiiggppiigg:
Rebased onto master `3c0043491` and added the `CMP0054` policy block to the root `CMakeLists.txt`.

**Regression test:** external `dyninst/testsuite` (`c694a8f7`), `runTests -all`, branch `f338207c9` vs. master `3c0043491` on each machine. The later head `7a65a866b` only adds quoting in an in-tree ctest `CMakeLists.txt`.

| host | arch | ctest | testsuite result |
|---|---|---|---|
| coriander | x86_64 | 136/136 | **No regressions.** Both runs completed (1546 rows each); no verdict differences. |
| zeroah | aarch64 | 134/134 | **No regressions.** The branch run hit the 600 s group limit in `pc_singlestep`. The 16 tests it never reached pass in isolation on both trees. `pc_singlestep` failed 4/5 isolated reruns on master and on the branch (pre-existing flake). |
| mahir | x86_64 (el8) | 136/136 | **No regressions in the 902 rows both runs reached.** Both runs stopped at the same 600 s group limit. The one `test1_27` crash moved between variants; isolated reruns: master 1/5 crashed, branch 0/5 (pre-existing flake). |
| cyclops | ppc64le | 110/110 | **Still running.** Master stopped at the 600 s group limit (743 rows); the branch run is in progress. |

PR #2395, opened 2026-09-19
dyninstAPI: check a library candidate's word size before parsing it

bbiiggppiigg:
`BinaryEdit::openResolvedLibraryName` (rewriter mode) resolves each DT_NEEDED library by opening every candidate path on the search path as a `BinaryEdit`, which parses the whole file, and only then compares its address width with the binary being rewritten. The search order lists the 64-bit system directories before `/usr/lib32`, so rewriting a 32-bit binary first fully parses the 64-bit `libdl.so.2` and `libc.so.6` and throws the result away.

Two consequences:
- In a `DYNINST_CODEGEN_ARCH=i386` build the mutator has no 64-bit ABI tables (`ABI::initialize64()` is compiled out), so that parse dereferences a null `bitArray` in `IA_x86::isTailCall` and the mutator segfaults: every 32-bit rewriter test group dies in group setup (328 of 329 rows on coriander). `DYNINST_DEBUG_STARTUP=1` shows `opening file /usr/lib/gcc/x86_64-linux-gnu/10/../../../x86_64-linux-gnu//libc.so.6` immediately before the crash.
- In a default build it is only wasted work: a complete CFG parse of the 64-bit libc for every 32-bit binary rewritten.

This change looks at the candidate's ELF class via `Symtab::openFile` first and skips wrong-width candidates before `BinaryEdit::openFile`. Archive members (static rewriting) keep the existing path; their word size is decided per archive.

**Validation** (branch rebased onto master 934edd115, which includes #2391)
- x86_64 default build (gcc 15, `-DDYNINST_WARNINGS_AS_ERRORS=ON`): clean; in-tree ctest 136/136; external-tests 7/7. clang-15 `-Werror` build: clean. CI container gate (ubuntu-25.04 and fedora-43 images, CI flags, INTEGRATION tests): PASS.
- 32-bit testsuite on coriander (`BUILD_RTLIB_32=ON`, `DYNINST_ENABLE_32BIT_TESTS=ON`, `runTests -32 -all`), same master tip on both sides: this branch as a `CODEGEN_ARCH=i386` build vs. master as a standard x86_64-codegen build (the usual way to test 32-bit). Both produce 1099 rows with the identical row set (create 449 / attach 320 / rewriter 329); 1096 rows have the same verdict class. The three differences: test_pt_ls create+rewriter (runs parseThat on the 64-bit system `ls`, which an i386-only code generator cannot parse) and one pc_hw_breakpoint row FAILED→CRASHED (the test that hangs and truncates both runs). Without this change every 32-bit rewriter group of the i386-codegen build crashed at group setup. The 32-bit rewriter failures that remain are identical on the standard build and pre-existing.
- Minimal reproducers (ParseAPI parse, `openBinary` with dependencies, `processCreate`, full rewrite) and their results across the four build combinations are in the comment thread below.


hainest:
> a code generator built for one word size (DYNINST_CODEGEN_ARCH=i386) cannot parse binaries of the other

Can you expand on this? You can definitely link a 64-bit binary against a 32-bit one. If Dyninst can't handle that case, then we should make that a known deficiency.

**EDIT**: I just noticed that the codegen arch here is 32-bit. I think we should change this to check if the target library's word size is _larger_ than the codegen target.

bbiiggppiigg:
@hainest 

There are two defects here: 

1) parsing a 64 bit binary on a 32 bit CODEGEN build will cause null pointer dereference because the gap in ABI initialization 

and 2) a 32 bit CODEGEN built dyninst, while trying to open a 32 bit binary, if there happens to be a 64 bit library with the same name on the search path, triggers the 64 bit parse and triggers the first bug

bbiiggppiigg:
**What is being compared.** `getAddressWidth()` here is the width of the binary being rewritten (`mapped_objects[0]`'s ELF class), not the host's or the mutator's. So for a 32-bit binary the check keeps 32-bit candidates and skips 64-bit ones, and for a 64-bit binary the reverse. It is the same predicate the function already applied (`temp->getAddressWidth() == getAddressWidth()`, three lines below); this PR only evaluates it before the candidate is parsed instead of after.

**Why equality is the right predicate.** A DT_NEEDED entry is only a soname, and `ld.so` resolves it by walking the search path and accepting the first file of the *same* ELF class; wrong-class files are skipped and only reported if nothing else matches. Mixed-width linking does not exist at any stage. On coriander with a 32-bit `lib32/libf.so` and a 64-bit `lib64/libf.so`:

```
$ gcc -m64 m.c -Llib32 -lf
/usr/bin/ld: skipping incompatible lib32/libf.so when searching for -lf
/usr/bin/ld: cannot find -lf
$ LD_LIBRARY_PATH=lib32:lib64 LD_DEBUG=libs ./m64      # 32-bit dir first
      trying file=lib32/tls/haswell/x86_64/libf.so  ... (skipped, wrong class)
      trying file=lib64/libf.so                      -> loaded, exit 7
$ LD_LIBRARY_PATH=lib32 ./m64
./m64: error while loading shared libraries: libf.so: wrong ELF class: ELFCLASS32
```

Dyninst's `getResolvedLibraryPath` builds a comparable candidate list (`LD_LIBRARY_PATH`, `ldconfig -p`, `/lib`, `/usr/lib`, `/usr/local/lib`), and unlike the loader's cache that list does contain both classes; `openResolvedLibraryName` then takes the first same-width one and returns null if there is none, which mirrors the loader's outcome. "Larger than the codegen target" would not be a resolution rule: a 32-bit candidate for a 64-bit binary would pass it, and Dyninst would model a library the process can never have loaded; in a standard build the 64-bit `libc.so.6` found first for a 32-bit binary is not larger than the codegen target, so it would still be parsed in full and rejected afterwards.

**So the bug is not *what* Dyninst decides but *when*.** The loader reads the ELF header and rejects before mapping anything; Dyninst mapped and fully parsed the candidate (`BinaryEdit::openFile` → `CodeObject::parse`) and only then looked at its class. That ordering costs something different in each build:

- *Standard x86_64-codegen build:* pure waste. Every 32-bit rewrite does a complete CFG parse of the 64-bit `libdl.so.2` and `libc.so.6` that get rejected a moment later; the final answer is still right.
- *`DYNINST_CODEGEN_ARCH=i386` build:* the parse itself is the crash. `ABI::initialize64()` is compiled out, so the parser has no 64-bit ABI tables and dereferences a null `bitArray` in `IA_x86::isTailCall`; it never reaches the rejection.

Minimal reproducers (coriander, both installs built the same way):

```cpp
// parse_probe: pure ParseAPI, no rewriting
#include "CodeObject.h"
#include "CodeSource.h"
int main(int, char **argv) {
  auto *sts = new Dyninst::ParseAPI::SymtabCodeSource(argv[1]);
  auto *co = new Dyninst::ParseAPI::CodeObject(sts);
  co->parse();
  printf("%s: %zu functions\n", argv[1], co->funcs().size());
}
// bpatch_probe: the rewriter path this PR touches
BPatch bp; BPatch_binaryEdit *be = bp.openBinary(argv[1], /*openDependencies=*/true);
```
`hello32` is `gcc -m32 hello.c`; its only DT_NEEDED is `libc.so.6`, whose first search-path candidate is the 64-bit `/usr/lib/x86_64-linux-gnu/libc.so.6`.

| Dyninst build | `parse_probe` 64-bit libc | `parse_probe` 32-bit libc | `bpatch_probe hello32` (open + deps) | `create_probe hello32` (processCreate, runs mutatee) |
|---|---|---|---|---|
| i386 codegen, master before #2391, neither fix | SIGSEGV `IA_x86::isTailCall` | 2746 functions | SIGSEGV | (crashes in group setup in the testsuite: trap-table heap overrun) |
| i386 codegen, #2391 only | SIGSEGV | 2746 functions | SIGSEGV, via `openBinary → getAllDependencies → openResolvedLibraryName("libc.so.6") → openFile(…/x86_64-linux-gnu/libc.so.6) → CodeObject::parse` | mutatee runs, exit 0 |
| i386 codegen, this PR only | SIGSEGV | 2746 functions | opened | mutator aborts: `free(): invalid pointer` (the #2391 bug) |
| i386 codegen, #2391 + this PR (current branch) | SIGSEGV (unchanged: defect 1) | 2746 functions | opened | mutatee runs, exit 0 |
| standard x86_64 codegen | 2997 functions | 2746 functions | opened | mutatee runs, exit 0 |

The first column is the deficiency you're pointing at: a codegen-restricted build segfaults rather than erroring when asked to parse the other width at all, through any API. Agreed that should be documented and fixed to fail cleanly; the place is `ABI::getABI` (it hands out an ABI object with null bitsets), and I'd rather do that as its own issue/PR than fold it in here. This PR is only about not asking it to.


hainest:
> 1. parsing a 64 bit binary on a 32 bit CODEGEN build will cause null pointer dereference because the gap in ABI initialization
> 
> and 2) a 32 bit CODEGEN built dyninst, while trying to open a 32 bit binary, if there happens to be a 64 bit library with the same name on the search path, triggers the 64 bit parse and triggers the first bug

I understand the bug, and this definitely fixes the case of accidentally parsing 64-bit code when doing 32-bit codgen. Right now, this is sufficient. However, if/when we get cross-platform codegen working, we'll need to make this more robust. After thinking about it more, my comment above assumes that we can rewrite 32-bit binaries when `DYNINST_CODEGEN_ARCH=x86_64`, but that's not true. So forget that comment. We can ship this PR, as-is.

I'll add some discussion about this in #1892.

hainest:
Well, it turns out gcc won't link a 32-bit library into a 54-bit executable. I guess ABI would be an issue. We can ignore all of my previous comments.

kupsch (approved):

PR #2396, opened 2026-09-19
cmake: enable Capstone for DYNINST_CODEGEN_ARCH=riscv64 cross builds

### LINKED ISSUE #2392: CODEGEN_ARCH=riscv64 cross builds never enable Capstone (duplicate `riscv64` arm in DyninstPlatform.cmake)
**Summary**
A build configured with `-DDYNINST_CODEGEN_ARCH=riscv64` on a non-RISC-V host gets the RISC-V code generator but no RISC-V instruction decoder: `DYNINST_ENABLE_CAPSTONE` stays at its default OFF, and `instructionAPI/src/decoder/riscv` builds as an empty INTERFACE library. Configure and build succeed, so nothing reports the problem. Native riscv64 hosts are not affected.

**Cause**
The codegen selection chain in `cmake/DyninstPlatform.cmake` tests `riscv64` twice:

```cmake
elseif(${_codegen_arch} STREQUAL "riscv64")     # matches first ...
  set(DYNINST_CODEGEN_ARCH_RISCV64 TRUE)
elseif(${_codegen_arch} STREQUAL "ppc64le")
  set(DYNINST_CODEGEN_ARCH_PPC64LE TRUE)
elseif(${_codegen_arch} STREQUAL "riscv64")     # ... so this arm is unreachable
  set(DYNINST_CODEGEN_ARCH_RISCV64 TRUE)
  set(DYNINST_ENABLE_CAPSTONE ON)                # and Capstone is never enabled
```

Only the second arm enables Capstone, mirroring the host-arch branch a few lines above (`elseif(${_host_arch} STREQUAL "riscv64") ... set(DYNINST_ENABLE_CAPSTONE ON)`). Because the first arm already satisfied the chain, the second is dead code. Downstream, `CMakeLists.txt` only includes `DyninstCapstone` when `DYNINST_ENABLE_CAPSTONE` is true, and `instructionAPI/src/decoder/riscv/CMakeLists.txt` returns an empty `instructionAPI_riscv` target when it is not. The duplicate dates from #2058 (Conditionally compile Capstone), which added the Capstone line to the second copy.

**Reproduce** (any x86_64 host, configure only)
```
cmake -S dyninst -B build -DDYNINST_CODEGEN_ARCH=riscv64
grep DYNINST_ENABLE_CAPSTONE build/CMakeCache.txt          # OFF
grep -c capstone <configure output>                        # 0
find build/instructionAPI/src/decoder/riscv -name flags.make | wc -l   # 0: no decoder sources
```
With the first `riscv64` arm deleted, the same configure enables Capstone and reaches `find_package(capstone)` (which then fails on a machine without Capstone installed, as it should).

**Fix**
Delete the first `riscv64` arm (the two lines without the Capstone `set`).

**Why CI did not catch it**
`.github/workflows/codegen-cross-build.yaml` builds the matrix `[i386, x86_64, aarch64, ppc64le, amdgpu_gfx908]`; riscv64 is not in it. Adding it would have failed the configure the moment Capstone became required for the decoder. Suggest adding `riscv64` to that matrix along with the fix.

Related: #2391 fixes a different bug in the same file (the `set(${_codegen_arch} ...)` override). This one is independent of it.

bbiiggppiigg:
Fixes #2392.

The codegen-arch selection chain in `cmake/DyninstPlatform.cmake` tested `"riscv64"` twice. The first arm only set `DYNINST_CODEGEN_ARCH_RISCV64`; the second, which also does `set(DYNINST_ENABLE_CAPSTONE ON)` exactly like the riscv64 *host* arm above it, was unreachable. Since the RISC-V decoder builds as an empty INTERFACE library when Capstone is off, a `-DDYNINST_CODEGEN_ARCH=riscv64` build on another host got no RISC-V instruction decoder and no error. Native riscv64 hosts were not affected. The shadowing arm is removed; the surviving arm is unchanged.

**Validation**
- Configure-only on x86_64 (no Capstone installed): for x86_64, i386, aarch64, ppc64le and amdgpu_gfx908 the exported `DYNINST_PLATFORM` and define sets are identical to master; for riscv64 the configure now enables Capstone (it stops at `find_package(capstone)` on that machine, where master silently configured without it).
- In the ubuntu-25.04 CI image (which has Capstone 6.0.0), `-DDYNINST_WARNINGS_AS_ERRORS=ON -DDYNINST_ENABLE_TESTS=REGRESSION -DDYNINST_CODEGEN_ARCH=riscv64`: Capstone is found, `instructionAPI_riscv` compiles its four decoder sources, and common/symtabAPI/instructionAPI/parseAPI/patchAPI/proccontrol/stackwalk build `-Werror`-clean with their 131 regression tests passing. Same image, same flags, master as control: Capstone never mentioned, zero decoder objects, 93 targets built vs 103 with the fix. Both trees then stop in dyninstAPI with the same 166 `unknown architecture` errors (see below), so the fix changes exactly the Capstone/decoder part and nothing else.

**What this does not fix, and why riscv64 is not added to the cross-codegen CI matrix here:** `dyninstAPI/src/codegen/codegen.h` has no RISC-V branch and ends in `#error "unknown architecture"`, so dyninstAPI cannot be built for `CODEGEN_ARCH=riscv64` with or without this change (the same holds for a native riscv64 host). RISC-V support is analysis-only today (decoder, parser, symtab, proccontrol, stackwalk). A riscv64 matrix entry would need the workflow to build that subset rather than the whole tree; that is a separate decision.


hainest:
Heh. Not sure how that got missed. This clearly indicates that we have no tests that parse riscv binaries.

hainest (approved):

PR #2399, opened 2026-09-22
Fix loop index comparison in AnnotatableSparse::annotationsReport

hainest:
This was reported by CodeQL.


common/h/Annotatable.h
```diff
@@ -700,7 +700,7 @@ class DYNINST_EXPORT AnnotatableSparse
 		  std::vector<AnnotationClassBase *> atypes;
 		  annos_t &l_annos = *getAnnos();
 
-		  for (AnnotationClassID id = 0; id < l_annos.size(); ++id)
+		  for (AnnotationClassID id = 0; id < static_cast<AnnotationClassID>(l_annos.size()); ++id)
```
bbiiggppiigg, line 703: Shouldn't it be a loop over size_t, and then static_cast on line 705 ?

hainest: I considered that, but went with this since it's a "loop over all IDs". The problem is that they conflate an ID with a position in `l_annos`. It's just a bad design.

As an aside, I'm passively working on tearing out annotations. They are an awkward solution to a fairly minimal problem. I've looked through the commit logs and can't find a discussion of why they were added.

hainest:
No regressions on one machine of each arch.

bbiiggppiigg (approved):
I assume this is just fixing some sort of compile warning as error on type comparison?

hainest:
> No regressions on one machine of each arch.

Yup. `AnnotationClassID` is an `unsigned short` and `size()` is `size_t`. It was the only useful thing reported by CodeQL.

kupsch (approved):

PR #2400, opened 2026-09-22
GitHub CI: Add automatic CodeQL scanning

### LINKED ISSUE #1213: Regular static analysis of the dyninst code base to improve code quality
**Is your feature request related to a problem? Please describe.**
There was mention on https://github.com/dyninst/dyninst/pull/1207 to have static analysis:
`Static analysis is definitely something we want to incorporate into our CI workflow
`
**Describe the solution you'd like**
One possible way to get some some static analysis for dyninst would be through

https://scan.coverity.com/faq#how-get-project-included-in-scan

**Describe alternatives you've considered**
Red Hat has an internal instance of coverity, but that is only available for employees.  RH can sometimes provide some of the results for dyninst generated to the dyninst developers, but it would be better if the upstream dyninst developers could more directly run and review the results.

hainest:
GitHub's built-in CodeQL integration doesn't actually build the code, so it doesn't cover all of the important source file _and_ it scans abandoned Windows and FreeBSD code.

closes #1213

---

I'm adding this both as part for the OpenSSF Scorecard and as a first step toward sustained static analysis. However, CodeQL isn't nearly as complete as other tools. For example, my initial scan of a basic x86_64 build showed a single vulnerability (see #2399). Yet, a scan with cppcheck found dozens of legitimate issues. I'll fix those and then add cppcheck to the GitHub CI.


hainest:
@kupsch @bbiiggppiigg 

For some reason, the action didn't update so I can't test this before merging. I may have to make subsequent PRs, if it fails. :angry:

bbiiggppiigg (commented):
Reviewed `git diff $(git merge-base origin/master origin/thaines/add_codeql_config)...origin/thaines/add_codeql_config` — one new file, `.github/workflows/codeql.yaml` (+64), across 2 commits. Scope is clean and single-purpose. Findings are concentrated in correctness and in divergence from the repo's own workflow conventions.

---

## Must-fix

**1. `.github/workflows/codeql.yaml:56-59` — the `c-cpp` leg cannot build; `compiler` is a required input and isn't passed.**

`.github/actions/build/action.yaml` declares `compiler: required: true`, and every other caller supplies it (`pr-tests.yaml:89`, `binary-tests.yaml:95`, `build-opts.yaml:159/236/283/326`, and the README example at `README.md:409`). A missing required input on a composite action is only a runner *warning*, so the job proceeds with `inputs.compiler == ''`. In `build-ubuntu/action.yaml` the `case "${{ inputs.compiler }}"` then matches no branch, `cc`/`cxx` stay empty, and the build runs with `-DCMAKE_C_COMPILER= -DCMAKE_CXX_COMPILER= ` — plus an `apt install -qq -y` with no package.

```yaml
        with:
          os: ubuntu-25.04
          compiler: "gcc"
          src-dir: "${GITHUB_WORKSPACE}/src"
```

(`os: ubuntu` at line 58 happens to work because `build/action.yaml` only does `startsWith(inputs.os, 'ubuntu')`, but every other caller passes the full `<distro>-<version>`; passing `ubuntu-25.04` keeps it correct if that action ever grows version-dependent logic.)

**2. `.github/workflows/codeql.yaml:21-23` — `packages: read` is missing, so the container pull will be denied.**

Once a `permissions:` block is present, everything not listed is `none`. The job pulls `ghcr.io/dyninst/amd64/ubuntu-25.04-base:latest` with `password: ${{ secrets.github_token }}`, and that package is not anonymously pullable (an unauthenticated manifest fetch returns 403). Every other container job in this repo pairs the `credentials:` block with `packages: read` — `pr-tests.yaml:71-72`, `system-libs.yaml:37-38`, `consumers.yaml:35-36`. Add it, and add `contents: read` for `actions/checkout`:

```yaml
    permissions:
      contents: read
      packages: read
      # Results are posted to the Security tab
      security-events: write
```

**3. The workflow has never executed — there is no evidence it works.**

`gh run list --workflow=codeql.yaml` returns 404 (not on the default branch), and `gh pr checks 2400` reports no checks. The branch lives on `dyninst/dyninst`, so `workflow_dispatch` can't reach it either — `schedule`/`workflow_dispatch` workflows only dispatch from the default branch. Combined with 1 and 2, this merges a job whose first real run will be a scheduled Wednesday failure. Before merging, either run it once from a branch where it is default-branch-visible, or add a `pull_request:` trigger with `paths: ['.github/workflows/codeql.yaml']` so it self-tests on this PR. The scan described in the PR body ("my initial scan of a basic x86_64 build showed a single vulnerability") was not produced by this file as written.

---

## Should-fix

**4. Commits — fold the two into one.**

`d341029` adds the workflow with the `schedule:` block commented out; `d0b8a8b` ("Re-enable cron job", empty body) uncomments it. The first commit ships commented-out scaffolding, and "Re-enable" is meaningless in merged history since nothing was ever enabled upstream. Squash into `d341029`'s message.

**5. `.github/workflows/codeql.yaml:15` — the comment is wrong.**

`0 4 * * 3` is 04:00 **UTC**, i.e. 4 AM, not 4 pm. Neighbouring workflows get this right (`build-opts.yaml:5` → `'0 3 * * 1'  # 3AM on Monday`). Change to `# every Wednesday at 4AM UTC`.

**6. `.github/workflows/codeql.yaml:12-17` — missing top-level `permissions: {}`.**

`binary-tests.yaml:33`, `build-opts.yaml:16`, `consumers.yaml:14`, `system-libs.yaml:14`, `libabigail.yaml:26` all declare `permissions: {}` at workflow level and grant per-job. Given the stated OpenSSF Scorecard motivation, the Token-Permissions check specifically wants the top-level default.

**7. `.github/workflows/codeql.yaml:17` — missing `concurrency:` block.**

Every other scheduled/dispatchable workflow here has one:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

A multi-hour CodeQL job is exactly the case where two overlapping manual dispatches waste the most runner time.

**8. `.github/workflows/codeql.yaml:35-39` — the container is pulled for all three matrix legs.**

Only `c-cpp` builds anything; the `actions` and `python` legs (`build-mode: none`) pull the full base image to do nothing with it. Move the image into the matrix so the none-build legs run bare:

```yaml
      matrix:
        include:
        - language: actions
          build-mode: none
        - language: c-cpp
          build-mode: manual
          container: ghcr.io/dyninst/amd64/ubuntu-25.04-base:latest
        - language: python
          build-mode: none
    container:
      image: ${{ matrix.container }}
      ...
```

**9. `.github/workflows/README.md` — the new workflow isn't documented.**

The `## Workflows` section carries an entry for 11 of the repo's workflows. Add a `### CodeQL scanning ([workflows/codeql](workflows/codeql.yaml))` entry stating the cadence, the single scanned configuration, and where results land.

---

## Nits (optional)

- `codeql.yaml:4` — "permuation" → "permutation"; `:4-6` "all of the important source file" → "source files".
- `codeql.yaml:64` — `${{matrix.language}}` is missing the inner spaces used everywhere else in the file.
- `codeql.yaml:36` — `ubuntu-25.04` is hardcoded while `.github/actions/os-versions` already publishes `latest` (`["ubuntu-25.04", "fedora-43"]`). When `latest` advances this file will silently keep scanning an older distro. Either derive it from `os-versions`, or note in the header comment that the pin is deliberate and must be bumped by hand — the "Adding a new OS version" checklist at `README.md:266-280` is where that reminder belongs.
- `codeql.yaml:57-59` — the build action's `num-jobs` defaults to 2. A traced CodeQL build of all of Dyninst at `-j2` will be the longest job in this repo by a wide margin; `num-jobs: 4` matches the runner's vCPU count and buys headroom against the 6-hour job limit.
- `codeql.yaml:47-52` — no `queries:` is set, so this runs the `default` suite only. If the intent is "a first step toward sustained static analysis", `queries: security-and-quality` is closer to the cppcheck-class findings you describe wanting.
- Triggers are `schedule` + `workflow_dispatch` only, so a vulnerability introduced in a PR isn't surfaced until the following Wednesday. Defensible as a cost tradeoff, but the header comment explains only why *one configuration* is scanned, not why it's weekly-only — worth one more sentence there.

---

Nothing here belongs in a separate branch; the change is correctly scoped to one file and one purpose, and the cppcheck follow-up is already planned as separate work.

bbiiggppiigg:
@hainest again for something like this I have no knowledge and I only give you the AI feedback as a reference.

bbiiggppiigg (dismissed):

hainest:
> 1. .github/workflows/codeql.yaml:56-59 — the c-cpp leg cannot build; compiler is a required input and isn't passed.

Derp. Fixed.

> 2. .github/workflows/codeql.yaml:21-23 — packages: read is missing, so the container pull will be denied.

It _used_ to be there. I must have deleted it. Fixed.

> 3. The workflow has never executed — there is no evidence it works.

See my previous comment.

> 6. .github/workflows/codeql.yaml:12-17 — missing top-level permissions: {}.

Bah. I'm not yet accustomed to remembering that. Fixed.

> 7. .github/workflows/codeql.yaml:17 — missing concurrency: block.

This won't be run by PRs or pushes, so it doesn't need one.

> 8. .github/workflows/codeql.yaml:35-39 — the container is pulled for all three matrix legs.

I'd rather all tests run in a known Dyninst environment.

> 9. .github/workflows/README.md — the new workflow isn't documented.

Derp. Fixed.

> codeql.yaml:36 — ubuntu-25.04 is hardcoded

That's intentional. There is no use in running these checks on both Ubuntu and Fedora. I could probably
use the os-version action and then filter it, but I'd rather get this running now.

> :4-6 "all of the important source file" → "source files".

That text doesn't exist. The chatbox is hallucinating.

> codeql.yaml:47-52 — no queries: is set, so this runs the default suite only. If the intent is "a first step toward sustained static analysis", queries: security-and-quality is closer to the cppcheck-class findings you describe wanting.

The [docs](https://docs.github.com/en/code-security/concepts/code-scanning/codeql/codeql-query-suites#default-query-suite) not that "The queries in the default query suite are highly precise and return few false positive code scanning results. Relative to the security-extended query suite, the default suite returns fewer low-confidence code scanning results." I'd rather have fewer checks with fewer false positives.

> Triggers are schedule + workflow_dispatch only, so a vulnerability introduced in a PR isn't surfaced until the following Wednesday. Defensible as a cost tradeoff, but the header comment explains only why one configuration is scanned, not why it's weekly-only — worth one more sentence there.

Once ready, cppcheck will run on PR/push.

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,68 @@
...
+# here just scans a basic x86_64 build using Ubuntu 25.04.
+#
+# Results are reported in the Security tab.
+
+name: "CodeQL"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 4 * * 3'  # every Wednesday at 4pm
```
bbiiggppiigg: I think it should be 4 a.m.?

hainest: Derp. Fixed.

bbiiggppiigg (approved):

kupsch (approved):

PR #2398, opened 2026-09-21
[SymtabAPI] Remove unused functions in Symtab

hainest:

PR #2401, opened 2026-09-22
instructionAPI: decode aarch64 SIMD&FP store-pair with 64-bit (D) operands

bbiiggppiigg:
## Problem

`InstructionDecoder` cannot decode **stores of 64-bit SIMD&FP (D) register pairs** on aarch64.
All four forms — signed offset, pre-index, post-index and non-temporal — come back
`*** INVALID ***`, while every `LDP`/`LDNP` form and the 32-bit (S) and 128-bit (Q) stores
decode correctly:

```
stp d8,d9,[x0,#112]  (real glibc word)  0x6d072408 -> *** INVALID ***
stp  d, signed-offset                   0x6d010402 -> *** INVALID ***
stp  d, pre-index                       0x6d810402 -> *** INVALID ***
stp  d, post-index                      0x6c810402 -> *** INVALID ***
stnp d, non-temporal                    0x6c010402 -> *** INVALID ***
ldp  s / d / q                                     -> decode fine
stp  s / q                                         -> decode fine
stp  x0, x1, [x2,#16]  (GPR pair)                  -> decodes fine
```

## Root cause

`OPRopc()` rejected `opc=01` with `L=0` for **every** load/store pair. That is right for
general-purpose pairs, where `opc=01` is `LDPSW` and so has no store form, but wrong for
SIMD&FP pairs, where `opc=01` selects the 64-bit D-register operand size and is allocated for
loads and stores alike in all four addressing modes. The neighbouring non-temporal check was
already guarded by `isSIMDInsn`; this one was not.

The opcode tables are fine — all four D-pair encodings already reach the correct
`aarch64_op_stp_fpsimd` / `aarch64_op_stnp_fpsimd` leaves. The rejection happens later, in
operand decoding.

## Why it matters: a rewritten libc that never calls `main`

glibc aarch64 `__sigsetjmp` saves the callee-saved FP registers with four
`stp d<n>, d<n+1>, [x0, #imm]`. Those are the **only** four `stp d`-pair instructions in the
whole of `libc-2.31.so`, and they are all in that one function. So:

1. the undecodable word truncates the parse of `__sigsetjmp` → classified `NORETURN`;
2. `_setjmp` and `setjmp` tail-branch into it and inherit that;
3. `__libc_start_main`'s `bl _setjmp` loses its `CALL_FT` edge;
4. the block containing the call to `main` is dropped from the CFG;
5. `relocateFunction()` emits only CFG blocks, so the **rewritten `libc.so.6` never calls
   `main`** and spins forever in a cold block reached by accident.

This hangs `test_reloc` on aarch64 and leaves a complete-but-poisoned `libc.so.6` in the
testsuite's `binaries/`, which then hangs every attach-mode test that picks it up.

## Verification (aarch64, Ubuntu 20.04, gcc-11, base `7e53b6c88`)

**Decode** — all four previously-invalid encodings now decode with correct mnemonic, registers
and scaled immediate; every previously-valid form is unchanged, `opc=11` stays unallocated for
both V values, GPR `opc=01` stores stay invalid, and `ldpsw` stays valid.

**Parse** — on `libc-2.31.so`, `__sigsetjmp` goes from `NORETURN` to `RETURN` and its block
reaches the tail branch to `__sigjmp_save`; `__libc_start_main`'s `bl _setjmp` block regains
its `CALL_FT` edge, recovering the tail of the function including the call to `main`.

**End to end** — same command, same machine, differing only in which Dyninst rewrote the
library:

| libc rewritten by | `timeout 20 env LD_LIBRARY_PATH=<binaries>:<install>/lib /bin/echo HELLO` |
|---|---|
| master | no output, exit 124 (hung) |
| this branch | `HELLO`, exit 0 |

**`test_reloc`** goes from 2 result rows and a hang to **12/12 PASSED**.

**No regression** — in-tree ctest **134/134**, matching the baseline for this machine. A
rewriter-mode spot check of `test1_1`, `test1_2`, `test5_1` across both compilers and both
PIC settings gives 20 rows, all PASSED, byte-identical to master.

## Test

`tests/integration/InstructionAPI/decoder/aarch64/ldst.cpp` gains a SIMD&FP pair block: the
four previously-broken D forms plus an S and a Q store to pin the whole operand-size axis.
It is a real regression test — those four words decode as invalid without this change, so the
expected read sets fail against master.

## Not verified

Built and tested with gcc-11 only; no clang build and no CI-image gate was run. `OPRopc` is
aarch64-only, so no other architecture is reachable from this change.


bbiiggppiigg:
## Cross-architecture verification

Filling in two gaps from the description: a full per-test run on aarch64 (it previously showed only
a 3-test rewriter spot check), and confirmation that the change disturbs nothing on the other two
architectures. The clang gap noted at the bottom of the description is covered by CI passing.

### aarch64 — full per-test run, single variable

`test_driver -test T -all` per test, 300 s cap, every test in the platform table.
Both sides use testsuite `017e5e0e`; the only difference is dyninst `7e53b6c88` → `ac05e59e1`.

| | master | this branch |
|---|---|---|
| result rows | 1784 | **1802** |
| PASSED | 1778 | **1796** |
| HUNG | 3 | **2** |
| FAILED | 2 | 3 |
| SKIPPED | 1 | 1 |

**0 regressions** — no test that passed on master got worse.

`test_reloc`, the test this PR is about, goes from **2 rows then a hang** to **all 12 variants
PASSED**:

```
master:        after:
create  PASSED create  PASSED   (x4 compiler/PIC combinations)
attach  PASSED attach  PASSED
HUNG           rewriter PASSED
```

**On the FAILED count moving 2 → 3** — this is not a regression, and it is the same pattern the
`pc_irpc` fix produced on ppc64le. `pc_singlestep` hung on master partway through its matrix; with
this branch it gets *further* and reaches static-linkage cells master never executed. Four of those
newly-reached cells pass, two fail, and one failure master did have (`attach MPMT dynamic nonPIC`)
now passes. I would not attribute `pc_singlestep`'s movement to this patch in either direction: that
test has a moving boundary independent of any change — five isolated reps on unmodified master gave
32 / 25 / 32 rows.

The two remaining hangs are unrelated to this change: `pc_irpc` (fixed separately by
dyninst/testsuite#280, already merged — this run deliberately pins the pre-merge testsuite to keep
the comparison single-variable) and `pc_singlestep` above.

### x86_64 and ppc64le — in-tree ctest

The aarch64 decoder is compiled into the libraries on both, and the in-tree suite contains aarch64
decoder tests, so this is not merely a build check — `instructionAPI_aarch64_ldst_test`, the file
this PR extends with six new cases, **executes and passes on both**:

| host | result | the modified test |
|---|---|---|
| x86_64 (Ubuntu 20.04, gcc-11) | **136/136 passed, 0 failed** | `instructionAPI_aarch64_ldst_test` Passed |
| ppc64le (POWER9, RHEL8) | **110/110 passed, 0 failed** | `instructionAPI_aarch64_ldst_test` Passed |

Fourteen aarch64 tests ran on each host; all passed.

### Summary

| arch | check | result |
|---|---|---|
| aarch64 | full per-test | 1802 rows, 0 regressions, `test_reloc` hang → 12/12 PASSED |
| x86_64 | in-tree ctest | 136/136 |
| ppc64le | in-tree ctest | 110/110 |

---

### Update: re-confirmed on current master

The per-test run above was taken at dyninst `7e53b6c88`. Master has since advanced to
`c041b5317` (`[SymtabAPI] Remove unused functions in Symtab` #2398, which does not touch the
decoder). A fresh per-test baseline of unpatched master at that tip still shows the bug:

```
test_reloc ALL_REMAINING_VARIANTS HUNG(timeout 300s)
```

so the PR is not stale. For context from the same baseline set, all four hosts pinned to
`c041b5317` + testsuite `d6445ce9`:

| host | rows | PASSED | non-PASSED on unpatched master |
|---|---|---|---|
| x86_64 | 1546 | 1544 | **none — fully clean** |
| x86_64 (2nd host) | 1535 | 1532 | `test_reloc` only (a *different*, host-specific relocation/liveness spin — not this decoder bug) |
| aarch64 | 1799 | 1793 | `test_reloc` (this bug), `pc_singlestep`, `test_pt_ls`, `pc_irpc` |

On aarch64, `test_reloc` is the only one of those four this PR addresses; the other three are
pre-existing and unrelated (two are documented flaky there, and `pc_irpc` is slow rather than
hung — it completes 32/32 given a 900 s cap instead of the harness default 300 s).

bbiiggppiigg:

kupsch:
The size of the memory operand is incorrect in some cases, should this be fixed here also:
```
  29000861  stp w1, w2, [x3]      reported  8 bytes, actual  8  ok              
  a9000861  stp x1, x2, [x3]      reported 16 bytes, actual 16  ok              
  69400861  ldpsw x1, x2, [x3]    reported  8 bytes, actual  8  ok              
  2d000861  stp s1, s2, [x3]      reported  8 bytes, actual  8  ok              
  6d0127e8  stp d8, d9, [sp,#16]  reported  8 bytes, actual 16  WRONG           
  6d400861  ldp d1, d2, [x3]      reported  8 bytes, actual 16  WRONG           
  6c000861  stnp d1, d2, [x3]     reported  8 bytes, actual 16  WRONG           
  ad000861  stp q1, q2, [x3]      reported 16 bytes, actual 32  WRONG           
  ad400861  ldp q1, q2, [x3]      reported 16 bytes, actual 32  WRONG           
```

bbiiggppiigg:
> The size of the memory operand is incorrect in some cases, should this be fixed here also:
> 
> ```
>   29000861  stp w1, w2, [x3]      reported  8 bytes, actual  8  ok              
>   a9000861  stp x1, x2, [x3]      reported 16 bytes, actual 16  ok              
>   69400861  ldpsw x1, x2, [x3]    reported  8 bytes, actual  8  ok              
>   2d000861  stp s1, s2, [x3]      reported  8 bytes, actual  8  ok              
>   6d0127e8  stp d8, d9, [sp,#16]  reported  8 bytes, actual 16  WRONG           
>   6d400861  ldp d1, d2, [x3]      reported  8 bytes, actual 16  WRONG           
>   6c000861  stnp d1, d2, [x3]     reported  8 bytes, actual 16  WRONG           
>   ad000861  stp q1, q2, [x3]      reported 16 bytes, actual 32  WRONG           
>   ad400861  ldp q1, q2, [x3]      reported 16 bytes, actual 32  WRONG           
> ```

probably.
I'm running the regression test on the fixes for these right now.

bbiiggppiigg:
Good catch — fixed in e3fa48327, kept as a separate commit so the change since your review stands
on its own.

`getMemRefPair_RT` decodes `opc` from bits 31/30 but never consults the V bit, so it applies the
general-purpose size table to SIMD&FP pairs. Your table, re-measured before and after:

| encoding | instruction | before | after | actual |
|---|---|---|---|---|
| `29000861` | `stp w1, w2, [x3]` | 8 | 8 | 8 |
| `a9000861` | `stp x1, x2, [x3]` | 16 | 16 | 16 |
| `69400861` | `ldpsw x1, x2, [x3]` | 8 | 8 | 8 |
| `2d000861` | `stp s1, s2, [x3]` | 8 | 8 | 8 |
| `6d0127e8` | `stp d8, d9, [sp,#16]` | **8** | **16** | 16 |
| `6d400861` | `ldp d1, d2, [x3]` | **8** | **16** | 16 |
| `6c000861` | `stnp d1, d2, [x3]` | **8** | **16** | 16 |
| `ad000861` | `stp q1, q2, [x3]` | **16** | **32** | 32 |
| `ad400861` | `ldp q1, q2, [x3]` | **16** | **32** | 32 |

Two things worth noting for anyone adding coverage here. The `stp s1, s2` row is correct only by
coincidence — two 4-byte registers happen to match what the general-purpose table returns for
opc=00 — so a test covering only the S case passes against the broken code and must include D and
Q. And this affects loads as well as stores: `ldp`/`ldnp` report the wrong size despite having
always decoded correctly, so it is wider than the store-side decode bug this PR started as.

The neighbouring `makeMemRefPair_offset7` already checks `isSIMDInsn` for its scale, so this was
an inconsistency inside the same function pair rather than a missing feature.

**Verification so far:** the nine encodings above; the decode-validity fix from the first commit
is unaffected (no new INVALIDs, all four D-pair forms still decode); in-tree ctest **134/134** on
aarch64, matching that host's baseline.

**Not yet run:** the external per-test testsuite for this second commit. The first commit has it
(1802 rows, 0 regressions on aarch64), but operand sizes feed analysis, so this one deserves the
same treatment and I have not done it yet — the machines are tied up on an unrelated run. I will
post those numbers when they exist rather than imply they are already covered.

bbiiggppiigg:
Added coverage for this in 956d1e47d — the previous commit fixed the sizes but left them
unprotected, since the decoder test only checked register read/write sets.

It checks your nine encodings directly rather than extending the 142-row table, which keeps the
cases together and leaves the table's indices alone.

Verified it actually catches the bug: with the fix reverted and everything else unchanged, it
fails with exactly your five rows —

```
Verifying memory operand size of 'stp d8, d9, [sp,#16]'
  expected 16 bytes, got 8
Verifying memory operand size of 'ldp d1, d2, [x3]'
  expected 16 bytes, got 8
Verifying memory operand size of 'stnp d1, d2, [x3]'
  expected 16 bytes, got 8
Verifying memory operand size of 'stp q1, q2, [x3]'
  expected 32 bytes, got 16
Verifying memory operand size of 'ldp q1, q2, [x3]'
  expected 32 bytes, got 16
```

and passes with it restored.

The `stp s1, s2` row is in there deliberately, as a guard against a future "simplification": two
4-byte registers coincidentally match what the general-purpose table returns for opc=00, so a test
covering only the S case would pass against the broken decoder. The general-purpose rows are
controls.

Still outstanding on this PR: the external testsuite run for the two commits after the first. The
first commit has one (aarch64 per-test, 1802 rows, 0 regressions); the size fix and this test have
only the in-tree suite so far. The machines are busy with an unrelated run; I will post those
numbers when they exist.

kupsch (approved):

bbiiggppiigg:
## External testsuite on aarch64 for the two post-review commits

This is the run promised above for `e3fa48327` (SIMD&FP pair operand sizes) and `956d1e47d` (their test). It found no regressions.

**Setup:** each test of the dyninst testsuite at `d6445ce9` ran in isolation with a 300 s timeout, on zeroah (aarch64, Ubuntu 20.04). The PR head `956d1e47d` was compared with a same-day master baseline at `c041b5317`, using the same testsuite commit. The only other source difference between the two trees is #2398, which removes unused Symtab functions and touches no file in this PR.

| | master | this PR |
|---|---|---|
| PASSED | 1811 | 1811 |
| `test_reloc` | 1 row: HUNG | **12/12 PASSED** |
| unchanged non-PASSED | `pc_irpc` HUNG (needs ~540 s against the 300 s cap), `test_pt_ls` create FAILED | same |
| `pc_singlestep` | 4 cells FAILED | 2 cells FAILED, then the rest of the test HUNG |

**The `pc_singlestep` delta is not from this PR:**
- It went 2 cells PASSED→FAILED and 1 FAILED→PASSED, then hung before its last 12 cells.
- Its complete shared-library chain is libpcontrol, libsymtabAPI, libcommon, libelf and libdw. It never loads libinstructionAPI, which contains `InstructionDecoder-aarch64.C`.
- It is unstable on master alone. In 5 isolated master reps on this host earlier today, 2 had failures and 1 of those stopped early the same way.

So on aarch64 the full PR now matches the first commit's result: `test_reloc` goes from hung to 12/12 PASSED, with 0 regressions. With the in-tree ctest already reported above, all three commits now have external evidence.

Still not done for this PR: a clang build, and the newer-toolchain CI container check.

PR #549, opened 2019-02-10
Attempts to fix legacy test failures in Dyninst test suite

mxz297:
This PR attempts to fix test failures listed in #550.

Issue #914, opened 2020-11-05
Crash in maybe_split_for_insert when processing symtab when using OpenMP


woodard:
When processing the symtab for /lib64/libgfortran.so.5.0.0 from libgfortran-10.2.1-6.fc33.x86_64 when using OpenMP (On a 12 thread 6 core machine) I get the following crash:

```
Thread 2 "repo1" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff1d02640 (LWP 149495)]
0x00007ffff78f89ab in maybe_split_for_insert (parentp=parentp@entry=0x29be060, gparentp=gparentp@entry=0x29bdd90, p_r=p_r@entry=1, gp_r=gp_r@entry=1, mode=mode@entry=1, rootp=<optimized out>, rootp=<optimized out>) at tsearch.c:229
229           if (parentp != NULL && RED(DEREFNODEPTR(parentp)))
#0  0x00007ffff78f89ab in maybe_split_for_insert (parentp=parentp@entry=0x29be060, gparentp=gparentp@entry=0x29bdd90, p_r=p_r@entry=1, gp_r=gp_r@entry=1, mode=mode@entry=1, rootp=<optimized out>, rootp=<optimized out>) at tsearch.c:229
#1  0x00007ffff78f8beb in __GI___tsearch (key=0x7fffecf74fc0, vrootp=<optimized out>, compar=0x7ffff776eb90 <findcu_cb>) at tsearch.c:359
#2  0x00007ffff776ee21 in __libdw_intern_next_unit (dbg=dbg@entry=0x4836a0, debug_types=debug_types@entry=false) at libdw_findcu.c:216
#3  0x00007ffff776f137 in __libdw_findcu (dbg=0x4836a0, start=41087368, v4_debug_types=<optimized out>) at libdw_findcu.c:250
#4  0x00007ffff775f9de in __libdw_offdie (dbg=<optimized out>, offset=<optimized out>, result=0x7ffff1d00a70, debug_types=<optimized out>) at dwarf_offdie.c:61
#5  0x00007ffff7765326 in dwarf_formref_die (attr=0x7ffff1d00b00, result=0x7ffff1d00a70) at dwarf_formref_die.c:75
#6  0x00007ffff77610c0 in dwarf_attr_integrate (die=0x7ffff1d00c00, search_name=3, result=0x7ffff1d00b00) at dwarf_attr_integrate.c:54
[...]

PR #2402, opened 2026-09-22
PoC/RFC: thread_db is never enabled on glibc-hwcaps systems (library matched by link-map basename, not SONAME)

bbiiggppiigg:
> **Draft / proof of concept.** The diagnosis below is solid and reproducible; the patch is
> deliberately the cheap version and its regression risk is **unmeasured**. Opening it as a
> draft to surface the bug and invite a view on which fix is wanted.

## Problem

**On a `glibc-hwcaps` system, Dyninst never enables `libthread_db`**, so for every
multithreaded debuggee it silently loses `UserThreadCreate`/`UserThreadDestroy` events, TLS
resolution (`td_thr_tls_get_addr`) and `int_thread::getStartFunction()`. The failure is
swallowed — nothing is reported to the caller.

## Root cause

`libthread_db` asks the debugger for objects by **SONAME**: `td_ta_new()` begins with
`td_lookup(ps, LIBPTHREAD_SO /* "libpthread.so.0" */, …)` through `ps_pglobal_lookup`.

`thread_db_process::getSymbolAddr()` answers by comparing that name against the **link-map
name** (`l_name`) — the path the loader actually opened — reduced to its basename:

```c
if (stripLibraryName(l->getName().c_str()) == name) { lib = l; break; }
```

Those agree only when the loader resolved the dependency through the SONAME-named symlink.
With `glibc-hwcaps` it does not:

```
$ ldd <mutatee>
    libpthread.so.0 => /lib64/glibc-hwcaps/power9/libpthread-2.28.so
    libc.so.6       => /lib64/glibc-hwcaps/power9/libc-2.28.so
```

`ldconfig` records the hwcaps entry under its **real filename**, so the link-map basename is
`libpthread-2.28.so`, the compare fails, `getSymbolAddr` returns `PS_ERR`, `td_ta_new()`
returns `TD_NOLIBTHREAD`, and thread_db is never enabled. The `TD_CREATE` event breakpoint is
therefore never planted and no thread-create event can ever be delivered.

Note the inconsistency this exposes: `isSupportedThreadLib()` finds the very same library with
a tolerant `libName.find("libpthread") != npos`, while `getSymbolAddr()` loses it with an
exact compare. **The two predicates disagreeing is the bug.**

### Regression point

`2a4bbf763` (#549, Mar 2019) simultaneously replaced a hand-rolled `stripLibraryName` that
truncated at the first `.` or `-` with `boost::filesystem::path::filename()`, **and** replaced
`strstr(...)` with `==`. Both tolerances went at once. Pre-#549 this would have matched.

## Evidence

Same unmodified Dyninst `7e53b6c88`, same binaries, **only the library search order differs**
(`pc_breakpoint`, create/SPMT — not the test this was found through, so it is not test-specific):

| `LD_LIBRARY_PATH` | loader resolves to | result |
|---|---|---|
| stock | `libpthread-2.28.so` | **stalled, killed at 150 s, no verdict row** |
| `/lib64` first | `libpthread.so.0` symlink | **PASSED in 1 s** |

Debug trace on the failing side:

```
[sysv.C]           Creating new library object for /lib64/glibc-hwcaps/power9/libpthread-2.28.so
[int_thread_db.C]  Creating threadAgent
[int_thread_db.C]  Looking up symbol nptl_version in libpthread.so.0
[int_thread_db.C]  Didn't yet find loaded thread library
[int_thread_db.C]  Debuggee isn't multithreaded at this point, libthread_db not enabled
```

With the patch, on the stock path, the agent is created and 132 thread-create events flow.

## Blast radius

This is not architecture-specific — it is a **glibc-layout** trigger. RHEL 8/9 ships
`glibc-hwcaps/power9` by default on ppc64le; an x86_64 host with `x86-64-v3` hwcaps libraries
installed would hit it too.

In our testsuite it takes out any proccontrol test running `create` with a multithreaded group.
`pc_stat` is the sole survivor only because `setupStatTest()` turns off thread tracking and the
startup thread check for it. With the patch, `pc_irpc`, `pc_breakpoint`, `pc_tls`, `pc_thread`
and `test_thread_2` all pass on that machine — i.e. a real share of what we had recorded as a
"flaky hang cluster" is this single defect.

## The patch, and what I would rather do

This adds a stem fallback (truncate at the first `.`/`-`) after the exact-basename compare
fails, restoring pre-#549 behaviour and re-aligning `getSymbolAddr()` with
`isSupportedThreadLib()`.

**The principled fix is to match the real `DT_SONAME`**, which is what `libthread_db` is
actually asking for. That is blocked today: `common/h/SymReader.h` has no SONAME accessor,
though SymtabAPI already reads it. I did not add that interface unprompted.

Worth considering separately:
- `plat_supportThreadEvents()` returns true merely because `libthread_db.so` dlopened in the
  mutator; it never checks whether *this process* has an agent. Making it per-process would
  turn this class of failure into "unsupported" rather than an infinite wait.
- The `TD_NOLIBTHREAD` at `int_thread_db.C:678` is swallowed silently even when the process
  demonstrably has a libpthread loaded.
- A complete fix must cover `libpthread.so.0`, `libc.so.6` **and** the rtld — libc is on the
  hwcaps path too.

## Not verified

- **Regression risk unmeasured**: only 5 test cells were run with the patch; no full-matrix
  comparison against a same-tip master baseline.
- The stem compare is looser than an exact one and could in principle match an unrelated
  library sharing a stem. This is the main reason it is a proof of concept.
- That pre-#549 code actually worked on a hwcaps box is inferred from reading it, not measured.


bbiiggppiigg:
## Per-test validation on three hosts

Isolated per-test runs (`test_driver -test T -all` per test, 300 s cap), patched vs master,
**single variable**: dyninst `7e53b6c88` → `d0f6b6b56`, testsuite pinned at `017e5e0e` on **both**
sides everywhere. All three runs completed; none truncated.

### The host that has the bug — cyclops (POWER9, RHEL8, `/lib64/glibc-hwcaps/power9` present)

| | master `7e53b6c88` | patched `d0f6b6b56` |
|---|---|---|
| result rows | 865 | **1061** |
| PASSED | 832 | **1058** |
| FAILED | 14 | **0** |
| HUNG | 19 | **2** |

**17 of the 19 hangs and all 14 failures clear**, for 226 additional passing rows. Verdict-class
diff: **0 regressions**, 14 fixes, 0 other class changes.

Cleared, and every variant now passes: `pc_addlibrary`, `pc_breakpoint`, `pc_detach`, `pc_fork`,
`pc_fork_exec`, `pc_groups`, `pc_launch`, `pc_library`, `pc_singlestep`, `pc_temp_detach`,
`pc_terminate`, `pc_terminate_stopped`, `pc_thread`, `pc_thread_cont`, `pc_tls`, `test_thread_2`,
`test_thread_6`. The 14 failures that cleared: `pc_tls` ×2, `test_thread_3` ×4, `test_thread_8` ×8.

Still not passing on the patched side, none of them this change:
- `pc_irpc` — a separate testsuite bug, fixed by dyninst/testsuite#280 (merged). Its fix was
  deliberately excluded here to keep the diff single-variable, so it still hangs in this run.
- `test_pt_ls` — documented flake, hangs on both trees.
- `test1_27` `DRIVER_DIED(signal 11)` — flake-classified 5 isolated reps on **both** trees:
  **1/5 on master, 1/5 patched**. Identical rate, matches the documented "1 in 5 on clean master".

### The hosts that do not have the bug — regression gate

Neither has a `glibc-hwcaps` directory, so the exact-basename compare already succeeds and the new
fallback should never execute. That is the prediction, and it holds:

| host | master | patched | diff |
|---|---|---|---|
| coriander (x86_64) | 1512 P / 1 H / 2 S | **identical** | **empty** |
| zeroah (aarch64) | 1778 P / 2 F / 3 H / 1 S | 1796 P / 1 F / 2 H / 1 S | 0 regressions, 1 fix |

coriander is a byte-identical verdict set. Every zeroah delta is inside `pc_singlestep` and in the
*improving* direction (a hang replaced by passing rows).

### Mechanism, measured rather than inferred

With `DYNINST_DEBUG_PROCCONTROL=1` on the patched tree:
- **cyclops: 2466 fallback hits, exactly ONE distinct match** —
  `libpthread.so.0` → `/lib64/glibc-hwcaps/power9/libpthread-2.28.so` by stem `libpthread`.
  No spurious stem collisions on any other object.
- **coriander and zeroah: 0 fallback hits.** The added path provably never runs there.

### Honest caveats

- A 5-rep `pc_singlestep` A/B on **zeroah** had the patched side terminating abnormally more often
  (4/5 vs 1/5). I am not calling it a regression, because the probe shows the new code never
  executes on that host, the reps were not interleaved, and the full per-test run went the other
  way — but `pc_singlestep` is unstable there on **both** trees (master alone gave 3, 2, 0, HUNG, 1
  failures across 5 reps). An interleaved n≥20 A/B would settle it; worth doing independently of
  this PR.
- Only ppc64le hwcaps is exercised. No lab machine has `glibc-hwcaps/x86-64-v3`, so that layout is
  untested.

### Why this is still a draft

The measured weaknesses are in the patch, not the diagnosis:
1. The principled key is the real `DT_SONAME`, not a filename stem — needs a SONAME accessor on
   `SymReader`. That is the actual work item.
2. Stem matching is ambiguous **and order-dependent**: the loop takes the first match while
   iterating a pointer-ordered `std::set`, so with two stem-equal candidates the winner is
   arbitrary and can differ run to run. It should at least detect multiple matches and refuse,
   or prefer an exact-SONAME candidate.
3. The fallback is unconditional — it runs for every object request that misses the exact compare,
   not only SONAME-shaped ones.
4. No test exercises a hwcaps link-map layout, so today only cyclops would catch a regression here.

### One side finding worth recording

The long-standing cyclops "19-test flaky hang cluster" — which our tooling excludes wholesale from
grouped runs — **was not flaky**. 17 of the 19 were this one deterministic bug. That exclude list
should be revisited once this (or its principled successor) lands.

bbiiggppiigg:
## External testsuite, per-test isolated, four machines

Branch `8ddc5ca1c` against each host's master baseline (dyninst `c041b5317`), testsuite
`d6445ce9` on both sides everywhere. Per-test isolation (`test_driver -test T -all` per test,
300 s cap) rather than grouped `runTests -all`, because a grouped run truncates at the first
group over the 600 s limit and that is exactly the proccontrol cluster this change moves.

| host | arch | ctest | baseline | branch |
|---|---|---|---|---|
| cyclops | ppc64le POWER9, **glibc-hwcaps** | 114/114 | 834 P / 14 F / 19 H | **1084 P / 0 F / 1 H** |
| coriander | x86_64 | 140/140 | 1544 P / 2 S | 1544 P / 2 S (identical) |
| zeroah | aarch64 | 138/138 | 1811 P / 5 F / 2 H / 1 S | 1786 P / 1 F / 3 H / 1 S |
| mahir | x86_64 el8 | 140/140 | 1532 P / 1 H | 1517 P / 1 H / 1 DRIVER_DIED |

### cyclops — the host where this code path actually executes

**Every failure and 18 of the 19 hangs clear.** The only survivor is `test_pt_ls`, a
documented flake there. No genuine subtest failures in the detail logs at all.

Coverage rises too, which matters — a run can look better simply by testing less:

| mode | baseline rows | branch rows |
|---|---|---|
| create | 394 | **498** |
| attach | 238 | **370** |
| rewriter | 180 | 180 |
| disk | 36 | 36 |

A hung test is killed at the cap, so its remaining variants never run. With a working thread
agent those tests finish and their variants execute: +250 passing rows of work the baseline
never reached, concentrated in `create` and `attach`, which is where thread events matter.

The cleared set is the predicted one — `test_thread_8` (8 rows), `test_thread_3` (4),
`pc_tls` (3), then one row each across `pc_breakpoint`, `pc_fork`, `pc_fork_exec`,
`pc_detach`, `pc_launch`, `pc_groups`, `pc_library`, `pc_terminate`,
`pc_terminate_stopped`, `pc_temp_detach`, `pc_thread`, `pc_thread_cont`, `pc_irpc`,
`pc_singlestep`, `test_thread_2`, `test_thread_6`. That is the signature of **one** missing
thread agent, not twenty unrelated defects.

### zeroah — the new code provably never runs there

Failures drop 5 → 1. The one class change is `pc_singlestep`, which moves from 4 failing
variants to hanging as a whole; it is unstable on that host on both trees and flips run to
run.

Rather than argue from plausibility, I instrumented it. Under
`DYNINST_DEBUG_PROCCONTROL=1`, over **1269** thread-library lookups:

* SONAME fallback executions: **0**
* `glibc-hwcaps` directory on the host: **absent**

The fallback is guarded by the ordinary name comparison failing, so on a normal library
layout it is dead code. It cannot be the cause of anything observed on this host.

### coriander

Byte-identical verdict sets. Same conclusion as zeroah, and the same reasoning: no hwcaps
layout, so the fallback never fires.

### mahir — one open item, possibly mine

`test1_27` killed `test_driver` with SIGSEGV once in the full run. It is recorded as a
pre-existing flake (non-reproducing on this host across two ~900-config runs; 1-in-5 on
both trees on ppc64le). But isolated repetitions do **not** match that description:

| tree | driver crashes |
|---|---|
| branch | **4 / 15** |
| master, same host, same cap | **0 / 15** |

4/15 against 0/15 is not something I am willing to wave through on the strength of the
flake record, even though the record is real and the rate may simply have moved. There is
also a candidate mechanism: this PR makes `Symtab::extractInfo` read `DT_SONAME` for
**every** object, and that pointer aims into the mapped ELF string table, so it is a new
unconditional dereference on a path that previously only ran in narrower circumstances.

I am reproducing it with cores now and will post the stack. **Treat this PR as not ready
until that is resolved** — if it is mine, it gets fixed before anything else happens here.

### Not covered

No test exercises a link map where the loader redirected a dependency, because building one
needs a hwcaps directory and therefore root. Only cyclops would catch a regression in the
matcher itself. The in-tree tests added here cover the name plumbing beneath it.

bbiiggppiigg:
## Rebased onto current master; matched four-host rerun

Rebased onto `d1f8126d9` (after #2401 merged). Diff unchanged: 242 insertions across 15 files.

The earlier numbers compared branch and baseline on testsuite `d6445ce9`. Since then the
baselines moved to current master and testsuite `c694a8f7`, so **every host has been rebuilt
and rerun on both sides** — branch `fb547ce3a` against master `d1f8126d9`, testsuite
`c694a8f7` everywhere, per-test isolated, 300 s cap.

| host | arch | ctest | baseline (master) | branch |
|---|---|---|---|---|
| cyclops | ppc64le POWER9, **glibc-hwcaps** | 114/114 | 834 P / 14 F / 19 H | **1084 P / 0 F / 1 H** |
| mahir | x86_64 el8 | 140/140 | 1532 P / 1 H / 2 S | 1532 P / 1 H / 2 S |
| coriander | x86_64 | 140/140 | 1541 P / 1 DRIVER_DIED / 2 S | 1544 P / 2 S |
| zeroah | aarch64 | 138/138 | 1824 P / 2 F / 1 H / 1 S | 1824 P / 2 F / 1 H / 1 S |

### cyclops reproduced exactly

Same tally as the first run and the same per-mode row counts (create 498, attach 370,
rewriter 180, disk 36, against a baseline of 394 / 238 / 180 / 36). A result that repeats
across a rebuild, a rebase and a testsuite change is about as firm as this harness gets.
Every failure and 18 of 19 hangs clear; the survivor is `test_pt_ls`, a documented flake
there.

### The three non-hwcaps hosts are flat, and the two deltas are not this PR

**mahir** — byte-identical. The single hang is `test_reloc`, the pre-existing quadratic
liveness cost, on both sides.

**coriander** — the branch is *cleaner* than its own baseline, because the baseline run
happened to hit `test_thread_1`, a known master use-after-free that a separate PR has been
measured to fix. Flake, not signal.

**zeroah** — identical tallies. The one difference is *which* `pc_singlestep` variant fails:
baseline `gcc/attach/MPMT/static`, branch `g++/create/MPMT/dynamic`. Same count, different
cell — the signature of an unstable test, not a regression. Independently: that test also
moved under #2401, which only touches the aarch64 decoder and which `pc_singlestep` never
loads. It is under investigation separately and belongs to neither PR.

### The driver crash reported earlier is resolved, and it is not this PR

My previous comment flagged `test1_27` segfaulting `test_driver` at 4/15 on the branch
against 0/15 on master, and said this PR was not ready until that was explained. It is
explained, and the explanation clears this change:

| tree | driver crashes |
|---|---|
| branch, default threads | 4 / 15 |
| branch, `OMP_NUM_THREADS=1` | 0 / 12 |
| **untouched master tree, freshly built, default threads** | **2 / 20** |
| older master build, default threads | 0 / 30 |

It is a **data race in parallel DWARF parsing**: `dwarf_getlocations` → `is_constant_offset`
→ `tsearch`, from `DwarfWalker::decodeLocationList`, on an OpenMP worker. The rate moves with
build timing alone, which is why the first master control looked clean. An unmodified tree
reproduces it, so it is pre-existing. It is being handled separately (#2407 and #914) and
another investigator has since shown, in a period-correct container, that it is a *distinct*
defect from the one #914 originally reported.

Nothing about that crash is specific to this branch, and it did not recur in either x86_64
rerun above.

### Unchanged from before

No test exercises a link map where the loader redirected a dependency; building one needs a
hwcaps directory and therefore root. Only cyclops would catch a regression in the matcher
itself. The in-tree tests added here cover the name plumbing beneath it.

